PR #10, opened 2015-04-22
Fix typo, path, missing semicolon.

shurcooL:
Hi there.

First off, thanks for making this! It seems very helpful and needed.

I'm still learning about grpc and protobufs, so I may be mistaken, but I think there were a few issues in the README that I had to figure out in order to get it to work. Please take a look.

I think they are valid fixes.

The first change:

```diff
@@ -73,14 +73,14 @@ Make sure that your `$GOPATH/bin` is in your `$PATH`.
    +    option (gengo.grpc.gateway.ApiMethodOptions.api_options) = {
    +      path: "/v1/example/echo"
    +      method: "POST"
-   +    }
+   +    };
    +  }
     }
    ```
```

I needed to add that semicolon, otherwise an error was being generated when running `protoc` command:

```
sample.proto:46:3: Expected ";".
```

Next change:

```diff
 3. Generate gRPC stub
    
    ```sh
-   protoc -I/usr/local/include -I. -I$GOPATH \
+   protoc -I/usr/local/include -I. -I$GOPATH/src \
      --go_out=plugins=grpc:. \
      path/to/your_service.proto
    ```
    
    It will generate a stub file `path/to/your_service.pb.go`.
    Now you can implement your service on top of the stub.
 4. Generate reverse-proxy
    
    ```sh
-   protoc -I/usr/local/include -I. -I$GOPATH \
+   protoc -I/usr/local/include -I. -I$GOPATH/src \
      path/to/your_service.proto
    ```
```

I needed to make that change, otherwise `protoc` command generated this error:

```
sample.proto: Import "github.com/gengo/grpc-gateway/options/options.proto" was not found or had errors.
```

This change makes sense, as `options.proto` file is indeed located at `$GOPATH/src/github.com/gengo/grpc-gateway/options/options.proto` and not `$GOPATH/github.com/gengo/grpc-gateway/options/options.proto`.

The rest are just typo fixes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,14 +73,14 @@ Make sure that your `$GOPATH/bin` is in your `$PATH`.
    +    option (gengo.grpc.gateway.ApiMethodOptions.api_options) = {
    +      path: "/v1/example/echo"
    +      method: "POST"
-   +    }
+   +    };
    +  }
     }
    ```
 3. Generate gRPC stub
    
    ```sh
-   protoc -I/usr/local/include -I. -I$GOPATH \
+   protoc -I/usr/local/include -I. -I$GOPATH/src \
      --go_out=plugins=grpc:. \
      path/to/your_service.proto
    ```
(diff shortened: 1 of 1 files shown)


yugui:
Thank you!

PR #12, opened 2015-05-04
Replace our own custom option with the one defined by Google

yugui:
# Background
* https://groups.google.com/forum/#!msg/grpc-io/Xqx80hG0D44/VNCDHjeE6pUJ

The new option brings:
* More expressiveness of method mapping
* Compatibility to api definitions released by Google

# Overview
* Import the definition files of the custom option from github.com/google/googleapis.
* Deprecate our own custom option
* Extract the parse process of CodeGeneratorRequest from code template execution
* Add a parser of path pattern rule defined in googleapis
* Add a stack machine which executes the parsed rule
* Add a compiler which compiles the parsed rule into opcodes in the stack machine
* Redesign the code template on top of this new stuff
* Update examples


shurcooL:
Hi @yugui,

Thank you for your excellent work on this!

I've been testing out an earlier version of this branch. I'll be updating to the latest and giving it a try next.

README.md
```diff
@@ -70,7 +70,7 @@ Make sure that your `$GOPATH/bin` is in your `$PATH`.
     service YourService {
    -  rpc Echo(StringMessage) returns (StringMessage) {}
    +  rpc Echo(StringMessage) returns (StringMessage) {
-   +    option (gengo.grpc.gateway.ApiMethodOptions.api_options) = {
+   +    option (google.api.http) = {
    +      path: "/v1/example/echo"
    +      method: "POST"
```
shurcooL: Doesn't this part of the example need to be updated to match the Google api definitions?

If I understand correctly, it should now be:

```
post: "/v1/example/echo"
body: "*"
```

examples/echo_service.proto
```diff
@@ -10,15 +10,15 @@ message SimpleMessage {
 
 service EchoService {
 	rpc Echo(SimpleMessage) returns (SimpleMessage) {
-		option (gengo.grpc.gateway.ApiMethodOptions.api_options) = {
-			path: "/v1/example/echo/:id"
-			method: "POST"
+		option (google.api.http) = {
+			post: "/v1/example/echo/{id}"
+			body: "*"
```
shurcooL: Is it right that this example has `body: "*"` while id is already specified as part of the URL path?

The `rpc EchoBody` also has `body: "*"` which makes more sense since that's where the `id` of `SimpleMessage` woudl be provided.

I may not have a great understanding of how `body` is used exactly, is this described anywhere? **Edit:** I see it is described [here](https://github.com/google/googleapis/blob/3ee9064570fef8838792dda7da9bd0ef973597c8/google/api/http.proto#L184-L187). I think my comment is valid and this was probably a copy paste error. Can you confirm if that's the case?

protoc-gen-grpc-gateway/gengateway/template.go
```diff
@@ -0,0 +1,198 @@
...
+package {{.GoPkg.Name}}
+import (
+	{{range $i := .Imports}}{{if $i.Standard}}{{$i | printf "%s\n"}}{{end}}{{end}}
+
+	{{range $i := .Imports}}{{if not $i.Standard}}{{$i | printf "%s\n"}}{{end}}{{end}}
+)
+
+var _ codes.Code
+var _ io.Reader
+var _ = runtime.String
```
shurcooL: You may want to insert `var _ = json.Marshal` or similar here, otherwise it will potentially generate Go code that does not compile:

```
....pb.gw.go:13: imported and not used: "encoding/json"
```

third_party/googleapis/google/api/annotations.proto
```diff
@@ -0,0 +1,29 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package google.api;
+
+import "google/api/http.proto";
```
shurcooL: Would you consider changing this import?

```diff
-import "google/api/http.proto";
+import "github.com/gengo/grpc-gateway/third_party/googleapis/google/api/http.proto";
```

The motivation is that, for generating .pb.gw.go files in user programs, users would typically do something like this, as described in README:

```
protoc -I/usr/local/include -I. -I$GOPATH/src \
 path/to/your_service.proto
```

That would cause the `google/api/http.proto` import to not be found. It would require users to add an additional include path directive:

```
protoc -I/usr/local/include -I. -I$GOPATH/src \
 -I$GOPATH/src/github.com/gengo/grpc-gateway/third_party/googleapis \
 path/to/your_service.proto
```

Which is extra work that can be eliminated. What do you think?

To make it work, all imports `google/api/http.proto` would be need to be updated to `github.com/gengo/grpc-gateway/third_party/googleapis/google/api/http.proto`. It also seems to be more in spirit of Go to use a full import path rather than a relative one.

slimsag: I think that changing the import path is perhaps incorrect: `google/api/http.proto` seems to be the canonical import path as [files like these](https://github.com/google/googleapis/blob/master/google/api/annotations.proto#L19-L20) in the official Google repository import it that way.

> It also seems to be more in spirit of Go to use a full import path rather than a relative one.

In Go I would certainly agree, but with `protoc` I don't think there is any such preference.

yugui: I agree with @slimsag. `google/api/http.proto` looks to be more appropriate as a canonical path than `github.com/gengo/grpc-gateway/third_party/googleapis/google/api/http.proto`.

BTW, I should have added `-I$GOPATH/src/github.com/gengo/grpc-gateway/third_party/googleapis` in the example in `README.md`.

examples/echo_service.proto
yugui: Done. Corrected.

README.md
yugui: Done.

third_party/googleapis/google/api/annotations.proto
shurcooL: That sounds good to me, thanks for thinking about it. `google/api/http.proto` indeed seems better if it's the canonical path even after vendoring.

protoc-gen-grpc-gateway/gengateway/template.go
yugui: Done.

README.md
```diff
@@ -60,9 +60,9 @@ Make sure that your `$GOPATH/bin` is in your `$PATH`.
    ```diff
     syntax = "proto3";
     package example;
-    
-   +import "github.com/gengo/grpc-gateway/options/options.proto";
-    
+   +
+   +import "github.com/gengo/grpc-gateway/third_party/googleapis/google/api/annnotations.proto";
```
rstrong-pica9: One too many `n`'s in annotations

yugui: done. thx

shurcooL:
Is the implementation of query parameters complete?

shurcooL:
By the way, thank you very much for your excellent work on this @yugui, it is very helpful and much appreciated!

yugui:
@shurcooL Not yet. But it was enough to replace the old implementation.
I am working on query parameters and other updates from Wolfgang in another branch.
https://github.com/gengo/grpc-gateway/pull/14

PR #21, opened 2015-06-18
Add support for non-nullable nested messages.

### LINKED ISSUE #20: Feature Request: Support for non-nullable nested messages.
[gogo protobuf generator](https://github.com/gogo/protobuf) has an option to generate non-nullable nested messages via an `[(gogoproto.nullable) = false]` option.

Could support for that be added to grpc-gateway in its `populateQueryParameter` func?

Since the code change is really small, I will make a PR that resolves this issue, but of course this it will need discussion/review.

shurcooL:
Nested messages with "[(gogoproto.nullable) = false]" option are generated as structs, rather than pointers to structs. Add support for such struct values to populate parameter functionality.

Add test case.

Resolves #20.

This PR is in need of review and discussion before merging.

### DIFF
--- a/runtime/query.go
+++ b/runtime/query.go
@@ -59,6 +59,9 @@ func populateQueryParameter(msg proto.Message, fieldPath []string, values []stri
 			}
 			m = f.Elem()
 			continue
+		case reflect.Struct:
+			m = f
+			continue
 		default:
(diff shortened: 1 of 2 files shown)


yugui:
LGTM.  Thank you for your contribution.

PR #29, opened 2015-07-08
Format error message in JSON

### LINKED ISSUE #25: Format Errors to JSON
Would it make sense to have some way to format the errors received from the RPC call into JSON? Currently, this [function](https://github.com/gengo/grpc-gateway/blob/f748a4771dce6d6974393fdace000eba85fd24a3/runtime/errors.go#L57) is called when an error occurs. A client of the HTTP server would be expecting JSON in the response though.

The response code is available which could indicate the error type but passing along an error message in JSON would be useful for the client.

Thoughts?

yugui:
Make the error format customizable.  Fixes #25.


hayate:
LGTM

PR #30, opened 2015-07-08
Add f.Flush() to runtime.ForwardResponseStream

vvakame:
I think Flush is needed on receive JSON streaming.


hayate:
there is already a `f.Flush()` before the loop (https://github.com/gengo/grpc-gateway/pull/30/files#diff-c90e18d9d959604ec95823fb307c9f65R31)

yugui:
lgtm

PR #31, opened 2015-07-13
[Experimental] Make the response forwarder function customizable

yugui:
Sometimes RESTful API needs something does not fit to just call-return semantics -- e.g. redirection based on the result.
It looks to be natural to support such a thing in an the gateway rather than to emit extra information from gRPC backend.

Users can do that by customizing the forwarder function.


trinchan:
LGTM! thanks for this!

PR #34, opened 2015-07-27
Reflection fix of proto3 nested messages.

mwitkow-io:
This doesn't require any changes to golang/protobuf.
And reuses the code already written for query string filtering.


mwitkow-io:
This fixes https://github.com/gengo/grpc-gateway/issues/32

PR #38, opened 2015-08-13
rename internal to utilties for 1.5 compatibility

tmc:
fixes #36

PR #39, opened 2015-08-13
fix examples to work with go1.5

### LINKED ISSUE #37: Examples break with 1.5 because of import of "main" examples package
```
can't load package: $GOPATH/src/github.com/gengo/grpc-gateway/examples/server/main.go:10:2: import "github.com/gengo/grpc-gateway/examples" is a program, not an importable package
```

tmc:
Fixes #37


yugui:
LGTM, but IIUC it depends on #38 thus it is not ready to merge?

PR #42, opened 2015-08-17
Fix undefined variable error in generated codes

### LINKED ISSUE #35: Gateway code generation broken when rpc method with a streaming response has an input paramter
For example, for this proto file

```proto
syntax = "proto3";

package grpc.testing;

option java_package = "grpc.testing";

message Price {
	string symbol = 1;
	int32 price =2;
	// adding this later to existing code
	repeated string additionalSymbols = 3;
	// again added after the fact
	map<string,int32> symbolIds = 4;
}

message Id {
	int32 clientId = 1;
}

service TestService {
	rpc Ping (Price) returns (Price) {
		option (google.api.http) = {
			get: "/v1/test/ping/{symbol}/{price}"
		};
	}

	rpc StreamingPing (Price) returns (stream Price) {
		option(google.api.http) = {
			post: "/v1/test/streamingping/{symbol}"
		};
	}

	rpc PubSubPing (Id) returns (stream Price);

	rpc BiDirPing (stream Id) returns (stream Price);
}
```

The gateway code generated for the StreamingPing method does not compile. This only happens when the URL has a input param for a method with a streaming response.

yugui:
Fixes #35


examples/main.go
```diff
@@ -13,6 +13,7 @@ import (
 var (
 	echoEndpoint = flag.String("echo_endpoint", "localhost:9090", "endpoint of EchoService")
 	abeEndpoint  = flag.String("more_endpoint", "localhost:9090", "endpoint of ABitOfEverythingService")
+	flowEndpoint = flag.String("flow_endpoint", "localhost:9090", "endpoint of ABitOfEverythingService")
```
trinchan: `endpoint of ABitOfEverythingService` -> `endpoint of FlowService`

trinchan:
LGTM except small comment :)

examples/main.go
yugui: done

PR #47, opened 2015-08-30
Consider tail segments after deep wildcard

### LINKED ISSUE #23: Values containing slashes in URL template?
From reading https://github.com/google/googleapis/blob/master/google/api/http.proto, it suggests that some parameters may be a part of the URL template, for example:

```
get: "/v1/messages/{message_id}"
get: "/v1/users/{user_id}/messages/{message_id}"
```

HTTP | RPC
-----|-----
GET /v1/messages/123456 | GetMessage(message_id: "123456")
GET /v1/users/me/messages/123456 | GetMessage(user_id: "me" message_id: "123456")

However, if I'm [reading this spec right](https://github.com/google/googleapis/blob/9b937a277493f73c531f517c340ac63f97b0d029/google/api/http.proto#L182-L189), those values may not contain slashes, can they?

For example, suppose I wanted to define:

```
get: "/v1/dirs/{dir_path}"
```

And have the following HTTP requests match:

HTTP | RPC
-----|-----
GET /v1/dirs/foo | GetDirs(dir_path: "foo")
GET /v1/dirs/foo/bar | GetDirs(dir_path: "foo/bar")
GET /v1/dirs/foo/bar/baz | GetDirs(dir_path: "foo/bar/baz")

Only the first will get matched, the other two will not. Is that so?

Is it true that there's no way to have all 3 of the above paths get matched, as long as grpc-gateway follows that `google.api.http` spec? Is the only alternative to use URL queries, like
`GET /v1/dirs?dir_path=foo/bar/baz`?

If so, that seems to put a heavy restriction on one's ability to create RESTful APIs using `grpc-gateway`, since it's often desirable to have some slash separated paths as values.

yugui:
Fixes #23.
c.f. https://groups.google.com/d/msg/grpc-io/51-kF_Fof2o/ddalFby8gh8J

### DIFF
--- a/runtime/pattern.go
+++ b/runtime/pattern.go
@@ -173,8 +173,13 @@ func (p Pattern) Match(components []string, verb string) (map[string]string, err
 			stack = append(stack, c)
 			pos++
 		case utilities.OpPushM:
-			stack = append(stack, strings.Join(components[pos:], "/"))
-			pos = len(components)
+			end := len(components)
+			if end < pos+p.tailLen {
+				return nil, ErrNotMatch
+			}
+			end -= p.tailLen
+			stack = append(stack, strings.Join(components[pos:end], "/"))
+			pos = end
 		case utilities.OpConcatN:
 			n := op.operand
 			l := len(stack) - n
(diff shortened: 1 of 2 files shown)


hayate:
:+1:  LGTM

PR #48, opened 2015-09-01
Follow a change of go package name convention in protoc-gen-go

### LINKED ISSUE #22: Multiple .proto files generates invalid import statements.
passing multiple .proto files that include an import generates invalid import lines.

ex: compile a.proto and b.proto together with a importing b:
```go
import (
proto_0 "b.pb"
)
```
would be generated.

yugui:
Follows the change golang/protobuf@dded9133a99a3cd7c3a9d24a9f85c2b8ef76ff31.

This fixes #22 as a side-effect.


hayate:
LGTM

PR #49, opened 2015-09-03
Configure Travis CI

yugui:
* Upgrade protoc version
* Upgrade go version
* Make CI faster and strict

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,8 @@ before_install:
 install:
 - go get github.com/gengo/grpc-gateway/protoc-gen-grpc-gateway
 script:
-- make examples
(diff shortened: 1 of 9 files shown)


.travis.yml
```diff
@@ -1,26 +1,29 @@
+branches:
+  only:
+    - master
 language: go
-
+sudo: false
 go:
-- 1.3
+- 1.4
```
shurcooL: Why 1.4 and not 1.5, which is the latest official release?

yugui: certainly. fixed.

hayate:
LGTM

.travis.yml
shurcooL: Thanks!

PR #50, opened 2015-09-03
Fix a build error with the latest protoc-gen-go

yugui:
Follows golang/protobuf@59b73b37c1e45995477aae817e4a653c89a858db,
which introduces an incompatible change in generated codes


hayate:
:+1: LGTM

PR #53, opened 2015-09-15
Add runtime.WithForwardResponseOption

peter-edge:
This one may be a bit more work back and forth. The idea is that sometimes, you want to be able to modify values prior to sending the response over the wire. An example of this is https://go.pedge.io/dockervolume - I'd like to be able to modify the VolumeDriver interface to be a Protocol Buffers interface, and then use grpc-gateway to handle the API calls, but to do so, I need to be able to set a custom content type for API versioning with Docker.

This change should be backwards compatible. It also includes the changes from https://github.com/gengo/grpc-gateway/pull/52 so that the tests pass.


peter-edge:
I'm not sure on the Travis build failure - advise?

peter-edge:
Tests passing on travis now

peter-edge:
Ping :)

peter-edge:
Hey @yugui, what do you think of this PR? I need this ASAP, can we pursue this?

yugui:
@peter-edge sorry for being late.  I'll take a look.

runtime/handler.go
```diff
@@ -68,8 +74,32 @@ func ForwardResponseMessage(ctx context.Context, w http.ResponseWriter, req *htt
...
 	}
 
-	w.Header().Set("Content-Type", "application/json")
 	if _, err = w.Write(buf); err != nil {
 		glog.Errorf("Failed to write response: %v", err)
 	}
 }
+
+func handleForwardResponseOptions(ctx context.Context, w http.ResponseWriter, resp proto.Message, opts []func(context.Context, http.ResponseWriter, proto.Message) error) error {
+	if opts == nil || len(opts) == 0 {
```
yugui: `opts == nil` is redundant.

runtime/handler.go
```diff
@@ -60,7 +60,13 @@ func ForwardResponseStream(ctx context.Context, w http.ResponseWriter, req *http
 }
 
 // ForwardResponseMessage forwards the message "resp" from gRPC server to REST client.
-func ForwardResponseMessage(ctx context.Context, w http.ResponseWriter, req *http.Request, resp proto.Message) {
+func ForwardResponseMessage(ctx context.Context, w http.ResponseWriter, req *http.Request, resp proto.Message, opts ...func(context.Context, http.ResponseWriter, proto.Message) error) {
```
yugui: Define a type `type ForwardResponseOption func(context.Context, http.ResponseWriter, proto.Message)`

yugui:
Could you resolve the merge conflict?

runtime/handler.go
peter-edge: removed

runtime/handler.go
peter-edge: how do you create an instance of this then? I originally did this, but is it legal to do:

```go
someOption = runtime.ForwardResponseOption(ctx context.Context, w http.ResponseWriter, message proto.Message) { ... }
```

I was not sure how to do this, to be honest.

peter-edge:
merge conflict fixed

PR #54, opened 2015-09-15
Add log package

peter-edge:
This is probably not ready to merge (but does pass tests), but is more for discussion as to what to do here.

I generally use other log implementations than glog, and others do too. It is nice to be able to choose an alternative logger, so I added a simple interface here for a Logger that takes care of most functionality. Instead of the V system that glog has, if you like this direction, I can add Debugf/Debugln to somewhat take care of that situation.

The package github.com/gengo/grpc-gateway is currently imported in runtime/mux.go, so that glog is still the default. Probably not the best place for this (again, this PR probably needs work, but wanted initial feedback).

Reason for doing *ln: all logging generally does newlines anyways, or handles newlines, so it makes more sense to me. Also this means that my https://github.com/peter-edge/go-protolog/blob/master/protolog.go#L59 Logger is a drop-in, which is what I want to use this for anyways :)

I did a similar thing for https://github.com/grpc/grpc-go in https://github.com/grpc/grpc-go/tree/master/grpclog

### DIFF
--- a/examples/examplepb/a_bit_of_everything.pb.gw.go
+++ b/examples/examplepb/a_bit_of_everything.pb.gw.go
@@ -15,9 +15,9 @@ import (
 	"net/http"
 
 	"github.com/gengo/grpc-gateway/examples/sub"
+	"github.com/gengo/grpc-gateway/log"
 	"github.com/gengo/grpc-gateway/runtime"
 	"github.com/gengo/grpc-gateway/utilities"
-	"github.com/golang/glog"
 	"github.com/golang/protobuf/proto"
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
@@ -218,7 +218,7 @@ func request_ABitOfEverythingService_CreateBody_0(ctx context.Context, client AB
 func request_ABitOfEverythingService_BulkCreate_0(ctx context.Context, client ABitOfEverythingServiceClient, req *http.Request, pathParams map[string]string) (proto.Message, error) {
 	stream, err := client.BulkCreate(ctx)
 	if err != nil {
-		glog.Errorf("Failed to start streaming: %v", err)
+		log.Errorf("Failed to start streaming: %v", err)
 		return nil, err
 	}
 	dec := json.NewDecoder(req.Body)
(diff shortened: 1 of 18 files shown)


shurcooL:
Travis CI build is failing because of `-logtostderr` flag still being present.

In general, I am also not a fan of the `glog` package being used for logging, I wish it were just the standard Go `log` package because it's simpler. But I can imagine other people their own preferred log libraries, so having an interface is the only way to allow that.

I think it's fair to follow the same strategy as `grpc-go` then. That would be consistent.

peter-edge:
Ah yea, I can fix that depending on the owner's feelings on this.

Yeah, there's a few options out there and it's inconsistent. I've started using my own https://go.pedge.io/protolog for structured logging in github.com/pachyderm/pachyderm (which is the initial reason I did this PR, so that logging would be consistent for the pachyderm repository), but I would go as far to say that most people use github.com/Sirupsen/logrus (including github.com/docker/docker), which this Logger interface is a drop-in for https://godoc.org/github.com/Sirupsen/logrus#Logger.

glog is way faster though, for the record...which is why it is nice to have the option :)

peter-edge:
@shurcooL @yugui I put together this https://github.com/peter-edge/go-dlog that might satisfy everyone, this would cover grpc-go too so ping @iamqizhao, we could make this common between both grpc-go and grpc-gateway, which would be a big win. I'll extend the documentation tomorrow morning, but it makes it simple, just `import _ "go.pedge.io/dlog/glog". Thoughs?

yugui:
Although I'm a fan of `glog`, but I like the overall direction of this PR.

I have some comments on implementation details.
* I don't want you to add dependency on `go-dlog` for now for two reasons.
  * I don't want to add much dependency on things which neither `grpc-go` or `protoc-gen-go` depends on.
* Some `glog.V(1)` logs are too verbose for `Info` level.  How about adding `Debug` level?  Is it too far from `grpclog`?

peter-edge:
My hope would be I could get both you and @iamqizhao to agree on a common implementation, and I figured dlog might be a good answer haha :) If you add a Debug level to grpclog, it's basically dlog, but dlog makes it easier to bring in other implementations and provides other functionality (tomorrow morning I'd send a pull request to grpc-go with it for consideration).

But honestly it really doesn't matter much to me - dlog clearly only took 30 minutes to put together.

Also of note, I've benchmarked glog via proxy on an interface with protolog, and the difference is not measurable as I can see.

peter-edge:
I thought I might have commented on this last night: re: the logging levels

In dlog (or whatever we go with), we could add something like:

```
package glog // the one in go.pedge.io/dlog/glog

var (
  debugVLevel = 1
)

func SetDebugVLevel(level int32) {
  debugVLevel = level
}

...

func (l *logger) Debug(args ...interface{}) {
  glog.V(debugVLevel).Info(args...)
}
```

For reference, this is what I'm referring to: https://github.com/peter-edge/go-dlog/blob/master/glog/glog.go

yugui:
Closing this PR due to this reason:
> I don't want you to add dependency on go-dlog for now. It is because I don't want to add much dependency on things which neither grpc-go or protoc-gen-go depends on.

and because https://github.com/grpc/grpc-go/pull/341 was closed.

PR #55, opened 2015-09-15
Fix parsing of verb and final path component.

hbchai:
The final path component was being considered the verb, and vice versa.


yugui:
LGTM

PR #56, opened 2015-09-15
Add option to format JSON response with an indent.

hbchai:
This would support human-readable responses, when desired, by setting the runtime.ResponseIndent global. Just the non-streaming response can be indented for now.

### DIFF
--- a/runtime/handler.go
+++ b/runtime/handler.go
@@ -1,6 +1,7 @@
 package runtime
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -9,6 +10,11 @@ import (
 	"github.com/golang/glog"
 	"github.com/golang/protobuf/proto"
 	"golang.org/x/net/context"
+)
+
+var (
+	// The indent to use for JSON responses.
+	ResponseIndent = ""
 )
 
 type responseStreamChunk struct {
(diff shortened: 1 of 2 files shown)


yugui:
Interesting.  But can I consider this PR after #53?  I guess this feature gets easier to implement without adding hacky flag once we complete #53.

hbchai:
Sounds reasonable. I think options are probably better in the long run, as long as they are easy to set.

runtime/handler.go
```diff
@@ -11,6 +12,11 @@ import (
 	"golang.org/x/net/context"
 )
 
+var (
+	// The indent to use for JSON responses.
+	ResponseIndent = ""
```
peter-edge: Generally I see public variables in golang as constants, but there are exceptions - I would generally do something like:

```
var (
  responseIndent = ""
)

func SetResponseIndent(s string) {
  responseIndent = s
}
```

But just my opinion.

shurcooL: @peter-edge, what would the benefit of having `SetResponseIndent` be?

mikeatlas: @shurcooL the `SetResponseIndent` could easily do things like strip invalid indentation characters, particularly via a whitelist of simply `[" ","\t"]`, for example. Getter/Setter pattern hides implementation details and can help protect invalid usage of exported variables in a package.

yugui:
Now we can indent JSON outputs with JSONPb marshaler.

PR #57, opened 2015-09-15
Generate Docker image gengo/grpcgateway, use CircleCI/Docker instead of Travis

peter-edge:
This is another one of those "suggestion" PRs, I'll try to explain:

For a while, I've been compiling protos using https://hub.docker.com/r/pedge/proto3grpc/, and more recently https://hub.docker.com/r/pedge/protolog/ which is based on pedge/proto3grpc but includes protoc-gen-protolog. The README on pedge/proto3grpc explains it - if you link your working directory as a volume inside the Docker container, you can compile protos without needing protoc/grpc/etc installed on your local computer. It makes it easy to get other team members on board. I also took this a step further with https://github.com/peter-edge/go-tools/tree/master/protoc-all, but that's less relevant.

But anyways, I've been maintaining pedge/proto3grpc, but it naturally fits as something produced from grpc-gateway. So I moved it over here, along with a few modifications, but also if that was being done, might as well just make it how grpc-gateway does CI. I do this in https://github.com/peter-edge/go-protolog/blob/master/circle.yml for comparison. On every push to master, CircleCI will build the image, run the tests, and if they pass, push a new image version to the Docker Hub. You have to set up CircleCI and put a docker user/password/email, but I can help you with that, it's easy (and free).

Note that the tests do not pass right now because there are pre-existing golint and go vet failures (not sure why they do not get picked up right now, maybe some weird versioning issue)?

```
golint github.com/gengo/grpc-gateway/protoc-gen-grpc-gateway/...
/Users/peter/go/src/github.com/gengo/grpc-gateway/protoc-gen-grpc-gateway/gengateway/generator.go:27:36: exported func New returns unexported type *gengateway.generator, which can be annoying to use
golint github.com/gengo/grpc-gateway/runtime/...
golint github.com/gengo/grpc-gateway/utilities/...
go vet github.com/gengo/grpc-gateway/protoc-gen-grpc-gateway/...
go vet github.com/gengo/grpc-gateway/runtime/...
runtime/errors_test.go:31: constant 5 not a string in call to Errorf
```

Let me know if this is something you would be interested in. If so, this way of compiling should probably be additionally documented in grpc-gateway's README.


yugui:
Personally I don't think it's worth to have a docker image because it is not so hard to build protoc-gen-grpc-gateway if the client has a dev environment of golang, and the client must have the dev environment in any way to build generated go files.

Also, we have no plan to switch from Travis CI to CircleCI for consistency to other gengo repositories.

PR #64, opened 2015-10-21
Fix `error` field always in chunk response

mwitkow-io:
Because the `responseStreamError` was an embedded object and not a pointer, it was always added to the response (the `omitempty` didn't work).

Moving it to be a pointer solves the problem and makes good results always have only `result` field, while errors only have the `error` field.


yugui:
LGTM

PR #68, opened 2015-12-02
Add swagger support

achew22:
Yuki,

I started to create a grpc-gateway swagger client which generates a .swagger.json file for every proto that you load in. This is what I have right now. It is broken into two commits. The first clones protoc-gen-grpc-gateway into protoc-gen-swagger and the 2nd modifies it to emit a really terrible swagger definition.

Right now it has a number of shortcomings.

1. It doesn't take parameters in requests
2. It generates both a post and a get (and no other verbs)
3. It's super hacky
4. I have no idea what I'm doing
5. Doesn't load any comment information into the swagger spec which would make it MUCH easier to understand the output

I was wondering if you could provide some general feedback. This is the first thing of any size I've ever done in go so sorry if it isn't idiomatic. I'm seeking comments on places that you think could be improved but please do the delta from the 2nd commit on or it will be very painful to review. "This looks like it's on the right track" or "try again with this strategy" is ultimately what I would like to get from you.

Thanks so much for creating this project! It is making my life so much less painful.


yugui:
Amazing!

I'm afraid, but it takes a bit long to take a look at this PR since it is large.
Maybe I have some time to do that this weekend.

yugui:
@achew22 
Thank you.  I really like the overall direction, e.g.
1. Map proto message to predefined message schema
2. Construct Swagger definition in process and marshal it in JSON afterward.

There are some comments.
1. You don't have to copy entire the protoc plugin. Code generator part of `protoc-gen-grpc-gateway` is pluggable by intention.  So you can reuse the existing message/option parser. Feel free to propose some extension of the parser if is functionality is not sufficient for your purpose.
2. I'm not sure, but sometimes it might be better to define `parameters` of paths with `parameterObject` rather than `referenceObject` because it is allowed to place some fields in request paths.

achew22:
I have been on vacation for the past week. I intend to start taking a look at this in my spare time over this week.

yugui:
:+1:

achew22:
@yugui, I have done a bit of cleanup on this PR. Just for reference, once you're happy with it I would like to squash it into a single commit, so don't just hit the merge button :smile:.

I think your suggestion about using a parameter object is a great one. I looked at the docs and agreed with you when I was writing but I couldn't find a compelling way to do union types in Golang. Do you know what the best way to do a union type is? That seems to me like the only way to do it unfortunately.

I have updated the branch with changes to make it reuse the existing protoc-gen-grpc-gateway instead of my clone. Turns out almost all the stuff I wanted was there.

protoc-gen-swagger/genswagger/generator.go
```diff
@@ -0,0 +1,113 @@
+package genswagger
+
+import (
+	"errors"
+	"fmt"
+	// Don't use the formatter
```
yugui: Could you just remove this if you don't need "go/format"?

protoc-gen-swagger/genswagger/generator.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+type generator struct {
+	reg         *descriptor.Registry
+	baseImports []descriptor.GoPackage
+}
+
+// New returns a new generator which generates grpc gateway files.
+func New(reg *descriptor.Registry) *generator {
+	var imports []descriptor.GoPackage
```
yugui: I guess you don't need to manage go packages

protoc-gen-swagger/main.go
```diff
@@ -0,0 +1,115 @@
+package main
```
yugui: Originally I thought about adding a flag to protoc-gen-grpc-gateway to let it generate swagger schema.
What do you think about that?  Do you think it is better to make it an independent protoc plugin?

protoc-gen-grpc-gateway/httprule/compile.go
```diff
@@ -105,10 +107,11 @@ func (t template) Compile() Template {
...
-		OpCodes: ops,
-		Pool:    pool,
-		Verb:    t.verb,
-		Fields:  fields,
+		Version:  opcodeVersion,
+		OpCodes:  ops,
+		Pool:     pool,
+		Verb:     t.verb,
+		Fields:   fields,
+		Template: t.template,
```
yugui: Strictly speaking, the template format of `google.api.HttpRule` and the one of Swagger are different.   But your approach is good enough as a first step.

IIUC, `/path/{param=foo/*}/` in `google.api.HttpRule` needs to be `/path/foo/{param}` in Swagger.

yugui:
https://travis-ci.org/gengo/grpc-gateway/builds/97933023

The test is broken because master branch was broken.
Maybe you just need to rebase onto the current master to fix the test.

protoc-gen-swagger/main.go
achew22: That's definitely an interesting idea. From what I've seen in the proto community is that people generally expect different plugins to protoc. Do you know of any examples of people passing flags to change the output to a different language?

protoc-gen-grpc-gateway/httprule/compile.go
achew22: Yeah, really good tip. Having the template available in the swagger side will allow me to split it and iterate over it to fix those entries up.

achew22:
The thing I'm stuck on right now is that you can import arbitrary protos into your api. They don't necessarily have to be included in the file that you're currently compiling. My plan thus far had been to use "#/definitions/{{message.FQMN()}}" for all the definitions. This is great, except when it comes to iterating over all the messages, I get stuck because they may not be in the file.

Right now I'm expecting that I will have to create a registry of messages, like you did, then recurse through all the fields creating a set of messages that have been seen. I was planning on using a hashmap from FQMN to the actual message object then in the stage where I generate the definitions I can just emit each entry as a definition. Looking at the go code it solves this by importing 1 layer deep and then using the protoc-go library's output to solve the problem of transitive dependence, but you might have a cool trick up your sleeve so I thought I would ask.

achew22:
Now I remember why this was hard! Field [https://godoc.org/github.com/gengo/grpc-gateway/protoc-gen-grpc-gateway/descriptor#Field] has a message attached, but it isn't the message I would expect. It is the message of it's parent. This is, no doubt, useful but not in the context that I need it. Lemme go dig in to see if I can figure out how to add a FieldMessage field to the Field struct. Meta-programming is hard!

achew22:
After a bit of playing, I just plumbed the registry through to the generator and now I'm able to resolve these. Thanks for writing that!

achew22:
Sorry, I accidentally closed this by pushing your master to github. It's all fixed now.

I think this is getting a lot closer. I have created a non-streaming example file which is the a bit of everything with the word streaming removed https://raw.githubusercontent.com/achew22/grpc-gateway/master/examples/examplepb/streamless_everything.swagger.json

If you want to see how their API viewer sees it, check out http://petstore.swagger.io/#/ and type that URL in at the top. If you want to see what linter checks it is failing you can go to https://editor.swagger.io and do the same thing.

Things that still need to be done:
- [ ] EmptyProto is still an interesting edge case. Right now it is not being included in output and I haven't figured out why.
- [ ] Even if empty proto is fixed in the swagger codegen to emit an empty dictionary, their editor will still barf on it https://github.com/swagger-api/swagger-spec/issues/452
- [ ] Repeated fields need to be marked as arrays
- [ ] Support for enums
- [ ] Parameters should not all be typed as "String". They have type information and we should represent that.
- [ ] More tests

yugui:
I see. I'm looking forward to seeing it.

achew22:
Alright, I think that the swagger output from this generator is valid (in spite of the bug in their validator) and I am able to generate swagger json definitions for all the test cases that you provided/I created. I will follow up with some unit tests but I would appreciate if you could take a look at it in its current form. I didn't intend for this to take so long, sorry about that but it's all there for the most part now.

protoc-gen-swagger/main.go
yugui: I have no good example.  But it is reasonable for me to put them together because the two generators must be consistent.

For me, they look to be similar to the relationship between protoc-gen-go and gprc-plugin rather than plugins for different languages.

yugui:
@achew22 
I want to hear your opinion about https://github.com/gengo/grpc-gateway/pull/68/files#r50193073.
Besides the comment and other small coding-level issues, I like the overall direction.

protoc-gen-swagger/main.go
achew22: I think you're referring to https://github.com/golang/protobuf/tree/master/protoc-gen-go which is the only protoc-gen-go and grpc-plugin I'm aware of. If I'm talking past you, please disregard.

It sounds to me like you're trying to ensure that the two generators are going to be versioned the same. Respectfully, I think this concern is not something that can be solved at this level. If you're a repository maintainer, you need to be tracking all the code generators you depend upon and staying up to date. I've several times gotten broken by drift between protoc-gen-go and grpc-gateway's gw definitions (not to mention the other generators). I think that keeping things in sync is a known issue with the entire ecosystem.

This is especially true because the code generation doesn't stop here. With the introduction of swagger support I am hooking up a JS code generator based on swagger-codegen (I will push that upstream). There are now 4 independent code generators in my project (protoc, protoc-gen-go, protoc-gen-grpg-gateway, and swagger-codegen). I expect to become familiar with the upgrade process for each.

Further, using the protoc-gen-go has a fairly obtuse parameter passing syntax --gen-go_out=plugins=grpc:. is puzzling at first, to say the least. I now know the syntax but to learn it I had to put log.Printf statements into the code, recompile it and waste a day or two on digging through code to learn why gRPC bindings weren't being created. Having these multiple modes which each produce valid outputs may be "easier" for you/me, but I think that it raises the barrier to entry for code generation which will reduce adoption. In summary, I feel the principle of UNIX is in effect here, "do one thing and do it well."

If your concern is keeping the code behind each generator in sync, I think that is orthogonal to the issue of having multiple binaries. We could have one folder that produces two binaries or two folders that produced a single one, it is just a matter of configuration. What code do you have in mind as being difficult to keep in sync between the two? Maybe we can find a way to abstract that logic into something a bit more pleasant to work with.

Looking forward to your responses and thanks for reviewing my code.

achew22:
PTAL

protoc-gen-swagger/genswagger/generator.go
```diff
@@ -0,0 +1,74 @@
...
+		var formatted bytes.Buffer
+		json.Indent(&formatted, []byte(code), "", "  ")
+
+		name := file.GetName()
+		ext := filepath.Ext(name)
+		base := strings.TrimSuffix(name, ext)
+		output := fmt.Sprintf("%s.swagger.json", base)
+		files = append(files, &plugin.CodeGeneratorResponse_File{
+			Name:    proto.String(output),
+			Content: proto.String(string(formatted.Bytes())),
```
yugui: Use `bytes.Buffer.String()`

protoc-gen-swagger/genswagger/generator.go
```diff
@@ -0,0 +1,74 @@
...
+		})
+		glog.V(1).Infof("Will emit %s", output)
+	}
+	return files, nil
+}
+
+func (g *generator) generate(file *descriptor.File) (string, error) {
+	pkgSeen := make(map[string]bool)
```
yugui: `pkgSeen` is not used.  So I don't think you need this wrapper method.

protoc-gen-swagger/genswagger/template.go
```diff
@@ -0,0 +1,432 @@
...
+
+import (
+	"bytes"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"regexp"
+	"strings"
+
+	//"github.com/golang/glog"
```
yugui: remove this line.

protoc-gen-swagger/genswagger/template.go
```diff
@@ -0,0 +1,432 @@
...
+
+	"github.com/gengo/grpc-gateway/protoc-gen-grpc-gateway/descriptor"
+	pbdescriptor "github.com/golang/protobuf/protoc-gen-go/descriptor"
+)
+
+// Discover all possible enums that can be generated and insert them in the
```
yugui: It is better to make the comment a full sentence(s) to follow the standard godoc convention, although this comment is not extracted by default.

protoc-gen-swagger/genswagger/template.go
```diff
@@ -0,0 +1,432 @@
...
+		// TODO: When grpc-gateway supports enums, extract the possible values for the enums.
+		// At this time I can't determine the proper way to go from enum.GetName to "FQEN".
+		d[enum.GetName()] = swaggerSchemaObject{
+			Type: "string",
+			Enum: enumNames,
+		}
+	}
+}
+
+// Discover all possible Messages that can be generated by the services.
```
yugui: ditto.

protoc-gen-swagger/genswagger/template.go
```diff
@@ -0,0 +1,432 @@
...
+		for _, meth := range svc.Methods {
+			m[meth.RequestType.FQMN()] = meth.RequestType
+			findNestedMessages(meth.RequestType, reg, m)
+			m[meth.ResponseType.FQMN()] = meth.ResponseType
+			findNestedMessages(meth.ResponseType, reg, m)
+		}
+	}
+}
+
+// Recursively discover all possible Messages that can be generated by the services.
```
yugui: ditto.

yugui:
The design looks good to me.
I'll add some coding-level comments.

protoc-gen-swagger/main.go
yugui: Thank you.  It sounds reasonable.  Let's keep the binaries separated.

protoc-gen-swagger/genswagger/template.go
```diff
@@ -0,0 +1,432 @@
...
+
+	// Find all the service's messages that are defined (recursively) and then
+	// write their request and response types out as definition objects.
+	m := messageMap{}
+	findServicesMessages(p.Services, p.reg, m)
+	renderMessagesAsDefinition(m, s.Definitions)
+
+	// We now have rendered the entire swagger object. Write the bytes out to a
+	// string so it can be written to disk.
+	w := bytes.NewBuffer(nil)
```
yugui: You don't have to allocate the buffer in heap.
```go
var w bytes.Buffer
enc := json.NewEncoder(&w)
```

protoc-gen-swagger/genswagger/template.go
```diff
@@ -0,0 +1,432 @@
...
+	pbdescriptor "github.com/golang/protobuf/protoc-gen-go/descriptor"
+)
+
+// Discover all possible enums that can be generated and insert them in the
+// definitions object.  Right now gRPC gateway doesn't support enums during the
+// parse so this is effectively a no-op since the enums object is never
+// created.
+func findEnumerations(enums []*pbdescriptor.EnumDescriptorProto, d swaggerDefinitionsObject) {
+	for _, enum := range enums {
+		enumNames := []string{}
```
yugui: You don't have to allocate zero-length underlying buffer.
```
var enumNames []string
```

protoc-gen-swagger/genswagger/template.go
```diff
@@ -0,0 +1,432 @@
...
+}
+
+// Swagger expects paths of the form /path/{string_value} but grpc-gateway paths are expected to be of the form /path/{string_value=strprefix/*}. This should reformat it correctly.
+func templateToSwaggerPath(path string) string {
+	// It seems like the right thing to do here is to just use
+	// strings.Split(path, "/") but that breaks badly when you hit a url like
+	// /{my_field=prefix/*}/ and end up with 2 sections representing my_field.
+	// Instead do the right thing and write a small pushdown (counter) automata
+	// for it.
+	parts := []string{}
```
yugui: don't have to allocate zero-length buffer.
```go
var parts []string
```

protoc-gen-swagger/genswagger/template.go
```diff
@@ -0,0 +1,432 @@
...
+	}
+
+	return strings.Join(parts, "/")
+}
+
+func renderServices(services []*descriptor.Service, paths swaggerPathsObject) error {
+	for _, svc := range services {
+		for _, meth := range svc.Methods {
+			if meth.GetClientStreaming() || meth.GetServerStreaming() {
+				return errors.New(`Service uses streaming, which is not currently supported. Maybe you would like to implement it? It wouldn't be that hard and we don't bite. Why don't you send a pull request to https://github.com/gengo/grpc-gateway?`)
```
yugui: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

protoc-gen-swagger/genswagger/template.go
```diff
@@ -0,0 +1,432 @@
...
+					case "TYPE_BOOL":
+						paramType = "boolean"
+						paramFormat = "boolean"
+						break
+					case "TYPE_STRING":
+						paramType = "string"
+						paramFormat = "string"
+						break
+					case "TYPE_GROUP":
+						panic("Groups are not allowed in the path for a HTTP RPC. Please use a primitive type instead (string, int64, etc)")
```
yugui: https://github.com/golang/go/wiki/CodeReviewComments#dont-panic

It is sufficient to return error even in this case.

protoc-gen-swagger/genswagger/template.go
```diff
@@ -0,0 +1,432 @@
...
+						break
+					case "TYPE_MESSAGE":
+						panic("Groups are not allowed in the path for a HTTP RPC. Please use a primitive type instead (string, int64, etc)")
```
yugui: ditto.

protoc-gen-swagger/genswagger/template.go
```diff
@@ -0,0 +1,432 @@
...
+						paramType = "integer"
+						paramFormat = "int32"
+						break
+					case "TYPE_SINT64":
+						paramType = "integer"
+						paramFormat = "int64"
+						break
+					default:
+						paramType = parameter.Target.FieldDescriptorProto.Type.String()
+						panic("Detected field type of '" + paramType + "' which is unknown. Please use a primitive type (Ex: string, int64)")
```
yugui: ditto.

protoc-gen-swagger/genswagger/template_test.go
```diff
@@ -0,0 +1,457 @@
...
+						MethodDescriptorProto: meth,
+						RequestType:           msg,
+						ResponseType:          msg,
+						Bindings: []*descriptor.Binding{
+							{
+								HTTPMethod: "POST",
+								PathTmpl: httprule.Template{
+									Version:  1,
+									OpCodes:  []int{0, 0},
+									Template: "/v1/echo", // TODO(achew): Figure out what this hsould really be
```
yugui: hsould --> should ?

protoc-gen-swagger/genswagger/generator.go
achew22: Thank you

protoc-gen-swagger/genswagger/generator.go
achew22: Great eye

protoc-gen-swagger/genswagger/template.go
achew22: Done

protoc-gen-swagger/genswagger/template.go
achew22: Good style note

protoc-gen-swagger/genswagger/template.go
achew22: Thanks

protoc-gen-swagger/genswagger/template.go
achew22: Thanks

protoc-gen-swagger/genswagger/template.go
achew22: Saving memory :+1:

protoc-gen-swagger/genswagger/template.go
achew22: Thank you

protoc-gen-swagger/genswagger/template.go
achew22: Sorry, this is my first real go code. I'm still learning

protoc-gen-swagger/genswagger/template.go
achew22: Thank you

protoc-gen-swagger/genswagger/template.go
achew22: Thank you

protoc-gen-swagger/genswagger/template.go
achew22: Thank you

protoc-gen-swagger/genswagger/template_test.go
achew22: Good eyes :smile:

achew22:
Thank you so much for the review. This is my first bout with Golang and I'm happy to have someone reviewing it. I think I have addressed all your concerns and would appreciate if you could take another look at the code.

yugui:
LGTM.

Thank you for your excellent work!

PR #74, opened 2015-12-15
Add WithValidateRequestOption to offer external validation on requests

gdm85:
Use-case: add CORS support for example, or extra validation/redirects for some requests. I am practically going to use this for the former but I can see it can be used for other purposes too.

It's a bit like the forward options, but runs before anything else as you can see from the template changes; this is also a cheap middleware solution.

### DIFF
--- a/protoc-gen-grpc-gateway/gengateway/template.go
+++ b/protoc-gen-grpc-gateway/gengateway/template.go
@@ -245,6 +245,9 @@ func Register{{$svc.GetName}}Handler(ctx context.Context, mux *runtime.ServeMux,
 	{{range $m := $svc.Methods}}
 	{{range $b := $m.Bindings}}
 	mux.Handle({{$b.HTTPMethod | printf "%q"}}, pattern_{{$svc.GetName}}_{{$m.GetName}}_{{$b.Index}}, func(w http.ResponseWriter, req *http.Request, pathParams map[string]string) {
+		if !runtime.HandleRequestValidateOptions({{$m.GetName | printf "%q"}}, ctx, w, req, pathParams, mux.GetValidateRequestOptions()) {
+			return
+		}
 		resp, err := request_{{$svc.GetName}}_{{$m.GetName}}_{{$b.Index}}(runtime.AnnotateContext(ctx, req), client, req, pathParams)
 		if err != nil {
 			runtime.HTTPError(ctx, w, err)
--- a/runtime/handler.go
+++ b/runtime/handler.go
@@ -87,6 +87,16 @@ func ForwardResponseMessage(ctx context.Context, w http.ResponseWriter, req *htt
 	}
 }
 
+// HandleRequestValidateOptions will run all request validation options before parsing the request and accessing the gRPC server.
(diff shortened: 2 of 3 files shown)


yugui:
I agree that there are such use cases.
But I am wondering if validation just shows an example of more generic interception mechanism.

What do you think about having a more generic middleware (interceptor) system?
e.g.
```go
type Handler interface {
  Handle(context.Context, http.ResponseWriter, *http.Request, map[string]string)
}
type HandlerFunc func(context.Context, http.ResponseWriter, *http.Request, map[string]string)

func (f HandlerFunc) Handle(ctx context.Context, w http.ResponseWriter, r *http.Request, pathParams map[string]string) {
  f(ctx, w, r, pathParams)
}

type Interceptor interface {
  Intercept(next Handler) Handler
}

type InterceptorFunc func(ctx context.Context, w http.ResponseWriter, r *http.Request, pathParams map[string]string, next Handler)

func (f InterceptorFunc) Intercept(next Handler) Handler {
  return HandlerFunc(func(ctx context.Context, w http.ResponseWriter, r *http.Request, pathParams map[string]string) {
    f(ctx, w, r, pathParams, next)
  })
}
```

gdm85:
@yugui yes, that would frame it better. Shall I update the PR or you want to do one yourself?

I also noticed documentation about this feature should probably be added in README.md.

yugui:
@gdm85 I'd appreciate it if you update the PR. But let me thing about if the interface `Handler` is good enough to expose.
I am suspecting if `pathParams` should not be there.

yugui:
@gdm85 I'm sorry but let me change my opinion.

I don't think it is useful to add the interceptor mechanism to grpc-gateway itself because `ServeMux` is just an ordinal implementation of `http.Handler`.  So you can wrap `ServeMux` in the standard way of golang, like https://godoc.org/github.com/goji/httpauth.

This PR or my proposal about interceptor can be still useful in one of the following cases, but none of them is enough to justify the idea of interceptor.
1. Your interceptor needs to depend on `pathParams`; or
2. Your interceptor needs to depend on gRPC method name as your PR does
3. You need to apply different set of interceptors per url pattern.

For 1 and 2, I don't want to make the interceptor aware of `pathParams` nor method names because it is too intrusive -- gateways should focus on HTTP-level and anything in gRPC-level should happen in the actual gRPC services behind the gateways.

For 3, I am not sure if there is such a real use-case. But I guess you don't have such a use case because your PR does not cover the functionality of 3. If so, I'd like to revisit this idea when we see an actual use-case.

gdm85:
@yugui no problem. The use case I was referring to is basically direct consumption of the JSON gateway API via a browser, which in turn needs [CORS support](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing).

The interceptor I need to implement would basically perform this pseudo-Go logic:
```go
func CORSInterceptor(ctx context.Context, w http.ResponseWriter, r *http.Request, next Handler) {
    // note that there is no pathParams, although I was considering passing `methodName` in my original PR to give some context information

    if origin header is not set {
        next(ctx, w, r)
        return
    }

    if method is "OPTIONS" {
         write CORS allowed methods
         return
    }

    if origin and method are allowed {
        write CORS headers
        next(ctx, w, r)
        return
    }

     fail with a 4xx error // browser will not process this response anyway
     return // no call to next() here
}
```

I will update this PR and add an example for the above.

PR #83, opened 2016-01-06
 Support bytes fields in path parameters 

yugui:


yugui:
Just %-encoding is not a safe representation of an arbitrary byte sequence within path due to golang/go#10887.
We need to confirm the expected spec at first.

PR #88, opened 2016-01-08
Add request in error handlers

daniellowtw:
We have a use case where we want to customise the error handler, for example, adding html templates and sending back the error depending on whether the client allows html or not. The library currently allows us to override the `HTTPError` caused by the grpc request but the not errors caused by the server muxer. Furthermore, the incoming client request is not propagated to `HTTPError` so it is not possible to figure out the `Accept` header of the request.

This PR adds these two functionalities: a `OtherErrorHandler` variable which can be overridden like `HTTPError`, and forwarding the incoming request to the error handlers so they can have more flexibility in determining how to handle the errors.


yugui:
LGTM

PR #89, opened 2016-01-16
make grpc.DialOption configurable

kazegusuri:
`grpc.WithInsecure()` is used implicitly as `grpc.DialOptions` currently. When TLS is enabled at gRPC server, we have to use `grpc.WithTransportCredentials(creds)`.
This patch makes `grpc.DialOption` configurable by passing options  via args of `RegisterXXXHandlerFromEndpoint`.

NOTE: this is an interface change of public method.


yugui:
Good. Could you re-generate example files with `make example`?

kazegusuri:
Regenerated examples. PTAL.

yugui:
LGTM

PR #85, opened 2016-01-06
Improve PascalFromSnake behavior

tmc:
Fixes #84 (and doesn't bring in new dependency)

PR #96, opened 2016-01-28
Better compatibility to field names generated by protoc-gen-go

yugui:
Stop emulating the conversion rule.  Use their implementation of the name conversion instead.

c.f. #85


shurcooL:
Nice change!

achew22:
Will this have any affect on the field names in the swagger documentation?

yugui:
> Will this have any affect on the field names in the swagger documentation?

I don't think so because swagger generator does not directly use name conversion, and the currently `gogen.CamelCase` and `utilities.PascalFromSnake` are exactly same.

achew22:
LGTM

PR #98, opened 2016-01-28
Fix the swagger section of the README.md

FabianHinz:
For me to create a *.swagger.json I had to call it with --swagger_out=...

Alternative: change path from protoc-gen-swagger to protoc-gen-grpc-swagger


yugui:
lgtm

peter-edge:
I'd argue to change to grpc-swagger, other projects are working on swagger plugins as well heh :)

achew22:
@peter-edge Really? That's great! Can you point me to some of those projects that are generating swagger from protos?

PR #102, opened 2016-02-02
Cancel context when HTTP connection is closed

### LINKED ISSUE #101: Closing the HTTP connection does not cancel the Context
If you have e.g. a streaming endpoint sending results from the server in an infinite stream and close the HTTP connection, the grpc server doesn't know about that and keeps waiting for the client to accept more data. To prevent that, the gprc gateway should cancel the context when the HTTP connection is closed.

floridoo:
Fixes #101


yugui:
LGTM

floridoo:
Thanks. Great project btw!

yugui:
np. Thank you for your contribution.

Issue #103, opened 2016-02-04
undefined: proto.SizeVarint


coyle:
Getting the following error when attempting to run ```go get github.com/gengo/grpc-gateway/protoc-gen-grpc-gateway```

![image](https://cloud.githubusercontent.com/assets/2541605/12821529/c7e611ca-cb31-11e5-9199-0eb6cf37a6a1.png)

shurcooL:
One of your dependencies is out of date. Using `-u` flag will fix that and allow it to work.

```bash
go get -u github.com/gengo/grpc-gateway/protoc-gen-grpc-gateway
```

coyle:
Thanks @shurcooL !

PR #104, opened 2016-02-05
README: Suggest go get -u by default.

shurcooL:
Suggesting `go get -u` is a safer default. It results in a more predictable action, the user will end up with the latest version of `grpc-gateway` and all its dependencies. Otherwise, the versions users end up with will be arbitrary, which may lead to issues. If users prefer not to update dependencies, they can always `go get` without `-u` flag.

As an example of the issue this would help resolve, see #103.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,11 +33,11 @@ make check
 sudo make install
 ```
 
-Then, `go get` as usual.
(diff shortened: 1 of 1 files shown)


yugui:
LGTM.

Thank you for your suggestion.

PR #105, opened 2016-02-06
README: add protoc-gen-swagger too

philips:
Otherwise the optional swagger generation later in the README will fail.


yugui:
LGTM

PR #106, opened 2016-02-09
Generate shorter description names for swagger output.

achew22:


yugui:
Let me confirm.  Why is the shorter name better than the current name?
I am worrying about if it causes name conflicts because request/response types for a service can come from several packages.

t-yuki:
swagger-ui and swagger-editor uses the name as a schema name in UI
http://petstore.swagger.io/?url=https://raw.githubusercontent.com/gengo/grpc-gateway/master/examples/examplepb/streamless_everything.swagger.json#!/ABitOfEverythingService/List
http://editor.swagger.io/

yugui:
@t-yuki Thank you.  But I don't think it justifies the change if the change causes name conflicts.

tmc:
Maybe it'd be nice if types were prefixed if they aren't defined in the proto file(s) that are being used to render the swagger output.

achew22:
@yugui I was very careful to make sure that it wouldn't cause name conflicts. I believe that the test cases demonstrate that. The thing I dislike about this patch is that you were iterating on your API you might create a new Foo proto when you have a preexisting one. That would change the name of both protos to be unique but break your preexisting code.

I figured I would write this up and just see what people thought about it. Doesn't have to go in

achew22:
I switched to doing it your way Yugui and I'm much happier. Thanks!

PR #108, opened 2016-02-10
Add parser and swagger support for enum, no gengateway yet

t-yuki:
This change adds enum feature to parser and gen-swagger.
But no gen-grpc-gateway/gengateway yet.

One of problems is enum json encoding.
proto3 spec says enum should be encoded using string name.
However, since golang/protobuf generated codes have no MarshalJSON/UnmarshalJSON methods, it is encoded as a int, even if proto.Unmarshal/MarshalJSONEnum methods exist.
So json.Unmarshal on request will cause error if string name value is provided.

Note that Swagger support in this change uses string name as spec.

Refs: https://developers.google.com/protocol-buffers/docs/proto3#json


yugui:
Almost LGTM.

* Could you change `examples/examplepb/streamless_everything.proto` as well so that you change is applied to the example of Swagger JSON?
* Could you re-generate `.pb.go` files with the latest `protoc-gen-go` because the test is failing due to the difference?

t-yuki:
Oops, I didn't `make realclean` after update.
It's done.

t-yuki:
Currently json Marshal/Unmarshal of enum string for gengateway requires jsonpb.
Actually proto2 files may work with enum string because protoc-gen-go generates if it is not proto3 (why??): https://github.com/golang/protobuf/blob/master/protoc-gen-go/generator/generator.go#L1407

yugui:
LGTM

PR #109, opened 2016-02-24
fixes missing swagger operation objects

t-yuki:
It fixes missing operation objects when multiple methods defined on the same path.


yugui:
LGTM

PR #130, opened 2016-04-09
Pluggable JSON interface on ServeMux

willtrking:
Attempts to address https://github.com/gengo/grpc-gateway/issues/79  by providing a way to configure what JSON library will be used on the `ServeMux`.

Introduces the `JSONAdapter` interface to solve this, and adds the `JSONHandler` property to `ServeMux`. Handler is used for input and output, although it would be trivial to separate them.

The grpc-gateway generator has been modified so that the generated functions use whatever `JSONHandler` is attached to the `ServeMux` used to register the handlers. Also slight modifications to runtime/handler in order to support those changes. 

Runtime includes 2 handlers by default: `JSONBuiltin` just wraps `encoding/json` and `JSONPb` wraps `jsonpb`. When `NewServeMux` is called the default handler is set to `JSONBuiltin`. `JSONPb` allows the standard `jsonpb.Marshaler` options.

New handlers can be introduced by following the `JSONAdapter` interface.

This implementation is naive as different handlers can't choose different JSON interfaces, they all have to use whatever is supplied to `ServeMux`

Haven't really touched any of the tests or the swagger generator (never used swagger personally), but if people are interested in this PR I'm happy to do so.


yugui:
@willtrking, thank you for your job.

I like the overall direction. I have several comments.
* I was thinking of implementing this feature as a generic abstraction of marshaling mechanism.
  It doesn't need to be extensible for other MIME types for now. But what do you think about naming `JSONAdapter` -> `Marshaler` or `Codec`?
* Unfortunately there is an edge case. Even when you use jsonpb, you still need to marshal/unmarshal primitive types or JSON arrays when `google.api.HttpRule.body` is not `*`. You would need to combine it with `encoding/json` in this case.

willtrking:
@yugui 

Name change makes a lot of sense to me, I think I'll probably go with `Marshaler`. How are you thinking of doing the MIME type abstraction? Based on the `Content-Type` header from the client? If so it seems like that would be pretty easy for me to take care of. I was considering some sort of system to determine if the protocol buffers being used required `jsonpb` as well, but this would of course take longer. It would also hide potential performance impacts of using `jsonpb`. 

In what cases can the `google.api.HttpRule.body` not be a protocol buffers? Looking through the examples it looked liked the `body` sections of the options still deserialized into a protocol buffer in the generated code, although I'm sure I'm missing something given how new I am to this project. Any particular source files I should look through for this?

yugui:
> Name change makes a lot of sense to me, I think I'll probably go with Marshaler.
Good to hear.

> How are you thinking of doing the MIME type abstraction? Based on the Content-Type header from the client? If so it seems like that would be pretty easy for me to take care of.

It'd be better if you kindly do it. But we can do it later.

> I was considering some sort of system to determine if the protocol buffers being used required jsonpb as well, but this would of course take longer. It would also hide potential performance impacts of using jsonpb.

Yeah, you are right.  So I prefer to leaving it with the users.  They can use `encoding/json` version if they need more performance and their protobuf is simple enough.  And we use `jsonpb` unless they explicitly configure.

> In what cases can the google.api.HttpRule.body not be a protocol buffers? Looking through the examples it looked liked the body sections of the options still deserialized into a protocol buffer in the generated code, although I'm sure I'm missing something given how new I am to this project. Any particular source files I should look through for this?

At [examples/examplepb/a_bit_of_everything.proto#L97](https://github.com/gengo/grpc-gateway/blob/d456ebcbd881b0c9af55f608323e764ef79cd3bd/examples/examplepb/a_bit_of_everything.proto#L97), the request body points a primitive leaf field [`gengo.grpc.gateway.examples.sub.StringMessage.value`](https://github.com/gengo/grpc-gateway/blob/d456ebcbd881b0c9af55f608323e764ef79cd3bd/examples/sub/message.proto#L6), whose type is `string`.

So the marshaler must be able to unmarshal a JSON string in the generated code [a_bit_of_everything.pb.gw.go#L354](https://github.com/gengo/grpc-gateway/blob/d456ebcbd881b0c9af55f608323e764ef79cd3bd/examples/examplepb/a_bit_of_everything.pb.gw.go#L354).

tmc:
It'd be great if the swagger generation could easily use this custom Marshaler.

willtrking:
Basic changes:

- Name change to `Marshaler`
- `ServeMux` now defines an `InboundMarshaler` and an `OutboundMarshaler`. This allows you to use different marshallers for inbound requests from the client, and outbound responses to the client.
- Each `Marshaler` also now defines a `ContentType` function which is used when responding to the client. This would (theoretically, untested) provide support for XML, and other responses besides JSON.
- The `DefaultError` function now uses a `Marshaler` when sending an error back.
- The default `Marshaler` for the `ServeMux` is a wrapper around `encoding/json`. If we switched to `jsonpb` this could break existing codebases.

MIME registry is in here as well, details as follows:

-  A registry can be created with `NewMarshalerMIMERegistry` and you can add inbound/outbound marshallers for different (case-sensitive) mime types. 
- `Content-Type` request header is used to determine marshallers.
- The `*` MIME type is a catch all for any inbound `Content-Type` headers. Specifically set MIME types take precendence over `*`. So if you have both `application/json` and `*` in the registry, and the client sends `application/json` the marshallers for `application/json` will be used. 
- If nothing is set in the registry, or the specified `Content-Type` is not found, and `*` is not set, it defaults to the ServeMux's `InboundMarshaler` and `OutboundMarshaler`. If those are nil, it has compile time defaults as well. Take a look at the `runtime/marshaler.go` source for more details.
- Code will `panic` if you attempt to setup a MIME type in the registry with an empty string.
- The registry is attached to the `ServeMux` with the `MIMERegistry` attribute. 


@yugui 

In terms of the edge case when `google.api.HttpRule.body` is not `*`, I have not touched this yet. Considering type assertions as a good go between, or perhaps generate new types that support standard `encoding/json` marshaling, marshal into those types, and then load the data from the marshaled type into the proper protocol buffer field.

@tmc 

Would love to help with this, not quite sure how I would go about it though. As it stands right now the swagger generation would have to have a reference to the `ServeMux` used by the server, as all the `Marshaler` interfaces are attached to that. I've never used swagger, so I'm not too sure what the output/usage pattern looks like here.

tmc:
@willtrking please don't break the convention of the context.Context parameter being the first argument to context-aware calls.

kazegusuri:
What's the remaining work? I want this feature, any help?

kellrott:
I think this would be a nice feature to have. It looks like the Travis testing errors are because of a dependency error: `The command "go get github.com/gengo/grpc-gateway/examples" failed and exited with 2`

tmc:
@kazegusuri you could move the context.Context parameter to consistently be the first argument

yugui:
Merged in #144

PR #133, opened 2016-04-16
Swagger definitions now have `type` set to `object`.

ivucica:
While the spec does not seem to require this, the Swagger editor at
http://editor.swagger.io/ gets confused and does not display the
'Try this operation' UI correctly.

Many of the examples found around also specify `"type": "object"`,
so this *should* be the right thing to do.


achew22:
Ivan,

Thanks for contributing to grpc-gateway! Your patch looks interesting. I think that in a second the build is going to fail because you need to regenerate the examples. Could you go ahead and do that to satisfy the build monster? After that happens an example of this will be available on github on a public URL and we can create a link to test it out.

Looking forward to helping get this merged!

ivucica:
(Un)amusingly, when I fire up

    make realclean
    make examples

`protoc-gen-go` seems to be dumping out protos without `protobuf:"....:,json=protojsonfieldnamehere"`. That should be the old behavior -- so I'm not sure why that's going on.

However, after a bit of fighting with the gopath set up just for this purpose :-) the golden JSONs have been rebaked and they seem similar to the output from the failed Travis job. There seems to be no other differences related to the aforementioned `protoc-gen-go` issue. Update coming up.

ivucica:
Alright, I'm not sure how to tackle this. Travis build on Go 1.5 passes, but fails on tip, with a diff on gzipped `FileDescriptorProto`s.

I'm almost certain that trying to push my local copy won't help much, given the aforementioned even-larger diff (completely lacking the `json` field in the `protobuf` annotations of generated Go structs).

Thoughts?

achew22:
My understanding of that test is that if you git add those files and upload it that it should start working. Would you mind trying that really quick since it shouldn't be too painful to try?

ivucica:
This is the same symptom witnessed by @kazegusuri in pull request #129. I'd say this is a broken test?

ivucica:
Pushed. I'll be going to sleep now; however I would expect that both 1.5 and tip will be broken with b31478d.

achew22:
@yugui I have been playing with this for a couple of hours and I'm unable to get it to compile the same on tip vs 1.5.

Would you be willing to take a look at that test and see if there is a better way to do it? Can we maybe whitelist the byte arrays or something? I'm not sure how to fix it

yugui:
@achew22, @ivucica 
Thank you for your contributions. I've fixed the build issue in ivucica/grpc-gateway#1. Could you take a look?

achew22:
@yugui Would you be willing to propose that as a PR to the grpc-gateway repo? That way travis will fire on it and we can get it committed

yugui:
@achew22 

I made a PR #136.  It contains only Travis configuration change.
You still need to update .pb.go with protoc-gen-go after golang/protobuf@001690d39bd620847bb265d93a7c5e1bd3737308.

Do you want me to create another PR which contains the .pb.go changes?

achew22:
I don't think it is worthwhile to do the .pb.go changes since everyone will have do them anyway. Looks like it passes on both 1.5 and tip. I LGTM'd your PR over there. @ivucica this should clear the path for you. Thanks for pushing back on me. @yugui thanks for your timely response!

ivucica:
Thanks, all, for your work on this! I've updated this pull request by merging master. Looks like it passes now.

achew22:
LGTM 👍

yugui:
Thank you for your contributions.

PR #134, opened 2016-04-17
Generate Swagger description for service methods using proto comments.

### LINKED ISSUE #128: protoc-gen-swagger comment parsing for documentation gen
Is there, by chance, any documentation on how to properly format comments so that the swagger fields in the JSON will be nicely populated for documentation purposes?

ivucica:
This PR adds support for extracting Swagger `description`s from protobuf comments.

This fixes #128.

---

**ORIGINAL OUTDATED TEXT FOLLOWS** 

While this is a first step in resolving gengo/grpc-gateway#128, this
needs to be cleaned up, and the same approach needs to be used for
messages, message fields, et al.

`echo_service.proto` has been annotated with extra comments in order to
demo the new descriptions.

Only the Swagger example has been regenerated, as my local generator
does not output all the expected fields in proto struct tags.

---

Aside from some uncleanliness and overzealous `panic()`ing, I'm also (obviously) unhappy with line 444. Other generators seem to be producing and comparing paths by joining them into comma-separated strings. They're also putting all the `SourceCodeInfo_Location`s into a `map` indexed by paths. 

(However, one thing that line 444 is doing good is dynamically looking up path IDs for 'message' type, etc. Even though it's using reflection for it, other generators written in Go seem to like to copy these constants into their own source code, which feels even worse.)

(For interested readers: More info on how to assemble a `SourceCodeInfo_Location` is [in google/protobuf's descriptor.proto](https://github.com/google/protobuf/blob/3b3c8ab/src/google/protobuf/descriptor.proto#L700).)

---

Similar to #133, I was unable to regenerate protobufs usefully. I fully expect Travis build to fail at this point.

### DIFF
--- a/examples/examplepb/a_bit_of_everything.pb.go
+++ b/examples/examplepb/a_bit_of_everything.pb.go
@@ -132,7 +132,7 @@ var _ grpc.ClientConn
 
 // This is a compile-time assertion to ensure that this generated file
 // is compatible with the grpc package it is being compiled against.
-const _ = grpc.SupportPackageIsVersion1
+const _ = grpc.SupportPackageIsVersion2
 
 // Client API for ABitOfEverythingService service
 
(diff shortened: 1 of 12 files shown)


ivucica:
Please let me know if something else is needed for this PR.

tmc:
@ivucica IMO you should separate the comment handling from this new `<!-- swagger extras start` concept as they're distinct and the latter deserves more consideration. Do you mind pulling that concept out into another PR? Thanks!

tmc:
What do you think about simply taking the last portion of the comment and if it's a json document integrate it.

ivucica:
@tmc That sounds like a great idea. I'll work on that, including separating it into a separate PR.

ivucica:
@tmc See 6d157cc.

yugui:
LGTM.

I can merge this PR as soon as I or you resolve merge conflicts.

tmc:
@yugui see #156

PR #136, opened 2016-04-18
Conditionally stops checking if generated file are up-to-date

yugui:
Golang changed the way of gzip compression between go1.6.1 and 1.6.1+.
So the gzipped proto.FileDescriptorProto always have diff.
Thus we cannot assume that generated files have the same contents as the
ones regenerated in Travis CI.

c.f. https://github.com/gengo/grpc-gateway/pull/133#issuecomment-211221715


achew22:
LGTM 👍  Thanks for the quick fix

PR #142, opened 2016-05-05
Link to LICENSE.txt

sun-kuet:
Link to LICENSE.txt added. I think this decrease the searching time of developers.


ivucica:
Wouldn't a relative link that doesn't encode full repo path make more sense?

PR #143, opened 2016-05-07
Allow period in path URL templates when generating Swagger templates.

ivucica:
This allows for the templates to contain references to request proto
messages' fields.

Some test code that I played with seems to indicate that
gRPC-Gateway does support dots to reach values
inside values.

In my interpretation, RFC 6570 does not forbid this, but
does not seem to explain exactly what happens.

Swagger spec also does not seem to forbid it. (It's very
nonspecific about how templates are to be parsed,
generally.)

Example proto:

```protobuf
syntax = "proto3";

package toy;

service Toy3 {
  rpc Build (Toy3BuildRequest) returns (Toy3BuildReply) {
    option (google.api.http) = {
      post: "/toy3/v1/{what.shape}/{what.color}/{what.class}"
      body: "where"
    };
  }
}

message Toy3BuildRequest {
  Toy3BuildWhat what = 1;
  string where = 2;
}

message Toy3BuildWhat {
  string shape = 1;
  string color = 2;
  string class = 3;
}

message Toy3BuildReply {
  string construction_id = 1;
}
```

Broken output snippet:

```json
    "/toy3/v1/{what}/{what}/{what}": {
      "post": {
        "summary": "Toy3.Build",
        "operationId": "Build",
        "responses": {
          "default": {
            "description": "Description",
            "schema": {
              "$ref": "#/definitions/toyToy3BuildReply"
            }
          }
        },
        "parameters": [
          {
            "name": "what.shape",
            "in": "path",
            "required": true,
            "type": "string",
            "format": "string"
          },
          {
            "name": "what.color",
            "in": "path",
            "required": true,
            "type": "string",
            "format": "string"
          },
          {
            "name": "what.class",
            "in": "path",
            "required": true,
            "type": "string",
            "format": "string"
          },
          {
            "name": "body",
            "in": "body",
            "required": true,
            "schema": {
              "$ref": "#/definitions/toyToy3BuildRequest"
            }
          }
        ],
        "tags": [
          "Toy3"
        ]
      }
    }
```

Incidentally, this pull request would also update the examples to {be compatible with,use} a newer Go proto package.


yugui:
LGTM.

Issue #79, opened 2015-12-29
allow use of jsonpb for marshaling


ericchiang:
[github.com/golang/protobuf/jsonpb](https://godoc.org/github.com/golang/protobuf/jsonpb) is a more compliant protobuf -> JSON package than `encoding/json`, and is the only way github.com/golang/protobuf plans to support JSON encoding in the future.

See this [comment](https://github.com/golang/protobuf/issues/44#issuecomment-117385573)

> The newly added jsonpb package (see 67cbcad) is the official vehicle for JSON support...

It would be great to be able to using this package for marshaling. It might even be wise to switch to it entirely.

peter-edge:
Using it would be nice, but note that in my benchmarks, jsonpb is
approximately 6x slower at marshaling then the standard library package

ericchiang:
A lot of the overhead comes from checking to ensure the result is actually compliant with protobuf's [defined JSON mapping](https://developers.google.com/protocol-buffers/docs/proto3#json). It's slower, but it's correct :)

peter-edge:
Ya, I know :) Ugh. Programming is no fun :)

philips:
So, someone needs to build a codegen ala ffjson

yugui:
@peter-edge Thank you for your information.  I didn't know that it is that slower.
Then, it sounds better to have an option to keep using `encoding/json`, although `jsonpb` is mandatory for users who wants `Any`, `Timestamp` types, `oneof` fields and others.

I am thinking of a mechanism which allows users to register/overwrite marshaler per MIME type.
With this mechanism, `jsonpb` would be a good default marshaler for application/json but you could keep using `encoding/json` if you don't need new proto3 types of fields.

tmc:
@yugui can you sketch out what sort of approach you were thinking? I may hack on this soon.

hbchai:
The `jsonpb.Marshaler` [recently changed its default behavior](https://github.com/golang/protobuf/commit/001690d39bd620847bb265d93a7c5e1bd3737308) to output camelCase, so not using jsonpb within grpc-gateway is now a compatibility issue. jsonpb users will have to set `OrigName: true` when creating the `Marshaler` to work around this, which took me a while to figure out and is less than ideal. I would like to see support land soon.

peter-edge:
@yugui @tmc @hbchai don't want to repeat work, but I'd propose what looks like (based on an overly-simple grep) a very small PR:

For code generation:

[...]

Issue #124, opened 2016-03-04
support protobuf-over-HTTP


tamird:
It'd be cool to provide encoded protobuf support in addition to JSON from the endpoints generated by grpc-gateway. This can be done by switching on the incoming request's content type as is done in `MarshalResponse` and `UnmarshalRequest` in https://github.com/cockroachdb/cockroach/blob/master/util/http.go

philips:
@tamird that is a cool use case. But, I think you missed a line number anchor in your example.

tamird:
Didn't intend to line anchor - look for the functions I mentioned by name.

tamird:
This is sufficiently supported now. See https://github.com/cockroachdb/cockroach/pull/6754 for an example.

PR #144, opened 2016-05-07
Implement custom Marshaler support, add jsonpb implemention.

tmc:
This extends upon the work in #130 but keeps net/context conventions in place and updates tests to reflect all the changes.

Addresses #79


yugui:
Thank you, @willtrking, @tmc .
It looks good to me as an initial implementation.  I'll add some small fixes later.

tamird:
This might also enable fixing #124

willtrking:
@tamird Yes it does allow that, we did that for a bit in development but never in production.

PR #145, opened 2016-05-07
add custom options to allow more control of swagger/openapi output

ivucica:
Continuation of #134, pending review of that.


tmc:
I'm personally welcoming of the switch of the naming around swagger to open api spec. I don't have a better label/tag offhand but putting "Swagger: " into this api means a slower transition to OpenAPISpec.

ivucica:
While OpenAPI is more accurate, Swagger is a very 'googleable' name -- it's a shame it's being dropped. Still, I've switched to the `OpenAPI: ` prefix. This looks a bit off, given that absolutely everything else in this project refers to 'swagger'; nonetheless, you're right, if it'll live in .protos, it deserves a more long-term name.

I also think using JSON here in the first place is a short-term cop-out fix (i.e. better than nothing), even though the resulting JSON looks just awful everywhere else. Long term, something that's still hierarchical such as:

    OpenAPI-Info-Version: 2.0
    OpenAPI-ExternalDocs-Name: Find out more at
    OpenAPI-ExternalDocs-URL: https://www.example.net/

would be more readable, while not much harder to parse. Perhaps even with Debian-style multilines:

    OpenAPI-Description: Hello world.
     This is a text over multiple lines,
     with a blank line, too.
     .
     This is a new paragraph.

That wouldn't look too horrible when seen by tooling of other languages in the generated code, either.

That's a topic for some other PR, however.

tmc:
I agree that it feels off in context. When considering longer term I do think your change makes more sense.

yugui:
I wonder if we can use custom options to annotate patches.
What do you think?

tmc:
@yugui I like that idea -- could you propose how that might look?

ivucica:
@yugui It's interesting.  I guess we need to describe the Swagger schema as a .proto, in which case we might be able to do away with (what will then be) the duplicate definitions in `types.go`.

I probably won't have a chance to work on this any time soon, though.

yugui:
@tmc, @ivucica 
In my personal opinion, it is not too bad even if we only have a single string custom option which stores schema patch in JSON because at least we could avoid parsing comments.

I am worrying about parsing comments because it is not very common in protobuf convention and it has less power than custom options.
With custom option, we could deprecate the simple string option and introduce a well-organized messages to describe schema in future.

tmc:
I like this as a middle ground. @ivucica thoughts?

ivucica:
It'll be way uglier, but we can give it a go.

ivucica:
Rebased and pushed. No other changes, i.e. still not based around proto options.

ivucica:
Rebased and pushed once again. From: 4ea3bbf 3525ecb e3b11c1

ivucica:
@yugui @tmc  Could you PTAL and let me know if the direction I'm taking is worth pursuing? I can also switch to just a single option.

For now I've kept the parsing of the comment, too.

I have not tried to properly update all the tests, or squashing the commits, given we might change the direction.

Random thought: There is an interesting side effect if there is an option to put JSON in the comment: generated code would include this metadata in its docstrings. It might not be a bad thing, either?

tmc:
@ivucica I think the custom option to supply additional openapi to inject is the way to go here.

ivucica:
@tmc Yes, but which variant? I'm currently creating a larger proto descriptor instead of just accepting a piece of JSON. See PR for details :)

tmc:
@ivucica sorry I missed this comment. -1 to overloading comments, +1 to rich options.

```
(option (grpc_gateway.protoc_gen_swagger.options.swagger_swagger) = {
externalDocs: {}
```
over
```
//   "externalDocs": {)
```

Question -- can we avoid the options.swagger_swagger stutter?

ivucica:
What I meant to ask is, should the proto option accept a JSON string, or as I started, describe the Swagger doc schema in the proto language :)

I'll see what I can do to avoid the "stutter". :)

tmc:
Unless it presents a serious technical challenge I think the richer option is better.

I think it might be worth revisiting the ```s/swagger/openapi/g``` issue -- do you have thoughts there?

eddiezane:
Any update on this?

Looks like Swagger has [officially been renamed](https://buildazure.com/2017/10/20/swagger-is-now-the-openapi-specification/) to the OpenAPI Specification.

I'm looking to open a PR to add some additional options that will allow for filtering out private/internal fields as well as marking things as [Output only](https://cloud.google.com/apis/design/design_patterns#output_fields).

Looking to this PR as guidance.

ivucica:
@eddiezane This surprises me, but apparently, technically:

> As a result, the last version of the Swagger Specification is v2.0, and the first OpenAPI Specification is version 3.0.

(Taken from the post you linked to)

On multiple occasions, I was trying to figure out what to reply in relation to the rename, and my general observation is that a lot of the v2 spec still uses "Swagger" terminology (see ["Swagger Object"][1], for example). Also, proto generator that this PR applies to is called `protoc-gen-swagger`. So I'll do the rename where reasonable! :)

I can try to tackle this this weekend. If you see updates here by, say, mid-next week, you can take my work and run with it. 

(On a separate note, I think OpenAPI v3 generator should be a separate generator. A quick glance at the spec about a week ago made me think it's distinct enough that it deserves a `protoc-gen-openapi` or so. That's a totally separate issue.)

[1]: https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#swagger-object

ivucica:
I have eliminated JSON processing and added support for loads more fields. I settled on changing reference to specification from 'Swagger' to 'OpenAPI v2'. This eliminated a class of confusion such as "swagger_swagger".

It also eliminates pieces of JSON from the Go comments that included OpenAPI-specific data, such as external docs URLs. If those are useful, we can bring them back in a future PR.

Remaining before this PR can be merged:

- First... someone, please **confirm whether it seems like this is on the right track**. I won't mind a round of code review :) @tmc? @yugui?

- I can't run the failing tests in `examples/browser` locally at this time. I should be able to do it over the weekend. I won't mind help with them, either, if there's anything obvious I messed up.
- I want to move the demonstration from `echo_service.proto` to `a_bit_of_everything.proto`.
- I should totally email `protobuf-global-extension-registry@google.com` and get us four extension IDs for attaching metadata to files, methods, messages and services (patching swagger object, operation object, schema object in definitions object, and tag object, respectively). We should not risk future binary incompatibility in on-disk protos which may merely import `annotations.proto`.
- I will squash most / all of the commits.

What can be punted to a future PR:

- Process remaining OpenAPI v2 properties that have already been added to the `openapiv2.proto`.
- Support remaining OpenAPI v2 objects ('security', 'security definitions', etc)
- Document individual fields in openapiv2. (Or remove the TODOs, as the `openapiv2.proto` option definitions generally won't be visible/autocompletable in any sort of smart code editor. Hence documenting fields might be an overkill, and referencing OpenAPI is enough.)
- Use existing 'deprecated' option on a method, not (just) OpenAPI field.

ivucica:
Looks like overriding "host" is a no-no, and breaks tests. :)

achew22:
> First... someone, please confirm whether it seems like this is on the right track. I won't mind a round of code review :) @tmc? @yugui?

I know I'm not on your list, but I'm happy to give you a review. I think you're on the right track. I would venture to say you're probably 1 or 2 back/forths on review before we can merge this.

> I want to move the demonstration from echo_service.proto to a_bit_of_everything.proto.

Do it. I think that's a better place for this.

> I should totally email protobuf-global-extension-registry@google.com and get us four extension IDs for attaching metadata to files, methods, messages and services (patching swagger object, operation object, schema object in definitions object, and tag object, respectively). We should not risk future binary incompatibility in on-disk protos which may merely import annotations.proto.

+1, emailing them is a great idea.

> I will squash most / all of the commits.

You're welcome to do that but the default when we merge is to squash rebase so if you wanna ignore that then you're fine.

>What can be punted to a future PR:
> Process remaining OpenAPI v2 properties that have already been added to the openapiv2.proto.

I would prefer if you omitted them from the openapiv2.proto file if there is no support for them.

> Support remaining OpenAPI v2 objects ('security', 'security definitions', etc)

These would be nice but the perfect shouldn't be the enemy of the good (which this clearly is very good).

> Document individual fields in openapiv2. (Or remove the TODOs, as the openapiv2.proto option definitions generally won't be visible/autocompletable in any sort of smart code editor. Hence documenting fields might be an overkill, and referencing OpenAPI is enough.)

Yep, that's a good idea. I think giving links to the OpenAPI definitions of these things (or copy/paste into the proto) would be good.

> Use existing 'deprecated' option on a method, not (just) OpenAPI field.

That's also a good idea but it can wait.

protoc-gen-swagger/genswagger/template.go
```diff
@@ -10,10 +10,14 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/golang/protobuf/proto"
 	pbdescriptor "github.com/golang/protobuf/protoc-gen-go/descriptor"
 	"github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway/descriptor"
+	swagger_options "github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger/options"
 )
 
+var swaggerExtrasRegexp = regexp.MustCompile(`(?s)^(.*[^\s])[\s]*<!-- swagger extras start(.*)swagger extras end -->[\s]*(.*)$`)
```
achew22: swaggerExtrasRegexp... what's that? Could you add  a comment here explaining it's purpose? It doesn't seem to get used anywhere else

protoc-gen-swagger/genswagger/template.go
```diff
@@ -175,6 +179,25 @@ func renderMessagesAsDefinition(messages messageMap, d swaggerDefinitionsObject,
 		if err := updateSwaggerDataFromComments(&schema, msgComments); err != nil {
 			panic(err)
 		}
+		opts, err := extractSchemaOptionFromMessageDescriptor(msg.DescriptorProto)
+		if opts != nil {
+			if err != nil {
```
achew22: Can you promote error checking to immediately after the call to extractSchemaOptionFromMessageDescription?

protoc-gen-swagger/genswagger/template.go
```diff
@@ -858,3 +973,63 @@ func protoPathIndex(descriptorType reflect.Type, what string) int32 {
...
+	if !proto.HasExtension(msg.Options, swagger_options.E_Openapiv2Schema) {
+		return nil, nil
+	}
+	ext, err := proto.GetExtension(msg.Options, swagger_options.E_Openapiv2Schema)
+	if err != nil {
+		return nil, err
+	}
+	opts, ok := ext.(*swagger_options.Schema)
+	if !ok {
+		return nil, fmt.Errorf("extension is %T; want an Schema", ext)
```
achew22: Nit: "want a Schema"

protoc-gen-swagger/options/annotations.proto
```diff
@@ -0,0 +1,25 @@
...
+
+package grpc_gateway.protoc_gen_swagger.options;
+
+option go_package = "github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger/options";
+
+import "protoc-gen-swagger/options/openapiv2.proto";
+import "google/protobuf/descriptor.proto";
+
+extend google.protobuf.FileOptions {
+  // TODO(ivucica): ask protobuf-global-extension-registry@google.com to assign ID
```
achew22: Let's do this soon so we can merge

protoc-gen-swagger/options/openapiv2.proto
```diff
@@ -0,0 +1,223 @@
+syntax = "proto3";
```
achew22: Nice touch on reserving all the numbers for the order they appear in the swagger spec!

protoc-gen-swagger/genswagger/template.go
```diff
@@ -619,6 +669,73 @@ func applyTemplate(p param) (string, error) {
...
+				}
+			}
+		}
+		if spb.Host != "" {
+			s.Host = spb.Host
+		}
+		if spb.BasePath != "" {
+			s.BasePath = spb.BasePath
+		}
+		if len(spb.Schemes) > 0 {
```
achew22: What is the correct default here if schemes is empty? Right now I don't think it emit anything but is that the correct behavior? Should it emit "http" only so at least there is a value?

achew22:
LGTM. Let's get the extension numbers reserved and merge it.

Thanks so much for doing this! Awesome work

ivucica:
Email has been sent. Pending reply.

tmc:
Super happy to see this progress!

achew22:
@ivucica got the email issuing us the extension IDs. When you're ready please add the new IDs and we can merge this as soon as that happens.

ivucica:
New IDs have been assigned. However let's block this on move to `a_bit_of_everything.proto`. I'll try to do this soon.

examples/examplepb/a_bit_of_everything.proto
```diff
@@ -8,10 +8,43 @@ import "google/protobuf/duration.proto";
 import "examples/sub/message.proto";
 import "examples/sub2/message.proto";
 import "google/protobuf/timestamp.proto";
+import "protoc-gen-swagger/options/annotations.proto";
+
+option (grpc_gateway.protoc_gen_swagger.options.openapiv2_swagger) = {
```
johanbrandhorst: Don't you need to import the relevant proto file to access options?

johanbrandhorst: Ah, that's annotations.proto, I see. I was confused by the package here, importing from `protoc-gen-swagger` and then accessing a `grpc_gateway` namespace. Might just be me but feels like a disconnect?

protoc-gen-swagger/options/annotations.proto
```diff
@@ -8,18 +8,18 @@ import "protoc-gen-swagger/options/openapiv2.proto";
 import "google/protobuf/descriptor.proto";
 
 extend google.protobuf.FileOptions {
-  // TODO(ivucica): ask protobuf-global-extension-registry@google.com to assign ID
-  Swagger openapiv2_swagger = 123456788;
+  // ID assigned by protobuf-global-extension-registry@google.com for grpc-gateway project.
```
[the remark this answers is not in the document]
ivucica: Done.

examples/examplepb/a_bit_of_everything.proto
johanbrandhorst: Sounds good to me :)

zxy198717:
@ivucica Thanks for your works!
 
Have you been working on `Support remaining OpenAPI v2 objects ('security', 'security definitions', etc)`

ivucica:
No, I haven't. I wanted to get this merged in case someone else wants to continue, but otherwise I have not yet needed things such as oauth2 definitions in the proto.

FWIW, I'll take the opportunity to speculate. I don't think I'll be working on the following, but it's an interesting thought, I think:

I can imagine a solution integrating security definitions inside the grpc-gateway-compatible and swagger-annotated .proto file with an oauth2 identity provider.

There may also be a complementary approach to this PR, #145 (which is just overriding Swagger fields) which is: use more definition files to describe the API, rather than just the .proto itself. That is: integrate with and read things such as:

* https://github.com/googleapis/googleapis/blob/master/google/api/service.proto
* https://github.com/googleapis/googleapis/blob/master/google/api/endpoint.proto
* https://github.com/googleapis/googleapis/blob/master/google/api/auth.proto
* https://github.com/googleapis/googleapis/blob/master/google/api/usage.proto

The actual gateway would thus be complemented by an extra lib to support the above and by an oauth2 provider. That is, grpc-gateway's runtime+generator wouldn't be the only thing covered by protoc-gen-swagger, which would use this input to generate a more complete swagger description, too.

achew22:
I'm also in favor of the techniques described by @ivucica. Since grpc-gateway isn't going to be generating the entirety of your http server for you it is unlikely that we will be able to handle all cases. Oauth, to pick an example, is not going to be included in grpc-gateway, it is too complex and there are too many specific requirements for it to make a generalized form of it. Additionally, I know that people are already doing oauth based security on services by using interceptors that live outside the grpc-gateway codebase. Since it isn't a hard requirement that it be in grpc-gateway I think it should be in another repo. We have this today with @tmc's [grpc-websocket-proxy](https://github.com/tmc/grpc-websocket-proxy). It is a simple 1 liner function call at handler registration and it makes things easy to handle.

Just to demarcate the lines in the sand that I will be using to determine if functionality should be included in grpc-gateway proper.

1.  Is the functionality a part of the grpc to http spec? Internal
1.  Is the functionality possible to implement as an interceptor? External
1.  Is the functionality possible to implement as an interceptor with some kind of modification to the generated code (e.g. passing request context through the http handlers into grpc calls #265)? Internalize the functionality necessary to allow a hook to perform the action but externalize the functionality itself.

Does anyone have any opinion on additional tests that should be included in order to prevent goalpost moving? I'm happy to codify these in the `README` file if people think that's of value.

PR #152, opened 2016-05-09
custom marshaler: handle `Accept` headers correctly

tamird:
Also greatly simplifies existing logic to avoid special casing of
defaults.

### DIFF
--- a/runtime/marshaler_registry.go
+++ b/runtime/marshaler_registry.go
@@ -5,12 +5,15 @@ import (
 	"net/http"
 )
 
-const mimeWildcard = "*"
+// MIMEWildCard is the fallback MIME type used for requests which do not match
+// a registered MIME type.
+const MIMEWildcard = "*"
 
 var (
-	defaultMarshaler = &JSONPb{OrigName: true}
-
+	acceptHeader      = http.CanonicalHeaderKey("Accept")
 	contentTypeHeader = http.CanonicalHeaderKey("Content-Type")
+
+	defaultMarshaler = &JSONPb{OrigName: true}
 )
 
 // MarshalerForRequest returns the inbound/outbound marshalers for this request.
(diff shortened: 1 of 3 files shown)


tamird:
@yugui @willtrking @tmc

tamird:
This failed on `go tip` because `go vet` crashed, but is otherwise green.

tmc:
why the MakeMarshalerMIMERegistry naming change?

tamird:
Because it no longer returns a pointer.

yugui:
I have another similar cleanup and bug fixes in #151.
I'll merge this PR into #151.

yugui:
It is unlikely that I will merge this PR because
* I have a similar cleanup #151.
* I don't think this is a "proper" handling of Accept header since it ignores media range and quality factor of the header.  https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.1
* Accept header is not the right way to identify MIME type of request body.

It would be even better if we can property handle Accept header, however, I don't think it is mandatory.
* c.f. http://programmers.stackexchange.com/a/170409

If we handle the header, it would require:
* interpret media range and quality factor
* split `Marshaler` into `Marshaler` and `Unmarshaler`.
  * Independent registry for each

tamird:
Rebased.

> I don't think this is a "proper" handling of Accept header since it ignores media range and quality factor of the header. https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.1

Yeah, it's probably incomplete, but still better than the current handling which conflates "content-type" (which specifies the incoming content) with "accept" (which specifies the outgoing content).

> Accept header is not the right way to identify MIME type of request body.

This is not what this PR does. It uses "content-type" to identify the MIME type of the incoming request, but uses "accept" to determine the MIME type that should be used to encode the outgoing response.

> If we handle the header, it would require:

> - interpret media range and quality factor
>   - split Marshaler into Marshaler and Unmarshaler.
>       - Independent registry for each

Yes, we could do these things, but it doesn't seem "required" to me. I'm not sure splitting Marshaler and Unmarshaler is even worth it, but I'll leave that decision to you. Nevertheless, I don't think that refactoring should block this change.

runtime/marshaler_registry.go
```diff
@@ -5,132 +5,87 @@ import (
...
+			break
 		}
-		if inbound != nil && outbound != nil {
-			// Got them both, return
-			return inbound, outbound
+	}
+
+	for _, contentTypeVal := range r.Header[contentTypeHeader] {
+		if m, ok := mux.marshalers.mimeMap[contentTypeVal]; ok {
+			outbound = m
```
yugui: maybe you mean `inbound = m`?

yugui:
@tamird 
Thank you.  I confused something in the original commits.

I'll add some comments.

runtime/marshaler_registry.go
```diff
@@ -5,132 +5,87 @@ import (
...
 		}
 	}
+
 	if inbound == nil {
-		inbound = defaultMarshaler
+		inbound = mux.marshalers.mimeMap[MIMEWildcard]
 	}
 	if outbound == nil {
-		outbound = defaultMarshaler
+		outbound = mux.marshalers.mimeMap[MIMEWildcard]
```
yugui: Could you try falling back to `mux.marshalers.mimeMap[contentTypeHeader]` before `MIMEWildcard`?

runtime/marshaler_registry.go
```diff
@@ -5,132 +5,87 @@ import (
...
 
-// marshalerRegistry keeps a mapping from MIME types to mimeMarshalers.
-type marshalerRegistry map[string]*mimeMarshaler
-
-type mimeMarshaler struct {
-	inbound  Marshaler
-	outbound Marshaler
+// marshalerRegistry is a mapping from MIME types to Marshalers.
+type marshalerRegistry struct {
+	mimeMap map[string]Marshaler
```
yugui: unnecessary change.

runtime/marshaler_registry.go
```diff
@@ -5,132 +5,87 @@ import (
...
-		m := mux.marshalers.lookup(val)
-		if m != nil {
-			if inbound == nil {
-				inbound = m.inbound
-			}
-			if outbound == nil {
-				outbound = m.outbound
-			}
+	for _, acceptVal := range r.Header[acceptHeader] {
+		if m, ok := mux.marshalers.mimeMap[acceptVal]; ok {
```
yugui: I prefer to keeping `lookup` method to encapsulate internal structure.

runtime/marshaler_registry.go
```diff
@@ -5,132 +5,87 @@ import (
...
-		return nil
+// MakeMarshalerMIMERegistry returns a new registry of marshalers.
+// It allows for a mapping of case-sensitive Content-Type MIME type string to runtime.Marshaler interfaces.
+//
+// For example, you could allow the client to specify the use of the runtime.JSONPb marshaler
+// with a "applicaton/jsonpb" Content-Type and the use of the runtime.JSONBuiltin marshaler
+// with a "application/json" Content-Type.
+// "*" can be used to match any Content-Type.
+// This can be attached to a ServerMux with the marshaler option.
+func MakeMarshalerMIMERegistry() marshalerRegistry {
```
yugui: With the new API with `WithMarshalerOption`, you don't need to export this function because the only caller of this function is `NewServeMux`.
Also the function should be named `newMarshalerRegistry` by convention.

runtime/marshaler_registry.go
```diff
@@ -5,132 +5,87 @@ import (
...
+	acceptHeader      = http.CanonicalHeaderKey("Accept")
 	contentTypeHeader = http.CanonicalHeaderKey("Content-Type")
+
+	defaultMarshaler = &JSONPb{OrigName: true}
 )
 
 // MarshalerForRequest returns the inbound/outbound marshalers for this request.
 // It checks the registry on the ServeMux for the MIME type set by the Content-Type header.
 // If it isn't set (or the request Content-Type is empty), checks for "*".
+// If that isn't set, uses the ServerMux's InboundMarshaler/OutboundMarshaler.
```
yugui: remove this line.

runtime/marshaler_registry_test.go
```diff
@@ -34,27 +34,27 @@ func TestMarshalerForRequest(t *testing.T) {
 		wantOut runtime.Marshaler
 	}{
 		{
```
yugui: Could you add scenario for `Accept` header?

tamird: Done.

runtime/marshaler_registry.go
tamird: Unexported. I left it `make` because it returns a value, not a pointer.

runtime/marshaler_registry.go
tamird: what's unnecessary? there's no longer a mapping to two marshalers.

runtime/marshaler_registry.go
tamird: Done.

runtime/marshaler_registry.go
tamird: Yep, done.

runtime/marshaler_registry.go
tamird: none of this is exported and the tuple form of map lookup is more correct.

runtime/marshaler_registry.go
tamird: Done.

tamird:
@yugui ping, could you take another look?

yugui:
LGTM

tamird:
Thanks!

PR #155, opened 2016-05-12
Implement gRPC timeout support for inbound HTTP headers

mwitkow:
This PR implements support for Grpc-Timeout support https://github.com/gengo/grpc-gateway/issues/107
It is based ontop of changes in https://github.com/gengo/grpc-gateway/pull/65 so that they can be merged one after the other.


yugui:
Thank you.  Merged with some modification.

Issue #168, opened 2016-05-26
Enable RFC6455 for streaming endpoints


tmc:
The most reliable way to support streaming (including bidi) to web clients is RFC6455 websockets. Go has a rich implementation for websocekts and grpc-gateway should enable grpc streaming endpoints with it.

joeblew99:
I notices that the swaggger and openAPi group as thinking of supporting WS as part of the swagger spec.

I am currently using GRPC internally, and wrapping with the GRPC gateway, to produce a swagger spec for the outside world.
then from the swagger spec i generate Javascript API.

So, if swagger starts to offer WS, then things will be complete.

tmc:
preliminary work and kind of hacked up but https://github.com/tmc/grpcutils/blob/master/websocket_proxy.go is an example of wrapping grpc-gateway with a transparent websocket upgrade

joeblew99:
@tmv 
Thank you for pointing that out. This is really something i have been wanting for a while. 

The code i saw referring to it was in the go-swagger repo. They are these days part of the open-api initiative. I cant remember the exact file.
But i tried that stack, and now am trying our the GRPC, GRPC-Gateway stack. I feel this is better for many reasons we dont need to go into.

Also i am building a React and React native stack on top of all this and saw your code for generating thats part of the stack at https://github.com/pwmckenna/swagger-models-to-react-proptypes.
This is awesome work, and something on my critical path.

yugui:
I'm happy to accept streaming support with websockets, but I expect two things in the support.

1. Concrete use cases
    * The main reason why I have not yet implemented it was that I didn't have a concrete use case.
2. Good support in client-code generation.
    * At least, we need to be able to expect that we can have a good support in future.

tmc:
@yugui I think there's a basic argument of just fully expressing grpc's semantics to browser clients.

I have starts of basic implementations of flowtype and elmlang generators here https://github.com/tmc/grpcutils and first-class streaming support in grpc-gateway will only motivate more effort in this area.

shaxbee:
@tmc Your code was extremely helpful! I had slightly different requirements and couldn't use gorilla. Shameless plug: https://github.com/shaxbee/go-wsproxy ;-)

shaxbee:
@yugui Currently websocket wrapper is used in application I work on to handle user chat and data changes notifications.

tmc:
FWIW I separated that code: https://github.com/tmc/grpc-websocket-proxy

[...]

PR #170, opened 2016-05-27
Add bidirectional streaming support by running Send() and Recv() concurrently

### LINKED ISSUE #169: Bidirectional streams don't concurrently Send and Recv
I was writing a websocket proxy on top of grpc-gateway but realized for bid streaming endpoints all the Send()s run before Recv() is first run. We should interleave these.

tmc:
Fixes #169

### DIFF
--- a/protoc-gen-grpc-gateway/gengateway/template.go
+++ b/protoc-gen-grpc-gateway/gengateway/template.go
@@ -258,7 +258,12 @@ var (
 		}
 		return nil
 	}
-	firstResult := handleSend()
+	if err := handleSend(); err != nil {
+		if err := stream.CloseSend(); err != nil {
+			grpclog.Printf("Failed to terminate client stream: %v", err)
+		}
+		return nil, metadata, err
+	}
 	go func() {
 		for {
 			if err := handleSend(); err != nil {
@@ -275,7 +280,7 @@ var (
 		return nil, metadata, err
 	}
 	metadata.HeaderMD = header
-	return stream, metadata, firstResult
+	return stream, metadata, nil
 }
 `))
 
--- a/protoc-gen-grpc-gateway/gengateway/template.go
+++ b/protoc-gen-grpc-gateway/gengateway/template.go
@@ -245,54 +245,37 @@ var (
 		return nil, metadata, err
 	}
 	dec := marshaler.NewDecoder(req.Body)
-	firstSend:= make(chan error, 1)
-	go func(errs chan<- error) {
-		defer func(){
-			if err := stream.CloseSend(); err != nil {
-				grpclog.Printf("Failed to terminate client stream: %v", err)
-			}
-		}()
+	handleSend := func() error {
+		var protoReq {{.Method.RequestType.GoType .Method.Service.File.GoPkg.Path}}
+		err = dec.Decode(&protoReq)
+		if err != nil {
+			grpclog.Printf("Failed to decode request: %v", err)
+			return err
+		}
+		if err = stream.Send(&protoReq); err != nil {
+			grpclog.Printf("Failed to send request: %v", err)
+			return err
+		}
+		return nil
+	}
+	firstResult := handleSend()
+	go func() {
 		for {
-			var protoReq {{.Method.RequestType.GoType .Method.Service.File.GoPkg.Path}}
-			err = dec.Decode(&protoReq)
-			if err == nil {
-				select {
-				    case errs <- err:
-					default:
-				}
-			}
-			if err == io.EOF {
-				select {
-				    case errs <- err:
-					default:
-				}
-				return
-			}
-			if err != nil {
-				grpclog.Printf("Failed to decode request: %v", err)
-				select {
-					case errs <- grpc.Errorf(codes.InvalidArgument, "%v", err):
-					default:
-				}
-				return
-			}
-			if err = stream.Send(&protoReq); err != nil {
-				grpclog.Printf("Failed to send request: %v", err)
-				select {
-					case errs <- err:
-					default:
-				}
-				return
+			if err := handleSend(); err != nil {
+				break
 			}
 		}
-	}(firstSend)
+		if err := stream.CloseSend(); err != nil {
+			grpclog.Printf("Failed to terminate client stream: %v", err)
+		}
+	}()
 	header, err := stream.Header()
 	if err != nil {
 		grpclog.Printf("Failed to get header from client: %v", err)
 		return nil, metadata, err
 	}
 	metadata.HeaderMD = header
-	return stream, metadata, <-firstSend
+	return stream, metadata, firstResult
 }
 `))
 
--- a/protoc-gen-grpc-gateway/gengateway/template.go
+++ b/protoc-gen-grpc-gateway/gengateway/template.go
@@ -245,8 +245,13 @@ var (
 		return nil, metadata, err
 	}
 	dec := marshaler.NewDecoder(req.Body)
-	sendErrs := make(chan error, 1)
+	firstSend:= make(chan error, 1)
 	go func(errs chan<- error) {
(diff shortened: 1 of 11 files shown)


protoc-gen-grpc-gateway/gengateway/template.go
```diff
@@ -234,6 +236,66 @@ var (
...
+			}
+		}
+	}(sendErrs)
+	header, err := stream.Header()
+	if err != nil {
+		grpclog.Printf("Failed to get header from client: %v", err)
+		return nil, metadata, err
+	}
+	metadata.HeaderMD = header
+	return stream, metadata, <-sendErrs
```
yugui: * it blocks forever unless you close `sendErrs`?
* it still blocks until the end of the goroutine even if you close `sendErrs`.  So it is not interleaving?

tmc: my intent was to block until the message is sent by the client as a sort of middle ground between just ignoring errors and interrupting the stream on any error. This was written mid-migraine so I'm happy to admit it's not as thought out as it could be. Open to suggestions.

yugui: I overlooked L260-L263.  So actually it does not block forever.

I have two more questions.
* Maybe I'm missing something, but it does not seem to be something like "interleave" for me.  What do you mean by "interleave"?
* If it is intended to block until the message is sent, why did you need to run the function in a goroutine?
  IIUC, the only difference made by the goroutine is that the goroutine function runs in parallel to `stream.Header()`.
  But I don't think the difference makes much sense.

tmc: This is for the many messages situation. Before this change all Send()s have to run before Recv() gets called. by putting Send() in a goroutine concurrent Send() and Recv() calls can be made.

tmc:
@yugui do you need further clarification?

To restate, before this change a streaming handler did not start a goroutine and ran all Send()s before returning and having the caller call Recv(). This proposal is a middle ground of blocking until the first Send() to aid in surfacing issues with initial payloads.

Perhaps setting errs to nil would be a nicer way to express the semantics of block until first execution. I'm open to suggestion here.

yugui:
> a streaming handler did not start a goroutine and ran all Send()s before returning and having the caller call Recv().

I don't think it was changed with this PR because `<-sendErrs` waits for an error from the goroutine or the termination of the goroutine. Also on error from `<-sendErrs`, the caller returns wihtout calling `Recv`.  So `Send` and `Recv` never interleave, IIUC.

So in my understanding, the change made by this PR was
* Read `req.Body` until its end even if there was an error
* Send streams to the upstream grpc server even if there was an error

I think your intention

> to block until the message is sent by the client as a sort of middle ground between just ignoring errors and interrupting the stream on any error.

makes sense. But if my understanding above was correct, you didn't need the new goroutine. You could do the same thing in the same goroutine as caller.

tmc:
How would you accomplish this without a separate goroutine?

tmc:
Did you miss that the first Send if successful will send nil and stop blocking?

yugui:
> Did you miss that the first Send if successful will send nil and stop blocking?

Yes, I did. Sorry.

Then, why do you treat the first item in the stream specially?
It returns `400 Bad Request` if it fails to decode the first item but it keeps going when it fails to decode later items.

Also I'm not sure if it is safe to start writing response body before reading the request body until the end.

https://www.w3.org/Protocols/rfc2616/rfc2616-sec6.html#sec6
> After receiving and interpreting a request message, a server responds with an HTTP response message.

tmc:
@yugui I was missing some things here (it was written in a rush). This now ensures CloseSend is called and cleans up the semantics a bit. 

One alternative would be to move the Send handling out and invoke it once to capture the initial error. Thoughts?

tmc:
@yugui I changed the implementation to be more clear.

yugui:
@tmc Thank you. That's much clearer.
I still have small coding-level comments and I am worrying about compatibility to the current HTTP/1.1 version of response forwarder.

https://www.w3.org/Protocols/rfc2616/rfc2616-sec6.html#sec6
> After receiving and interpreting a request message, a server responds with an HTTP response message.

In my understanding, it requires us to receive the request before sending any response in the TCP session.  So #168 looks to be a requirement of this PR.  What do you think?

protoc-gen-grpc-gateway/gengateway/template.go
```diff
@@ -234,6 +236,49 @@ var (
...
+	var metadata runtime.ServerMetadata
+	stream, err := client.{{.Method.GetName}}(ctx)
+	if err != nil {
+		grpclog.Printf("Failed to start streaming: %v", err)
+		return nil, metadata, err
+	}
+	dec := marshaler.NewDecoder(req.Body)
+	handleSend := func() error {
+		var protoReq {{.Method.RequestType.GoType .Method.Service.File.GoPkg.Path}}
+		err = dec.Decode(&protoReq)
```
yugui: `if err := dec.Decode(&protoReq); err != nil` would be more consistent?

protoc-gen-grpc-gateway/gengateway/template.go
```diff
@@ -234,6 +236,49 @@ var (
...
+			grpclog.Printf("Failed to decode request: %v", err)
+			return err
+		}
+		if err = stream.Send(&protoReq); err != nil {
+			grpclog.Printf("Failed to send request: %v", err)
+			return err
+		}
+		return nil
+	}
+	firstResult := handleSend()
```
yugui: why don't you return here on error?

protoc-gen-grpc-gateway/gengateway/template.go
tmc: i was following existing style from the 'client-streaming-request-func' template.

protoc-gen-grpc-gateway/gengateway/template.go
tmc: I did that to allow the goroutine to start so CloseSend would be called. would you prefer that be done explicitly and return before launching the goroutine?

tmc:
@yugui updated with feedback taken into consideration.

This SO post deals with the state of response-before-request (early 401s in particular) http://stackoverflow.com/questions/14250991/is-it-acceptable-for-a-server-to-send-a-http-response-before-the-entire-request

The most common case here would be clients that don't process the response until they've sent the request fully. For my use case (a websocket proxy on top of grpc-gateway) this isn't a concern as I control the http communication on both sides.

I don't think having some misbehaving http clients should block this as it enables writing websocket proxies. Having this work in grpc-gateway closes the gap of bringing grpc semantics to browsers (with a little extra effort).

protoc-gen-grpc-gateway/gengateway/template.go
yugui: No, I don't. it's ok

yugui: Yes.  Could you do so?

I am worrying about that the error from `stream.Header()` overwrites `firstResult`.

yugui:
@tmc.  Thank you.  LGTM.

karlkfi:
New to grpc-gateway. Have some questions related to this work.

I take it by "enables writing websocket proxies" that grpc-gateway does not itself use websockets for streaming. Is that correct?

Is there an external websocket proxy project? 

If not, what would one look like? Would it be generic or have to be generated from proto files?

tmc:
@karlkfi that is correct.

I have an example websocket proxy here: https://github.com/tmc/grpcutils/blob/master/websocket_proxy.go

it's a generic http.Handler that will perform a ws upgrade on the same url as the http endpoint.

example use here:
https://github.com/tmc/grpcutils/tree/master/examples/cmd/wsechoserver

PR #174, opened 2016-06-03
Include HTTP req.remoteAddr in gRPC ctx

mikeatlas:
This annotates the gRPC request context with the HTTP remote address, which is a solution the problem of the gRPC service not being able to correctly identify the remote peer address due to the proxy-forwarding nature of grpc-gateway, which I described in https://github.com/gengo/grpc-gateway/issues/173.

### DIFF
--- a/runtime/context.go
+++ b/runtime/context.go
@@ -26,7 +26,7 @@ func AnnotateContext(ctx context.Context, req *http.Request) context.Context {
 
 	for key, vals := range req.Header {
 		for _, val := range vals {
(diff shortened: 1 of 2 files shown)


yugui:
Thank you.
I merged it, but I did not choose the metadata name "RemoteAddr" because there was a similar PR #155.

mikeatlas:
The PR #155 is more comprehensive than my changeset. Thanks for the merge and pointer :+1:

PR #190, opened 2016-07-12
Fix asynchronous send in gateway codegen

shaxbee:
I'd like to check if it was expected behavior since i hit an issue when stream will not become bidirectional until first send occurred.

It seems to me that the sync send was left in template - is that intended behavior?

### DIFF
--- a/examples/examplepb/a_bit_of_everything.pb.go
+++ b/examples/examplepb/a_bit_of_everything.pb.go
@@ -8,7 +8,7 @@ import proto "github.com/golang/protobuf/proto"
 import fmt "fmt"
 import math "math"
 import _ "github.com/gengo/grpc-gateway/third_party/googleapis/google/api"
-import google_protobuf1 "github.com/golang/protobuf/ptypes/empty"
+import google_protobuf1 "google/protobuf"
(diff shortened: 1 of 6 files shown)


shaxbee:
@yugui Seems like protobuf ptypes for Empty type are being used now but ```make examples``` target has not been updated. Should I fix it as well?

yugui:
cc: @tmc 

I think it is intentional, so that HTTP response headers are available.

shaxbee:
I've run ```go get -u github.com/golang/protobuf/protoc-gen-go``` beforehand.
It might be related to homebrew packaged version of protoc (I'm at 3.0.0-beta-3).

tmc:
Waiting to Send() until the first message was an intentional tradeoff to attempt to fail fast when invalid messages are sent on the stream. I think we should revisit this tradeoff because it's an impedance mismatch between this and straight grpc.

tmc:
https://github.com/gengo/grpc-gateway/compare/master...tmc:tmc.bidi_send_without_recv pushed here.

shaxbee:
@yugui Feel free to close this PR in favor of @tmc change, as I'm unable to regenerate examples correctly on MacOS.

yugui:
@shaxbee Sure.
Thank you for your interest in improving grpc-gateway.

PR #200, opened 2016-07-25
Avoid Internal Server Error on zero-length input for bidi streaming

### LINKED ISSUE #195: marshaler runtime.Marshaler does not handle io.EOF when decoding
Hi,

after resolving `golang/protobuf/jsonpb` compatibility issue, we found another issue with json decoder

Details can be found at https://github.com/coreos/etcd/pull/5969#issuecomment-233384529.

To summarize

`dec.Decode(&protoReq)` returns `io.EOF` in this code (full generated code can be found https://github.com/coreos/etcd/blob/master/etcdserver/etcdserverpb/rpc.pb.gw.go#L92-L93):

```go
func request_Watch_Watch_0(ctx context.Context, marshaler runtime.Marshaler, client WatchClient, req *http.Request, pathParams map[string]string) (Watch_WatchClient, runtime.ServerMetadata, error) {
	stream, err := client.Watch(ctx)
	if err != nil {
		grpclog.Printf("Failed to start streaming: %v", err)
		return nil, metadata, err
	}
	dec := marshaler.NewDecoder(req.Body)
	handleSend := func() error {
		var protoReq WatchRequest
		err = dec.Decode(&protoReq)
		if err != nil {
			grpclog.Printf("Failed to decode request: %v", err)
			return err
		}
		if err = stream.Send(&protoReq); err != nil {
			grpclog.Printf("Failed to send request: %v", err)
			return err
		}
		return nil
	}
```

But if we manually make it handle `io.EOF` as below:

```diff
        dec := marshaler.NewDecoder(req.Body)
        handleSend := func() error {
                var protoReq WatchRequest
                err = dec.Decode(&protoReq)
-               if err != nil {
+               if err != nil && err != io.EOF {
```

where `req.Body` is `{"create_request": {"key": "Zm9v"}}`, everything works as expected.

Is there any way to make this decoder handle `io.EOF`?

Thanks.

yugui:
Fixes #195

### DIFF
--- a/examples/examplepb/flow_combination.pb.gw.go
+++ b/examples/examplepb/flow_combination.pb.gw.go
@@ -119,8 +119,11 @@ func request_FlowCombination_StreamEmptyStream_0(ctx context.Context, marshaler 
 		return nil
 	}
 	if err := handleSend(); err != nil {
-		if err := stream.CloseSend(); err != nil {
-			grpclog.Printf("Failed to terminate client stream: %v", err)
+		if cerr := stream.CloseSend(); cerr != nil {
+			grpclog.Printf("Failed to terminate client stream: %v", cerr)
+		}
+		if err == io.EOF {
+			return stream, metadata, nil
 		}
 		return nil, metadata, err
 	}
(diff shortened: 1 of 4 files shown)


yugui:
@tmc Would you mind taking a look at this PR?

tmc:
@yugui LGTM

Issue #161, opened 2016-05-17
Support specifying which schemes should be output in swagger.json


johanbrandhorst:
It is not possible to modify what the "schemes" list will contain when output from protoc-gen-swagger, as per 

``` go
Schemes:     []string{"http", "https"},
```

(https://github.com/gengo/grpc-gateway/blob/03164ec50aaf90b115daec838d3ce18f4b0a34ea/protoc-gen-swagger/genswagger/template.go#L432).

This means that on HTTPS only servers, the swagger "Try it out!" buttons will attempt to use HTTP, and thus fail. Currently the workaround is to edit the output of protoc-gen-swagger to remove HTTP from the schemes list.

I'm not sure what would be the best way to fix this, but I would like some kind of way to configured the schemes at "compile time", without having to edit the output.

https://github.com/swagger-api/swagger-ui/issues/1006 describes my problem.

edrex:
I hit this too.

Reversing the order (`"https"` before  `"http"`) seems to be sufficient to cause swagger-ui to make https requests, so that might be a quick fix.

Is running grpc-gateway without HTTPS even supported?

achew22:
There is no need for it to be specified as https. It could be specified as http. You can edit the swagger output file by hand and serve it with that line saying just "http" if you don't want to bother to modify the protoc-gen-swagger binary.

edrex:
@achew22 I don't really understand your comment. In our config grpc-gateway is served under https-only, so ideally for us schemes would just have one item, "https". 

Are there people running it http-only, for whom changing the order would break the swagger "try it out" buttons?

edrex:
I realized there is another option, just dropping the "schemes" key. If it's missing it defaults to the scheme the UI is being served from.

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#schema

edrex:
Agree with @achew22's thinking this is a bug in swagger-ui. Filed https://github.com/swagger-api/swagger-ui/issues/2342.

c4milo:
Yeah, it seems to be a bug in swagger-js https://github.com/swagger-api/swagger-js/issues/841

achew22:
swagger-api/swagger-js#841 is now closed. Closing this issue with the assumption that is fixed. If that's not the case please feel free to reopen the issue.

PR #214, opened 2016-08-12
protoc-gen-swagger: output https first in schemas key

edrex:
Workaround for #161

This assumes that (most) everyone is running grpc-gateway under https. If that's not a valid assumption, maybe adding a `schemas` option to `protoc-gen-swagger` is needed.

Also, a better default might be to omit the `schemas` key, which should cause swagger-ui to use whichever schema it's being accessed through.

https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#schema

### DIFF
--- a/protoc-gen-swagger/genswagger/template.go
+++ b/protoc-gen-swagger/genswagger/template.go
@@ -443,7 +443,7 @@ func applyTemplate(p param) (string, error) {
 	s := swaggerObject{
 		// Swagger 2.0 is the version of this document
 		Swagger:     "2.0",
-		Schemes:     []string{"http", "https"},
+		Schemes:     []string{"https", "http"},
 		Consumes:    []string{"application/json"},
 		Produces:    []string{"application/json"},
 		Paths:       make(swaggerPathsObject),
--- a/protoc-gen-swagger/genswagger/template_test.go
+++ b/protoc-gen-swagger/genswagger/template_test.go
@@ -101,7 +101,7 @@ func TestApplyTemplateSimple(t *testing.T) {
 	if want, is, name := "", got.BasePath, "BasePath"; !reflect.DeepEqual(is, want) {
 		t.Errorf("applyTemplate(%#v).%s = %s want to be %s", file, name, is, want)
 	}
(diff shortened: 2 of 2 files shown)


tmc:
@edrex can you sign the CLA and re-trigger the build (it looks like an intermittent failure).

christianvozar:
gRPC requires HTTP/2. While TLS is not a requirement of HTTP/2, and therefore gRPC, as pointed out in the [HTTP/2 FAQ](https://http2.github.io/faq/) `However, some implementations have stated that they will only support HTTP/2 when it is used over an encrypted connection, and currently no browser supports HTTP/2 unencrypted.`

So it seems the assumption connections would be more or less TLS is valid and it is hard to imagine connection not being TLS unless under development.

achew22:
@christianvozar You are definitely right about gRPC being on HTTP/2, but gRPC-gateway is a gateway between gRPC and HTTP. gRPC-gateway runs a normal HTTP/1.1 server which can operate in both http and https mode depending on how it is configured.

@edrex This looks reasonable to me, but I think I'm even more fond just omitting the schema from the output. @yugui, what do you think?

edrex:
I'm happy to update to omit the key. I guess it might cause problems if the UI and API are on different origins, with one HTTPS and the other HTTP, but that is not generally well supported by browsers (and in fact if the UI is on HTTPS I believe the browser will block subrequests over HTTP, so only the other way is possible).

I think the build is failing because I didn't update some of the test data. I'll try to run the full suite before submitting again.

@tmc I looked around for the CLA link but failed to find it so far. I've signed a Google CLA for Camlistore, not sure if it's the same.

achew22:
@edrex, you bring up very valid points. I don't know that there will ever be a way to solve this that makes absolutely every situation happy. It is too bad the swagger-ui doesn't have a "baseUrl" override on it.

Would that be a better way to approach this? Is this really a bug in swagger-ui?

edrex:
@achew22 I think you're right that there is a bug in swagger-ui. Filed https://github.com/swagger-api/swagger-ui/issues/2342

tmc:
This is fixed in https://github.com/swagger-api/swagger-js/issues/841

PR #216, opened 2016-08-16
Use a released version of protoc

yugui:


achew22:
LGTM

yugui:
The test is failing due to an issue of the devel version of go compiler

> ```
> # google.golang.org/grpc/transport
> ../../../google.golang.org/grpc/transport/http_util.go:248: internal compiler error: schedule does not include all values
> ```

PR #222, opened 2016-09-07
readme: added links to additional documentation

### LINKED ISSUE #212: documentation of semantics of endpoint declarations
After working with this project for a little while, I finally found [Google's description of the semantics](https://github.com/googleapis/googleapis/blob/master/google/api/http.proto) that are used in this package. It would be super helpful if that (or a better version of that) were linked to prominently in the README.md. It would be even better if the specific way that this library implemented that description was documented, so that someone learning how to use this library doesn't have to just mess around with the library to see how it works. The implicit generation of query parameters especially threw me, I didn't realize that this library could even do that until someone mentioned it in another issue.

sdemos:
This commit adds links to the google cloud platform's documentation on the custom options used by grpc-gateway to generate your http proxy. This documentation has information about how path, query, and body parameters are handled, as well as specific assumptions about the arguments present for different http methods.

I added the link in a couple of places in the readme that seemed like places where you might want additional documentation. Let me know if they should be linked to differently or something like that. 

Fixes #212

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,7 @@
 grpc-gateway is a plugin of [protoc](http://github.com/google/protobuf).
 It reads [gRPC](http://github.com/grpc/grpc-common) service definition,
 and generates a reverse-proxy server which translates a RESTful JSON API into gRPC.
+This server is generated according to [custom options](https://cloud.google.com/service-management/reference/rpc/google.api#http) in your gRPC definition.
 
 It helps you to provide your APIs in both gRPC and RESTful style at the same time.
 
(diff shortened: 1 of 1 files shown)


yugui:
@sdemos Thank you for your help.

The automatic process is failing to check if you have signed [Google CLA](https://cla.developers.google.com/clas) for a technical reason.  Have you signed?

sdemos:
I hadn't done that. It should be all set to go now.

yugui:
@sdemos Thanks. LGTM.

PR #231, opened 2016-10-07
README - Add CoreOS example

### LINKED ISSUE #230: Serving swagger.json from runtime mux?
It seems that the best practice for serving a swagger.json is to do it at the apex of the API. However, the runtime does not seem to easily support this. 

Is there a best practice for this? Does it make sense to create a custom handler for the swagger here?

philipithomas:
Show users how to combine these different projects in a unified way, e.g. how to serve the generated `swagger.json` on the API

Closes #230

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,6 +196,8 @@ More examples are available under `examples` directory.
 * `server/main.go`: service implementation
 * `main.go`: entrypoint of the generated reverse proxy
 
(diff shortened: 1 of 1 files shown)


philips:
LGTM :)

yugui:
LGTM. Thank you for your contribution!

PR #235, opened 2016-10-15
populate from both path fields and body as per spec

### LINKED ISSUE #234: Support combination of query params and body for POSTs
Currently, according to [http.proto](https://github.com/googleapis/googleapis/blob/master/google/api/http.proto), it's not possible to combine both query parameters and a body. I don't see any technical reason why this is not possible; manually tweaking the genertaed `.pb.gw.go` with a call to `PopulateQueryParameters` seems to confirm that it works fine.

For example, assume that every RPC request accepts a parameter `accessToken`. It's convenient to say that the parameter may occur either in the body or in the query. These should be equivalent:

```
$ curl -XPOST http://localhost:20000/ -d'{
  "accessToken": "foo",
  "message": "hello world"
}'
```

and:

```
$ curl -XPOST http://localhost:20000/?accessToken=foo -d'{
  "message": "hello world"
}'
```

However, grpc-gateway only seems to offer the choice between `body: "*"` and `body: "message"`. I'd actually be happy with a compromise that mandates `accessToken` always be a query-string param, but you can't choose a subset of fields, either.

tmc:
related to #234

### DIFF
--- a/examples/examplepb/a_bit_of_everything.pb.gw.go
+++ b/examples/examplepb/a_bit_of_everything.pb.gw.go
@@ -219,11 +219,19 @@ func request_ABitOfEverythingService_Create_0(ctx context.Context, marshaler run
 
 }
 
+var (
+	filter_ABitOfEverythingService_CreateBody_0 = &utilities.DoubleArray{Encoding: map[string]int{"": 0}, Base: []int{1, 1, 0}, Check: []int{0, 1, 2}}
+)
+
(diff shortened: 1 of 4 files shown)


tmc:
.

achew22:
@tmc, is this still valid?

achew22:
I closed the issue associated with this and will do the same to the PR. Please reopen or resend the PR if this is still a thing you want to pursue :smile:

PR #236, opened 2016-10-16
runtime - export prefix constants

philipithomas:
Some clients that map HTTP to GRPC requests (e.g. for tracing purposes or  for having a load balancer denote an external ip) need to know how this library maps HTTP to gRPC. Rather than hardcoding the prefix in their code, I think that users should be able to import this variable so that future gprc-gateway changes do not (transparently) break this functionality.

(and I'm writing a client right now that would benefit from this)

### DIFF
--- a/runtime/context.go
+++ b/runtime/context.go
@@ -15,8 +15,8 @@ import (
 	"google.golang.org/grpc/metadata"
 )
 
-const metadataHeaderPrefix = "Grpc-Metadata-"
-const metadataTrailerPrefix = "Grpc-Trailer-"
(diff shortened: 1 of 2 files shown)


yugui:
LGTM.

But cla/google bot failed to check if you have signed a [CLA].(https://cla.developers.google.com/clas). 
@philipithomas Would you let me know if you have signed? And if you haven't, could you sign?

philipithomas:
@yugui  - I have signed (and had merged code already). CLA is under philip at staffjoy.com or mail at philipithomas.com

yugui:

PR #247, opened 2016-11-03
Render the swagger request body properly

dprotaso:
The rendering of the http body was always the methods request type.

This is _only_ correct if the body is "*" which is not always true.


tmc:
@dprotaso thank you for your contribution -- could you look at the test failure and sign the cla?

dprotaso:
Take a look at the failure it's unrelated to the diff. Something in
examples is importing a package that's main.

dprotaso:
I rebased but gotip is so fragile - it's failing to fetch and install golint

tmc:
@dprotaso I re-ran the tests and the flakiness is solved.

PR #252, opened 2016-11-05
Pass permanent HTTP request headers

tmc:
This is a rework/revival of #213 with the following changes:

* 'grpcgateway-' metadata prefix is specified and documented.
* the 'authorization' header is passed as it was before, preserving backwards compatibility.
* MetadataHeaderPrefix comment is changed to be more accurate.

Further work would include the ability to explicitly whitelist additional parameters but we should do that in a separate diff.


tmc:
@willnorris any info on the cla/google check hanging?

tmc:
Ah. @christianvozar can you fill out the google cla at https://cla.developers.google.com ?

willnorris:
@tmc, as noted in the comment from @googlebot, @christianvozar has already signed the CLA (emphasis added):

> We found a Contributor License Agreement for you (the sender of this pull request) **and all commit authors**

Them problem is that you are submitting a pull request including code authored by someone else, so we just want that someone else to confirm that they're okay with their code being contributed to the project.  The CLA status won't actually change from it's current state, but it's okay to merge once @christianvozar has indicated agreement.

tmc:
@willnorris this project uses GitHub's protected branches feature to prevent direct pushes to master -- how likely is it that the googlebot will learn how to deal with this situation?

@yugui if the answer is no then we I/we need another path to get these types of PRs in.

yugui:
@christianvozar Can I merge your contribution in this PR into this repository?

PR #262, opened 2016-11-16
template: don't CloseSend stream on http request io.EOF

heyitsanthony:
curl will POST with a single message then EOF. If there is more than one
response from the stream, it wouldn't send because of the CloseSend().


heyitsanthony:
Problem looks to be on our side. Sorry for the noise. Closing

PR #267, opened 2016-11-20
Handle methods with no bindings more carefully

### LINKED ISSUE #263: protoc-swagger-generator messes up the comments if there is rpc method that does not have rest
In case there is rpc method that does not support rest api (or does not have option (google.api.http)) the json generation gets confused and misaligns the comment for the methods.

tmc:
Fixes #263


yugui:
LGTM

PR #269, opened 2016-11-21
Add `go_package` option

remohammadi:
So when `annotations.proto` is imported in another proto, the generated code of that proto import the right package.


tmc:
Thanks for your report, could you supply an example and/or failing test demonstrating this issue?

tmc:
this is replaced by #325

Issue #279, opened 2016-12-11
Option to Allow Swagger for DELETEs with a body


msample:
Retro-specifying some legacy REST services in GRPC+REST annotations causes swagger generation from the GRPC to fail when rpc methods use http delete and have a request body.   

Using a body with delete is certainly bad style so this check should not be implicitly disabled.  I have a PR  #280 that adds a bool "allow_delete_body" plugin parameter that defaults to false.

tmc:
The related issue is merged now.

PR #280, opened 2016-12-11
add plugin param 'allow_delete_body' 

msample:
Allows  GRPC REST mappings for legacy services that inappropriately use HTTP DELETE with a body
Issue #279


msample:
I filled in the google contributor license agreement just after I submitted the PR.  Not sure how to make the cla/google check go faster or retry.

tmc:
@msample thank you for your contribution. This code looks good -- what are your thoughts on if this flag should be commented?

msample:
@tmc hopefully most people won't need it, but I'd be happy to update wherever flags are externally documented if you want.  Please point me at right thing to edit.

Alternatively, we could upgrade the error message that is currently printed when you try to use a DELETE with a body to mention this option (gets to be a bit of a drag if we add more cases that care if DELETE has a body though).

tmc:
I'm fine with leaving it undocumented. Other thoughts @achew22, @yugui, @t-yuki?

achew22:
I like it as an undocumented feature. @msample could you add a couple of tests to verify that it does what you need when the flag is set?

msample:
@achew22 sounds good. I'll try to get to it this weekend.

msample:
Two things I left as-is that seem suspect in protoc-gen-swagger/main.go.  Some might depend on this behaviour:

```	if *file != "stdin" {
		f, _ = os.Open("input.txt") // <---" input.txt"? not *file?
	}
```

2. support of pkg map parameters. "Mfoo/bar/baz.proto=github.com/proj/with/gen/baz-pb-go" seems unneeded for generating swagger.  Could break some makefiles if it becomes a  param error tho.

tmc:
@msample that does appear suspect, thanks for pointing that out. Can you comment again to kick googlebot?

msample:
CLA check kick.

tmc:
@msample I like the kvvar refactor but can we separate that into a separate pr?

msample:
@tmc sure, not a problem. Should probably be applied to both protoc_gen_* anyway.

msample:
@tmc kvvar removed.

msample:
last commit under my work email.  Have added CLA to that.

tmc:
@msample thanks!

PR #281, opened 2016-12-15
Added support for Timestamp in URL.

JohanSJA:


tmc:
@JohanSJA thank you for your contribution. This change looks good -- can you sign your Google CLA?

Are there other values than "null" that need to be special cased? NaN

JohanSJA:
I thought I have signed one before? If not, can you point me to where I should sign for it?

Not sure if we need to handle the other null value though. The code is actually referred from the protobuf project.

tmc:
@JohanSJA looks like googlebot just hadn't reported correctly -- can you resolve the query_test.go conflict? I'd love to see this land.

tmc:
ahem. googlebot?

JohanSJA:
Wake up, googlebot.

achew22:
I want to use this so I'm going to squash and merge. Thanks for your contribution @JohanSJA

Issue #276, opened 2016-12-05
client declared and not used compilation error, after recent upgrade


c4milo:
I'm getting compilation errors with the latest changes. 

```
...
// RegisterAuthzHandler registers the http handlers for service Authz to "mux".
// The handlers forward requests to the grpc endpoint over "conn".
func RegisterAuthzHandler(ctx context.Context, mux *runtime.ServeMux, conn *grpc.ClientConn) error {
	client := NewAuthzClient(conn)

	return nil
}

var ()

var ()
...
````

```proto
...
service Authz {
	rpc SignIn (SignInRequest) returns (SignInResponse) {}
	rpc SignOut (SignOutRequest) returns (SignOutResponse) {}
}
...
```

c4milo:
I can confirm 199c40a060d1e55508b3b85182ce6f3895ae6302 to be the culprit

PR #282, opened 2016-12-19
If there are no http rules or options, don't register methods

kirk91:
Fix: https://github.com/grpc-ecosystem/grpc-gateway/issues/276

@tmc @yugui


tmc:
@kirk91 thank you for your contribution. We can't just skip those as we manage to mess up swagger comment attachment if we do so. I have an alternate solution here: https://github.com/grpc-ecosystem/grpc-gateway/pull/286

kirk91:
@tmc sorry, i haven't notice the swagger comment.

PR #286, opened 2017-01-04
Prevent lack of http bindings from generating non-building output

### LINKED ISSUE #276: client declared and not used compilation error, after recent upgrade

tmc:
Fixes #276

The change to allow empty bindings was incomplete in that it would produce invalid code if a service existed with no bindings at all (versus having a subset of methods without bindings).


tmc:
googlebot kick

PR #293, opened 2017-01-09
properly respect file flag for protoc-gen-swagger

tmc:


protoc-gen-swagger/main.go
```diff
@@ -43,8 +43,12 @@ func main() {
 
 	glog.V(1).Info("Processing code generator request")
 	f := os.Stdin
-	if *file != "stdin" {
-		f, _ = os.Open("input.txt")
+	if *file != "-" {
+		var err error
+		f, err = os.Open(*file)
```
tmc: @achew that would define a new f instead of assigning to the one created previously.

tmc:
ping googlebot

achew22:
The error is `main_test.go:118: Test 0: file misparsed, expected 'stdin', got '-'`. That can be fixed by editing https://github.com/grpc-ecosystem/grpc-gateway/blob/b15a928074001d2729c56cdac319c907648856e6/protoc-gen-swagger/main_test.go#L25 to be `-` instead of `stdin`.

PR #300, opened 2017-01-15
Convert the first letter of method name to upper

lipixun:
It'll generate wrong method name if the method name starts with a lower-case letter in the proto file of gRPC.

The code generation rule of protoc-go will convert the first letter to upper-case but grpc-gateway do not.

So I just slightly changed the method name before applying the template.

### DIFF
--- a/protoc-gen-grpc-gateway/gengateway/template.go
+++ b/protoc-gen-grpc-gateway/gengateway/template.go
@@ -76,6 +76,8 @@ func applyTemplate(p param) (string, error) {
 		var methodWithBindingsSeen bool
 		for _, meth := range svc.Methods {
 			glog.V(2).Infof("Processing %s.%s", svc.GetName(), meth.GetName())
+			methName := strings.Title(*meth.Name)
+			meth.Name = &methName
 			for _, b := range meth.Bindings {
 				methodWithBindingsSeen = true
 				if err := handlerTemplate.Execute(w, binding{Binding: b}); err != nil {


tmc:
Thanks for your contribution. Can you sign the CLA and demonstrate this behavior in an example or a test?

lipixun:
All right. I have signed the CLA.

It seems a little hard to write test code into current test cases, the tests are ran on a mock output of protoc but this pr focus on fix the issue of integration with protoc-go.

Here's the test case:

Write a protobuf file:

```
syntax = "proto3";

package test;

option go_package = "testpkg";

message SomeRequest {
    int32 start = 1;
    int32 size = 2;
}

message SomeResponse {
    repeated string items = 1;
}

service SomeService {
    rpc list(SomeRequest) returns (SomeResponse) {
        option (google.api.http) = {
            post: "/resource/list"
            body: "*"
        };
    }
}
```

Pay attention to the method `list` not `List`, and run protoc `protoc -I . -I $GOPATH/src/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis --go_out=plugins=grpc:. test.proto`

This will generate the go source code file with the following codes:

```go
func (c *someServiceClient) List(ctx context.Context, in *SomeRequest, opts ...grpc.CallOption) (*SomeResponse, error) {
	out := new(SomeResponse)
	err := grpc.Invoke(ctx, "/test.SomeService/list", in, out, c.cc, opts...)
	if err != nil {
		return nil, err
	}
	return out, nil
}
```

The method name is `List` not `list`.

Run the grpc gateway compile command: `protoc -I=. -I $GOPATH/src/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis --grpc-gateway_out=logtostderr=true:. test.proto`

The generated grpc gateway code:

```
func request_SomeService_list_0(ctx context.Context, marshaler runtime.Marshaler, client SomeServiceClient, req *http.Request, pathParams map[string]string) (proto.Message, runtime.ServerMetadata, error) {
	var protoReq SomeRequest

	if err := marshaler.NewDecoder(req.Body).Decode(&protoReq); err != nil {
		return nil, metadata, grpc.Errorf(codes.InvalidArgument, "%v", err)
	}

	msg, err := client.list(ctx, &protoReq, grpc.Header(&metadata.HeaderMD), grpc.Trailer(&metadata.TrailerMD))
	return msg, metadata, err

}
```

Take a look at `client.list`, it should be `client.List`. And what I've committed is to convert the first letter from lower case to upper case.

The protobuf official document says:

```
The first letter is capitalized for export. If the first character is an underscore, it is removed and a capital X is prepended.
```

lipixun:
I signed it!

lipixun:
I signed it!

PR #312, opened 2017-02-11
Do not add imports from methods with no bindings.

### LINKED ISSUE #310: Package imported but not used
Hi!
When a service method uses a type `T` from an imported proto and does not define the option `google.api.http` to get exposed by the grpc-gateway, `protoc-gen-grpc-gateway` tries anyway to import the package, where the type `T` is defined, and therefore I get the following message when I try to compile the gateway:
```
./main.pb.gw.go:17: imported and not used: "github.com/golang/protobuf/ptypes/empty"
```

Here's an example of such service:
```protobuf
syntax = "proto3";

import "github.com/golang/protobuf/ptypes/empty/empty.proto";

package tester;

service Service {
  rpc MethodExposed (Message) returns (Message) {
    option (google.api.http) = {
      post: "/exposed"
      body: "*"
    };
  }

  rpc MethodImportingEmpty (google.protobuf.Empty) returns (google.protobuf.Empty) {}
}

message Message {
	string value = 1;
}
```

fische:
Fixes #310


tmc:
@fische thanks for your contribution -- can you sign the google CLA please?

fische:
Normally I've already signed it on https://cla.developers.google.com/. Is it the good one?

tmc:
@fische ah, sometimes it takes another comment, as demonstrated.

tmc:
@fische I like this change but would like the behavior covered by a test case to prevent regression in the future.

fische:
No problem. Just trying to fix travis tests and it should be fine.

fische:
I still don't understand why 7f2e9507539a786703a38d25ad801b04b0dd833b did not pass the tests on travis...
Anyway, it should be alright now @tmc.

PR #314, opened 2017-02-17
enable parsing enums from query parameters

tzneal:
- parses enumeration strings and integer values
- range checks both the string and integer form

PR #315, opened 2017-02-23
Look up enum value maps by their proto name

nilium:
This fixes parsing of the string form of protobuf enum values in query
parameters and fixes the enum value map lookup done by query.go,
originally introduced in #314.

Previously, the runtime looked up the the enum value map via the Go
type name of the enum, derived from its reflect.Type. The Go type name
may not always map 1:1 to the protobuf name, though. For example, if
the proto file declares a package such as `google.protobuf`, it won't
match since the enum's name is fully qualified when registered for the
proto package. To make sure tests reflected the disparity, the query
test was amended to use a different package name from the Go package
when registering the enum.

To support this, it's necessary to pass the *proto.Properties for
fields forward when populating them, in order to allow the
populateField functions to determine if the field is an enum and look
up its value map by the props.Enum field.

As a result of this, the tests now pass, using the protobuf enum name
as registered (i.e., fully qualified package name) instead of the Go
package name derived by calling (reflect.Type).String. This should have
no real impact on existing use, but it may be justified to re-add
a fallback case for that reflect.Type name for hand-written protobuf
types. Even then, however, it seems unlikely that it should be expected
for enums-by-name to work when registering the enum map under a name
other than the one it's referred to in protobuf.

### DIFF
--- a/runtime/query.go
+++ b/runtime/query.go
@@ -40,13 +40,15 @@ func populateFieldValueFromPath(msg proto.Message, fieldPath []string, values []
 	if m.Kind() != reflect.Ptr {
 		return fmt.Errorf("unexpected type %T: %v", msg, msg)
 	}
+	var props *proto.Properties
 	m = m.Elem()
 	for i, fieldName := range fieldPath {
 		isLast := i == len(fieldPath)-1
 		if !isLast && m.Kind() != reflect.Struct {
 			return fmt.Errorf("non-aggregate type in the mid of path: %s", strings.Join(fieldPath, "."))
 		}
-		f := fieldByProtoName(m, fieldName)
+		var f reflect.Value
+		f, props = fieldByProtoName(m, fieldName)
 		if !f.IsValid() {
 			grpclog.Printf("field not found in %T: %s", msg, strings.Join(fieldPath, "."))
 			return nil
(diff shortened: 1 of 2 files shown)


runtime/query.go
```diff
@@ -82,26 +84,26 @@ func populateFieldValueFromPath(msg proto.Message, fieldPath []string, values []
...
+func fieldByProtoName(m reflect.Value, name string) (reflect.Value, *proto.Properties) {
 	props := proto.GetProperties(m.Type())
 	for _, p := range props.Prop {
 		if p.OrigName == name {
-			return m.FieldByName(p.Name)
+			return m.FieldByName(p.Name), p
 		}
 	}
-	return reflect.Value{}
+	return reflect.Value{}, nil
```
echlebek: Can this nil propagate to calls to populateRepeatedField or populateField?

nilium: Haven't seen it do so in tests. I could add a nil check around its uses, but an enumeration with no properties seems like bad code gen more than anything else. I'll poke around a bit and see what I can find.

tmc:
@nilium thanks for catching this -- have you signed the CLA?

nilium:
@tmc Yes, I have. I guess the checker isn't working right now?

runtime/query.go
nilium: It looks like the only case where it's nil would be if the reflect.Value is invalid. If the reflect.Value is invalid, then the caller returns an error (and in this case, there's only one place that fieldByProtoName is called). Could also add a prop == nil check to where it's being called, though? That'd make doubly sure we've got a proto field, I just don't know if it's paranoid vs. practical.

tmc:
Thank you.

Issue #82, opened 2016-01-05
Support oneof types of fields


yugui:
Extracted from #5.

Currently oneof fields are not supported. They should be allowed in request body and query string as well as other types.
* This feature requires #79 because `"encoding/json".Marshal` returns error for oneof fields
* No plan to allow them in path parameters because path parameters are "mandatory" parameters.  So they don't fit to `oneof` very well.

JohanSJA:
Perhaps use of https://godoc.org/github.com/golang/protobuf/jsonpb instead of encoding/json can help.

yugui:
@JohanSJA Thank you.  Yes, it is #79.

gyuho:
@yugui etcd is hitting this issue, though it's happening only in experimental feature.

Do we have any workaround for this? Or ETA to support this?

Thanks!

JohanSJA:
You should be able to use `one_of` with the help of `runtime.WithMarshalerOption`.

yugui:
Now the default marshaller uses 'jsonpb'.  So I think it is supported.
https://travis-ci.org/gengo/grpc-gateway/jobs/137727616#L1049

@gyuho Could you re-confirm with the HEAD of the master branch?
Let me know if you still have the issue.
I'll close this issue if otherwise.

gyuho:
@yugui First thanks for helping debugging this.

etcd still complains `{"Error":"json: cannot unmarshal object into Go value of type etcdserverpb.isWatchRequest_RequestUnion","Code":2}`. And tried the latest master branch of grpc-gateway and golang/protobuf, and regenerate the gateway files, as here https://github.com/coreos/etcd/pull/5891/commits/87592304b2c095425e224f289e122143c8e27487.

I added `reflect.TypeOf(marshaler)` to this line https://github.com/gyuho/etcd/blob/3108859828f851e9a46772fe8e16d257b40f3639/etcdserver/etcdserverpb/rpc.pb.gw.go#L91 and confirm that `*runtime.JSONPb` typed marshaler is being used, but still seeting that error.

```
*runtime.JSONPb
2016-07-06 22:07:13.861689 I | v3rpc/grpc: Failed to decode request: json: cannot unmarshal object into Go value of type etcdserverpb.isWatchRequest_RequestUnion
```

Maybe we generate gateway files in the wrong way? I tried to regenerate `*.gw.go` files but don't see any diff.

Thanks.

yugui:
@gyuho I see. I'll take a look.

[...]

PR #321, opened 2017-03-03
Support oneof fields in query params

nilium:
This modifies fieldByProtoName to support two things:

1. It may now return an error. It only does this for the oneof case
   right now, as the not-found case only causes the field to be
   skipped.

2. It will also look for the field name in the message's OneofTypes
   map. If the field matches a OneOf type, it will check first that the
   field is nil -- if it isn't, the field is already set and it returns
   an error. If nil, it allocates a new oneof type for the field and
   returns the first field of the oneof type.

The reason for rejecting multiple oneof fields is that the runtime
iterates over url.Values and gets pseudo-random key order, meaning that
one field is selected out of several. We could address this by sorting
the url.Values and picking the last seen, in which case the code is far
simpler (in that there's no new error case) but involves allocating and
sorting an array of field names before walking them (only to get
predictable field selection).

This change amends the error result tested for in
TestPopulateParameters to include the name of oneof_value.

Related to issue #82.

Commit 24b4e8c adds tests that intentionally fail without oneof support
in place. Commit 3ab9905 includes changes to support oneof fields
(allowing the tests to pass).

### DIFF
--- a/runtime/query.go
+++ b/runtime/query.go
@@ -48,8 +48,11 @@ func populateFieldValueFromPath(msg proto.Message, fieldPath []string, values []
 			return fmt.Errorf("non-aggregate type in the mid of path: %s", strings.Join(fieldPath, "."))
 		}
 		var f reflect.Value
-		f, props = fieldByProtoName(m, fieldName)
-		if !f.IsValid() {
+		var err error
+		f, props, err = fieldByProtoName(m, fieldName)
+		if err != nil {
+			return err
+		} else if !f.IsValid() {
 			grpclog.Printf("field not found in %T: %s", msg, strings.Join(fieldPath, "."))
 			return nil
 		}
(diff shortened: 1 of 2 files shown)


nilium:
In case anyone (or the bot) needs it stated, I have signed the Google CLA.

runtime/query.go
```diff
@@ -43,13 +43,16 @@ func populateFieldValueFromPath(msg proto.Message, fieldPath []string, values []
 	var props *proto.Properties
 	m = m.Elem()
 	for i, fieldName := range fieldPath {
-		isLast := i == len(fieldPath)-1
+		var isLast = i == len(fieldPath)-1
```
nilium: Considering making a commit to revert this particular line. It got changed as part of hacking on this a bit (before fieldByProtoName returned an error). I could also amend the commit and submit a new pull request. Does it seem worthwhile to do either?

tmc: Feel free to amend and push

nilium: Amended commit force-pushed. Didn't realize GitHub supported that now. (Has it always?)

tmc:
@nilium this looks great, thank you for your contribution.

tmc:
I think being explicit over allowing random behavior is prefereable here. LGTM.

PR #323, opened 2017-03-09
add support for adding context annotators, expose runtime.AddContextAnnotator

ilius:
For services that need to read from `http.Request` (for example take something from cookie) and modify GRPC context (`ctx`).
A very useful (or maybe essential) use case is reading JWT token from cookie and puttng it in GRPC context (with or without validation)


tmc:
Thank you for your contribution! let's move forward with the approach in #336.

PR #325, opened 2017-03-13
Update upstream proto files and add google.golang.org/genproto support.

tmc:
This updates the vendored proto files which point to google.golang.org/genproto as the generated go location.

- default CI build to go1.8
- expect committed go files to match protoc+go1.8


tmc:
x

PR #335, opened 2017-03-23
Ignore Intellij IDEA folder

TamalSaha:


tmc:
@TamalSaha thank you for your issue. Let's keep editor specific configuration int local git configuration. You can extend your local ignore list with the file .git/info/excludes

PR #336, opened 2017-03-23
Support user configurable header forwarding & context metadata

### LINKED ISSUE #311: Why does not gateway forward headers as-is?
Hi,
Why does not gateway forward headers as-is? Is there a technical reason for this?

Thanks.

TamalSaha:
Fixes #311 

@tmc / @yugui , if the general idea looks ok to you, I will add tests for this.

cc:@sadlil

### DIFF
--- a/runtime/context.go
+++ b/runtime/context.go
@@ -44,7 +44,7 @@ At a minimum, the RemoteAddr is included in the fashion of "X-Forwarded-For",
 except that the forwarded destination is not another HTTP service but rather
 a gRPC service.
 */
-func AnnotateContext(ctx context.Context, req *http.Request) (context.Context, error) {
+func AnnotateContext(ctx context.Context, mux *ServeMux, req *http.Request) (context.Context, error) {
 	var pairs []string
 	timeout := DefaultContextTimeout
 	if tm := req.Header.Get(metadataGrpcTimeout); tm != "" {
@@ -61,12 +61,10 @@ func AnnotateContext(ctx context.Context, req *http.Request) (context.Context, e
 			if strings.ToLower(key) == "authorization" {
 				pairs = append(pairs, "authorization", val)
 			}
-			if isPermanentHTTPHeader(key) {
-				pairs = append(pairs, strings.ToLower(fmt.Sprintf("%s%s", MetadataPrefix, key)), val)
-				continue
-			}
-			if strings.HasPrefix(key, MetadataHeaderPrefix) {
-				pairs = append(pairs, key[len(MetadataHeaderPrefix):], val)
+			if mux.incomingHeaderMatcher != nil {
+				if h, ok := mux.incomingHeaderMatcher(key); ok {
+					pairs = append(pairs, h, val)
+				}
 			}
 		}
 	}
@@ -94,7 +92,11 @@ func AnnotateContext(ctx context.Context, req *http.Request) (context.Context, e
 	if len(pairs) == 0 {
 		return ctx, nil
 	}
-	return metadata.NewOutgoingContext(ctx, metadata.Pairs(pairs...)), nil
+	md := metadata.Pairs(pairs...)
+	if mux.metadataAnnotator != nil {
+		md = metadata.Join(md, mux.metadataAnnotator(ctx, req))
+	}
+	return metadata.NewOutgoingContext(ctx, md), nil
 }
 
 // ServerMetadata consists of metadata sent from gRPC server.
--- a/runtime/mux.go
+++ b/runtime/mux.go
@@ -1,12 +1,14 @@
 package runtime
 
 import (
+	"fmt"
 	"net/http"
 	"strings"
 
 	"golang.org/x/net/context"
 
 	"github.com/golang/protobuf/proto"
+	"google.golang.org/grpc/metadata"
 )
 
 // A HandlerFunc handles a specific pair of path pattern and HTTP method.
@@ -19,6 +21,9 @@ type ServeMux struct {
 	handlers               map[string][]handler
 	forwardResponseOptions []func(context.Context, http.ResponseWriter, proto.Message) error
 	marshalers             marshalerRegistry
+	incomingHeaderMatcher  HeaderMatcherFunc
+	outgoingHeaderMatcher  HeaderMatcherFunc
+	metadataAnnotator      func(context.Context, *http.Request) metadata.MD
 }
 
 // ServeMuxOption is an option that can be given to a ServeMux on construction.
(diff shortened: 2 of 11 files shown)


tmc:
@TamalSaha thanks for your contribution.

A couple notes:
1. Mutating a global to modify behavior should be avoided, I'd prefer passing some sort of Option.
2. It's not clear to me that we should supply a set of Matcher implementations.

TamalSaha:
> Mutating a global to modify behavior should be avoided, I'd prefer passing some sort of Option.
Do you have any suggestion about that? The only passed thing I can see if the Mux

https://github.com/grpc-ecosystem/grpc-gateway/blob/686368427ddb9a51628d63db6c74fbc96a206e1f/protoc-gen-grpc-gateway/gengateway/template.go#L340

> It's not clear to me that we should supply a set of Matcher implementations.
I have no options about this. I thought it could be handy.

tmc:
I would prefer the Register methods taking a variadic options slice.

TamalSaha:
@tmc, I have looked into the code and template. It seems to me that extending ServeMux might be easier way. I can also turn my matchers into `ServeMuxOption`. So, mux sounds good? Then I will try to update this pr. Now, this will be a pretty big change.

TamalSaha:
@tmc, I think it is ready for review. I have extended ServeMux to add the matchers.

tmc:
$DAYJOB beckons but I can review this sometime soon. In the mean time you might take a look at somewhat related efforts in https://github.com/grpc-ecosystem/grpc-gateway/pull/323 and https://github.com/grpc-ecosystem/grpc-gateway/pull/305

TamalSaha:
@tmc, I have signed the CLA. 

I think https://github.com/grpc-ecosystem/grpc-gateway/pull/323 is taken care by this pr.

runtime/matchers.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+}
+
+// PrefixFoldMatcher performs a case-insensitive prefix match for request metadata keys
+func PrefixFoldMatcher(h string) ServeMuxOption {
+	h = strings.ToLower(h)
+	return func(mux *ServeMux) {
+		mux.headerMatchers = append(mux.headerMatchers, func(key string) bool {
+			return strings.HasPrefix(strings.ToLower(key), h)
+		})
```
nilium: I think this could be handled without the ToLower by doing an EqualFold on the head of the key. For example, https://play.golang.org/p/lh_JIYnZ0s

I only mention this because EqualFold is a bit more involved than comparing two lowercase strings.

TamalSaha: Thanks @nilium . I updated accordingly.

TamalSaha:
@tmc, I have added support for context annotators in my pr. Looks like we need this too!

TamalSaha:
@ilius, I have added support for adding metadata to context via Mux options. Does this solve your use-case?

ilius:
@TamalSaha Not sure. Does it support returning error (in REST response) before reaching grpc client?

We need to read request cookie, set grpc context or return error, on each request

TamalSaha:
I have few questions:
- When you read cookie, do you need to send it via a direct key on Context or as part of metadata in Context? From looking at the metdata code, it seems to perform some encoding, decoding based on the data.

- If your intention is to reject a call if it is missing Cookie? If so, why not just pass the cookie from http request to context metadata. Then in the grpc side, you can add an interceptor that will reject requests that are missing cookie. Does that work for you?

ilius:
- We use metadata for putting context item (JWT token), because otherwise it won't be transferred via grpc

- We reject the call if the cookie is missing, or invalid (invalid signed JWT token)

TamalSaha:
@ilius, in that case, you will be able to handle your use case using this pr, like below:
```
gwMux := runtime.NewServeMux(runtime.EqualFoldMatcher("Cookie"))
```

Then on the grpc side, you can check for JWT token in cookie and reject the request if it fails in a grpc interceptor. IMO, this is better than failing at the gateway layer for 2 reasons:
- You keep all your application in one place instead of passing some to the gateway.
- If/when applications can direct http/2 calls to the grpc service, it will also work if JWT token in passed via cookie.

runtime/matchers.go
```diff
@@ -0,0 +1,39 @@
+package runtime
+
+import "strings"
+
+// EqualMatcher performs a case-sensitive equality match for request metadata keys
+func EqualMatcher(s string) ServeMuxOption {
```
tmc: I would prefer a WithHeaderMatcher Option that accepts a ```func(string) bool``` to both support more scenarios but also to have a smaller api surface area.

runtime/mux.go
```diff
@@ -36,12 +39,21 @@ func WithForwardResponseOption(forwardResponseOption func(context.Context, http.
 	}
 }
 
+// WithMetadata returns metadata to be passed with context.
```
tmc: this comment should elaborate a bit more

runtime/mux.go
```diff
@@ -36,12 +39,21 @@ func WithForwardResponseOption(forwardResponseOption func(context.Context, http.
...
+}
+
 // NewServeMux returns a new ServeMux whose internal mapping is empty.
 func NewServeMux(opts ...ServeMuxOption) *ServeMux {
 	serveMux := &ServeMux{
 		handlers:               make(map[string][]handler),
 		forwardResponseOptions: make([]func(context.Context, http.ResponseWriter, proto.Message) error, 0),
 		marshalers:             makeMarshalerMIMERegistry(),
+		headerMatchers:         make([]func(string) bool, 0),
+		metadataAdders:          make([]func(context.Context, *http.Request) metadata.MD, 0),
```
tmc: nit: can we call this metadataAnnotators?

runtime/matchers.go
tmc: Thinking further, perhaps the signature should be ```func(string) (string, bool)``` to allow transformations in addition to simple predicate filtering.

TamalSaha: @tmc , I am not sure what you mean by `WithHeaderMatcher` option? Mind elaborating a bit?

tmc: I mean to something like:

```
func WithHeaderMatcher(fn HeaderMatcherFunc) ServeMuxOption {
    return func(s *ServeMux) {
        s.headerMatcher = fn
   }
}
```

TamalSaha:
@tmc  PTAL.

runtime/mux.go
```diff
@@ -36,12 +39,37 @@ func WithForwardResponseOption(forwardResponseOption func(context.Context, http.
...
+	}
+}
+
 // NewServeMux returns a new ServeMux whose internal mapping is empty.
 func NewServeMux(opts ...ServeMuxOption) *ServeMux {
 	serveMux := &ServeMux{
 		handlers:               make(map[string][]handler),
 		forwardResponseOptions: make([]func(context.Context, http.ResponseWriter, proto.Message) error, 0),
 		marshalers:             makeMarshalerMIMERegistry(),
+		headerMatchers:         make([]HeaderMatcherFunc, 0),
```
tmc: I think we should allow just one -- the semantics of multiple are a bit unclear and it is simpler to have one.

runtime/mux.go
```diff
@@ -36,12 +39,37 @@ func WithForwardResponseOption(forwardResponseOption func(context.Context, http.
...
+		metadataAnnotators:     make([]func(context.Context, *http.Request) metadata.MD, 0),
```
tmc: Similarly, I think we should allow just one metadata annotating func.

runtime/mux.go
TamalSaha: There are 2 separate use-cases:

- headerMatchers: Whether to pass header from http request to gRPC context (with optional transformation). This is either 1->1 or 1->0.

- metadataAnnotator: This is used to add additional metdata to gRPC context. This is 0->1 . This can't be done using header matcher.

Here is code snippet from our prod server:
```
gwMux := gwrt.NewServeMux(
	gwrt.EqualFoldMatcher("Cookie"),
	gwrt.PrefixFoldMatcher("x-csrf-"),
	gwrt.PrefixFoldMatcher("access-control-"),
	gwrt.WithMetadata(func(c context.Context, req *http.Request) metadata.MD {
		return metadata.Pairs("x-forwarded-method", req.Method)
	}))
```

runtime/context.go
```diff
@@ -61,6 +61,12 @@ func AnnotateContext(ctx context.Context, req *http.Request) (context.Context, e
 			if strings.ToLower(key) == "authorization" {
 				pairs = append(pairs, "authorization", val)
 			}
+			if mux.headerMatcher != nil {
+				if h, ok := mux.headerMatcher(key); ok {
+					pairs = append(pairs, h, val)
+					continue
+				}
+			}
 			if isPermanentHTTPHeader(key) {
```
achew22: It seems to me that isPermanentHTTPHeader and the following MetadataHeaderPrefix check are just the default forms of the header matcher. What would you think about moving them into a HeaderMatcher and providing them an overridable default headermatcher? That way we can unblock a lot of the bugs around headers not behaving the way people expect?

tmc: I like that. @TamalSaha thoughts?

TamalSaha: Updated.

tmc:
@TamalSaha thanks for your contribution

TamalSaha:
@tmc, I added doc. Just to be clear, mux's header matcher is applied in both direction (gw -> grpc and grpc -> gw). isPermanentHTTPHeader and Grpc-Metadata- check checks will always fail in grpc -> gw direction (handleForwardResponseServerMetadata), since all keys are small letter in md.HeaderMD (as per http/2). My question is that intentional?

TamalSaha:
@tmc, tests are passing.

tmc:
I wonder if we should expose the direction to the header matcher?

TamalSaha:
Like a IncomingHeaderMatcher, OutgoingHeaderMatcher ?

tmc:
@TamalSaha I think those names are fine.

tmc:
https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md FWIW

TamalSaha:
Updated pr to add separate incoming/outgoing HeaderMatcher. IncomingHeadermatcher is set to DefaultheaderMatcher to maintain current behavior.

tmc:
LGTM

ilius:
@TamalSaha I don't understand. We have no access to cookie in our grpc server handlers. Only context and input(request). We must use context for storing token. Cookie is only for REST (http1). So we must copy the token from cookie to GRPC context when translating REST to GRPC.

TamalSaha:
@ilius , if you want to forward the Cookie in http request to grpc, you can add an incomingHeaderMatcher for that.
```
runtime.NewServeMux(runtime.WithIncomingHeaderMatcher(func(h string) (string, bool) {
		if strings.EqualFold(h, "Cookie") {
			return h, true
		}
		return "", false
	}));
```
If you want to forward a specific key from Cookie, you can also do that by returning that in the `if` block.

ilius:
And where to put this header in the grpc context?

TamalSaha:
It will be automatically added to the Metadata in grpc context. See LN # 66 & # 95 https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/context.go#L66

ilius:
Thanks

ilius:
@TamalSaha Sorry, I think `runtime.WithIncomingHeaderMatcher` transforms **headers**, **not cookies**

TamalSaha:
@ilius , I am not sure I understand you. Using `runtime.WithIncomingHeaderMatcher`, you can pass the full http request's `Cookie` header to grpc. You can also, just a specific key from the Cookie, if you prefer that.  And custom application specific logic for validation that Cookie/Cookie-key should be done in your grpc server (directly in the method or using a interceptor). Does that make sense?

ilius:
Thanks, I get it now (the name mistakened me)

fiibbb:
This is awesome. Been waiting for 2+ months for this feature. Would like to see this in a new release. (It's kinda unfortunate this missed `1.2.2` by just one commit)

rogchap:
For anyone reading this and looking for a way to get a `*http.Cookie` from the request:

In a normal REST request you would access Cookies via `req.Cookies()` or `req.Cookie("my-cookie")` 

With all the suggestions above we are only mapping the request headers to the gRPC metadata (FYI: cookies are transmitted in the request header)

This means you have access to the raw cookie string in the metadata, not the `http.Cookie` struct.

There is no `ParseCookie()` function in the standard library 😞 see:  https://github.com/golang/go/issues/25194

So as a workaround you may need to do something like this (code not tested):

```go
	md, _ := metadata.FromIncomingContext(ctx)
	rawCookies := md.Get("Grpc-Metadata-Cookie")
	dummyReq, _ := http.NewRequest("", "", nil)
	for _, rawCookie := range rawCookies {
		dummyReq.Header.Add("Cookie", rawCookie)
	}
	myCookie, _ := dummyReq.Cookie("my-cookie")
```

PR #344, opened 2017-04-01
Add charset=utf-8 to json marhalers

ezotrank:
It's just minor change, but it should remove some worries for new users who uses  Chrome. Because when Chrome got content with "Content-Type: application/json" it doesnt set utf-8 charset as default.


tmc:
@ezotrank did you mean to close this?

ezotrank:
I realized that is not necessary.

Issue #219, opened 2016-08-21
--swagger_out plugin translates proto type int64 to string in Swagger specification


tartale:
The --swagger_out plugin yields a swagger definition that translates an int64 protobuf type to a string, even though the [Swagger 2.0 OpenAPI specification](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#data-types) allows for a signed 64-bit type (type `integer`; format `int64`).  This might be an oversight and/or copy/paste error, as [this code section](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/protoc-gen-swagger/genswagger/template.go#L148) seems to indicate that this was done intentionally for `uint64` values.

Example proto file:
```proto
syntax = "proto2";
package testmessage;

import "google/api/annotations.proto";

message TestMessage {
    optional int32 int32Field = 1;
    optional int64 int64Field = 2;
    optional uint32 uint32Field = 3;
    optional uint64 uint64Field = 4;
}

service TestService {
  rpc DoItNow(TestMessage) returns (TestMessage) {
    option (google.api.http) = {
      post: "/do/it/now"
      body: "*"
    };
  }
}
```

Running `protoc` with the `grpc-gateway` plugin and the `--swagger_out` output currently yields this swagger definition:
```json
{
  "swagger": "2.0",
  "info": {
    "title": "TestService.proto",
    "version": "version not set"
  },
  "schemes": [
    "http",
    "https"
  ],
  "consumes": [
    "application/json"
  ],
  "produces": [
    "application/json"
  ],
  "paths": {
    "/do/it/now": {
      "post": {
        "operationId": "DoItNow",
        "responses": {
          "200": {
            "description": "",
            "schema": {
              "$ref": "#/definitions/testmessageTestMessage"
            }
          }
        },
        "parameters": [
          {
            "name": "body",
            "in": "body",
            "required": true,
            "schema": {
              "$ref": "#/definitions/testmessageTestMessage"
            }
          }
        ],
        "tags": [
          "TestService"
        ]
      }
    }
  },
  "definitions": {
    "testmessageTestMessage": {
      "type": "object",
      "properties": {
        "int32Field": {
          "type": "integer",
          "format": "int32"
        },
        "int64Field": {
          "type": "string",
          "format": "int64"
        },
        "uint32Field": {
          "type": "integer",
          "format": "int64"
        },
        "uint64Field": {
          "type": "string",
          "format": "uint64"
        }
      }
[...]

PR #351, opened 2017-04-06
Set int64 fields as integers in Swagger doc

utrack:
Since jsonpb happily accepts int64s as both strings and integers from JSON - I think it's more intuitive to report them as integers in Swagger docs.


tmc:
@utrack thanks for your contribution! I agree with your assessment. Have you signed the google CLA?

TamalSaha:
I believe this has been discussed before https://github.com/grpc-ecosystem/grpc-gateway/issues/219#issuecomment-251250029

utrack:
@TamalSaha ah, I see. Thanks!

PR #359, opened 2017-04-17
bugfix: reflect upstream api change.

### LINKED ISSUE #358: reflect upstream grpc metadata api change
https://github.com/grpc/grpc-go/issues/1178

tmc:
metadata.NewContext now is a synonym for metadata.NewOutgoingContext. Call NewIncomingContext instead.

Fixes #358


tmc:
we don't currently have a more "e2e" test to show this context propagation over an actual call so this wasn't caught in test. I'd like for us to have that but also want to get a release out that is compatible with upstream grpc-go.

achew22:
+1 to e2e testing. Also LGTM.

PR #361, opened 2017-04-17
metadata: fix properly and change to Outgoing

tmc:
I was hasty in my modification before. As we're pre-call here this should use outgoing not incoming.


tmc:
clabot

tmc:
It was posted in haste. Thanks.

PR #369, opened 2017-04-19
Use canonical header form in default header matcher.

TamalSaha:
- I also removed strings.ToLower(), since metadata method itself performs this transformation for http/2.
 - This will also strip `Grpc-Metadata-` prefix regardless of the original casing used in http header.

### DIFF
--- a/runtime/mux.go
+++ b/runtime/mux.go
@@ -1,13 +1,12 @@
 package runtime
 
 import (
-	"fmt"
 	"net/http"
+	"net/textproto"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


tmc:
ci failure

TamalSaha:
Fixed ci

TamalSaha:
Code that shows that `metdata.Pairs()` encoded keys to lower case.
https://github.com/grpc/grpc-go/blob/master/metadata/metadata.go#L110
https://github.com/grpc/grpc-go/blob/master/metadata/metadata.go#L53

PR #374, opened 2017-05-01
Return response headers from grpc server

TamalSaha:
This bug was introduced in #336.


TamalSaha:
cc: @tmc

tmc:
I'd like to have this exercised under test.

tmc:
I opened #362 previously and this may fit within that definition.

achew22:
2+1 to having a test. @tmc, what do you think about merging to get it fixed and then having @TamalSaha do a followup?

tmc:
@achew22 I think that's acceptable.

achew22:
LGTM once CI approves, I think this better fits into #362

TamalSaha:
Thank you!

PR #378, opened 2017-05-05
Use status package for error and introduce WithProtoErrorHandler option

kazegusuri:
grpc-go introduced status package to handle errors with concrete types instead of interface and obsoleted some functions like grpc.Errorf.

First, this PR changed code to use status package. This will require grpc-go v1.3+
And, add `WithProtoErrorHandler` option and `ProtoErrorHandlerFunc` for handle errors as `Status` message in grpc-gateway instead of original error type a.k.a `errorBody`.
This also unifies errors returned by gRPC server and errors happened inside grpc-gateway, which helps error handling in client.

I think discussion required for response status of errors inside grpc-gateway and how to handle errors for streaming.
If needed, I will separate this PR into two PRs.

### DIFF
--- a/examples/examplepb/a_bit_of_everything.pb.gw.go
+++ b/examples/examplepb/a_bit_of_everything.pb.gw.go
@@ -23,6 +23,7 @@ import (
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/grpclog"
+	"google.golang.org/grpc/status"
 )
 
 var _ codes.Code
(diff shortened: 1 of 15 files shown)


tmc:
@kazegusuri thanks for this! this looks great. It looks like your generated swagger/openapi is a bit out of date.

kazegusuri:
It was actually caused by using newer version of protoc v3.3.0. I regenerated files with protoc v3.1.0.

TamalSaha:
Hi, we would like to see this pr merged. The new status stuff looks great!

kazegusuri:
ready to be merged if no one opposes the response status.

PR #387, opened 2017-05-13
Reference import grpc Status to suppress unused errors.

TamalSaha:
Here is an example proto file that resulted in unused errors:

```
syntax = "proto3";

package appscode.ci.v1beta1;

option go_package = "v1beta1";

import "appscode/api/annotations.proto";

service Metadata {
  rpc ServerInfo(dtypes.VoidRequest) returns (ServerInfoResponse) {
    option (google.api.http) = {
      get : "/ci/v1beta1/metadata/server-info/json"
    };
  }
}

message ServerInfoResponse {
  string provider = 1;
  string server_url = 2;
  string ca_cert = 3;
}
```

### DIFF
--- a/examples/examplepb/a_bit_of_everything.pb.gw.go
+++ b/examples/examplepb/a_bit_of_everything.pb.gw.go
@@ -28,6 +28,7 @@ import (
 
 var _ codes.Code
 var _ io.Reader
+var _ status.Status
 var _ = runtime.String
 var _ = utilities.NewDoubleArray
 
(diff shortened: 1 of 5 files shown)


TamalSaha:
Basically if you have a service that does not use any path/query parameters, this error happens.

cc: @kazegusuri @tmc

tmc:
@TamalSaha thanks for the patch.

tmc:
some files need to be re-generated.

TamalSaha:
Regenerated example protos

PR #397, opened 2017-05-19
Return codes.InvalidArgument to rather return HTTP 400 instead of HTTP 500

pieterlouw:
Hi,

Currently if there's a conversion error on the request then the gateway return an HTTP 500 with the Go error in the HTTP Body.

Won't it be clearer to the API client user if a `codes.InvalidArgument` (that translate to HTTP 400) is returned rather?

Pieter


achew22:
Pieter, this seems reasonable to me. Could you sign the CLA and add a test to verify the new behavior?

pieterlouw:
Hi,
I've signed the CLA.

About the test, where exactly should I add one ?
https://github.com/grpc-ecosystem/grpc-gateway/blob/master/examples/integration_test.go
or
https://github.com/grpc-ecosystem/grpc-gateway/blob/master/protoc-gen-grpc-gateway/gengateway/template_test.go

achew22:
Generally I believe you should test it in the smallest size test that completely validates behavior. Since template.go is generating a string that is going to be compiled by the go compiler I don't know that we can fully test the functionality in template.go. Therefore, I think the test should go in integration.go. If you have a nifty trick for testing it in template.go I would love to learn.

achew22:
Oh, I should also note that since you're modifying a template you will need to run `make examples` in the root directory to update all the example code with your change.

Thanks!

pieterlouw:
Closing this as #409 have test case and examples

Issue #400, opened 2017-05-28
allow protobuf well known types in params


tmc:
example: allow a google.protobuf.Duration in a get param.

AlekSi:
Right now, support for `google.protobuf.Duration` is very confusion.

1. https://developers.google.com/protocol-buffers/docs/proto3#json says:
> Generated output always contains 0, 3, 6, or 9 fractional digits, depending on required precision, followed by the suffix "s". **Accepted are any fractional digits (also none) as long as they fit into nano-seconds precision and the suffix "s" is required.**

2. https://github.com/go-openapi/strfmt/blob/v0.19.2/duration.go contains support for various suffixes.

3. But in practice, grpc-gateway always returns duration as seconds with `s` suffix, and uses `time.Parse` for parsing, so suffixes like `d` are not supported.

johanbrandhorst:
Is this not implemented already? Also not sure I understand your comment @AlekSi, are you saying it should support the full range required by the openapi definition of a duration? It might be possible with some pre-json unmarshalling parsing.

AlekSi:
All I'm saying is that current handling is inconsistent.

I see several options:
1. Do nothing in code, just document that behavior. :)
2. Extend parsing (only) with strfmt mapping, so suffixes like `d` are accepted.
3. Use protobuf JSON mapping consistently: accept and return only seconds with `s` suffix.
4. Use strfmt mapping consistently: accept and return other suffixes like `d`.

Options 3 and 4 are breaking changes, so probably some generator flag / protobuf option / runtime parameter would be required to use a new behavior.

Personally, I would prefer option 4 with a generator flag.

johanbrandhorst:
What are we doing right now? I thought it was #3.

AlekSi:
Right now, grpc-gateway accepts suffixes `m` and `h`. It uses `time.ParseDuration`.
Additionally, I just discovered that if no suffix is given, value is silently ignored. I think that's bad.

AlekSi:
That's how it works now:
```go
"59s": "59s",
"60s": "60s",
"61s": "61s",
"61" : "", // no suffix => error
"2m" : "120s",
"1h" : "3600s",
"1d" : "", // d suffix => error
```

Suffixes `m` and `d` are accepted, but never returned.

johanbrandhorst:
I see. So #2 but without `d` support?

AlekSi:
Yes, right now it is #2 without `d` and `w`.

[...]

PR #404, opened 2017-05-30
improve WKT handling in gateway and openapi output

tmc:
This improves Timestamp and Duration support in field marshaling and openapi/swagger output.

### DIFF
--- a/runtime/convert.go
+++ b/runtime/convert.go
@@ -2,6 +2,10 @@ package runtime
 
 import (
 	"strconv"
+
+	"github.com/gogo/protobuf/jsonpb"
+	"github.com/golang/protobuf/ptypes/duration"
+	"github.com/golang/protobuf/ptypes/timestamp"
 )
 
 // String just returns the given string.
@@ -55,4 +59,18 @@ func Uint32(val string) (uint32, error) {
 		return 0, err
 	}
 	return uint32(i), nil
+}
+
+// Timestamp converts the given RFC3339 formatted string into a timestamp.Timestamp.
+func Timestamp(val string) (*timestamp.Timestamp, error) {
+	var r *timestamp.Timestamp
+	err := jsonpb.UnmarshalString(val, r)
+	return r, err
+}
+
+// Duration converts the given string into a timestamp.Duration.
+func Duration(val string) (*duration.Duration, error) {
+	var r *duration.Duration
+	err := jsonpb.UnmarshalString(val, r)
+	return r, err
 }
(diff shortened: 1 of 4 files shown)


tmc:
x

tmc:
refs #400

runtime/convert.go
```diff
@@ -56,3 +60,17 @@ func Uint32(val string) (uint32, error) {
 	}
 	return uint32(i), nil
 }
+
+// Timestamp converts the given RFC3339 formatted string into a timestamp.Timestamp.
+func Timestamp(val string) (*timestamp.Timestamp, error) {
```
achew22: When would these functions be called?

runtime/convert.go
```diff
@@ -56,3 +60,17 @@ func Uint32(val string) (uint32, error) {
...
+	var r *timestamp.Timestamp
+	err := jsonpb.UnmarshalString(val, r)
+	return r, err
+}
+
+// Duration converts the given string into a timestamp.Duration.
+func Duration(val string) (*duration.Duration, error) {
```
achew22: Can you also add a couple of simple tests to validate this?

runtime/convert.go
tmc: these are dropped in as 'runtime.Timestamp' calls via '''ConvertFuncExpr'''

runtime/convert.go
tmc: I was considering that but kind of figured this code was well-tested in jsonpb

AlekSi:
Any update on this?

tmc:
@AlekSi this is waiting on some additional test coverage. I've been bogged down with ${dayjob} and haven't written said tests. This could land sooner if those were contributed.

achew22:
@AlekSi, if this is a thing that you need soon, I bet @tmc would be happy to have you clone this branch and write the tests so we can merge it.

ajalab:
I'm facing the similar problem about WKT handling now, so I would like to take over this task.
Where should I write tests for `Duration`?

tmc:
I think adding a convert_test.go file would be appropriate.

runtime/convert.go
achew22: I've been thinking on this for a bit. I think you're just exposing helper methods against jsonpb. That should mean they are well tested. If these were to become more complex then I would want tests but they are just so imple I think we should leave it.

achew22:
@ajalab I think I was wrong to request testing on those methods and block. I'm going to merge this. If you have any more needs beyond this can you send another PR?

Issue #405, opened 2017-06-01
customize the error return


jinleileiking:
If the input type is error, gateway will return a json.

I need to customize the message. Is there a way ?

```
➜  ~ git:(master) ✗ curl 'localhost:9998/v1/example/echo?name=dddd'
{"error":"strconv.ParseInt: parsing \"dddd\": invalid syntax","code":3}%
```

the name is int32

pieterlouw:
https://github.com/grpc-ecosystem/grpc-gateway/pull/397

jinleileiking:
I am working at ksyun openapi, we need to let the return format be the same. So we need to customize.

https://docs.ksyun.com/read/latest/116/_book/KLSAPI/CancelRecordTask.html

TamalSaha:
@jinleileiking , you might be out of luck. gRPC protocol defines its `status` using a `Status` proto. https://github.com/grpc/grpc-go/pull/1156
https://github.com/grpc/grpc-go/blob/master/status/status.go

Gateway just turns that into a JSON object.

TamalSaha:
@jinleileiking , I think you need to set the HTTPError for runtime.Servemux. See https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/mux.go#L98

vaporz:
+1
I'm having the same issue.
@pieterlouw  I was going to create a PR exactly the same with #397 , until I see this issue.

@TamalSaha Hi, our company is recently trying to build micro services using grpc-gateway, and I'm using HTTPError to handle errors. In this case, the error HTTPError receives is the error from "strconv.ParseInt()". I can not assert that, a ParseInt() error is an InvalidArgument error, and then a HTTP 4xx. More detail is needed, HTTPError has to know what exactly is happening.

vaporz:
Hi, I created a PR #409 , it's almost the same with #397 .  
For we are blocked by this issue, if there is anything I can help with, please let me know.  
Thank you!

ewang:
A workaround I've used is to have server pass the error details via a trailer metadata, the on the grpc-gateway side, override the HTTPError handler to read from the trailer metadata.

jinleileiking:
I will test lately.

jinleileiking:
The return message is: 
```
{"error":"type mismatch, parameter: num, error: strconv.ParseInt: parsing \"aaa\": invalid syntax","code":3}
```

We should completely customize this return.( Use template?)

[...]

PR #409, opened 2017-06-05
Return codes.InvalidArgument to rather return HTTP 400 instead of HTTP 500

vaporz:
1, Same change in "template.go" with #397 .  
2, Add a test case.  
3, Update examples


vaporz:
Ah...."Waiting for status to be reported", why....

vaporz:
I signed it!

vaporz:
zzzZZZ

vaporz:
ping @TamalSaha ,
Thank you.

pieterlouw:
I will close #397 as this PR have test case and examples

vaporz:
Hi, any comments on this PR?
Thanks!

TamalSaha:
LGTM.

jinleileiking:
#405

jinleileiking:
The return message is: 
```
{"error":"type mismatch, parameter: num, error: strconv.ParseInt: parsing \"aaa\": invalid syntax","code":3}
```

We should completely customize this return.( Use template?)

We should not change the core code. We need a config file  ??? to set this line:
```
+		return nil, metadata, status.Errorf(codes.InvalidArgument, "type mismatch, parameter: %s, error: %v", {{$param | printf "%q"}}, err)
```

PR #416, opened 2017-06-17
Supports oneof fields in more places

### LINKED ISSUE #413: oneof fields are not supported as the "body"?
Hi,

Here's a short sample of proto input:
```protobuf
syntax = "proto3";

package foo;

service FooService {
  rpc foo(FooRequest) returns (FooResponse) {
    option (google.api.http) = {
      post: "/foo-message"
      body: "msg"
      additional_bindings: {
        post: "/foo-struct"
        body: "str"
      };
    };
  }
}

message FooRequest {
  oneof payload {
    string msg = 1;
    google.protobuf.Struct str = 2;
  }
}

message FooResponse {}
```

It successfully generates the `pb.gw.go` file, but `go build` fails as:
```
./foo.pb.gw.go:40: protoReq.Msg undefined (type FooRequest has no field or method Msg)
./foo.pb.gw.go:61: protoReq.Str undefined (type FooRequest has no field or method Str)
```

The related line looks like:
```go
if err := marshaler.NewDecoder(req.Body).Decode(&protoReq.Msg); err != nil {
```

The generated go type has the name of field as `Payload` (the name of oneof) rather than individual oneof names, and that field accepts `FooRequest_Msg` or `FooRequest_Str` type. Maybe the codegen can't deal with oneof fields in the body parameter?

yugui:
* root of the request body: fixes #413 
* path parameters


yugui:
I noticed that the alloc function does not work when the target message is defined outside of the current package. I'll rewrite this PR.

mcos:
@yugui What's the current state of this?

abronan:
Although it's relatively easy to find a workaround when in need of such a message structure, supporting `oneof` at the root of the request body would be awesome 🤗 

Is there any help needed to push this PR forward?

tmc:
This is replaced by #634

PR #421, opened 2017-06-23
Memoise calls to fullyQualifiedNameToSwaggerName to speed it up for large registries

peterebden:
We're observing some very slow compile times (on the order of ~1 minute) for reasonably large protos, containing hundreds of messages or so.

This implements a very simple cache of calls to determine Swagger names inside `fullyQualifiedNameToSwaggerName` which appears to be taking most of the time; after this change compilation time is well under 1 second.

### DIFF
--- a/protoc-gen-swagger/genswagger/template.go
+++ b/protoc-gen-swagger/genswagger/template.go
@@ -8,6 +8,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"sync"
 
 	pbdescriptor "github.com/golang/protobuf/protoc-gen-go/descriptor"
 	"github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway/descriptor"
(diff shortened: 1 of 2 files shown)


tmc:
@peterebden thank you for your contribution, have you signed the CLA?

peterebden:
Yes I have - I can see it on cla.developers.google.com.

tmc:
@peterebden thanks, for whatever reason the cla bot needs a comment to kick in.

protoc-gen-swagger/genswagger/template.go
```diff
@@ -329,17 +329,27 @@ func renderEnumerationsAsDefinition(enums enumMap, d swaggerDefinitionsObject, r
...
+		lastRegistryMapping = resolveFullyQualifiedNameToSwaggerNames(append(reg.GetAllFQMNs(), reg.GetAllFQENs()...))
+		lastRegistry = reg
+	}
+	return lastRegistryMapping[fqn]
 }
 
+// lastRegistry is used to remember the last registry passed into fullyQualifiedNameToSwaggerName;
+// we use that to memoise calls to resolveFullyQualifiedNamesToSwaggerNames since it can be
+// significantly slow for registries of hundreds of entries.
+var lastRegistry *descriptor.Registry
```
tmc: can we instead do something like ```var fqmnCache = map[*descriptor.Registry]map[string]string``` so we can avoid the extra 'last' concept and just compute the cache once per registry seen?

peterebden: Sure, done. I did think about doing something like that at first but in the end got a little bit lazy with it :)

tmc:
@peterebden ping

peterebden:
I don't seem to be able to merge this - maybe cla-bot needs another message?

achew22:
Looks like the CLA bot is happy now. Merging (sorry for the delay)

PR #432, opened 2017-07-13
 add MethodDesciptions (slice of interfaces) to generated .pb.gw.go files

ilius:
So that REST-based client codes don't have to parse the .proto files (not even sure that's possible)

### DIFF
--- a/protoc-gen-grpc-gateway/gengateway/template.go
+++ b/protoc-gen-grpc-gateway/gengateway/template.go
@@ -335,6 +335,39 @@ func Register{{$svc.GetName}}HandlerFromEndpoint(ctx context.Context, mux *runti
 	return Register{{$svc.GetName}}Handler(ctx, mux, conn)
 }
 
+type methodDesciptionStruct struct {
+	serviceName string
+	methodName string
+	inputType string
+	outputType string
+	httpMethod string
+	pattern string
+}
+
+func (md methodDesciptionStruct) ServiceName() string {
+	return md.serviceName
+}
+
+func (md methodDesciptionStruct) MethodName() string {
+	return md.methodName
+}
+
+func (md methodDesciptionStruct) InputType() string {
+	return md.inputType
+}
+
+func (md methodDesciptionStruct) OutputType() string {
+	return md.outputType
+}
+
+func (md methodDesciptionStruct) HTTPMethod() string {
+	return md.httpMethod
+}
+
+func (md methodDesciptionStruct) Pattern() string {
+	return md.pattern
+}
+
 // Register{{$svc.GetName}}Handler registers the http handlers for service {{$svc.GetName}} to "mux".
 // The handlers forward requests to the grpc endpoint over "conn".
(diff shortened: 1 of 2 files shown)


tmc:
@ilius thanks for your contribution!

Can you elaborate on this use case a bit? Proto files are / should be parsable in a number of languages.

ilius:
@tmc I didn't find a library to parse .proto file. And parsing is not enough, we also need the input (request) and output (response) structs for each method.
Since grpc-gateway extends the proto format (by adding `option (google.api.http)`), i think it should ideally extend `grpc.ServiceDesc` as well, and add those REST-specific method properties (new fields, or simply a new function like `serviceDesc.GetMethodDescriptions()`)

achew22:
It looks to me this information is already all exported inside the [fileDescriptorApi](https://github.com/google/protobuf/blob/master/src/google/protobuf/descriptor.proto) which can be parsed into a [FileDescriptorProto](https://github.com/google/protobuf/blob/master/src/google/protobuf/descriptor.proto#L60) which has a file field and then a ServiceDescriptorProto available as "service". Can you help me understand what additional information you would need to get access to?

achew22:
I dug into this a little bit more and discovered another kind of cool thing. Looks like you can call [proto.FileDescriptor](https://godoc.org/github.com/golang/protobuf/proto#FileDescriptor) and pass in the proto filename and get back the ProtoFileDescriptor that you're interested in :smile:

ilius:
Let's say I have a proto file like this:

	syntax = "proto3";

	package ping;

	service Ping {
		rpc Ping(PingRequest) returns (PingResponse) {
			option (google.api.http) = {
				get: "/ping/ping"
			};
		}
	}

	message PingRequest {
	}

	message PingResponse {
	}

Can you give me some Go sample code that I can access the information in the `option` block of a given method ("Ping") ?

achew22:
Assuming your proto is called `ping.proto` and held in the directory `foo`, it would be something like this

```
import "github.com/golang/protobuf/proto"
import "github.com/golang/protobuf/protoc-gen-go/descriptor"

fd := &descriptor.FileDescriptorProto{}
err := proto.Unmarshal(proto.FileDescriptor("foo/ping.proto"), fd)
if err != nil {
  log.Fatalf("Oops %s", err)
}
for service := range fd.GetService() {
  for method := range service.GetMethod() {
    for option := range service.GetOptions() {
      log.Printf("Options: %s", option.String())
    }
  }
}
```

Shouldn't take much more than that to get what you want. However, if you're wanting to do some serious code generation work, I would recommend writing your own protoc plugin. It's surprisingly easy. It is just a program that takes in a FileDescriptorSet proto on stdin and outputs a GeneratedCodeInfo (or something like that). This whole project is an example of not one but two ways of doing this! Get your hands dirty, it's really fun!

ilius:
@achew22 Thank you for the code.
I get this error in `proto.Unmarshal`:
`bad wiretype for field descriptor.FileDescriptorProto.Dependency: got wiretype 7, want 2`

ilius:
Turns out `proto.FileDescriptor` returns gzipped bytes, and I had to uncompress it before passing `proto.Unmarshal`.

Issue #435, opened 2017-07-16
Support disabling manual Flush via header.


hexfusion:
In certain HTTP 1.1 streaming circumstances the runtime handler calls `Flush` when it should instead fallback to internal net/httpd handling of "flush after chunk" outlined in this commit.
https://github.com/golang/go/commit/e5febf957f5be9d3325c2d851bff6ec7c55e4662 which references issue https://github.com/golang/go/issues/6574

Example of how to replicate issue via newline delimited JSON request.  https://github.com/coreos/etcd/issues/8237

To allow this flexibility I propose a header ie `Grpc-Metadata-AutoFlush` (name?) which will conditionally disable manual `Flush`, thus falling back to the internal net/http handling. The opt-in nature of this should be unobtrusive yet allow this type of streaming request to properly return.  Also by using the Grpc-Metadata- naming it will provide useful access to the metadata key.

I was thinking something like below in runtime/handler.go

```
// replaces  f.Flush()
checkFlush(req,f)

// unless header is found manually Flush
func checkFlush(req *http.Request, f http.Flusher) {
    if af := req.Header.Get("Grpc-Metadata-AutoFlush"); af == "" {
        f.Flush()
    }
    return
}
```
Thank you for your consideration and input.
PR ref: https://github.com/grpc-ecosystem/grpc-gateway/pull/436

achew22:
Right now grpc-gateway doesn't support streaming over HTTP/1.1 the way you propose. @tmc and I have talked about adding the ability to do streaming to grpc-gateway, but we haven't yet landed on how to actually do it while being faithful to the spec. We will definitely take this effort into consideration when we work out streaming support, but for now this is a low priority item for me. If this is something you'd be interested in taking on (or any other open issue/FR) I would love to help point you to the right place. I'm going to close this to keep our backlog clean-ish but feel free to reopen it/turn it into a PR if you would like to take a swing at it.

hexfusion:
@achew22 thank you for the notes and kind invitation to provide recon. I hope to revisit this and will make sure to touch base with you before hand to compare notes.

hexfusion:
@achew22 I think this functionality would really benefit the project and would like to take this on if for nothing else to better understand how grpc-gateway works. If you have any notes you could pass my way or tickets, in particular, outlining your thoughts/requirements please forward them. Thanks!

nikkolasg:
+1 : registering my interest in this functionality as well !

Thank you for your work !

PR #436, opened 2017-07-16
runtime/handler: Support conditional Flush via header. (#435)

hexfusion:
* Support conditionally disabling manual Flush via header key, falling back to internal net/http handling. "flush after chunk"

If this idea makes sense I will add tests.


tmc:
Can you describe your use case?

The current check for simple lack of presence probably should be a little more strict, additionally this would need tests and documentation to be accepted.

hexfusion:
@tmc understood, I will get this hammered down and include tests. Thanks!

hexfusion:
@googlebot I signed it!

hexfusion:
Closing this for now while I rethink approach.

Issue #457, opened 2017-09-10
support google.api.HttpBody


theRealWardo:
I'd like my gateway to understand that I'm returning a `google.api.HttpBody` proto that should be the actual HTTP body of the response.

this ended up not being so easy because of the way the `Content-Type` handling is being done now. In order to do this properly, the following diff is necessary:

```
-       // ContentType returns the Content-Type which this marshaler is responsible for.
-       ContentType() string
+       // ContentType returns the response Content-Type for "v".
+       ContentType(v interface{}) string
```

further, in order to do this I'd propose a new Marshaler that wraps the current `JSONPb` Marshaler. the new Marshaler would be the new default as well as I believe most users want this behavior. does this seem reasonable?

achew22:
I'm going to consolidate this into just the PR #458

PR #458, opened 2017-09-10
add support for the google.api.HttpBody proto as a response

theRealWardo:
#457 fix proof of concept - probably needs some discussion on the bug as this is not very straight forward. I just hacked this together, no tests yet, so not ready for serious review until I hear from someone on the bug (if the general idea here is cool then I'll polish this PR up for submission).


tmc:
Thanks for your contribution! master has a build system fix, can you please rebase this?

theRealWardo:
@tmc sorry for the delay! rebased and pushed finally.

tmc:
@theRealWardo sorry, can you do so again?

runtime/handler.go
```diff
@@ -7,7 +7,7 @@ import (
 	"net/textproto"
 
 	"github.com/golang/protobuf/proto"
-	"github.com/grpc-ecosystem/grpc-gateway/runtime/internal"
+	"github.com/therealwardo/grpc-gateway/runtime/internal"
```
achew22: I don't think we should merge with this sourcing from another repo.

achew22:
@tmc, I found out something nifty about github the other day. You can `git clone git@github.com:theRealWardo/grpc-gateway`, check out the branch that has a PR coming your way and then push back onto it. The default PR behavior is to grant the PR recipient write into that branch (which is you). We don't need to impotently wait for a PR author to rebase so we can confirm everything works before merging... yay! :fireworks:

jinleileiking:
waiting for merge.

jinleileiking:
Needs example to show me how to use it...

runtime/handler.go
```diff
@@ -42,10 +42,11 @@ func ForwardResponseStream(ctx context.Context, mux *ServeMux, marshaler Marshal
 	if d, ok := marshaler.(Delimited); ok {
 		delimiter = d.Delimiter()
 	} else {
-	    delimiter = []byte("\n")
+		delimiter = []byte("\n")
 	}
 
 	var wroteHeader bool
+	ctSet := false
```
achew22: I'm all for short variable names, but this is too far away from the code using it to use a 2 letter abbreviation IMO. "contentTypeSet" would be preferred

runtime/marshal_httpbody.go
```diff
@@ -0,0 +1,74 @@
+package runtime
```
achew22: Do you think you could add a test for this?

runtime/marshaler.go
```diff
@@ -15,8 +15,8 @@ type Marshaler interface {
 	NewDecoder(r io.Reader) Decoder
 	// NewEncoder returns an Encoder which writes bytes sequence into "w".
 	NewEncoder(w io.Writer) Encoder
-	// ContentType returns the Content-Type which this marshaler is responsible for.
-	ContentType() string
+	// ContentType returns the response Content-Type for "v".
```
achew22: Can you elaborate on what "v" would be?

runtime/marshaler.go
```diff
@@ -15,8 +15,8 @@ type Marshaler interface {
...
+	ContentType(v interface{}) string
```
achew22: This is going to be a painful breaking change to the API. Is there ANY way to do this without passing in v?

runtime/marshaler_registry.go
```diff
@@ -13,7 +13,7 @@ var (
 	acceptHeader      = http.CanonicalHeaderKey("Accept")
 	contentTypeHeader = http.CanonicalHeaderKey("Content-Type")
 
-	defaultMarshaler = &JSONPb{OrigName: true}
+	defaultMarshaler = &HTTPBodyMarshaler{}
```
achew22: This seems incorrect. Why does the default marshaller need to change from this change?

jinleileiking:
this pull changes much for the core. I am going to work on this.

jinleileiking:
@theRealWardo Can I modify your codes? If you donot have time, I will make code changes. I like this to be merged.

runtime/marshaler_registry.go
theRealWardo: why shouldn't the default behavior be to treat HTTP body protos specially?

runtime/handler.go
theRealWardo: done.

runtime/marshaler.go
theRealWardo: hmm well the problem is that the content type right now is not dynamic. I suppose we could add another function like `ContentTypeForMessage(m proto.Message) string` and make that work as a preferred content-type setter. i.e. if `ContentTypeForMessage(m proto.Message) string` returns a non-empty string, we use it. otherwise we use the existing `ContentType() string`. now this has a similar problem as it changes a public interface, which would result in compilation errors for implementors anyway, so I'm afraid I don't have a clever solution. this feels like the lesser of two evils to me?

runtime/marshaler.go
theRealWardo: v is the same thing that `Marshal(v interface{})` is passed, I'm honestly just following convention here. should I update comments on `Marshal` as well? (happy to do that just lmk)

side note, I am not actually sure why all of these interfaces are not instead of type `proto.Message` which is what is actually being passed...

runtime/marshal_httpbody.go
theRealWardo: yea sure, but let's finish the other discussion first so I don't have to change things too much.

runtime/marshaler.go
theRealWardo: bumping this - any thoughts here?

runtime/marshaler_registry.go
achew22: It is a breaking change. I try not to make those except on major version bumps. If you want to be added to the list of 2.0 breaking changes and do it then, I'd be happy to do that.

runtime/marshaler.go
achew22: Unfortunately adding another method to the interface suffers from the same API breaking change problem so I would rather go with the simpler breaking change if it is possible. Just spitballing cause I don't think it's possible, there isn't a way to make a type union or something, is there?

theRealWardo: actually wow yea I think it is possible. so via a type union I think we can do something like this - https://play.golang.org/p/G4x7sEtEHbh - which in this case would mean we have something like this probably in this file:

```
// ContentType returns the Content-Type header value given a Marshaler or TypeAwareMarshaler and the interface to marshal v.
func contentType(m, v interface{}) string {
  switch t := m.(type) {
  case TypeAwareMarshaler:
    return t.ContentTypeForInterface(v)
  case Marshaler:
    return t.ContentType()
  }
  return ""
}
```

then we'd update all the `marshaler.ContentType()` calls in this package to `contentType(marshaler, v)`. what do you think of that? I think that is not technically API breaking, right?

johanbrandhorst:
This looks like it still needs some work, and may even be considered for v2 if the submitter wants to change the default marshaller.

johanbrandhorst:

wimspaargaren:
Hi all, Can someone tell me if there's still any work being done to support HttpBody proto as response?

johanbrandhorst:
This PR looks abandoned. Would you like to pick up the mantle @wimspaargaren?

wimspaargaren:
Sure, I know it needs a rebase, but could you enhance a bit if there need to be more adjustments before this can be merged?

johanbrandhorst:
Judging by the comments from @achew22 there still needs to be a decision made whether this should be a backwards compatible change (perhaps under a flag) and so mergable today or whether we want to make it a breaking change in which case it would have to wait until v2. As v2 is currently somewhat unclear, I would suggest the former. In any case, you personally couldn't take over this PR, you'd have to submit a new one because of how the CLA works.

wimspaargaren:
Allright! Flag or env var would work fine for me. I'll propose my own PR on Monday then.

johanbrandhorst:
Superceded by #904

theRealWardo:
awesome! thanks for picking this up @wimspaargaren

PR #459, opened 2017-09-12
Add proto marshaller for proto-over-http

MatthewDolan:


tmc:
Thanks for your contribution! master has a build system fix, can you please rebase this?

MatthewDolan:
@tmc thanks.

The build is passing now.

MatthewDolan:
@tmc Is there anything else I need to do here?

Mistobaan:
This seems interesting. Any more context / documentation on how to use it from a frontend?

MatthewDolan:
Once the `ProtoMarshaller` is added as a marshaller on the server side, you can do "proto over http." The client can send an http request with a content type "application/octet-stream" where the body is a serialized proto and the gateway will be able to handle that.

PR #417, opened 2017-06-19
Use go_package option if found.

notbdu:
Currently, `protoc-gen-grpc-gateway` does not respect the `go_package` option for output path. 

Check if a `go_package` option was specified and use that as the base output path.

PR #462, opened 2017-09-14
grpc-gateway/generator: respect full package

glerchundi:
Rebased on top of master, done by @notbdu.

Superseeds #417.

### DIFF
--- a/protoc-gen-grpc-gateway/gengateway/generator_test.go
+++ b/protoc-gen-grpc-gateway/gengateway/generator_test.go
@@ -1,6 +1,7 @@
 package gengateway
 
 import (
+	"path/filepath"
 	"strings"
 	"testing"
 
@@ -9,7 +10,16 @@ import (
 	"github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway/descriptor"
 )
 
-func TestGenerateServiceWithoutBindings(t *testing.T) {
+func newExampleFileDescriptor() *descriptor.File {
+	return newExampleFileDescriptorWithGoPkg(
+		&descriptor.GoPackage{
+			Path: "example.com/path/to/example/example.pb",
+			Name: "example_pb",
+		},
+	)
+}
+
+func newExampleFileDescriptorWithGoPkg(gp *descriptor.GoPackage) *descriptor.File {
 	msgdesc := &protodescriptor.DescriptorProto{
 		Name: proto.String("ExampleMessage"),
 	}
@@ -39,7 +49,7 @@ func TestGenerateServiceWithoutBindings(t *testing.T) {
 		Name:   proto.String("ExampleService"),
 		Method: []*protodescriptor.MethodDescriptorProto{meth, meth1},
 	}
-	file := descriptor.File{
+	return &descriptor.File{
 		FileDescriptorProto: &protodescriptor.FileDescriptorProto{
 			Name:        proto.String("example.proto"),
 			Package:     proto.String("example"),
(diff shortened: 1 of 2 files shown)


glerchundi:
Tested on our CI and it works as expected. Take into account that `file.GoPkg.Path` **cannot be empty**.

@tmc PTAL.

glerchundi:
do you agree @notbdu? :)

notbdu:
Fine with me. Thanks @glerchundi.

achew22:
Wouldn't this change the output path for every file generated, not just the ones using go_package?

glerchundi:
@achew22 yep, it will. It's a breaking change. I can refactor to switch on a parameter or whatever and keep the old behaviour but the question is, what is desired? IMHO this is what end-user expects , me included. 

I can create another PR after this one to override `go_package` based path for one that lets user choose which output folder to use.

WDYT?

glerchundi:
@tmc, @achew22 I changed the code to keep backward compatibility.

glerchundi:
I don't think this is related to the PR itself...Flaky tests? IWOMM...

glerchundi:
Any chance/free-slot to look at this, @achew22, @tmc? Thanks

achew22:
@tmc what do you think about this?

From my vantage it appears to be the correct behavior but I would like a 2nd set of eyes before we merge a breaking change. Also, once we merge this let's do a release.

@glerchundi, could you add some tests to verify the behavior and prevent regression?

achew22:
@glerchundi, I just tagged a release which means now is a GREAT time to introduce breaking changes like this. Please update the PR with some tests and let's get it merged.

glerchundi:
Ok, i don’t have much time but give me a day and will try to do ASAP.

glerchundi:
@achew22 PTAL

protoc-gen-grpc-gateway/gengateway/generator_test.go
```diff
@@ -9,7 +10,16 @@ import (
 	"github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway/descriptor"
 )
 
-func TestGenerateServiceWithoutBindings(t *testing.T) {
+func newExampleFileDescriptor() *descriptor.File {
+	return newExampleFileDescriptorWithGoPkg(
+		&descriptor.GoPackage{
+			Path: "example.com/path/to/example/example.pb",
```
achew22: Shouldn't the file name end in .go?

glerchundi: I don't think so.

I migrated an already existing example into a customisable function without doing any change to it but as you can observe in the two cases that I've implemented i used this descriptor as a 1-1 representation of the `go_package` proto option.

I tried to touch as less as possible :)

achew22:
I have manually verified CLAs on glerchundi@ and notbdu@. Merging

glerchundi:
👏🏻👏🏻

c4milo:
How can I disable this behavior? it broke my existing code :/

glerchundi:
@c4milo can you give us more details? Example proto file which was working and not now, as well as an expected behaviour explanation.

Although it should work as it was behaving before with just explicitly specifying the same `go_package` name as the one you're using in `package`.

c4milo:
I should have clarified a bit better. I have no issues with the generated code itself. I had code relying on the Swagger gateway generated code being placed in the same location as the code generated by `protoc-gen-go`. The piece of code relying on that broke.

PR #473, opened 2017-10-10
runtime: use r.Context()

srenatus:
Looks like we're only using 1.7+ (judging [from travis](https://github.com/grpc-ecosystem/grpc-gateway/blob/c323909dd41de182a54b4d1610f26034da5e097e/.travis.yml#L3-L7)) -- so, let's squash that TODO, shall we? 😉


srenatus:
> *cla/google* — Waiting for status to be reported 

Is there anything I can do to make this become green? 
- [X] My employer's CLA shows up in [gerrit](https://go-review.googlesource.com/#/settings/agreements)
- [X] I've used the same email address

PR #474, opened 2017-10-12
protoc-gen-grpc-gateway: flip request_context default to true

srenatus:
Follow up to #473.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,4 +35,4 @@ env:
   - "PATH=$PATH:$HOME/local/bin"
   matrix:
   - GATEWAY_PLUGIN_FLAGS=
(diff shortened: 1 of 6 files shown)


srenatus:
(Oh dear that CLA check really takes a while. 😃)

srenatus:
No love for this? 💔 
😉

achew22:
Not to imply that it isn't a desirable change but it is definitely a breaking change since anyone using go 1.7 (which is part of our testing matrix) is going to have a bad time with this.

srenatus:
@achew22 Thanks for the input. Is it still desirable to have 1.7 supported and as part of the testing matrix? 😃

achew22:
I just tagged a new release and as such it is a great time to be turning on new features like this. If people have issues they can use the tag v1.3

PR #489, opened 2017-11-27
Support mapping bytes to []byte

### LINKED ISSUE #5: Support bytes fields in path parameter
Should decode percent encoding to support bytes fields

loderunner:
This PR adds support for protobuf `bytes` by mapping it to `[]byte`. It uses URL-safe base64 encoding to serialize to URL query parameters.

Fixes #5


achew22:
Could you also note that this fixes #5 and remove the entry from https://github.com/grpc-ecosystem/grpc-gateway/blob/master/README.md where it says "bytes fields in path parameter. #5"

loderunner:
Done.

Could you help me out with Travis? I don't quite understand why CI is failing.

loderunner:
Travis is failing for Go 1.8.x, any idea why?

tmc:
For 1.8 (we should up it to 1.9) ci fails if generating the examples causes a diff. Please regenerate examples and push again. Could you also rebase and get rid of the merge commits? Thanks!

loderunner:
I'm getting different auto-generated files in the examples between `1.9` & `master`. I can't understand how to fix this. Is this a breaking change since 1.9? Or can I change this using some configuration of make?

achew22:
Looking at this, it seems to me that the test we have for the go version is just wrong and it is running our clean test against go master and go 1.9.

I fixed it in #520. I'm going to get lunch and let the CI run then merge that. Sorry for all the trouble. It was an interesting bug if you wanna stare at it.

achew22:
Could you rebase this onto the current HEAD? I updated the travis config so this shouldn't be a problem any more.

loderunner:
Done. Thanks for checking that out !

achew22:
No, thank you for contributing! This project is amazing for all the effort and help we've gotten from such a wide community. Thanks for joining us!

PR #494, opened 2017-12-06
Use golang/protobuf instead of gogo/protobuf

shouichi:
It seems gogo protobuf is mistakenly(?) used.


tmc:
Yes this was a mistake. Thank you.

chai2010:
greate 👍

PR #496, opened 2017-12-07
Fix gen-swagger to support more well known types

shouichi:
Some well know types are not handled correctly. Given we have the following proto message.

```
// proto
message User {
  google.protobuf.StringValue name = 1;
}
```

`name` is output as an object.

```
// swagger.json
"name": {
  "$ref": "#/definitions/protobufStringValue"
}
```

It should be output as a primitive.

```
// swagger.json
"name": {
  "type": "string"
}
```

This PR adds support for the following well know types.

- StringValue
- Int32Value
- Int64Value
- FloatValue
- DoubleValue
- BoolValue

### DIFF
--- a/protoc-gen-swagger/genswagger/template.go
+++ b/protoc-gen-swagger/genswagger/template.go
@@ -25,6 +25,29 @@ var wktSchemas = map[string]schemaCore{
 	".google.protobuf.Duration": schemaCore{
 		Type: "string",
 	},
+	".google.protobuf.StringValue": schemaCore{
+		Type: "string",
+	},
+	".google.protobuf.Int32Value": schemaCore{
+		Type:   "integer",
+		Format: "int32",
+	},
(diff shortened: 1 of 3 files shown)


achew22:
This looks like a great start! I'm excited to merge this. Could you add some tests to verify `schemaOfField` does the right thing for these new well known types? It looks like we are totally missing coverage on that method right now.

https://github.com/shouichi/grpc-gateway/blob/31e65b13c82edfcf1c78926c59fa829b7ed14841/protoc-gen-swagger/genswagger/template.go#L250

shouichi:
Sure, thanks for your advice!

shouichi:
Added test for `schemaOfField`. I'm not sure if I set up descriptor registry correctly. Feel free to comment if I'm doing something wrong.

achew22:
Sorry this took so long. Looking through the docs a bit I think this is correct. Merging. Thanks for your contribution!

PR #505, opened 2017-12-15
Fix #504 Missing Definitions

### LINKED ISSUE #504: protoc-gen-swagger missing definition issue
protoc-gen-swagger appears to have an issue where an entity is omitted from the definitions when a field is specified for the body rather than `*`.

Reproduce with the following code:

example.proto
```
syntax = "proto3";

package example;

service ExampleService {
    rpc DoSomething(DoSomethingRequest) returns (DoSomethingResponse) {
        option (google.api.http) = {
            post: "/do-something/{id}"
            body: "data"
        };
    }
}

message ExtraData {
    string foo = 1;
}

message DoSomethingRequest {
    string id = 1;
    ExtraData data = 2;
}

message DoSomethingResponse {
    string bar = 1;
}

```
Generate the code:
```
protoc -I. -I$GOPATH/src -I$GOPATH/src/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis --go_out=plugins=grpc:. --swagger_out=logtostderr=true:. ./example.proto
```

Results in:

```
{
  "swagger": "2.0",
  "info": {
    "title": "example.proto",
    "version": "version not set"
  },
  "schemes": [
    "http",
    "https"
  ],
  "consumes": [
    "application/json"
  ],
  "produces": [
    "application/json"
  ],
  "paths": {
    "/do-something/{id}": {
      "post": {
        "operationId": "DoSomething",
        "responses": {
          "200": {
            "description": "",
            "schema": {
              "$ref": "#/definitions/exampleDoSomethingResponse"
            }
          }
        },
        "parameters": [
          {
            "name": "id",
            "in": "path",
            "required": true,
            "type": "string"
          },
          {
            "name": "body",
            "in": "body",
            "required": true,
            "schema": {
              "$ref": "#/definitions/exampleExtraData"
            }
          }
        ],
        "tags": [
          "ExampleService"
        ]
      }
    }
  },
  "definitions": {
    "exampleDoSomethingResponse": {
      "type": "object",
      "properties": {
        "bar": {
          "type": "string"
        }
      }
    }
  }
}
```

Note that there is a $ref to `#/definitions/exampleExtraData` but it does not appear in the definitions.

edit: actually I don't know if the body attribute is relevant.

warmans:
This apparently fixes the issue, but I'm not 100% sure why. It seems to be something to do with the refs map missing entries for nested request types.

fixes #504


achew22:
Do you think you could add an example in https://github.com/grpc-ecosystem/grpc-gateway/blob/master/examples/examplepb/a_bit_of_everything.proto that exhibits this problem (and the solution) so that we don't regresson this in the future? Thanks so much!

warmans:
Hi, added something to the example which shows the problem (if the fix is reverted anyway). Let me know if anything is missing.

achew22:
> edit: erm, not sure how to make the tests run in CI. It imports the examples from the grpc-ecosystem path but my changes are only available in my own namespace.

I'm not sure I understand what you're saying here. Can you help me understand a little better?

warmans:
Yeah sure, AFAIK the travis tests fail because the new `GetWithBody` method is undefined on the server in `github.com/grpc-ecosystem/grpc-gateway/examples/server` implementation. But I cannot implement this method because I cannot import the request type from the imported package: 

`"github.com/grpc-ecosystem/grpc-gateway/examples/examplepb"`

it's only available in 

`"github.com/warmans/grpc-gateway/examples/examplepb"`

But I shouldn't important that. Additionally it looks like travis just `go gets` the grpc-ecosystem package so I don't know if I could import it even if I wanted to.

This wouldn't be an issue if the PR was a branch rather than a fork so maybe a maintainer could just copy the branch and replace this PR. Or perhaps there is another solution. Relative imports or something. I'm not sure :/

achew22:
I think those errors are probably only occurring in your local $GOROOT. What happens if you `mv` the `$GOROOT/src/github.com/warmans/grpc-gateway` directory into `$GOROOT/src/github.com/grpc-ecosystem/grpc-gateway`. Do you get different errors?

warmans:
Hah oh god. Why did I not think of that. Should be fixed now.

PR #508, opened 2017-12-23
Emit default values in marshalled JSON (with tests)

### LINKED ISSUE #233: Support emitting default values in JSON

ornithocoder:
Closes #233 and #242 cc: @philipithomas @achew22.


ornithocoder:
There are more tests to update. Closing it.

PR #512, opened 2018-01-02
Generate "Description" and "TermsOfService" fields

lukasmalkmus:
Both fields have been implemented but were not properly generated and thus not present in the resulting swagger spec.


lukasmalkmus:
I signed it!

tmc:
Ideally this would be covered in test but this is a small enough change that I’ll approve.

PR #514, opened 2018-01-02
ci: add codecov

### LINKED ISSUE #513: ci: set up codecov

tmc:
Fixes #513

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,10 +26,13 @@ before_script:
 - sh -c 'cd examples/browser && npm install'
 script:
 - make realclean && make examples SWAGGER_CODEGEN="java -jar $HOME/local/swagger-codegen-cli.jar"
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -4,6 +4,7 @@ go:
 - 1.7.x
```
achew22: I think we should drop 1.7.x right now and drop 1.8.x when 1.10 is released

tmc: +1

tmc:
Hm, oddly hitting an infinite recursion issue when trying to run coverage.

achew22:
I think we should remove the 1.10 beta from the test matrix. I'm sure we aren't the only people hitting that and we don't have release notes on how to fix this yet

tmc:
I wanted/needed it for coverage output consolidation. (-coverpkg=all)

PR #515, opened 2018-01-03
add error details to error json

srenatus:
I've tried to pick up where the discussion in #405 has trailed off.

I'm not sure about the import changes of the generated api_clients, but I'll see what happens when travis comes in.


srenatus:
Ok, I'm not sure I understand what is happening on travis. Could someone please have a look and share a hint? 😉

achew22:
@tmc do you have any thoughts on this?

loderunner:
I'm having the same problem in #489. It seems some commit between go 1.9 and `master` changes line breaks in autogenerated files, and the `git status --porcelain` fails.

If someone finds a fix to this, keep me posted on my PR, please.

achew22:
Looking at this, it seems to me that the test we have for the go version is just wrong and it is running our clean test against go master and go 1.9.

I fixed it in #520. I'm going to get lunch and let the CI run then merge that. Sorry for all the trouble. It was an interesting bug if you wanna stare at it.

achew22:
Could you rebase this onto the current HEAD? I updated the travis config so this shouldn't be a problem any more.

srenatus:
@achew22 rebased ✔️ Thanks -- #520 was a good catch! 😄

examples/server/a_bit_of_everything.go
```diff
@@ -248,6 +249,21 @@ func (s *_ABitOfEverythingServer) Timeout(ctx context.Context, msg *empty.Empty)
 	}
 }
 
+func (s *_ABitOfEverythingServer) ErrorWithDetails(ctx context.Context, msg *empty.Empty) (*empty.Empty, error) {
+	stat := status.New(codes.Unknown, "with details")
```
achew22: I think it would be better to collapse the next few lines

```
stat, err := status.New(codes.Unknown, "with details").
  WithDetails([]proto.message{
  ...
  })
```

PR #521, opened 2018-01-14
Add support for basic gRPC API Configuration YAML files

hacst:
Sometimes the requirement to have to annotate your .proto files to expose them using grpc-gateway can be inconvenient. For one all your backends and clients using the .proto now need to know the google api types for the annotation which can be a hassle. More importantly in a number of situations you cannot/do not want to change the .proto at all (e.g. if you don't own the .proto or only want to use the gateway for experiments).

Google offers a way to achieve this for services hosted with them called "gRPC API Configuration" files. This PR implements a small part of those for grpc-gateway:

> Add support for basic gRPC API Configuration YAML files
>
> Separate gRPC API Configuration files as described by google at
> https://cloud.google.com/endpoints/docs/grpc/grpc-service-config
> can be used to separate the description of the basic API surface
> in .proto files from their mapping to HTTP REST. In grpc-gateway
> this allows exposing gRPC services without having to add any
> annotations to their .proto files.
> 
> The format of gRPC API Configuration YAML files is defined by the
> proto message schema google.api.service which can be found as
> service.proto in the googleapis repository. This patch exposes
> the Http part of the service configuration to the gateway
> generator.
> 
> To enable this all HttpRules contained in the Http part of the
> description are loaded into the generator registry as a map from
> service method to HttpRules called externalHttpRules. During
> loading of the proto files we can now use these externalHttpRules
> to look for additional rules to apply to a service method in addition
> to the ones from the annotations found in the proto file. As the
> types are the same no fundamentally new codepaths are required in
> either of the generators.
> 
> google.api.service is quite a complex protobuf message with lots
> of dependencies on other google api. To not have to pull in all
> other dependent types this patch constructs a reduced service
> message type with only the Http field. This relies on protobufs
> backwards compatibility guarantees which allows us to remove all
> other fields. Also we do not need all protobuf features on this
> custom message.
> 
> To load the actual YAML file as a protobuf message this patch uses
> a new dependency on github.com/ghodss/yaml to convert the YAML into
> JSON which we can then unmarshal into protobuf using jsonpb.

The code already works quite nicely for what I had in mind but before doing the boring parts of completing test coverage and examples I wanted to gauge interest in integrating this feature as well as get some feedback on the approach taken.


hacst:
I signed it!

achew22:
@hacst, thanks so much for sending this in! I'm definitely intrigued by the concept here. Would the idea be to entirely leave this at code generation time, or would it become a runtime configuration option? As I understand it, the YAML files exist to modify configuration at runtime in the endpoints project. Can you help me understand a little better?

tmc:
Very cool! My read (haven't combed the diff yet) is that this is just an alternate http.proto specification avenue ala https://cloud.google.com/endpoints/docs/grpc/transcoding#configuring_transcoding_in_yaml

hacst:
Great to see there is interest.

@achew22 Now that you mention it, being able to have this a runtime configuration would be kinda neat. Google uses them this way but for grpc-gateway that would be a pretty large conceptual change I imagine. This PR really only adds another way to get http configuration into the grpc-gateway protoc plugins during generation time.

Taking the README.md echo example instead of doing step 2 you would create a ```your_api_config.yaml``` containing something like:
```yaml
type: google.api.Service
config_version: 3

http:
  rules:
  - selector: example.YourService.Echo
    post: /v1/example/echo
    body: "*"
```
That can then be passed to the gateway and swagger generator:
```
  -I$GOPATH/src \
  --grpc-gateway_out=logtostderr=true,grpc_api_configuration=path/to/your_api_config.yaml:. \
  path/to/your_service.proto

  -I$GOPATH/src \
  --swagger_out=logtostderr=true,grpc_api_configuration=path/to/your_api_config.yaml:. \
  path/to/your_service.proto
```

Hope that clarifies things.

achew22:
Could you comment on this

> Sometimes the requirement to have to annotate your .proto files to expose them using grpc-gateway can be inconvenient. For one all your backends and clients using the .proto now need to know the google api types for the annotation which can be a hassle. More importantly in a number of situations you cannot/do not want to change the .proto at all (e.g. if you don't own the .proto or only want to use the gateway for experiments).

As I understand it, you wouldn't need to include the annotation types in the outputted code. Since they are service definitions they would be available in the proto reflection as an extension, but wouldn't be included as a compiled component. Do you mean that dependent projects would need to configure the protoc to have those extensions available?

hacst:
@achew22 Yes. That and more. My understand is that as soon as my service proto includes annotations every user (whether they care about the http annotations or - imho more likely - not) not only has to make sure annotations.proto (and transitively http.proto) are available to protoc during generation but also has to have these files protoc'd for their implementation language for the client/service compilation. As these types are part of the google.api which is not shipped with protobuf this can create quite a lot of friction if the service proto is widely used.

As a concrete example in C++ I need ```google/api/annotations.pb.h``` and ```google/api/http.pb.h``` headers to be able to build as soon as the service proto includes annotations. I assumed it was the same for other languages. Is this not the case? Is there a way around this?

achew22:
Upon reflection you're totally 100% right about that. There is no way to even have that without importing the other libraries. This is also the only way we can go forward that allows someone to use grpc-gateway on a file they don't have edit permissions on.

Out of curiosity, do you just invoke this on the command line through protoc but with an additional argument grpc_api_configuration with a path to the file? For some reason in my mind it seems weird to load a file that isn't passed in on stdin for a protoc plugin. Does protoc do anything weird like sandboxing to prevent doing that?

hacst:
@achew22 Passing the file is pretty natural. You do it the same way you pass other parameters directly to the plugin right now.

E.g to make the gateway or swagger plugin itself log to stderr you currently do:
```
protoc --grpc-gateway_out=logtostderr=true:. my.proto
protoc --swagger_out=logtostderr=true:. my.proto
```
accoringly passing a yaml file is:
```
protoc --grpc-gateway-out=grpc_api_configuration=my.yaml:. my.proto
protoc --swagger_out=grpc_api_configuration=my.yaml:. my.proto
```
Everthing between ```*_out=``` and the ```:``` after which you specify the output location is handed to the plugin as part of the request. This is a protoc feature and the gateway and swagger plugin explicitely implement support for it (see https://github.com/grpc-ecosystem/grpc-gateway/blob/master/protoc-gen-grpc-gateway/main.go#L60 ff.). You can pass an arbitrary number of comma separate ```param=value``` which are interpreted like commandline parameters would be. No special magic needed.

Works a treat.

hacst:
With regards to this PR in general: I would be great to receive some guidance on what would be needed to make this ready to land. E.g. is the current implementation approach acceptable and what kind of tests and documentation is required for people to be comfortable landing this. Thanks.

protoc-gen-grpc-gateway/descriptor/grpc_api_configuration.go
```diff
@@ -0,0 +1,94 @@
...
+	"google.golang.org/genproto/googleapis/api/annotations"
+)
+
+// GrpcAPIService represents a stripped down version of google.api.Service .
+// Compare to https://github.com/googleapis/googleapis/blob/master/google/api/service.proto
+//
+// Note that for the purposes of the gateway generator we only consider a subset of all
+// available features google supports in their service descriptions and hence we do not
+// bother having the full service protobuf which has a lot of other dependencies. Thanks
+// to backwards compatibility this should be relatively safe.
```
achew22: "Thanks to backwards compatibility" doesn't seem to follow from the rest of the comment. It seems like "thanks to structural typing" may be a better way to phrase that?

protoc-gen-grpc-gateway/descriptor/grpc_api_configuration.go
```diff
@@ -0,0 +1,94 @@
...
+func (m *GrpcAPIService) String() string { return proto.CompactTextString(m) }
+
+func loadGrpcAPIServiceFromYAML(yamlFileContents []byte, yamlSourceLogName string) (*GrpcAPIService, error) {
+	jsonContents, err := yaml.YAMLToJSON(yamlFileContents)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to convert gRPC API Configuration from YAML in '%v' to JSON: %v", yamlSourceLogName, err)
+	}
+
+	// As our GrpcAPIService is incomplete accept unkown fields.
+	unmarshaler := jsonpb.Unmarshaler{
```
achew22: This is the same configuration as the default global unmarshaller. Is it important to keep them in sync or are you just using the unmarshaller to load the data into a proto shaped object for use during generation?

achew22: Can you confirm for me that this is only used during the protoc stage and not during runtime? If it is used during runtime, can you describe the circumstances when it is used?

protoc-gen-grpc-gateway/descriptor/grpc_api_configuration.go
```diff
@@ -0,0 +1,94 @@
...
+	"strings"
+
+	"github.com/ghodss/yaml"
+	"github.com/golang/protobuf/jsonpb"
+	"github.com/golang/protobuf/proto"
+	"google.golang.org/genproto/googleapis/api/annotations"
+)
+
+// GrpcAPIService represents a stripped down version of google.api.Service .
+// Compare to https://github.com/googleapis/googleapis/blob/master/google/api/service.proto
```
achew22: Could you add a comment in here that the file you're avoiding importing has 23 import statements and that is why it's being avoided as well as a note that says "If a significant subset of these (>50%) are being reproduced in this file, swap it out for the generated version.

protoc-gen-grpc-gateway/descriptor/grpc_api_configuration.go
```diff
@@ -0,0 +1,94 @@
...
+	"io/ioutil"
+	"strings"
+
+	"github.com/ghodss/yaml"
+	"github.com/golang/protobuf/jsonpb"
+	"github.com/golang/protobuf/proto"
+	"google.golang.org/genproto/googleapis/api/annotations"
+)
+
+// GrpcAPIService represents a stripped down version of google.api.Service .
```
achew22: Can you move this into its own file?

protoc-gen-grpc-gateway/descriptor/grpc_api_configuration.go
```diff
@@ -0,0 +1,94 @@
...
+
+	serviceConfiguration := GrpcAPIService{}
+	if err := unmarshaler.Unmarshal(bytes.NewReader(jsonContents), &serviceConfiguration); err != nil {
+		return nil, fmt.Errorf("Failed to parse gRPC API Configuration from YAML in '%v': %v", yamlSourceLogName, err)
+	}
+
+	return &serviceConfiguration, nil
+}
+
+func registerGrpcAPIService(registry *Registry, service *GrpcAPIService, sourceLogName string) error {
```
achew22: registerHTTPRulesFromGrpcAPIService

protoc-gen-grpc-gateway/descriptor/grpc_api_configuration_test.go
```diff
@@ -0,0 +1,164 @@
...
+	if second.GetSelector() != "some.other.service" {
+		t.Errorf("some.other.service has unexpected selector '%v'", second.GetSelector())
+	}
+	if second.GetDelete() != "foo" {
+		t.Errorf("some.other.service has unexpected delete '%v'", second.GetDelete())
+	}
+	if len(second.GetAdditionalBindings()) != 0 {
+		t.Errorf("some.other.service has %v additional bindings when it should not have any. Got: %v", len(second.GetAdditionalBindings()), second.GetAdditionalBindings())
+	}
+}
```
achew22: Solid tests. Good work!

protoc-gen-grpc-gateway/descriptor/grpc_api_configuration.go
```diff
@@ -0,0 +1,94 @@
...
+		return nil
+	}
+
+	for _, rule := range service.HTTP.GetRules() {
+		selector := "." + strings.Trim(rule.GetSelector(), " ")
+		if strings.ContainsAny(selector, "*, ") {
+			return fmt.Errorf("Selector '%v' in %v must specify a single service method without wildcards", rule.GetSelector(), sourceLogName)
+		}
+
+		registry.externalHTTPRules[selector] = append(registry.externalHTTPRules[selector], rule)
```
achew22: This feels like something that should be added to the public interface of the registry rather than mucking around in the internals of the registry object

protoc-gen-grpc-gateway/descriptor/registry.go
```diff
@@ -205,6 +210,11 @@ func (r *Registry) LookupFile(name string) (*File, error) {
 	return f, nil
 }
 
+// LookupExternalHTTPRules looks up external http rules by fully qualified service method name
+func (r *Registry) LookupExternalHTTPRules(qualifiedMethodName string) []*annotations.HttpRule {
```
achew22: You have a getter for external http rules, I think you should add a setter and use it instead of direct manipulation.

protoc-gen-grpc-gateway/descriptor/services.go
```diff
@@ -49,7 +53,7 @@ func (r *Registry) loadServices(file *File) error {
 	return nil
 }
 
-func (r *Registry) newMethod(svc *Service, md *descriptor.MethodDescriptorProto, opts *options.HttpRule) (*Method, error) {
+func (r *Registry) newMethod(svc *Service, md *descriptor.MethodDescriptorProto, optss []*options.HttpRule) (*Method, error) {
```
achew22: s/optss/options/

Let's be explicit that we are talking about plural of opts instead of it being a typo of opts.

protoc-gen-grpc-gateway/gengateway/generator.go
```diff
@@ -103,7 +102,7 @@ func (g *generator) generate(file *descriptor.File) (string, error) {
 	for _, svc := range file.Services {
 		for _, m := range svc.Methods {
 			pkg := m.RequestType.File.GoPkg
-			if m.Options == nil || !proto.HasExtension(m.Options, options.E_Http) ||
+			if len(m.Bindings) == 0 ||
```
achew22: Can you help me understand why this is preferential? Could you also add a test demonstrating why this is preferred

hacst: This is from 3356bb1 :

> Previously the generator generator checked whether the http extension
> was used to decide whether imports were required for a method. This
> patch makes it check whether a method has bindings instead as those
> are what code is actually generated for.

Basically the previous check skipped adding imports when the protobuf file wasn't using the Http annotation. That check is done on data protoc passed from the proto file. Since the whole point of this PR is to no longer have the Http annotations in the proto this check had to be amended. Now it checks whether we created bindings for the method (it should have a binding for each endpoint). This will happen based on annotation or the yaml.

What you see without this change is that the generated .gw.go file might be missing imports (e.g. if the method uses ```google.protobuf.Duration``` for which an import is required). In that case it just won't compile.

I will see if I can do a test for that.

ensonic:
Any update?

hacst:
@ensonic  Unfortunately I have been pretty busy lately. Will try to finish this up this weekend.

hacst:
@achew22 Ok. I think I addressed the remaining review feedback and rebased onto current master.

For the e2e test I derived a "unannotated" echo service from the echo service example. I added in a import of the duration well known type for good measure. This should break without my ```len(m.Bindings) == 0``` change. Imho this should exercise the new codepaths sufficiently. The functionality expressable by the HTTP rule annotations is directly re-used by the PR so I see no reason to test it again in this in this E2E.

I'm not 100% happy with my README.md changes. I tried to keep it as short as possible but it still is quite a chunk of text. Maybe putting the information in a separate .md file would be better?

In the makefile I did a bit of cargo-culting from the echo service (I've never seen a ExamplepbSimpleMessage.go or git_push.sh in my tree) so there might be some clutter that ought to be removed.

Also my generation run insisted on changing some occurances of "context" imports to "golang.org/x/net/context" which I understand is a legacy construct. I was using go 1.10 and a 3.0.0 protoc so I am not quite sure whether that is something that needs fixing. It does compile.

My bazel file edits were blind as I do not have a bazel environment to use. The bazel build on travis fails due to what looks like my new dependency missing which makes a certain amount of sense. I have no idea how to get a 3rd party go dependency into a bazel build so help is welcome there.

Also the travis builds with disabled context seem to fail. This looks like a makefile issue to me but I'm not quite sure. protoc line looks like I expect it to be but the error output is strange. Like a space slipped in somewhere where it shouldn't. If you have an idea I'm all ears. Otherwise I will have to investigate later.

hacst:
@achew22 I think I figured out the Bazel stuff. Also fixed the Makefile so all travis builds are green now. See my previous comment for the overall state of the PR but from my POV this is getting close to being finished. Would be great if you could take a look.

README.md
```diff
@@ -238,6 +238,48 @@ But patch is welcome.
 * HTTP headers that start with 'Grpc-Metadata-' are mapped to gRPC metadata (prefixed with `grpcgateway-`)
 * While configurable, the default {un,}marshaling uses [jsonpb](https://godoc.org/github.com/golang/protobuf/jsonpb) with `OrigName: true`.
 
+# Generating for unannotated protos
```
achew22: What would you think about making this a page in the `/docs` section and linking to it?

achew22:
> Also my generation run insisted on changing some occurances of "context" imports to "golang.org/x/net/context" which I understand is a legacy construct. I was using go 1.10 and a 3.0.0 protoc so I am not quite sure whether that is something that needs fixing. It does compile.

Totally cool. 1.10 `go fix` replaces everything and 1.9 has type aliasing. I'm going to do a big sweeping change in a few weeks as soon as swagger-codegen 2.4 comes out and probably do a 2.0 of grpc-gateway so no worries at all.

> My bazel file edits were blind as I do not have a bazel environment to use. The bazel build on travis fails due to what looks like my new dependency missing which makes a certain amount of sense. I have no idea how to get a 3rd party go dependency into a bazel build so help is welcome there.

Sorry 'bout that. It looks like you figured it out, but that is obviously a sub-par experience. How can we make this easier for contributors in the future?

examples/clients/unannotatedecho/docs/ExamplepbUnannotatedSimpleMessage.md
```diff
@@ -0,0 +1,12 @@
+# ExamplepbUnannotatedSimpleMessage
```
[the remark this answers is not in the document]
hacst: ok

README.md
hacst: Sgtm. Much better than littering the landing page.

achew22:
I think this is good to go! Thanks so much for all your work in here. Could you rebase this, fix the merge conflicts, and run `go fix`, then I will merge it?

PS: Thanks for looking into the Jekyl stuff. I'm also totally new to it so I don't know much much at all about how to fix things. I'm glad you took a look at it.

hacst:
@achew22 This was more involved than I expected. Quite a few relevant changes since the last rebase a few days ago. I hope it is ok that I merged instead of rebased but I really didn't feel rebasing all those commits was a sane thing to do now that the examplepb folder got moved.

Added new commits for:
* Adding a delete method to the unattended echo example as one was added to the stock echo one
* Fix the documentation to correctly point to the new location of the examplepb folder. Seems like that was forgotten before

Hoping for no more breaking changes till we manage to land this ;)

hacst:
Awesome. Thanks a lot for the guidance along the way. Having someone so responsive handle PRs is really awesome for a first time contributor to a project.

hacst:
@yugui No worries. Pretty much my own fault for letting this sit so long ;)

ensonic:
Now from reading the code, this seems to not do what I think it would :/

We have a proto with e.g.
```proto
service AbcService {
  rpc CreateAbc(Abc) returns (Abc) {
    option (google.api.http) = {
      post: "/apis/abc/v1/abcs"
      body: "*"
    };
  }
```
in the grpc service config, we have:
```yaml
type: google.api.Service
config_version: 3
name: abc.endpoints.${GCP_PROJECT_ID}.cloud.goog
endpoints:
- name:abc.endpoints.${GCP_PROJECT_ID}.cloud.goog
  allow_cors: true
title: Abc gRPC API
apis:
- name: domain.abc.v1alpha1.AbcService
  version: 1.0.0
authentication:
  providers:
  - id: xxx_auth
    ...
  rules:
  - selector: "*"
    allow_without_credential: false
    requirements:
    - provider_id: xxx_auth
usage:
  rules:
  - selector: "*"
    allow_unregistered_calls: true

```
Now with or without using `grpc_api_configuration`, I get the same resulting open-api.json. I was expecting:
1.) the 'title' from the grpc yaml to go to '.info.description'.
2.) security settings being configured

Should any of those work?

hacst:
@ensonic No. As described in https://grpc-ecosystem.github.io/grpc-gateway/docs/grpcapiconfiguration.html this PR only implements the HTTP rule part of the specification. The goal was not having to annotate the .proto file for basic use with grpc-gateway. I have not looked into how other parts of gRPC API Configurations might be mapped to grpc-gateway though I am sure there are some that could have meaningful mappings.

PR #522, opened 2018-01-15
Converts the first letter of service name to uppercase

### LINKED ISSUE #484: bug: uncapitalized service name causes runtime error unknown function in service.pb.gw.go
see title.

Steps to reproduce:

1. define a service with a name that has a lowercase first letter in your `service.proto`
1. generate the `service.pb.go` (notice that the generated code includes `func NewNameClient` -- it is capitalized `Name`)
1. generate the `service.pb.gw.go` (notice that the generated code attempts to call `NewnameClient` -- uncapitalized `name` which leads to the runtime error)

this looks similar to #300 and looks like it could be fixed in the same way.

the style guide, https://developers.google.com/protocol-buffers/docs/style, says you should capitalize service names, but seems it does not say it requires capitalized. So i think this is a valid bug

thurt:
fixes #484


thurt:
@achew22 im not sure what file you are referring to as the goldens file? 

I added a new service with a camelCase name and regenerated the `pb.go`, `pb.gw.go`, and `pb.swagger.json`.

You can see the the correction provided by the first commit changes the `pb.gw.go` here https://github.com/grpc-ecosystem/grpc-gateway/pull/522/commits/8f93341dfb8998833da93ecd0706fee29d9fac2c#diff-715cf1c1ab38e7007f330d8663ae2d5fR1080

achew22:
That's exactly what I was asking you to do. Sorry for not being more clear!

thurt:
cool! :+1:

PR #524, opened 2018-01-15
adds opts.Tags to operationObject.Tags

thurt:
this is a continuation of work in #145, which left as a TODO to check
remaining options fields for the operationObject during swagger generation.
This PR causes protoc-gen-swagger to check if opts.Tags exists and
overwrite the default value if it does exist.


achew22:
I just looked at https://github.com/grpc-ecosystem/grpc-gateway/pull/526 and I think I'm just going to merge that since it includes tags support as well. Thanks so much for sending in the PR!

PR #526, opened 2018-01-16
Add description, summary and tags fields in operationObject (swagger)

devnull-:
Hello all,

Small PR to add `description`, `summary` and `tags` fields in [operationObject](https://swagger.io/specification/#operationObject) (swagger -- #145). `Tags` field was already fixed by @thurt in https://github.com/grpc-ecosystem/grpc-gateway/pull/524 but not merge (yet).

Devnull-

### DIFF
--- a/examples/examplepb/a_bit_of_everything.proto
+++ b/examples/examplepb/a_bit_of_everything.proto
@@ -187,6 +187,7 @@ service ABitOfEverythingService {
 		};
 		option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
             description: "Description Echo";
+            summary: "Summary: Echo rpc";
 			external_docs: {
 				url: "https://github.com/grpc-ecosystem/grpc-gateway";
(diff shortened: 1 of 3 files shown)


devnull-:
I signed it!

devnull-:
@googlebot : Add alternate email to validate the CLA.

ivucica:
I like that this is in, but I think it worth adding a few caveats. For instance, I'd argue it's much more readable if `summary` and `description` are written as a proto comment. Special options seem a sane choice only if you have Swagger-specific text that doesn't apply to the gRPC service, or that should otherwise not end up in generated code.

`tags` field may be useful if same call appears in multiple services, or if the default of service name is not sufficiently human readable (e.g. developer wants spaces in the Swagger 'tag' name). But I would still suggest just using the default: the proto service name. Maybe we should have a proto option specifying a 'human readable' name for a service, instead?

Either way, this is neat, even if easy to misuse. Thumbs up 👍

Issue #320, opened 2017-03-01
Swagger-UI query parameters for enum types are sent as strings


johanbrandhorst:
When using a `GET` endpoint and an enum type in the query parameter, the swagger definition shows a selection of the string representation of the enum as input options, but using that representation returns an error:

```
enum MyEnum {
    DEFAULT = 0
    OTHER = 1
}

message Request {
    MyEnum value = 1;
}

service MyService {
    rpc Method(Request) returns (google.protobug.Empty) {
        option (google.api.http) = {
            get: "/api/method"
        };
    };
}
```

```
{
  "error": "strconv.ParseInt: parsing \"DEFAULT\": invalid syntax",
  "code": 3
}
```

The swagger definition needs to use the numeric representation of the enum for it to work.

tmc:
@johanbrandhorst thanks for the issue report. My preference would be to allow either representation.

johanbrandhorst:
The error is coming from https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/convert.go#L39.

johanbrandhorst:
Here's a stack trace (edited):
```
net/http.(*http2serverConn).runHandler.func1(0xc42031c010, 0xc420471faf, 0xc42034a000)
        /usr/lib/go/src/net/http/h2_bundle.go:4592 +0x190
panic(0x9b54c0, 0xc42012c8a0)
        /usr/lib/go/src/runtime/panic.go:489 +0x2cf
myrepo/vendor/github.com/grpc-ecosystem/grpc-gateway/runtime.Int32(0xc4202f45e0, 0x9, 0x0, 0x0, 0x0)
        /go/src/myrepo/vendor/github.com/grpc-ecosystem/grpc-gateway/runtime/convert.go:41 +0x63
reflect.Value.call(0x9d3b60, 0xab4658, 0x13, 0xa8c513, 0x4, 0xc420471420, 0x1, 0x1, 0xc42012c870, 0x9b54c0, ...)
        /usr/lib/go/src/reflect/value.go:434 +0x91f
reflect.Value.Call(0x9d3b60, 0xab4658, 0x13, 0xc420471420, 0x1, 0x1, 0xa076fc, 0x76, 0x30)
        /usr/lib/go/src/reflect/value.go:302 +0xa4
[...]

PR #529, opened 2018-01-18
runtime: support FieldMask as query param

glerchundi:
In order to support [`FieldMask`](https://developers.google.com/protocol-buffers/docs/reference/csharp/class/google/protobuf/well-known-types/field-mask) in query params, [as Google proposes](https://cloud.google.com/apis/design/standard_methods#update), this should be treated as another known type.

/cc @achew22 @tmc 

As a reference:
https://cloud.google.com/iam/reference/rest/v1/organizations.roles/patch
https://github.com/googleapis/googleapis/blob/f704d14a7224a140bca5cc26835fae471eaf7281/google/iam/admin/v1/iam.proto#L157
https://github.com/googleapis/googleapis/blob/f704d14a7224a140bca5cc26835fae471eaf7281/google/iam/admin/v1/iam.proto#L614-L626

### DIFF
--- a/runtime/query.go
+++ b/runtime/query.go
@@ -149,11 +149,13 @@ func populateRepeatedField(f reflect.Value, values []string, props *proto.Proper
 }
 
 func populateField(f reflect.Value, value string, props *proto.Properties) error {
-	// Handle well known type
+	i := f.Addr().Interface()
+
+	// Handle protobuf well known types
 	type wkt interface {
 		XXX_WellKnownType() string
 	}
-	if wkt, ok := f.Addr().Interface().(wkt); ok {
+	if wkt, ok := i.(wkt); ok {
 		switch wkt.XXX_WellKnownType() {
 		case "Timestamp":
 			if value == "null" {
@@ -214,6 +216,20 @@ func populateField(f reflect.Value, value string, props *proto.Properties) error
 				return fmt.Errorf("bad BytesValue: %s", value)
 			}
 			f.Field(0).SetBytes(bytesVal)
+			return nil
+		}
+	}
+
+	// Handle google well known types
+	if gwkt, ok := i.(proto.Message); ok {
+		switch proto.MessageName(gwkt) {
+		case "google.protobuf.FieldMask":
+			p := f.Field(0)
+			for _, v := range strings.Split(value, ",") {
+				if v != "" {
+					p.Set(reflect.Append(p, reflect.ValueOf(v)))
+				}
+			}
 			return nil
 		}
 	}
--- a/runtime/query_test.go
+++ b/runtime/query_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/golang/protobuf/ptypes/wrappers"
 	"github.com/grpc-ecosystem/grpc-gateway/runtime"
 	"github.com/grpc-ecosystem/grpc-gateway/utilities"
+	"google.golang.org/genproto/protobuf/field_mask"
 )
 
 func TestPopulateParameters(t *testing.T) {
@@ -24,6 +25,9 @@ func TestPopulateParameters(t *testing.T) {
 	if err != nil {
 		t.Fatalf("Couldn't setup timestamp in Protobuf format: %v", err)
 	}
+
+	fieldmaskStr := "float_value,double_value"
+	fieldmaskPb := &field_mask.FieldMask{[]string{"float_value", "double_value"}}
 
 	for _, spec := range []struct {
 		values  url.Values
@@ -46,6 +50,7 @@ func TestPopulateParameters(t *testing.T) {
 				"enum_value":            {"1"},
 				"repeated_enum":         {"1", "2", "0"},
 				"timestamp_value":       {timeStr},
+				"fieldmask_value":       {fieldmaskStr},
 				"wrapper_float_value":   {"1.5"},
 				"wrapper_double_value":  {"2.5"},
 				"wrapper_int64_value":   {"-1"},
@@ -71,6 +76,7 @@ func TestPopulateParameters(t *testing.T) {
 				EnumValue:          EnumValue_Y,
 				RepeatedEnum:       []EnumValue{EnumValue_Y, EnumValue_Z, EnumValue_X},
 				TimestampValue:     timePb,
+				FieldMaskValue:     fieldmaskPb,
 				WrapperFloatValue:  &wrappers.FloatValue{1.5},
 				WrapperDoubleValue: &wrappers.DoubleValue{2.5},
 				WrapperInt64Value:  &wrappers.Int64Value{-1},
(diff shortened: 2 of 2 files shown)


achew22:
Can you add a test to verify this functionality? I'd be very happy to merge once that happens. Thanks!

glerchundi:
Yeah, will do!

glerchundi:
ready to review @achew22!

achew22:
I love being reviewed!

On a more serious note. LGTM, let's ship it. Thanks for contributing!

johanbrandhorst:
I've tried playing around with this a bit, and I can't seem to get this to work the way I think it should work. Given [the example](https://cloud.google.com/apis/design/standard_methods#update) and the code in this PR, I'm assuming the field mask should be provided as a query parameter like so:

```bash
curl -k -X PATCH "https://localhost:11000/api/v1/user/1?update_mask=role" -H  "accept: application/json" -H  "Content-Type: application/json" -d "{\"role\": \"ADMIN\"}"
```

Given the following definitions:

```protobuf
rpc UpdateUser(UpdateUserRequest) returns (User) {
    option (google.api.http) = {
        patch: "/api/v1/user/{user.id}"
        body: "user"
    };
}

message UpdateUserRequest {
    User user = 1;
    google.protobuf.FieldMask update_mask = 2;
}
```

Unfortunately, this kind of definition doesn't even survive generation; the generated code tries to decode into a `**User`:

```go
if err := marshaler.NewDecoder(req.Body).Decode(&protoReq.User); err != nil {
```

Is there an example of this implemented a used successfully in the wild? The query test "example" is basically useless as far as I'm concerned.

glerchundi:
I've tested with this proto:

```
 // Update an existing user.
    rpc UpdateUser(UpdateUserRequest) returns (User) {
        option (google.api.http) = {
            patch: "/api/v1/{user.name=installations/*/users/*}"
            body: "user"
        };
    }

[...]

message UpdateUserRequest {
    // The user resource which replaces the resource on the server.
    User user = 2;

    // The update mask applies to the resource. For the `FieldMask` definition,
    // see https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#fieldmask
    google.protobuf.FieldMask update_mask = 3;
}

[...]

message User {
    // Relative resource name.
    string name = 1;

    // Display name.
    string display_name = 3;

    // Description.
    string description = 4;
}
```

And this worked:
```
curl -H "Content-Type: application/json" --request PATCH -d@./update_user.json localhost:8080/api/v1/installations/1/users/user2?update_mask=display_name
```

Where `update_user.json` is:
```
{
  "display_name": "user3" 
}
```

So i firmly believe that something is not working on your side. Are you using the standard go generator or something like gogo? Give us a full example and perhaps we can go further.

johanbrandhorst:
@glerchundi thanks for that, sounds like my use case _should_ work, I will do some more testing and report back.

johanbrandhorst:
Update: my initial problems were _indeed_ because I was using GoGo Protobuf. My intentions are actually to investigate whether this works with gogoprotobuf, but I wanted to get it working first, so thanks for pointing me in the right direction. I will see what would need to be done on the gogo/protobuf side, but I expect this will flat out not work with gogo/protobuf until an alternative implementation for https://github.com/grpc-ecosystem/grpc-gateway/blob/58f78b988bc393694cef62b92c5cde77e4742ff5/runtime/query.go#L275 can be made.

Out of interest; since the update_mask specifies fields as strings, what reflect magic do you use to figure out which fields of the struct to update?

glerchundi:
Good investigation.

> Update: my initial problems were indeed because I was using GoGo Protobuf. My intentions are actually to investigate whether this works with gogoprotobuf, but I wanted to get it working first, so thanks for pointing me in the right direction. I will see what would need to be done on the gogo/protobuf side, but I expect this will flat out not work with gogo/protobuf until an alternative implementation for

But I don't know why it's not working, `proto.MessageName` should work because gogo's fieldmask registers itself under the same name that the official fieldmak does:

official: https://github.com/google/go-genproto/blob/master/protobuf/field_mask/field_mask.pb.go#L249
gogo: https://github.com/gogo/protobuf/blob/master/types/field_mask.pb.go#L244

Also the structs have the same parameters (just one) with the same signature which means the reflection should work as well. And `proto.MessageName` seems to behave in the same way...

> Out of interest; since the update_mask specifies fields as strings, what reflect magic do you use to figure out which fields of the struct to update?

We're pre-generate everything, updateable fields which we then map to database fields, no reflection magic at all. With something like this:

```
message User {
string display_name = 3  [metadata.updateable = true];
}
```

Would be interesting to see your progress.

johanbrandhorst:
@glerchundi I'm afraid you're making an incorrect assumption about how `proto.MessageName` works. `golang/proto.MessageName` only returns messages that are _registered_ with the `golang/protobuf` type registry. `gogo/protobuf` is forced to maintain its own type registry, which leads to this problem as well as [many](https://github.com/grpc-ecosystem/grpc-gateway/issues/320) [other](https://github.com/gogo/protobuf/issues/178) issues. The links you're providing ignore the fact that one registeres with `gogo/protobuf.RegisterType` and one with `golang/protobuf.RegisterType`.

If you want more information about this kind of thing, you're welcome to read my blogpost on `gogo/protobuf` compatibility issues: https://jbrandhorst.com/post/gogoproto/.

There are basically two solutions to this problem:

1. Make all `gogo/protobuf` files register with both type registries.
    This is undesireable because `gogo/protobuf` shouldn't have to import `golang/protobuf` (for what I think should be obvious reasons).
1. We implement some way in the gRPC-Gateway that accesses both registries, presumably with `golang/protobuf` taking precedence.
2.1 Alternatively, allow users of the gRPC-Gateway runtime to select a `proto.MessageName` resolver.

Anyway, as I mentioned, there seems to be another problem here with the `gogo/protobuf` generated message, so I will look into that before suggesting any changes to the gRPC-Gateway.

glerchundi:
> @glerchundi I'm afraid you're making an incorrect assumption about how proto.MessageName works. golang/proto.MessageName only returns messages that are registered with the golang/protobuf type registry. gogo/protobuf is forced to maintain its own type registry, which leads to this problem as well as many other issues. The links you're providing ignores the fact that one registeres with gogo/protobuf.RegisterType and one with golang/protobuf.RegisterType.

Right.

> 2.1 Alternatively, allow users of the gRPC-Gateway runtime to select a proto.MessageName resolver.

What about `gogo` implementing `XXX_MessageName` method in its types? Would solve without messing the runtime with resolvers here and there.

https://github.com/golang/protobuf/blob/master/proto/properties.go#L846-L855

johanbrandhorst:
Hm, that is a fair point, I didn't know about that method. It would solve the problem with `proto.MessageName` specifically, but if that name was used to look up any types in `proto.MessageType`, it would still error.

glerchundi:
> Hm, that is a fair point, I didn't know about that method. It would solve the problem with proto.MessageName specifically, but if that name was used to look up any types in proto.MessageType, it would still error.

Better than nothing. I think the problem of how to interoperate is much bigger than these concrete methods. I still think that in this case using interfaces is the way to go and would fix your current issue, what do you think if start by fixing this in `gogo` types and file/comment in another issue(s) for the interoperability between libraries (i'm quite sure there are already issues in this regard).

johanbrandhorst:
I agree, and as I said there are other issues here with gogo/protobuf which I will investigate at the same time. Feel free to follow the discussion in https://github.com/gogo/grpc-example/issues/9 if you wish.

PR #535, opened 2018-01-30
Support for map type in query string

### LINKED ISSUE #316: Support map type in query string
Is it possible to support map[string]string in the query parameters?

adamstruck:
Adds support for the map type in query strings.  Filter strings take the form: `?map_field[key]=value`.

Only maps with primitive keys and values are supported. 

Closes #316

### DIFF
--- a/runtime/query.go
+++ b/runtime/query.go
@@ -96,7 +96,7 @@ func populateFieldValueFromPath(msg proto.Message, fieldPath []string, values []
 			continue
 		case reflect.Map:
 			if !isLast {
-				return fmt.Errorf("unexpected map field in %s", strings.Join(fieldPath, "."))
+				return fmt.Errorf("unexpected nested field %s in %s", fieldPath[i+1], strings.Join(fieldPath[:i+1], "."))
 			}
 			return populateMapField(f, values, props)
 		default:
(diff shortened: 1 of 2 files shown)


runtime/query.go
```diff
@@ -84,6 +94,11 @@ func populateFieldValueFromPath(msg proto.Message, fieldPath []string, values []
 		case reflect.Struct:
 			m = f
 			continue
+		case reflect.Map:
+			if !isLast {
+				return fmt.Errorf("unexpected map field in %s", strings.Join(fieldPath, "."))
```
achew22: Can you elaborate in this error message. As I understand it the conern here is that you could make the querystring `?foo[bar]` which doesn't have a value. Could you add a remark about what was expected and why this is an error condition?

runtime/query_test.go
```diff
@@ -60,6 +59,21 @@ func TestPopulateParameters(t *testing.T) {
...
+				"map_value6[key]":       {"3"},
+				"map_value7[3]":         {"value"},
+				"map_value8[key]":       {"4"},
+				"map_value9[4]":         {"value"},
+				"map_value10[key]":      {"1.5"},
+				"map_value11[1.5]":      {"value"},
+				"map_value12[key]":      {"2.5"},
+				"map_value13[2.5]":      {"value"},
+				"map_value14[key]":      {"true"},
+				"map_value15[true]":     {"value"},
```
achew22: Can you add some examples where you populate multiple map values at the same time?

runtime/query.go
adamstruck: This is to handle the case where an intermediate field in `fieldPath` refers to a map.  

i.e.  `obj.nested.map_field.foo` where `map_field` is a map.

achew22: Could you enhance the error message to reflect that intent?

adamstruck:
Thanks for the feedback! I think my latest commit addresses your comments.

runtime/query.go
adamstruck: sure no problem.

achew22:
Thanks for the new feature and thanks for contributing!

Issue #536, opened 2018-02-02
import_path option not working as intended


shynie:
This option doesn't work the same way it does with protoc-gen-go. The latter has option `M` that allows us to associate a set of *.proto files with package given in import_path, so all of the generated *.go files will have the same package name.

>Mfoo/bar.proto=quux/shme - declares that foo/bar.proto is associated with Go package quux/shme. This is subject to the import_prefix parameter.

protoc-gen-grpc-gateway should support that option too. Now importh_path only changes the path but doesn't affect package names.

achew22:
@shynie, thanks for the bug report!

Unfortunately, I'm not at all familiar with the argument you're attempting to use. Do you think you could point to some documentation that has more information? Additionally it would be extremely helpful if you could write a test that demonstrates this failure. With that it will be a lot easier to figure out what next steps should be.

PR #507, opened 2017-12-20
Add support for --Import_path

achew22:
Fixes #443

PR #537, opened 2018-02-04
Use importPath to set package name rather than package path.

rwlincoln:
In #536 @shynie correctly observes that the `import_path` flag was not implemented correctly in #507. Although it is called `import_path`, in `golang/protobuf` the flag is used to set the package identity name.

> import_path=foo/bar - used as the package if no input files declare go_package. If it contains slashes, everything up to the rightmost slash is ignored.

Currently it is used to set the package path:

https://github.com/grpc-ecosystem/grpc-gateway/blob/6658b3a4fd7017117532b54b5c6bf2fd2207ffc2/protoc-gen-grpc-gateway/descriptor/registry.go#L251

This PR moves the use of the `importPath` field to the `packageIdentityName` function.

Note that eventually the "Want to support" section of the README should be updated:

https://github.com/grpc-ecosystem/grpc-gateway/blob/6658b3a4fd7017117532b54b5c6bf2fd2207ffc2/README.md#want-to-support

### DIFF
--- a/protoc-gen-grpc-gateway/descriptor/registry.go
+++ b/protoc-gen-grpc-gateway/descriptor/registry.go
@@ -301,12 +301,6 @@ func (r *Registry) defaultGoPackageName(f *descriptor.FileDescriptorProto) strin
 // protoc-gen-grpc-gateway rejects CodeGenerationRequests which contains more than one packages
 // as protoc-gen-go does.
 func (r *Registry) packageIdentityName(f *descriptor.FileDescriptorProto) string {
-	if p := r.importPath; len(p) != 0 {
-		if i := strings.LastIndex(p, "/"); i >= 0 {
-			p = p[i+1:]
-		}
-		return p
-	}
 	if f.Options != nil && f.Options.GoPackage != nil {
 		gopkg := f.Options.GetGoPackage()
 		idx := strings.LastIndex(gopkg, "/")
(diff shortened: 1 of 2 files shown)


achew22:
Does this PR mean that importPath should be working as expected? Could you add a test by modifying the makefile to invoke protoc with this parameter set to a value you know to work and output to a different directory so we could have a more e2e test?

rwlincoln:
This PR makes `import_path` work in the same way that it does in `golang/protobuf`.

> Could you add a test by modifying the makefile to invoke protoc with this parameter set to a value you know to work and output to a different directory so we could have a more e2e test?

Unfortunately, all of the example `.proto` files include the `go_package` option. `import_path` is only used "as the package if no input files declare go_package". I will try to add some tests to `descriptor/registry_test.go`.

rwlincoln:
I have added tests for the `SetImportPath` method on the `registry` type. They test:

- that `importPath` is only used as the package if no input files declare `go_package` and
- if `importPath` contains slashes, everything up to the rightmost slash is ignored.

This is in accordance with the documentation:

https://github.com/grpc-ecosystem/grpc-gateway/blob/0a810034e45df6b5a119b0cd3a56150c459e3f9a/protoc-gen-grpc-gateway/descriptor/registry.go#L218

achew22:
Looking through this and some other documentation on import paths, I think this is the right way to do it but I wouldn't be surprised if someone came along and said "this totally breaks my use case" when we merge. If something goes when this is merged, do you mind if I @mention you in the issue and see if we can find a good resolution for everyone?

PR #541, opened 2018-02-07
Omit well-known type definitions from swagger output

alexleigh:
In #496 the swagger output was improved by adding support for more well-known types to the `wktSchemas` map. However, these types are not added to a parallel list of cases in `renderMessagesAsDefinition`, causing the swagger output to contain type definitions for types that will never be emitted by the schema generator. This change fixes the issue by omitting these well-known types from the generated swagger schemas.


achew22:
@alexleigh Do you think you could add an example proto in `examples/examplepb` that exhibits this new behavior (and modify the makefile to have it be built automatically?)

achew22:
@alexleigh, Looks great to me! Thanks so much for your contribution :smile:. I look forward to working more with you in the future!

PR #543, opened 2018-02-08
Set the defaultMarshaler to not use original pb

Kuqd:
Set the defaultMarshaler to not use original pb.

As per requested in : 

https://github.com/grpc-ecosystem/grpc-gateway/pull/540

This should fullfil requirements.


Kuqd:
I signed it!

achew22:
@Kuqd, as I'm sure you've noticed by the email from Travis.ci, we have some end to end tests that will need to be updated along with this. Potentially even merged at the same time as the change #540. If you want to go about making those changes it would be HUGELY appreciated.

Kuqd:
Ok let me close this for now

PR #545, opened 2018-02-09
Eliminate unnecessary linebreaks in swagger paths

lukasmalkmus:
This is a quick fix for a problem I encountered using `protoc-gen-swagger`. Imagine having an rpc definition like this:

```protobuf
service Users {
  // Search for users
  //
  // Searches for users that match specific filter criteria and returns a
  // (possibly paginated) list of user resources on success.
  rpc SearchUsers(SearchUsersRequest) returns (SearchUsersResponse) {
    option (google.api.http) = { post: "/v1/users:search" body: "*" };
  }
}
```

I prefer line breaks in my comments at a certain length (~80). But `protoc-gen-swagger` will also incorporate these line breaks into the `description` field.

bad

This fix performs a simple string replace on single line breaks, replacing them with a single whitespace:

good

Further comments? @tmc @achew22


achew22:
First, thanks for the contribution! Do you know what does this do on multi paragraph comments? e.g.

```
service Users {
  // Search for users
  //
  // Searches for users that match specific filter criteria and returns a
  // (possibly paginated) list of user resources on success.
  //
  // Additional contextual information is included here.
  rpc SearchUsers(SearchUsersRequest) returns (SearchUsersResponse) {
    option (google.api.http) = { post: "/v1/users:search" body: "*" };
  }
}
```

I suspect that would all get condensed into a single line with this patch. How do you think we could enhance it to provide a line break in the single paragraph case as well as in the multi-paragraph case?

Also, as an aside, we depend on "golden file" tests in order to ensure there haven't been any regressions in the creation of the files. Since this change modifies the output format for the .swagger.json files, you will need to run `make examples` and commit them before the CI tests will pass.

achew22:
Oh, an additional concern I have is in the case where someone wants to make a list in the comment:

```
service Users {
  // Search for users
  //
  // Returns a list of users and allows you to search by:
  // * Family Name
  // * First Name
  // * Height
  // * User ID
  //
  // More information that would be useful to a person calling this API.
  rpc SearchUsers(SearchUsersRequest) returns (SearchUsersResponse) {
    option (google.api.http) = { post: "/v1/users:search" body: "*" };
  }
}
```

Can you help me understand how your change would impact a usecase like that as well?

lukasmalkmus:
> Do you know what does this do on multi paragraph comments?

Yes. It makes a line break and thus a new paragraph. Take a look at the source code. The original code separated the paragraphs using `strings.Split("comment", "\n\n")`. I just hooked in after that and remove any `\n` from that paragraph only, not the whole comment. Furthermore I noticed, that the official Swagger Editor does not render a double line break `\n\n`, which is used [here](https://github.com/lukasmalkmus/grpc-gateway/blob/3cbc72a99ea77a79bf36ac57f3838a7375fccc3a/protoc-gen-swagger/genswagger/template.go#L849), to separate the paragraphs. It uses a single line break anyways.

> Also, as an aside, we depend on "golden file" tests in order to ensure there haven't been any regressions in the creation of the files. Since this change modifies the output format for the .swagger.json files, you will need to run make examples and commit them before the CI tests will pass.

Ups... my first commit build just fine. But sure, will do :)

> Oh, an additional concern I have is in the case where someone wants to make a list in the comment [...] Can you help me understand how your change would impact a usecase like that as well?

Good point 👍 will investigate and report back.

lukasmalkmus:
As you already guessed: My simple solution didn't work with bullet lists. So I had to improve my code:

Stuff like this...

```protobuf
service Users {
  // Search for users
  //
  // Returns a list of users and
  //  allows you to search by:
  // * Family Name
  //    is your last name
  // * First Name
  // * Height
  // * User ID
  //
  // More information that would be useful to a person calling this API.
  rpc SearchUsers(SearchUsersRequest) returns (SearchUsersResponse) {
    option (google.api.http) = { post: "/v1/users:search" body: "*" };
  }
}
```

... now renders nicely:

better

achew22:
Text formatting in these cases is a generally hard problem. If you look at the [CommonMark test suite](http://spec.commonmark.org/0.28/spec.json) for example, you can see some of the strange things that people do when writing text files.

The reason I'm hesitant to do this formatting automatically is because this changes the code from interpreting the literal text they wrote to interpreting the intent they had behind their words, whitespace, and punctuation. As another example, what if someone has a numbered list? They will run into similar problems. It almost seems to me the better solution is to tell the swagger UI to run a markdown formatter on the input text from the comment. I agree that it is mildly unsightly to have widow words and strange line breaks but it does come at a cost -- the ability to fully express your formatting.

Maybe we can look around for other options. Do you know how other tools handle the swagger generation WRT line breaks? Maybe we can learn from them?

lukasmalkmus:
To be honest, I had the same thoughts while writing the code.

> I agree that it is mildly unsightly to have widow words and strange line breaks but it does come at a cost -- the ability to fully express your formatting.
👍 

> Do you know how other tools handle the swagger generation WRT line breaks? Maybe we can learn from them?
I will dig around a little bit. And the issue I explained occurred in the swagger editor, I have not yet tested it in ReDoc.

achew22:
The more I think about this the less confident I am that it is the correct course of action. I very much appreciate you spending the time to contribute. Please stick around, there are a lot of fun projects embedded in the gateway that would be fun to work on if you're looking for a challenge. Feel free to email me (address in commit messages) if you're looking for something to spice up your life!

PR #547, opened 2018-02-11
Add security fields support to protoc-gen-swagger

### LINKED ISSUE #428: Adding authentication definitions to generated swagger files
My API requires authentication for all endpoints. This is defined in a `config.yml` file that I feed into Google Cloud Endpoints. 

I generated a swagger spec from my service's .proto file, and then generated a javascript library for use on the web. 

Do you have any plans to support swagger's [API keys](https://swagger.io/docs/specification/authentication/api-keys/) definitions, or should we monkey-patch them into generated libraries ourselves (if that's even an optionI)?

ivucica:
This adds security fields to the openapiv2 options proto. It adds them to the `protoc-gen-swagger`'s `types.go` as well. Finally, it adds support for dumping them out into the JSON file.

For testing purposes, these fields are (hopefully comprehensively!) added to `a_bit_of_everything.proto`.

This fixes #428.


examples/examplepb/a_bit_of_everything.proto
```diff
@@ -32,6 +32,68 @@ option (grpc.gateway.protoc_gen_swagger.options.openapiv2_swagger) = {
...
+	  scope: {
+	    key: "admin";
+	    value: "Grants read and write access to administrative information";
+	  }
+	}
+      }
+    }
+  }
+  security: {
+    security_requirement: {
```
achew22: I don't see an entry for bearer (in authorization header). Which one would that map to? Reading the docs I guess an apiKey with the header "authorization"?

PR #553, opened 2018-02-21
error details: add @type key by switching to any.Any

### LINKED ISSUE #551: proto: no slice oenc for *reflect.rtype = []*reflect.rtype
Hey there

With the introduction of error details in JSON responses, #515, we've also gotten us an extra line of output for every JSON error encoding process:

```
proto: no slice oenc for *reflect.rtype = []*reflect.rtype
```

...irregardless of whether this error contains error details or not. Looking back at [the PR's CI run](https://travis-ci.org/grpc-ecosystem/grpc-gateway/builds/327578972), it has been in there [right from the start](https://travis-ci.org/grpc-ecosystem/grpc-gateway/jobs/327578983#L916).

The line comes from [`proto.(*Properies).setEncAndDec`](https://github.com/golang/protobuf/blob/bbd03ef6da3a115852eaf24c8a1c46aeb39aa175/proto/properties.go#L342), called via [`jsonpb.jsonProperties`](https://github.com/golang/protobuf/blob/bbd03ef6da3a115852eaf24c8a1c46aeb39aa175/jsonpb/jsonpb.go#L1027).

However, since the tests pass, it seems like whatever is happening there wouldn't have to happen?

Have you got any idea how to get rid of that?

(While looking into this, I've also started wondering why grpc-gateway's error details don't have `@type` fields, but that's a different matter. But would _you_ think they should have them?)

srenatus:
This is arguably more correct than what was introduced in #515. So, any
error details no have a "@type" field indicating their underlying
proto.Message's type, following the Cloud API docs<sup>[1]</sup>.

[1]: https://cloud.google.com/apis/design/errors#http_mapping

Note that this does **not** help with #551. The error is still present.

### DIFF
--- a/examples/integration_test.go
+++ b/examples/integration_test.go
@@ -761,6 +761,13 @@ func TestErrorWithDetails(t *testing.T) {
 	if got, want := ok, true; got != want {
 		t.Fatalf("msg.Details[0] got type: %T, want %T", msg.Details[0], map[string]interface{}{})
 	}
+	typ, ok := details["@type"].(string)
+	if got, want := ok, true; got != want {
+		t.Fatalf("msg.Details[0][\"@type\"] got type: %T, want %T", typ, "")
+	}
+	if got, want := details["@type"], "type.googleapis.com/google.rpc.DebugInfo"; got != want {
+		t.Errorf("msg.Details[\"@type\"] = %q; want %q", got, want)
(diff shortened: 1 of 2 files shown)


achew22:
Good catch! Thanks for taking the time to fix it for everyone

PR #557, opened 2018-02-24
fix noenc error by fixing Details error field

### LINKED ISSUE #551: proto: no slice oenc for *reflect.rtype = []*reflect.rtype

srenatus:
As advised by @itizir in the discussion of #551.

However, I've gone with the minimal change to make the error go away, I
haven't introduced a proper, generated protobuf definition.

Also, I've noticed our error body message there looks suspiciously like
[Status](https://github.com/google/go-genproto/blob/2b5a72b8730b0b/googleapis/rpc/status/status.pb.go#L83-L93), except for the field name "Error" vs "Message".

Fixes #551.

### DIFF
--- a/runtime/errors.go
+++ b/runtime/errors.go
@@ -65,9 +65,9 @@ var (
 )
 
 type errorBody struct {
-	Error   string    `protobuf:"bytes,1,name=error" json:"error"`
-	Code    int32     `protobuf:"varint,2,name=code" json:"code"`
-	Details []any.Any `protobuf:"bytes,3,name=details" json:"details"`
+	Error   string     `protobuf:"bytes,1,name=error" json:"error"`
+	Code    int32      `protobuf:"varint,2,name=code" json:"code"`
+	Details []*any.Any `protobuf:"bytes,3,rep,name=details" json:"details,omitempty"`
 }
 
 // Make this also conform to proto.Message for builtin JSONPb Marshaler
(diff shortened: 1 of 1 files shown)


achew22:
> However, I've gone with the minimal change to make the error go away, I
> haven't introduced a proper, generated protobuf definition.

In that PR we decided to not include the complete proto. Turns out it is comically large and dramatically increases the complexity of vendoring making it kind of a logistical nightmare.

Thanks for sending in the PR!

srenatus:
@achew22 thanks for sharing your rationale :)

PR #560, opened 2018-02-26
Use the gRPC status Message when possible in stream errors

johanbrandhorst:
The default behaviour when returning errors in a stream is to use the error code from the gRPC status, but still use the error string instead of the gRPC status message. This PR changes it to use the gRPC status message.


johanbrandhorst:
This change + more is included in https://github.com/grpc-ecosystem/grpc-gateway/pull/561, so review that before accepting this.

johanbrandhorst:
Closed in favour of #561

PR #561, opened 2018-02-27
Add details to stream error response

johanbrandhorst:
Add the gRPC status details to the streamed response.

This is built on top of https://github.com/grpc-ecosystem/grpc-gateway/pull/560, but I can close that one if we want to merge these together instead.


johanbrandhorst:
Probably easiest to merge this and close #560 if we want both of these changes.

runtime/handler.go
```diff
@@ -168,17 +168,15 @@ func handleForwardResponseStreamError(wroteHeader bool, marshaler Marshaler, w h
 
 func streamChunk(result proto.Message, err error) map[string]proto.Message {
 	if err != nil {
-		grpcCode := codes.Unknown
-		if s, ok := status.FromError(err); ok {
-			grpcCode = s.Code()
-		}
-		httpCode := HTTPStatusFromCode(grpcCode)
+		s, _ := status.FromError(err)
```
srenatus: Hmm. I know it should never happen, but seeing the second return ignored is weird, too. Any thoughts about something like
```go
s, ok := status.FromError(err)
if !ok {
  panic("err is no status")
}
```

johanbrandhorst: I wanted to make use of the new status handling in [gRPC-Go 1.10](https://github.com/grpc/grpc-go/commit/d50734d1d6ca477a72646f3022216ec39639f4cd) which is equivalent to the old handling, but I accept gRPC-Gateway can't enforce that users have the latest gRPC version. I will change it back.

srenatus: Oh that's a lovely function. Thanks for the pointer.

johanbrandhorst:
I hate to bump this but I think it's a pretty straightforward, backwards compatible change with a new test case and I'd _really_ like to have this merged ASAP. Please advise if I can speed things up somehow.

achew22:
I just reviewed this and I think it is good to go. Marking as approved. Can you please close any other PRs that this obviates and then I'll merge it (probably tomorrow morning/early afternoon US time)

johanbrandhorst:
Closed #560 and #564.

achew22:
Today was a disaster. Sorry for the slow merge. Merging now

johanbrandhorst:
Any chance we could have a new release with this? 1.4.0?

PR #563, opened 2018-02-28
Add 3rd party rpc protos in order to have access to status and error

rvegas:


rvegas:
I signed it!

rvegas:
Any love from maintainers?

achew22:
Thanks so much for your contribution!

Can you help me understand a little more about this change? Why we should package these inside the grpc-gateway package instead of having clients vendor them in their own projects? Why is this necessary now so many years into the project when it wasn't needed before?

It's great to have more minds thinking about these problems so we, together, can trend toward better solutions

rvegas:
Hey @achew22 thanks for replying and reviewing.

The main objective here is to allow for the REST API definition to be able to handle possible error definitions without having to load the whole `googleapis` proto package entirely, just as it's already done by including the `http` and the `annotations` protos.

For example, it would allow you to have this in your protos:

```go
// This resource represents a long-running operation that is the result of a
// network API call.
message Operation {
  string name = 1;
  google.protobuf.Any metadata = 2;
  bool done = 3;

  // The operation result, which can be either an `error` or a valid `response`.
  // If `done` == `false`, neither `error` nor `response` is set.
  // If `done` == `true`, exactly one of `error` or `response` is set.
  oneof result {
    // The error result of the operation in case of failure or cancellation.
    google.rpc.Status error = 4;

    // The normal response of the operation in case of success.  
    google.protobuf.Any response = 5;
  }
}
```

What do you think? This allows for far better expressed APIs and its common practice in many standards such as JSONAPI to define errors fully.

Issue #538, opened 2018-02-04
proxy: inconsistent error messages between unary and stream response


thurt:
i am using `google.golang.org/grpc/status` to create errors sent in server responses. However, i noticed that my client is receiving different messages depending on whether the rpc has a unary or stream response.

example body response from a unary rpc: 
```
{
    "error":"Unauthenticated",
    "code":16
}
```

example body response from a stream rpc:
```
{
    "error":{
        "grpc_code":16,
        "http_code":401,
        "message":"rpc error: code = Unauthenticated desc = Unauthenticated",
        "http_status":"Unauthorized"
    }
}
```

I expected the response body for an error to be the same message shape for both unary and stream rpcs.

thurt:
it looks like the unary rpc error message is defined here (inside `DefaultHTTPError`) https://github.com/grpc-ecosystem/grpc-gateway/blob/7542c7659a3cd03f02153d9b9da331fb80c311e4/runtime/errors.go#L93

and the stream rpc error message defined here (inside `streamChunk`)
https://github.com/grpc-ecosystem/grpc-gateway/blob/8db8c1ac6f97d8ee1b7a0d7198c39f687e19a937/runtime/handler.go#L176

johanbrandhorst:
I agree that this is a problem. The reason for the streaming error being the way it is seems to be to make it fit into the `for loop` that writes to the `ResponseWriter`: https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/handler.go#L63.

I think actually the solution to this problem is to move away from our custom error types and just use the `status.Proto()`, which is perfectly marshallable and consistent. It does _not_ contain any HTTP translations of course, but tbh we're already translating the gRPC Code to a HTTP code in the actual response code, so I don't think it's necessary. It would mean we can remove the `stream_chunk.proto` type altogether.

The `{"error":<actual error>}` object key probably has to stay, to differentiate from normal messages (which use the `{"result": <actual message>` key.

@tmc @achew22 thoughts on this? Would obviously break backwards compatibility in terms of the message returned from errors, but some change here is obviously needed.

johanbrandhorst:
Making this change makes `DefaultHTTPProtoErrorHandler` and `DefaultHTTPError` _almost_ identical. In general, I think this area could do with an overhaul, and will necessitate a 2.0 release.

achew22:
[...]

PR #564, opened 2018-02-28
Add method for custom formatting of stream errors.

johanbrandhorst:
Alternative to discussion in https://github.com/grpc-ecosystem/grpc-gateway/issues/538#issuecomment-369139421. This would allow a user to get errors as marshalled `status.Proto()` types with the use of both `WithProtoStreamErrorFormatter` and `WithProtoErrorHandler`.

This is pretty experimental so I'm happy to make any changes.

### DIFF
--- a/runtime/handler.go
+++ b/runtime/handler.go
@@ -42,7 +42,7 @@ func ForwardResponseStream(ctx context.Context, mux *ServeMux, marshaler Marshal
 	if d, ok := marshaler.(Delimited); ok {
 		delimiter = d.Delimiter()
 	} else {
-	    delimiter = []byte("\n")
+		delimiter = []byte("\n")
 	}
 
 	var wroteHeader bool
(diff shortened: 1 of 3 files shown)


abursavich:
I would much prefer the handler's signature to be more in the style of (Default)HTTPError, so that a client can write the stream error however they wish. In my case, I would like to construct my own error status proto directly and avoid my marshalers having to handle maps.

johanbrandhorst:
Could you come up with a practical alternative? The stream is already established at this point, so a handler function may require rewriting of much of the logic.

abursavich:
Here's my first pass. It may need some work: https://github.com/abursavich/grpc-gateway/commit/47b315fad04ce2e68cc41661d2938bcbd643cd41

johanbrandhorst:
@abursavich Thanks for that! I think one problem with your approach is that we expose a new function type, used by the user, which has an empty interface in the signature. That's pretty poor interface design. It also rewrites quite a bit of the logic, and it seems to add an error case if the response is nil, which may break backwards compatibility? I think your suggestion is a step in the right direction, as I would also like more control over the writing of mesages, but maybe the configuration should be more fine grained? One option for formatting the error and one for formatting a successful reply? I'm not sure.

abursavich:
FWIW, the error on nil response was already there, I just lifted it out of streamChunk (which I deleted): https://github.com/grpc-ecosystem/grpc-gateway/blob/cde2f8f5bef06abdf9d0557f284f072f1cfa9f56/runtime/handler.go#L185-L187

I don't like the empty interface much either, but I was matching the existing interface for Marshaler: https://github.com/grpc-ecosystem/grpc-gateway/blob/e4b8a938efae14de11fd97311e873e989896348c/runtime/marshaler.go#L7-L20

johanbrandhorst:
My heart is not really in this change after merging #561, so I will close this to let someone else provide a better solution to the specific problem of customizing stream error returns.

Issue #423, opened 2017-06-27
Add Bytes and Enum handlers to `proto3ConvertFuncs`


gdbelvin:
This is an issue to track the TODO on this [line in types.go](https://github.com/grpc-ecosystem/grpc-gateway/blob/acebe0f9ff5993e130b141ee60e83e592839ca22/protoc-gen-grpc-gateway/descriptor/types.go#L290).

```go
var (
	proto3ConvertFuncs = map[descriptor.FieldDescriptorProto_Type]string{
		descriptor.FieldDescriptorProto_TYPE_STRING:  "runtime.String",
		// FieldDescriptorProto_TYPE_GROUP
		// FieldDescriptorProto_TYPE_MESSAGE
		// FieldDescriptorProto_TYPE_BYTES
		// TODO(yugui) Handle bytes
		descriptor.FieldDescriptorProto_TYPE_UINT32: "runtime.Uint32",
		// FieldDescriptorProto_TYPE_ENUM
		// TODO(yugui) Handle Enum
	}

)
```

lucasvo:
Bytes was implemented in: https://github.com/grpc-ecosystem/grpc-gateway/pull/489

lucasvo:
@loderunner, thanks for implementing the bytes support.

I've been wondering if anyone ever considered encoding bytes as a hex string instead of base64. I'd love to make this configurable. I've been digging through the different runtime.convert methods to see if there is any support for configuration in runtime.convert or in the ConvertFuncExpr (https://github.com/grpc-ecosystem/grpc-gateway/blob/e651a16acb9bd6028d18bce8115de407cb2aa5f4/protoc-gen-grpc-gateway/descriptor/types.go#L202) but it looks like there isn't really anything that would do something similar. 

I might go ahead and propose an implementation but I've wanted to open the discussion here first to hear what you think.

loderunner:
@lucasvo I went for base64 because it is the [standard mapping](https://developers.google.com/protocol-buffers/docs/proto3#json) for protobuf `bytes`. `bytes` were already handled for methods other than `GET` that supported sending JSON in the body. It did indeed use the standard base64 mapping from protobuf. I only added support for `bytes` fields in as URL variables.

[...]

PR #565, opened 2018-03-01
[]byte in query now uses base64.StdEncoding

lucasvo:
This fixes the issue with different base64 string encoding in query params and json representations in the request body discussed in #423

@loderunner Care to review?


lucasvo:
Signed the CLA.

loderunner:
Looks fine to me!

The burden will be on the client to correctly escape the query parameters to avoid `/`, `+` and `=`.  I wonder if this is worth documenting somewhere. Perhaps someone from the grpc-gateway project can help us here?

achew22:
Do either of you know if it is possible to use the same base64 en/decoder that is provided by golang/protobuf? How does the deserialization work when they are processing JSONified protos? Could you point me to the code where they do that?

loderunner:
golang/protobuf uses the stdlib json encoding/decoding. The go structs generated from the proto file are annotated for JSON encoding.

Encoding byte slices to base64 is part of the stdlib:

> Array and slice values encode as JSON arrays, except that []byte encodes as a base64-encoded string, and a nil slice encodes as the null JSON value.

[https://github.com/golang/go/blob/master/src/encoding/json/encode.go#L54-L56](Source)

You can find the source for the encoding [here](https://github.com/golang/go/blob/master/src/encoding/json/encode.go#L705-L725) and the decoding [here](https://github.com/golang/go/blob/master/src/encoding/json/decode.go#L967-L973)

I think there is a bit of [custom Marshal/Unmarshal code](https://github.com/golang/protobuf/blob/master/proto/lib.go#L417-L442) for handling enums and their string representations, but nothing for byte slices.

lucasvo:
@loderunner did the heavy lifting, he deserves the credit! Thanks for helping out with this.

Issue #431, opened 2017-07-11
RESOURCE_EXHAUSTED -> 503


jhspaybar:
Right now it appears that HTTPStatusFromCode converts a RESOURCE_EXHAUSTED into a 403.  I'd like to convert this to a 503, and it looks like I need to do this by calling withProtoErrorHandler and passing in my mux?  I will be more or less copy/pasting the DefaultProtoErrorHandler method just swapping out the status mapping function and before I went this route, wanted to see if there was a better way to pull this off.

Thanks.

achew22:
Unfortunately I'm not sure of a better way without changing things in grpc-gateway. I think a case could be made that 403 isn't the right code for this (that is forbidden, right?) and I'd be open to changing it to something different. Could you send a PR?

PR #569, opened 2018-03-06
runtime: return 503 not 403 with ResourceExhausted.

hexfusion:
ResourceExhausted returning a 403 is confusing. 503 makes a lot more sense here.

Per: https://github.com/grpc/grpc-go/blob/8a8ac82f1f7f141a47430dd8be7aa90b92141603/codes/codes.go#L73

Fixes https://github.com/grpc-ecosystem/grpc-gateway/issues/431

ref: https://groups.google.com/forum/#!topic/etcd-dev/nSK92tS__c8


achew22:
I believe this is the correct behavior and therefore is a bug fix, not a breaking change. Merging -- thank you!

PR #546, opened 2018-02-10
Massive breaking changes in preparation for 2.0

achew22:
This PR turned out to be an absolutely gigantic... thing. Since I'm making breaking changes why not break all the things.

Things I broke:

  * Emitting defaults
  * Now uses the jsonpb marshaller by default
  * Upgrade to swagger 2.4

Obviously this can't go in until swagger 2.4 is released, but we have a backlog of significant breaking changes that would improve the quality of the service dramatically. My plan is to cut one last 1.X branch release then merge merge merge and cut a 2.0

# TODO

 * [ ] I need to add an upgrading section to the docs

PR #570, opened 2018-03-09
support one-ofs in body

jhump:
I was curious if this was something that might be accepted as a patch, and (if so) if I'm on the right track.

I must admit, I don't really understand how the `FieldPath` stuff really works. I mean, I *think* I see how it's supposed to work. But it looks like actually using nested elements (e.g. `foo.bar.baz`) would always result in nil dereference panic since the actual request's fields have zero values. So, in this example, field `protoReq.Foo`, presumably a pointer to message struct, would be nil, and thus unmarshaling code would traverse a nil pointer trying to compute `&protoReq.Foo.Bar.Baz` (even if it were just one level of nesting: `&protoReq.Foo.Bar`).

So, in this patch, I didn't really bother examining the entire path for oneofs and only look at the first field. Does that seem sane?

BTW, the whole reason I was exploring this patch is to support an API idiom like so:

```proto
message IndividualId {
  oneof id {
    string uid = 1;
    int64 user_id = 2;
  }
}

service Users {
  rpc QueryIndividual (IndividualId) returns (IndividualDetails) {
    option (google.api.http) = {
      get: "/users/v1/individual/{uid}"
      additional_bindings {
        get: "/users/v1/user/{user_id}"
      }
    };
  }
}
```


achew22:
Hey @jhump, sorry for the very slow response on this. Let me give you an update on the state of affairs around here. I have [a PR](https://github.com/grpc-ecosystem/grpc-gateway/pull/546) that I'm going to be submitting soon that is the beginning of the breaking change window for a 2.0 release. As such I've been focusing on getting that done and not really tending to the PR queue. Sorry about that.

This looks like a really good PR to have included and I don't mind at all that it doesn't completely solve the recursive form of the problem -- the perfect is the enemy of the good after all. One thing I noticed is that you've hit a bug in our CI system. I am curious. Do you think you could rebase this PR after [that commit](https://github.com/grpc-ecosystem/grpc-gateway/commit/3189c0c0c638bd69bf223a1411ba20c4427647d5) and see if the CI system blows up?

I would have expected this to fail since you changed the template which should make the examples directory (which is a dir of goldens) not match. To fix this, you can run `make examples` to update the goldens and then the PR should pass CI.

Thanks so much for your hard work!

achew22:
Friendliest of pings

protoc-gen-grpc-gateway/gengateway/template.go
```diff
@@ -207,6 +207,7 @@ var (
 	var protoReq {{.Method.RequestType.GoType .Method.Service.File.GoPkg.Path}}
 	var metadata runtime.ServerMetadata
 {{if .Body}}
+	{{.Body.AssignableExprPrep "protoReq"}}
 	if err := marshaler.NewDecoder(req.Body).Decode(&{{.Body.AssignableExpr "protoReq"}}); err != nil && err != io.EOF  {
```
jhump: This is what generated bad code. The `AssignableExpr` would have emitted the prep statements, which would get incorrectly inlined in the call to `Decode`. By splitting out the prep statements, it now generates working code.

protoc-gen-grpc-gateway/descriptor/types.go
```diff
@@ -261,13 +267,46 @@ func (p FieldPath) IsNestedProto3() bool {
...
+		components = components + "." + c.ValueExpr()
+	}
+
+	return components
+}
+
+// AssignableExprPrep returns preparation statements for an assignable expression to assign a value
+// to the target field. The go expression of the method request object is "msgExpr". This is only
+// needed for field paths that contain oneofs. Otherwise, an empty string is returned.
+func (p FieldPath) AssignableExprPrep(msgExpr string) string {
```
jhump: I essentially forked `AssignableExpr`, with this method preserving just the `preparations` and the other only emitting the last line, the actual field ref.

jhump:
@achew22, PTAL

jhump:
The build breakage was bad timing - I pushed just after protobuf merged their dev branch to master. I will rebase this PR after #636 lands, to see if I can get the tests to go green.

achew22:
@yugui Have you seen this?

jhump:
@achew22 or @yugui: ping?

tmc:
My only comment is that I'd like a test case that demonstrates oneof population from body.

apeletz512:
@jhump @achew22  @yugui @tmc Any progress here?

johanbrandhorst:
Bump, need a rebase and a test case please.

jhump:
@johanbrandhorst, sorry, I've been busy, and we've since re-formulated our proto to not actually need this feature. So, if this seems like a valuable contribution, someone else will have to take it over.

PR #575, opened 2018-03-16
fix racy access of err variable

jhump:
Without this fix, the `handleSend` function is closing over `err` from the outer function. This trips up the Go race detector because `handleSend` gets invoked in a goroutine and there is no synchronization between that goroutine and a subsequent write to `err` after a call to `stream.Header`.

This appears to be an accidental closure -- `handleSend` doesn't actually need to share the `err` variable so can instead just create its own `err` local variable. Which clears it up.

### DIFF
--- a/examples/examplepb/flow_combination.pb.gw.go
+++ b/examples/examplepb/flow_combination.pb.gw.go
@@ -105,7 +105,7 @@ func request_FlowCombination_StreamEmptyStream_0(ctx context.Context, marshaler 
 	dec := marshaler.NewDecoder(req.Body)
 	handleSend := func() error {
 		var protoReq EmptyProto
-		err = dec.Decode(&protoReq)
+		err := dec.Decode(&protoReq)
 		if err == io.EOF {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


jhump:
Ping @achew22. This is causing failing tests for some of our stuff because every bidi stream invocation tickles the race detector. We've temporarily forked grpc-gateway. This seems like a non-controversial fix, so we were hoping to not have to maintain a fork for very long.

achew22:
Can you help me understand why changing the template doesn't change any of the generated code in the examples section? Could you run `make examples` and include any changes that are in there?

jhump:
@achew22: I didn't even realize that there was an examples folder.

I ran `make generate`, but it produced a *ton* of differences. I am guessing I have a different version of `swagger-codegen` than was used to produce the existing files. (I actually just installed it on OS X: `brew install swagger-codegen`. So maybe the existing swagger-generated files need to be freshened?)

I only `git add`'ed the ones relevant to my change.

achew22:
Could you rebase this change? I've fixed the test that was not running that I was expecting to fail in here.

Thanks!

jhump:
> Could you rebase this change?

@achew, done

PR #580, opened 2018-03-20
Changed to use more appropriate http status code for ResourceExhausted

eleniums:
HTTPStatusFromCode should be mapping codes.ResourceExhausted to http.StatusRequestEntityTooLarge (https://httpstatuses.com/413) as it is the more appropriate HTTP status. The current mapping of http.StatusServiceUnavailable seems to indicate a temporary problem with the server, when in reality it is an issue with the client sending a request that is too large.


johanbrandhorst:
Arguably the correct mapping is that defined by the [Google protobuf error code proto file](https://github.com/googleapis/googleapis/blob/aa14ef295443c528d6c02aa4ddb5998a09223c69/google/rpc/code.proto#L108); `429 Too Many Requests`.

achew22:
+1, it should be 429. Could you also add a link to https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto in the function comment?

johanbrandhorst:
Looks like there are more codes that should be chaged, FailedPrecondition should be `400 Bad Request`, not `412 PreconditionFailed`. This is probably a good time to review all the codes.

eleniums:
My initial reason for bringing this up was that the instances in grpc-go where ResourceExhausted is currently being used are related to message size, not the resource actually being exhausted. I thought that `413 PAYLOAD TOO LARGE` would be a more appropriate mapping, given the current implementation.

That being said, in order to be consistent, it does make sense to map the codes based on https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto. I'll make the necessary adjustments and push a commit soon.

PR #586, opened 2018-04-06
Support multiple metadata annotators

dmacthedestroyer:
Modify the `ServeMux` struct so multiple `runtime.WithMetadata` options can be provided, and all their results combined into a single metadata object


dmacthedestroyer:
I signed it!

achew22:
Thanks so much for your contribution!

PR #594, opened 2018-04-14
Added Bazel support

f0rmiga:
Sorry for the size of the PR. I would be up for a peer review if wanted.

Almost every target was generated by Gazelle.
Gazelle tried to index `//google/api/...` because of the third party protos being used. To solve that I excluded the `third_party` directory from Gazelle and created the `repositories.bzl` that at this point:
    - Downloads the `googleapis` repository.
    - Adds a `BUILD.bazel` file under `//google/api/` with targets for the protos.

Gazelle still tries to remove `"@com_github_googleapis_googleapis//google/api:api_proto"` and `"@com_github_googleapis_googleapis//google/api:go_default_library"` in favour of `//google/api/...`, which is wrong.

To check what Gazelle reports in check mode, run `bazel run //:gazelle`.

The `third_party` directory is not being used by any Bazel target at all. If the grpc-gateway team decides to move completely to Bazel, this directory can be removed.

Every `.pb.go` and `.pb.gw.go` can also be removed if this repo moves completely to Bazel. In order to have the swagger generator to work with Bazel, I would need to write a specific rule for that.

To compile everything:
```
$ bazel build //...
```

To test everything:
```
$ bazel test //...
```


achew22:
I'm not entirely sure how to express how supportive of this pr I am. You literally freed my weekend plans up because this was what I was going to do.

It's 1am for me so I'll be reviewing tomorrow but let me say thank you preemptively

f0rmiga:
@achew22 You are very welcome. It was a fun Friday evening task. 🤓

protoc-gen-grpc-gateway/BUILD.bazel
```diff
@@ -0,0 +1,40 @@
...
+)
+
+go_binary(
+    name = "protoc-gen-grpc-gateway",
+    embed = [":go_default_library"],
+)
+
+go_proto_compiler(
+    name = "go_gen_grpc_gateway",
+    options = ["logtostderr=true"],
```
f0rmiga: I'm not sure if we should use the option `logtostderr=true`. Bazel seems to succeed if the code gen fails.

achew22: Do you have any theories on that? I would expect `protoc` to propagate the nonzero exit code through.

f0rmiga: It works properly. I was having another issue that made me think this was the cause. Not true.

f0rmiga:
@achew22 I screwed up rebasing master into my branch :).

achew22:
I think you could have force pushed your change on top of the branch and it would have been okay, but feel free to send another PR :smile:

f0rmiga:
@achew22 Yes. I didn't force push. Damn. http://classicprogrammerpaintings.com/post/142586036029/junior-programmer-learns-git-rebase

PR #599, opened 2018-04-14
Added Bazel support

f0rmiga:
Sorry for the size of the PR. I would be up for a peer review if wanted.

Almost every target was generated by Gazelle.
Gazelle tried to index `//google/api/...` because of the third party protos being used. To solve that I excluded the `third_party` directory from Gazelle and created the `repositories.bzl` that at this point:
    - Downloads the `googleapis` repository.
    - Adds a `BUILD.bazel` file under `//google/api/` with targets for the protos.

Gazelle still tries to remove `"@com_github_googleapis_googleapis//google/api:api_proto"` and `"@com_github_googleapis_googleapis//google/api:go_default_library"` in favour of `//google/api/...`, which is wrong.

To check what Gazelle reports in check mode, run `bazel run //:gazelle_diff`.

The `third_party` directory is not being used by any Bazel target at all. If the grpc-gateway team decides to move completely to Bazel, this directory can be removed.

Every `.pb.go` and `.pb.gw.go` can also be removed if this repo moves completely to Bazel. In order to have the swagger generator to work with Bazel, I would need to write a specific rule for that.

To compile everything:
```
$ bazel build //...
```

To test everything:
```
$ bazel test //...
```


f0rmiga:
@achew22

achew22:
Would you be willing to edit `.travis.yaml` to include bazel in the CI process. You can review https://github.com/bazelbuild/rules_go#how-do-i-run-bazel-on-travis-ci for inspiration

f0rmiga:
Sure. I have Bazel setup for a couple projects of mine in Travis CI.

achew22:
To be a bit more explicit, could you add to the matrix something like `USE_BAZEL=true` and then put an early termination if statement in?

runtime/internal/BUILD.bazel
```diff
@@ -0,0 +1,23 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_library")
+load("@io_bazel_rules_go//proto:def.bzl", "go_proto_library")
+
+package(default_visibility = ["//visibility:public"])
```
yugui: Could you make the visibility of the package consistent with the one by go compiler?
i.e. it looks to be better to limit the default visibility to `//runtime:__subpackages__`.

BTW, as the original author of Gazelle, I'm sorry -- it should have been done by gazelle. ;-)

f0rmiga: You are excused hehe. Thanks for the Gazelle, it saves tons of time daily for me.
I also made changes to other package visibilities as I saw fit.

examples/examplepb/BUILD.bazel
```diff
@@ -0,0 +1,51 @@
...
+        "echo_service.proto",
+        "flow_combination.proto",
+        "stream.proto",
+        "wrappers.proto",
+    ],
+    deps = [
+        "//examples/sub:sub_proto",
+        "//examples/sub2:sub2_proto",
+        "//protoc-gen-swagger/options:options_proto",
+        "@com_github_googleapis_googleapis//google/api:api_proto",
```
f0rmiga: @yugui Any idea on how to have Gazelle to respect that?

f0rmiga:
@achew22 What do you think of the CI config? We don't need to test Bazel 3 times. Not sure how to solve that... any ideas?

examples/examplepb/BUILD.bazel
yugui: Sorry, I have no idea. Let me merge this PR as it is.

yugui:
It still makes sense to build with different versions of go in Bazel. But either is OK for me.

LGTM except for the redundant 3 builds.

f0rmiga:
@yugui We can do that in another PR. What if we merge the way it is right now and I focus on improving the CI build?

f0rmiga:
I actually need to rebase and add changes merged to master.

f0rmiga:
@yugui @achew22 CI done.

achew22:
@f0rmiga, thanks so much for your contribution! I look forward to adding more support for Bazel going forward :fireworks:

f0rmiga:
@achew22 Honestly, I tried to integrate it. I think we would need to create a rule similar to `go_proto_library`. The problem of trying to use it for swagger is that this rule expects `.go` files to come out of the compiler so it can compile and output a single `.a` file. I can give it a shot later this week.

f0rmiga:
@achew22 For reference: https://github.com/bazelbuild/rules_go/blob/c29b08c0b81a1bfdc358c6bd3a59c068165bfefa/proto/def.bzl#L105

achew22:
Fortunately there are other projects who have implemented it in a less go specific way. For example, https://github.com/bazelbuild/rules_closure/blob/dbb96841cc0a5fb2664c37822803b06dab20c7d1/closure/protobuf/closure_proto_library.bzl

f0rmiga:
Nice! Good stuff.

PR #600, opened 2018-04-14
Pulled out parseReq func into a generic package + tests

f0rmiga:
Pulled out the duplicated parseReq function into a more generic package and added tests to it.

### DIFF
--- a/codegenerator/parse_req_test.go
+++ b/codegenerator/parse_req_test.go
@@ -1,0 +1,66 @@
+package codegenerator_test
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"reflect"
+	"strings"
+	"testing"
+
+	"github.com/golang/protobuf/proto"
+	plugin "github.com/golang/protobuf/protoc-gen-go/plugin"
+	"github.com/grpc-ecosystem/grpc-gateway/codegenerator"
+)
+
+var parseReqTests = []struct {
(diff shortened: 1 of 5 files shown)


codegenerator/parse_req_test.go
```diff
@@ -0,0 +1,66 @@
...
+			}
+			if err == nil && !reflect.DeepEqual(*out, *tt.out) {
+				t.Errorf("got %v, want %v", *out, *tt.out)
+			}
+		})
+	}
+}
+
+func mustGetReader(pb proto.Message) io.Reader {
+	b, _ := proto.Marshal(pb)
```
achew22: I would prefer to panic on error instead of silencing in a must function

protoc-gen-grpc-gateway/main.go
```diff
@@ -29,33 +28,19 @@ var (
...
-
 func main() {
 	flag.Parse()
 	defer glog.Flush()
 
 	reg := descriptor.NewRegistry()
 
 	glog.V(1).Info("Processing code generator request")
-	req, err := parseReq(os.Stdin)
+	glog.V(1).Info("Parsing code generator request")
```
achew22: Duplicate logging

achew22: Can you remove the one on line 37

PR #613, opened 2018-04-24
feat(bazel): Add rule for generating .swagger.json files

mrmeku:
Wraps protoc and protoc-gen-swagger in a bazel rule so that other rules can depend upon the output


achew22:
@f0rmiga I don't think you have to write the swagger rule any more :wink:, great work @mrmeku!

f0rmiga:
That is neat!

achew22:
@f0rmiga, I mentioned you on the PR because I saw that message but I only got notified a few hours ago. Sorry if there was any overlapping work here.

f0rmiga:
@achew22 No worries! No overlapping work here. Nice work @mrmeku!

mrmeku:
@f0rmiga I think I'm done! Let me know if you have anything you can see that you'd like cleaned up

third_party/googleapis/BUILD.bazel
```diff
@@ -0,0 +1,6 @@
+package(default_visibility = ["//visibility:public"])
+
+filegroup(
+    name = "googleapis",
+    srcs = glob(["**/*.proto"])
```
f0rmiga: [nit] Missing `,`.

mrmeku: TY

protoc-gen-swagger/defs.bzl
```diff
@@ -0,0 +1,79 @@
...
+            default = "@com_google_protobuf//:well_known_protos",
+            allow_files = True,
+        ),
+        "_protoc_gen_swagger": attr.label(
+            default = Label("@grpc_ecosystem_grpc_gateway//protoc-gen-swagger:protoc-gen-swagger"),
+            executable = True,
+            cfg = "host",
+        ),
+        "_googleapis": attr.label(
+            default = Label("@grpc_ecosystem_grpc_gateway//third_party/googleapis:googleapis"),
```
f0rmiga: When I added the Bazel support initially I created the [repositories.bzl](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/repositories.bzl) that creates the `@com_github_googleapis_googleapis` repository target. The purpose of that repository target is to avoid using the `.proto` files under `third_party`. Could you try to use that repository target instead?

protoc-gen-swagger/defs.bzl
```diff
@@ -0,0 +1,79 @@
...
+    swagger_file = actions.declare_file(
+        "%s.swagger.json" % proto_service.basename[:-len(".proto")],
+        sibling = proto_service,
+    )
+
+    args = actions.args()
+    args.add("--plugin=%s" % protoc_gen_swagger.path)
+    args.add("--swagger_out=logtostderr=true:%s" % swagger_file.dirname)
+    args.add("-Iexternal/com_google_protobuf/src")
+    args.add("-Iexternal/grpc_ecosystem_grpc_gateway/third_party/googleapis/")
```
f0rmiga: It may be tricky to get the include directory for the protos from the repository target. Let me know if you have any questions regarding that.

protoc-gen-swagger/defs.bzl
```diff
@@ -0,0 +1,79 @@
...
+                ctx.actions,
+                ctx.executable._protoc,
+                ctx.executable._protoc_gen_swagger,
+            )
+        ])
+    )
+
+protoc_gen_swagger = rule(
+    attrs = {
+        "proto_service": attr.label(
```
f0rmiga: Instead of taking the `.proto` file, I think it would be better to take a `proto_library` so it is consistent on how the `go_proto_library` works. Take a look at the implementation [here](https://github.com/bazelbuild/rules_go/blob/master/proto/go_proto_library.bzl).

mrmeku: Donezo

protoc-gen-swagger/BUILD.bazel
```diff
@@ -19,6 +19,7 @@ go_library(
 go_binary(
     name = "protoc-gen-swagger",
     embed = [":go_default_library"],
+    visibility = ["//visibility:public"],
```
f0rmiga: Not sure why you need to set the visibility to be public here.

mrmeku: This binary is now an implicit dependency of protoc_gen_swagger since it is listed as the default label for the _protoc_gen_swagger attribute of that rule so it needs to be public since protoc_gen_swagger is itself publicly exposed

protoc-gen-swagger/defs.bzl
f0rmiga: You could rename to `proto` instead of `proto_service`.

mrmeku:
Good point

examples/examplepb/BUILD.bazel
```diff
@@ -49,3 +50,18 @@ go_library(
     embed = [":examplepb_go_proto"],
     importpath = "github.com/grpc-ecosystem/grpc-gateway/examples/examplepb",
 )
+
+protoc_gen_swagger(
+    name = "expamplepb_protoc_gen_swagger",
+    proto_service = "a_bit_of_everything.proto",
```
f0rmiga: You need to change to `proto = ":examplepb_proto",`. If you do, then you will get a bunch of errors, because the output of the `proto_library` is a list of proto descriptors in `.bin` files. So you will need to use a `Provider` to grab the `.proto` files from the `src` list in the `proto_library`.

protoc-gen-swagger/defs.bzl
```diff
@@ -0,0 +1,79 @@
...
+                ctx.executable._protoc_gen_swagger,
+            )
+        ])
+    )
+
+protoc_gen_swagger = rule(
+    attrs = {
+        "proto": attr.label(
+            mandatory = True,
+            allow_single_file = True,
```
f0rmiga: You will need to remove this to accept a Label instead of files. From that Label you will extract the `src` files using a Provider.

mrmeku: Donezo

protoc-gen-swagger/defs.bzl
```diff
@@ -0,0 +1,79 @@
...
+                ctx.actions,
+                ctx.executable._protoc,
+                ctx.executable._protoc_gen_swagger,
+            )
+        )
+    )
+
+protoc_gen_swagger = rule(
+    attrs = {
+        "deps": attr.label_list(
```
f0rmiga: I still think this should be called `proto`, not `deps`. You also don't need it to be a `label_list`.

f0rmiga:
Good job @mrmeku! Thank you for your patience in the review. I appreciate it.

ensonic:
Nice, is a support for proto options? E.g. https://github.com/grpc-ecosystem/grpc-gateway/pull/521 introduced: "grpc_api_configuration=path/to/your_api_config.yaml".

achew22:
No it isn't supported yet but it would be an easy pull request to add the functionality. Give it a go and send in whatever you can get working. I'll help you if you get stuck

ensonic:
Sound good, I am on it.

PR #617, opened 2018-04-24
Update dependency declarations in the Makefile

yugui:
This reduces unnecessary rebuilds by swagger-codegen


achew22:
Looks good to me but has some CI issues

tmc:
Looks like same intermittent failure on golint.

PR #619, opened 2018-04-25
Fixes file integrity errors on TravisCI

yugui:
Partially reverts 37660b1a4ff83f0ac3672ff7976223a859859e2f to fix
CI issues like
https://travis-ci.org/grpc-ecosystem/grpc-gateway/jobs/370802179.

Keep using the old "golang.org/x/net/context" in these files until
protoc-gen-go adopts the standard "context".


achew22:
Ah, good catch. Thanks for doing that.

PR #622, opened 2018-04-26
Skip unnecessary steps in USE_BAZEL builds on TravisCI

yugui:


.travis.yml
```diff
@@ -12,34 +12,55 @@ cache:
     - $HOME/.cache/_grpc_gateway_bazel
 before_install:
   - if [ "${USE_BAZEL}" = true ]; then ./.travis/install-bazel.sh 0.12.0; fi
```
achew22: Can you make this line match the test style you used below?

yugui: Do you mean just `if test "${USE_BAZEL} = true; then`?  Or are you thinking about something like `test "${USE_BAZEL}" = true && ./.travis/install-bazel.sh`?

I did not touch those steps for two reasons. 
* Bazel-related steps are less than others.  So they does not bloat up the configuration very much even if I left them as they are.
* I can't simply replace them with `test "${USE_BAZEL}" = true && ./.travis/install-bazel.sh ...` because whole of the step must have exit code 0 on `-z ${USE_BAZEL}`.

yugui: OK. Then, are you ok for leaving `if [ "${USE_BAZEL} = true ]; then` as is?

> WDYT about consolidating version numbers into the env section?

I think it is a good opportunity to do it. I'll push another commit later.

PR #627, opened 2018-04-27
[not ready] Port more build steps onto Bazel

yugui:
Generates example go clients with swagger in the build process with
Bazel and let the integration test use the generated clients.

### DIFF
--- a/examples/clients/abe/BUILD.bazel
+++ b/examples/clients/abe/BUILD.bazel
@@ -1,25 +1,51 @@
 load("@io_bazel_rules_go//go:def.bzl", "go_library")
+load(
+    "//examples/clients:clients.bzl",
+    "swagger_codegen",
+    "gen_file_integrity_test",
+)
 
 package(default_visibility = ["//visibility:public"])
 
+# keep
 go_library(
     name = "go_default_library",
     srcs = [
-        "a_bit_of_everything_nested.go",
-        "a_bit_of_everything_service_api.go",
-        "api_client.go",
-        "api_response.go",
-        "camel_case_service_name_api.go",
-        "configuration.go",
-        "echo_rpc_api.go",
-        "echo_service_api.go",
-        "examplepb_a_bit_of_everything.go",
-        "examplepb_body.go",
-        "examplepb_numeric_enum.go",
-        "nested_deep_enum.go",
-        "protobuf_empty.go",
-        "sub_string_message.go",
+        ":gen_client",
     ],
     importpath = "github.com/grpc-ecosystem/grpc-gateway/examples/clients/abe",
     deps = ["@com_github_go_resty_resty//:go_default_library"],
+)
+
+SRCS = [
+    "a_bit_of_everything_nested.go",
+    "a_bit_of_everything_service_api.go",
+    "api_client.go",
+    "api_response.go",
+    "camel_case_service_name_api.go",
+    "configuration.go",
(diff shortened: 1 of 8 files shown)


achew22:
@yugui, I am working with @mrmeku on a way to do this where you don't have to enumerate all the generated files. Do you mind if we hack on it over the weekend and see what we can come up with or would you rather we merge this in the interim?

yugui:
np. I'm looking forward to seeing it.
Then, I'll work on enhancement of the support of `oneof` fields this weekend.

Another thing I though about was a similar integrity check on `//examples/proto/...`. But it looks to be related to your plan.  So I'll explore other areas.

Have a good weekend.

achew22:
You bring up a good point. It might be worth it to explicitly enumerate these to do a golden check. I don't think the way we are going to do this will help with the golden check. We will try this out (maybe be unsuccessful) and then we can merge this when I give up and admit defeat 😉

examples/clients/clients.bzl
```diff
@@ -0,0 +1,53 @@
...
+    for out in outputs:
+      if not out.startswith(out_dir + "/"):
+        fail("%s must reside in the out_dir %s" % (out, out_dir), "outputs")
+
+  native.genrule(
+      name = name,
+      srcs = [spec],
+      outs = outputs,
+      cmd = " ".join([
+          "$(location //examples/clients:swagger-codegen-cli)",
```
achew22: I wrote something very similar to this for the project I'm working on. One thing I did that turned out to be extremely useful was to generate into a temp dir, then move the files one by one into out_dir_spec, then rm the files that are generated by swagger that we don't care about, then rmdir the temporary dir.

It seems like a lot of additional work, and it is, but it provides you with a compile time error if you missed one of the generated files in your `srcs` list. WDYT about doing that?

examples/clients/abe/BUILD.bazel
```diff
@@ -1,25 +1,51 @@
...
-        "nested_deep_enum.go",
-        "protobuf_empty.go",
-        "sub_string_message.go",
+        ":gen_client",
     ],
     importpath = "github.com/grpc-ecosystem/grpc-gateway/examples/clients/abe",
     deps = ["@com_github_go_resty_resty//:go_default_library"],
 )
+
+SRCS = [
```
achew22: Could this be done as a glob? If we do it as a glob and use he comment I left in the generator, then if something changes in the swagger definition and creates a new `.go` file it'll be updated automatically

examples/clients/clients.bzl
achew22: Github is being... weird. I typed out a whole comment in here but it wasn't included.

Here is approximately what I wrote:

What would you think about using a temp dir to generate these in along with a `.swagger-ignore-codegen` (or set of rm statements) and then a `mv` from the temp to the final `out_dir_spec` followed by an `rmdir` on the temp dir. This has the nice benefit of making it a compile time error if a new file is generated but not enumerated in the srcs section.

achew22:
I don't know if I mentioned my failure to get it working over the weekend. I think we should merge this or something like it. If you're interested in finishing it I would like to merge it

johanbrandhorst:
@yugui are you interested in finishing this work?

johanbrandhorst:
This ought to be easier to continue work on if we rebase on master.

johanbrandhorst:
Closing as stale

PR #628, opened 2018-04-28
Emit default values in marshalled JSON (with tests)

### LINKED ISSUE #233: Support emitting default values in JSON

flw-cn:
Repost for closes #233, #242 and #508. cc @ornithocoder @philipithomas @achew22.


flw-cn:
Sorry, I notice that there is already a PR #540 do the same thing, even more.

So, just close this.

PR #632, opened 2018-04-30
Add support for the grpc_api_configuration option in the bazel rule.

ensonic:
PR #521 added support for grpc service configurations. Add an option to let
users specify this config in the bazel rule.

Also add a simple doc blob to the rule.

### DIFF
--- a/protoc-gen-swagger/defs.bzl
+++ b/protoc-gen-swagger/defs.bzl
@@ -48,7 +48,6 @@ def _run_proto_gen_swagger(direct_proto_srcs, transitive_proto_srcs, actions, pr
 
 def _proto_gen_swagger_impl(ctx):
     proto = ctx.attr.proto.proto
-    grpc_api_configuration = None
     grpc_api_configuration = ctx.file.grpc_api_configuration
 
     return struct(
--- a/protoc-gen-swagger/defs.bzl
+++ b/protoc-gen-swagger/defs.bzl
@@ -1,3 +1,8 @@
+"""Generated an open-api spec for a grpc api spec.
+
+Reads the the api spec in protobuf format and generate an open-api spec. 
+Optionally applies settings from the grpc-service configuration.
+"""
 def _collect_includes(srcs):
     includes = ["."]
     for src in srcs:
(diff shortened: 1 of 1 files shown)


protoc-gen-swagger/defs.bzl
```diff
@@ -36,6 +48,8 @@ def _run_proto_gen_swagger(direct_proto_srcs, transitive_proto_srcs, actions, pr
 
 def _proto_gen_swagger_impl(ctx):
     proto = ctx.attr.proto.proto
+    grpc_api_configuration = None
```
achew22: rm unnecessary line

ensonic: Done.

ensonic:
> Could you also add a new target in //examples/proto/examplepb that runs with grpc_api_configuration set? Thanks!
I was trying to add:
```
protoc_gen_swagger(
    name = "expamplepb_protoc_gen_swagger_with_services",
    grpc_api_configuration = "unannotated_echo_service.yaml",
    proto = ":examplepb_proto",
)
```

But this will fail with:
```
bazel build :expamplepb_protoc_gen_swagger_with_services 
ERROR: file 'examples/proto/examplepb/unannotated_echo_service.swagger.json' is generated by these conflicting actions:
Label: //examples/proto/examplepb:expamplepb_protoc_gen_swagger, //examples/proto/examplepb:expamplepb_protoc_gen_swagger_with_services
RuleClass: protoc_gen_swagger rule
Configuration: b93351232485d7aae598752ea518de7e
Mnemonic: SkylarkAction
Action key: 497a439399336d44de55e732fde92588, ca9f888c655373c5b768919ee6510936
Progress message: SkylarkAction examples/proto/examplepb/unannotated_echo_service.swagger.json
PrimaryInput: File:[[<execution_root>]bazel-out/host/bin]external/com_google_protobuf/protoc
PrimaryOutput: File:[[<execution_root>]bazel-out/k8-fastbuild/bin]examples/proto/examplepb/unannotated_echo_service.swagger.json
Owner information: //examples/proto/examplepb:expamplepb_protoc_gen_swagger com.google.devtools.build.lib.skyframe.BuildConfigurationValue$Key@99896fcf false (1448708908), //examples/proto/examplepb:expamplepb_protoc_gen_swagger_with_services com.google.devtools.build.lib.skyframe.BuildConfigurationValue$Key@99896fcf false (431521607)
MandatoryInputs: Previous action contains artifacts not in attempted action (first 5): 
	examples/proto/examplepb/unannotated_echo_service.yaml
Outputs: are equal
```
If I uncomment the 'expamplepb_protoc_gen_swagger' target it works. I would like to investigate this separately, if that is okay with you. I think the bazel macro is not properly namespacing its outputs.

achew22:
Ah, that's totally fair, can you remove that entry from the BUILD file and I'll track doing it in https://github.com/grpc-ecosystem/grpc-gateway/issues/640

protoc-gen-swagger/defs.bzl
achew22: Not quite done?

ensonic:
> Ah, that's totally fair, can you remove that entry from the BUILD file and I'll track doing it in #640

You mean removing the example that works right now:
https://github.com/grpc-ecosystem/grpc-gateway/blob/master/examples/proto/examplepb/BUILD.bazel#L61

Can we not change it when fixing #640?

achew22:
I think this is good to go almost. I haven't merged due to https://github.com/grpc-ecosystem/grpc-gateway/pull/632/files#diff-256f5102c31109d1da9e7ebdbde6813cR51 where I asked for a change but I can't see the change. Am I missing something?

PR #637, opened 2018-05-01
provide access to underlying *json.Decoder from JSONPb.NewDecoder

jhump:
I am using the `JSONPb` marshaler in non-grpc-gateway server code to handle a mix of proto and non-proto types for JSON encoding/decoding.

I need to be able to control the underlying `*json.Decoder`, such as calling its `UseNumber()` method and being able to access its `Buffered()` data to resume reading from a stream after a JSON value has been decoded from it.

### DIFF
--- a/runtime/marshal_jsonpb.go
+++ b/runtime/marshal_jsonpb.go
@@ -14,6 +14,9 @@ import (
 // JSONPb is a Marshaler which marshals/unmarshals into/from JSON
 // with the "github.com/golang/protobuf/jsonpb".
 // It supports fully functionality of protobuf unlike JSONBuiltin.
+//
+// The NewDecoder method returns a DecoderWrapper, so the underlying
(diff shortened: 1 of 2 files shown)


runtime/marshal_jsonpb.go
```diff
@@ -84,16 +87,24 @@ func (j *JSONPb) marshalNonProtoField(v interface{}) ([]byte, error) {
 }
 
 // Unmarshal unmarshals JSON "data" into "v"
-// Currently it can marshal only proto.Message.
-// TODO(yugui) Support fields of primitive types in a message.
```
jhump: I removed these lines because they are not true anymore. There are even tests that verify that unmarshaling works for non-proto types.

jhump:
@achew22 or @yugui, if you have a sec, maybe you could review this small PR.

PR #641, opened 2018-05-04
enable generate single one swagger.json from many proto file

cy-zheng:
the cmd like this:
`protoc -I ${GOPATH}/src/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis -I src/pb/proto/ src/pb/proto/models.proto src/pb/proto/models2.proto --swagger_out=allow_merge=true,merge_file_name=myswagger,logtostderr=true:src/static/`


cy-zheng:
CLA fixed

PR #643, opened 2018-05-04
combine swagger file to one

cy-zheng:
the cmd like this:
`protoc src/pb/proto/models.proto src/pb/proto/models2.proto --swagger_out=allow_merge=true,merge_file_name=myswagger,logtostderr=true:src/static/`


wanghong230:
This is the feature I am looking for.

PR #646, opened 2018-05-09
Regenerate .pb.go code using the new Go protobuf package.

ivucica:
There have been incompatible changes in the underlying package, breaking at
least the OpenAPIv2 generator.

See https://groups.google.com/forum/#!topic/golang-nuts/F5xFHTfwRnY.

I did not run the tests locally, as I don't have a functioning setup of swagger codegen. I'll instead wait for the CI to complete, fix any problems and only spend time on setting it up locally if needed.

I regenerated the `.pb.go` files by just running `make test`.


ivucica:
Any thoughts on why only the build configuration Go 1.10.x without request_context plugin flag would be failing?

ivucica:
Any thoughts on why only the build configuration Go 1.10.x without request_context plugin flag would be failing here? @achew22

achew22:
@ivucica I think I've fixed the underlying problem, but there might be more that you've noticed. Can you try at HEAD?

ivucica:
I've rebased on HEAD, let's see if the tests go through now!

ivucica:
I just saw https://github.com/grpc-ecosystem/grpc-gateway/pull/656. Thanks, let's close this PR.

PR #647, opened 2018-05-09
fix #229 make Message use the name of the mapped import

### LINKED ISSUE #229: grpc-gateway_out does not entirely obey the -M parameter
Hi,

I'm using grpc-gateway_out with the github.com/gogo/protobuf project and I came across a case where the grpc-gateway generated code is invalid when using the -M parameter to the compiler.

When using the
`google/protobuf/empty.proto`
and want to replace the generated file with one from
`github.com/gogo/protobuf/types/`
via
`grpc-gateway_out=Mgoogle/protobuf/empty.proto=github.com/gogo/protobuf/types`
the generated code will import
`github.com/gogo/protobuf/types`
as expected, but the code still uses the
`empty.Empty`
to reference the Empty struct.

This results in both an unused import and an undefined reference compile error. The correct action should be to use `types.Empty` in the generated functions in this case.

Mistobaan:
this in particular fixes the situation when you use gogo.proto and your Message uses

`empty.Empty -> types.Empty`

fix #229

### DIFF
--- a/protoc-gen-grpc-gateway/descriptor/registry.go
+++ b/protoc-gen-grpc-gateway/descriptor/registry.go
@@ -316,6 +316,25 @@ func (r *Registry) defaultGoPackageName(f *descriptor.FileDescriptorProto) strin
 // protoc-gen-grpc-gateway rejects CodeGenerationRequests which contains more than one packages
 // as protoc-gen-go does.
 func (r *Registry) packageIdentityName(f *descriptor.FileDescriptorProto) string {
+	name := f.GetName()
+
+	if gopkg, ok := r.pkgMap[name]; ok {
+		idx := strings.LastIndex(gopkg, "/")
+		if idx < 0 {
+			gopkg = gopkg[idx+1:]
+		}
+
+		gopkg = gopkg[idx+1:]
+		// package name is overrided with the string after the
+		// ';' character
+		sc := strings.IndexByte(gopkg, ';')
+		if sc < 0 {
+			return sanitizePackageName(gopkg)
+
+		}
+		return sanitizePackageName(gopkg[sc+1:])
+	}
+
 	if f.Options != nil && f.Options.GoPackage != nil {
 		gopkg := f.Options.GetGoPackage()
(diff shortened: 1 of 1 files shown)


Mistobaan:
seems this is the reason for failing: https://travis-ci.org/grpc-ecosystem/grpc-gateway/jobs/377015212#L1007

```
runtime/context.go:88: the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak
```

achew22:
Looks like there are some CI issues. Do you think you could address them and ping the pr? Thanks!

Mistobaan:
properly fixing the error seems like a big refactor. The cancel function of
```
		ctx, _ /* cancel_fn */ = context.WithTimeout(ctx, timeout)
```
is never called which makes it useless (you can't cancel the timeout).

Mistobaan:
Is weird why it happens only when I edit go files. the lint is not ran always? needs more digging into the CI process.

protoc-gen-grpc-gateway/descriptor/registry.go
```diff
@@ -316,6 +316,25 @@ func (r *Registry) defaultGoPackageName(f *descriptor.FileDescriptorProto) strin
 // protoc-gen-grpc-gateway rejects CodeGenerationRequests which contains more than one packages
 // as protoc-gen-go does.
 func (r *Registry) packageIdentityName(f *descriptor.FileDescriptorProto) string {
+	name := f.GetName()
+
+	if gopkg, ok := r.pkgMap[name]; ok {
+		idx := strings.LastIndex(gopkg, "/")
+		if idx < 0 {
```
jdef: I don't understand the need for this if statement. The statement inside this if block is the same as the statement that follows it.

johanbrandhorst:
@Mistobaan Thanks for this! Could you rebase this and take a look at the comments?

johanbrandhorst:

ivanovaleksey:
@Mistobaan hi! Do you have any plans to push it to be merged?

PR #650, opened 2018-05-10
set grpc_gateway compiler public. change workspace name

Mistobaan:
adds the changes required by rules_go 
see https://github.com/bazelbuild/rules_go/pull/1495#discussion_r187386952

### DIFF
--- a/protoc-gen-grpc-gateway/BUILD.bazel
+++ b/protoc-gen-grpc-gateway/BUILD.bazel
@@ -1,7 +1,7 @@
 load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
 load("@io_bazel_rules_go//proto:compiler.bzl", "go_proto_compiler")
 
-package(default_visibility = ["//visibility:private"])
+package(default_visibility = ["//visibility:public"])
 
 go_library(
     name = "go_default_library",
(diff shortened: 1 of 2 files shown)


protoc-gen-grpc-gateway/BUILD.bazel
```diff
@@ -1,7 +1,7 @@
 load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
 load("@io_bazel_rules_go//proto:compiler.bzl", "go_proto_compiler")
 
-package(default_visibility = ["//visibility:private"])
+package(default_visibility = ["//visibility:public"])
```
achew22: Why do you need to make this public? the [go_proto_compiler](https://github.com/Mistobaan/grpc-gateway/blob/269da9ef1e0801d76d236e1cc60d7f03d9f142e5/protoc-gen-grpc-gateway/BUILD.bazel#L30) target is already public.

Mistobaan: mmh, I think I wasn't able to import the rule from the package without that change.

achew22: I use this in a project right now, so please forgive my skepticism. Can you create a small repo that reproduces that outcome? I would like to file a bug against the appropriate repo if it isn't working, then we can make the whole package public visibility.

johanbrandhorst:
I'm going to close this, could you please create an issue if this is still a problem?

Issue #661, opened 2018-05-24
proto-gen swagger: use service + method name as the opeartionId


wanghong230:
In the case, there are two services in the same proto file. If method names are duplicated, the swagger generated won't be valid, since operationId is not unique.

The idea is to use service name + method name to increate the uniqueness.

weisd:
+1

lpabon:
This is really important. Anyone working on this? @johanbrandhorst ?

achew22:
@lpabon, I don't think anyone is working on this. Would you be interested in taking a stab at it? What information do you need to get started on adding this as a flag?

johanbrandhorst:
I think there's an open PR for allowing the operation id to be configured. I'm not working on it myself.

johanbrandhorst:
See https://github.com/grpc-ecosystem/grpc-gateway/pull/739 and #40

lpabon:
Thanks @johanbrandhorst , but I think #739 is not exactly what I thought the solution would be. The solution in that PR is one way, but I was thinking of having the `OperationID` equal to the proto rpc Service name + the Method name. Adding the service name would make it 'like' the gRPC model.

What do you think?

johanbrandhorst:
I think we can have both, since the current default is to leave it empty, we can change that default in line with this issue but also allow it to be configured per #740. #739 has stalled so if you could pick that up in a separate PR I'd be happy to help you get that merged.

lpabon:
Sweet, will do

PR #662, opened 2018-05-24
#661 add service name to OperationID in swagger gen

wanghong230:


wanghong230:
I signed it!

wanghong230:
I signed it!

achew22:
This would be an interesting change to merge. Can you help me understand what the migration strategy would be for people who are using the project right now and aren't expecting their service names to change

wanghong230:
It is not breaking the existing logic. People will receive the new swagger spec which has a different opertional id. "operationId is an optional unique string used to identify an operation. If provided, these IDs must be unique among all operations described in your API." So, it shouldn't cause any issues in reality.

But existing logic definitly broken, since you can have multiple services in same proto file, but using same set of method names like Get, List, Delete, the swagger spec could be generated but cannot be used without manual fixing duplicated operation ids.

achew22:
@ivucica, thanks for the thoughts! Just a few things to add:

> There are ways out of this; how about adding a suffix only if the operation ID is determined to be duplicate? (Similar to what's done with bIdx.) This will address the immediate issue.

We have discussed proposals like this before, but I'm fairly reluctant to have behavior change based on duplication. My primary consideration is this case:

I have `a.proto` file that has a method `Get`. I use this for a long time. Later I add a Get method to a different service and now, all the code I wrote before needs to be updated to reflect the surprise changing of the original `Get` method's name to `packageGet`.

To me, it seems like explicitly renaming their method `packageGet` is more obvious about what is happening and has most (if not all) of the benefits.

What do you think?

ivucica:
> Later I add a `Get` method to a different service and now, all the code I wrote before needs to be updated to reflect the surprise changing of the original `Get` method's name to `packageGet`.

What I meant is 'leave the original alone', but then again, you can't really tell which method is the original. So ignore the suggestion.

I agree, renaming the Swagger operation explicitly is the only way to go about it, and right now it's the only way to do it. But renaming the Swagger operation can only be done by renaming the gRPC method. That breaks the gRPC interface. In gRPC, it's a valid expectation that different services have differently named methods. It's simply not right that Swagger's limitations force a change in how the methods must be named.

There's the possibility of using the OpenAPI proto-options to rename the operation (I don't know if this can currently be done, but surely it can be hacked in). That, however, seems like a manual workaround, and necessary only because the default behavior is incorrect.

So how about hiding the new operation ID proposed in this PR behind a flag, and announcing that the default value of the flag will change?

---

This makes me sad a bit. To state the obvious: Having the power of hindsight on our side, it feels like 'operation ID is global' and 'we'll just use tags to group methods' was a big mistake on Swagger's part. It's not even an issue on grpc-gateway's side that it declares service membership using tags; Swagger's own code generators use tags to generate service-like groupings of methods in client code.

johanbrandhorst:
It sounds like this would have to be part of a v2 release if it goes in at all.

dadgar:
I feel like this is a pretty important issue to revisit. As it stands, if you have multiple services with basic CRUD methods, you get operationId conflicts. It seems like the default behavior should be updated to append the service name to generate unique operationIds across services.

As a compatibility bridge, supporting setting the operationId explicitly via an annotation seems reasonable for those who want to continue using the unprefixed operationId.

johanbrandhorst:
> I feel like this is a pretty important issue to revisit. As it stands, if you have multiple services with basic CRUD methods, you get operationId conflicts. It seems like the default behavior should be updated to append the service name to generate unique operationIds across services.
> 
> As a compatibility bridge, supporting setting the operationId explicitly via an annotation seems reasonable for those who want to continue using the unprefixed operationId.

We welcome help in this area, and fixing clear bugs is an acceptable reason to break users. I don't think we necessarily need the annotation since if our users are going in to manually change their proto files to fix this change they can probably just update their generated files. Would you be interested in contributing a fix for this?

dadgar:
@johanbrandhorst Just to level set on the expectation for this. The functional change of the PR would be prefixing the operationId with the service name, and then some unit testing?

johanbrandhorst:
> @johanbrandhorst Just to level set on the expectation for this. The functional change of the PR would be prefixing the operationId with the service name (similar to this PR), and then some unit testing?

Sounds good to me. I know @ivucica was proposing a slightly less damaging change (new behaviour behind flag), but tbh this does seem like a bug to me, and I don't anticipate it's going to have that big of an impact on users. If I'm missing something, please let me know before you start working on this.

dadgar:
@johanbrandhorst So I implemented this (https://github.com/dadgar/grpc-gateway/compare/master...f-operation-id?expand=1) but it has a knock on effect I didn't consider and I wanted to discuss before opening the PR.

The code generators use operationId to generate the method names. So this change causes the generated clients to be quite verbose: https://github.com/dadgar/grpc-gateway/commit/bdeecff3783f66a4a0fbdc2b36c786add7b00ed8#diff-2f368f91f1b1c6435040ee606a82216eR818

This is clearly not nice. FWIW, this pattern is what Azure does with their swagger (https://github.com/Azure/azure-rest-api-specs/blob/master/specification/compute/resource-manager/Microsoft.Compute/stable/2019-12-01/compute.json#L42) and they fix it on the client generation (https://github.com/Azure/azure-sdk-for-go/blob/master/services/compute/mgmt/2019-12-01/compute/operations.go#L45).

Further the spec that is being generated today is not compliant. https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#fixed-fields-5 says the ID must be unique.

Let me know how you want to proceed.

johanbrandhorst:
Thanks for all that, it's making me more convinced that going ahead with this change is the right choice. Obviously we don't control the generators so I imagine the names will be different, but it's probably worthwhile.

One thought I didn have though, would it be possible for us to detect if this is necessary at generation time? So we could make a fix that only activates if necessary?

dadgar:
@johanbrandhorst I thought about only adding the prefix if a conflict was detected but decided against it for two reasons:

1) I'd rather have consistency of the operation ID so I can for example customize generators to strip the prefix.
2) Don't want operation IDs of existing endpoints to change if I add a future endpoint with the same name. That could break people using generated clients.

dadgar:
@johanbrandhorst https://github.com/grpc-ecosystem/grpc-gateway/pull/1193

johanbrandhorst:
Superceded by #1193

PR #668, opened 2018-06-11
Bazel expose protoc-gen-grpc-gateway

afking:
Exposes go_binary() rule for protoc-gen-grpc-gateway so dependencies can be vendored for the go_proto_compiler() rule when using //vendor dependencies.

### DIFF
--- a/protoc-gen-grpc-gateway/BUILD.bazel
+++ b/protoc-gen-grpc-gateway/BUILD.bazel
@@ -20,6 +20,7 @@ go_library(
 go_binary(
     name = "protoc-gen-grpc-gateway",
     embed = [":go_default_library"],
+    visibility = ["//visibility:public"],
 )
 
 go_proto_compiler(


afking:
Closing, recommendation is to remove all BUILD files for vendor/ dependencies.

afking:
Reopening, we are defining `go_proto_compilier(...)` internally to reference vendor/ deps and require this to be exposed.

afking:
So some of our deps for https://github.com/grpc-ecosystem/grpc-gateway/blob/master/protoc-gen-grpc-gateway/BUILD.bazel#L31 are registered in bazel under `//vendor/...`. This leads to compile errors as the package types are incompatible.  Not sure if there is a nicer way to do this? The current solution is to point the go library deps at the vendored packages and use the external repositories just for binaries.

afking:
@achew22 any idea why this is breaking?

achew22:
Looks like protoc-go changed their output again. Can you run `make examples` and upload?

afking:
I'm unable to get the correct versions without playing with it more, seems to edit all the examples/*.go files.

achew22:
I would expect that. Since the upstream protoc changed, probably all the `.pb.go` files will need to be updated.

achew22:
Out of curiosity, what would you think about adding a new `go_proto_compiler` entry in there that has the option you care about enabled as well. WDYT? I'm happy to merge if you think it's not the right way to go.

afking:
@achew22 our go_proto_compiler rule looks like:

```
go_proto_compiler(
    name = "go_gateway",
    options = ["logtostderr=true"],
    plugin = "@grpc_ecosystem_grpc_gateway//protoc-gen-grpc-gateway",
    suffix = ".pb.gw.go",
    visibility = ["//visibility:public"],
    deps = [
        "//vendor/github.com/golang/protobuf/proto:go_default_library",
        "//vendor/github.com/grpc-ecosystem/grpc-gateway/runtime:go_default_library",
        "//vendor/github.com/grpc-ecosystem/grpc-gateway/utilities:go_default_library",
        "//vendor/golang.org/x/net/context:go_default_library",
        "//vendor/google.golang.org/grpc:go_default_library",
        "//vendor/google.golang.org/grpc/codes:go_default_library",
        "//vendor/google.golang.org/grpc/grpclog:go_default_library",
        "//vendor/google.golang.org/grpc/status:go_default_library",
    ],
)
```

I think there will be nicer ways to do this in the future with the rules_go. Maybe something in https://github.com/bazelbuild/rules_go/issues/1548 will fix it.

achew22:
Oh, so is the major factor that you're vendoring in groc-gateway? If you weren't vendoring would the existing target work for you?

afking:
Yes and other grpc libraries that can cause version conflicts.

PR #681, opened 2018-06-24
gen-swagger: use GetJsonName for struct member names

utrack:
This PR fixed Swagger definition.

Currently `protoc-gen-swagger` uses `.GetName()` to get a field name.
It's wrong, since proto field `snake_case`'s generated name in `json` tag is `snakeCase` - however, generated swagger still has `snake_case` in the definition. 
`.GetJsonName()` returns field name in the right case and form.

### DIFF
--- a/protoc-gen-swagger/genswagger/template.go
+++ b/protoc-gen-swagger/genswagger/template.go
@@ -251,7 +251,7 @@ func renderMessagesAsDefinition(messages messageMap, d swaggerDefinitionsObject,
 				panic(err)
 			}
 
-			schema.Properties = append(schema.Properties, keyVal{f.GetName(), fieldValue})
+			schema.Properties = append(schema.Properties, keyVal{f.GetJsonName(), fieldValue})
 		}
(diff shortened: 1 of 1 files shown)


achew22:
Hey @utrack, thanks for sending in a PR! This is definitely a change I would like to make. The swagger generator is objectively wrong. Unfortunately it is wrong in a way that is paired with our default marshaller. In order to make this change we would have to change the default marshaller, which is a kind of big change.

What would you think about adding this as a flag controlled argument to the swagger generator? That would let you continue forward while I try to figure out a path to removing the bad marshaller.

utrack:
Oh dang, I just noticed I've already filled #375 :)

I see that the new flag should be added via common registry - it would be confusing for the flag to work w/ `-gen-swagger`, but not with `-gen-gateway`.

I'll maintain my fork w/ `-gen-go`-style JSON tags and wait for the bad marshaller replacement meanwhile.

What's the issue related to the marshaller replacement, if there's any? I might be able to help with that.

johanbrandhorst:
This is already being considered in #546

PR #682, opened 2018-06-25
Api

ithinker1991:
add api


rogerhub:
https://help.github.com/articles/blocking-a-user-from-your-organization/

achew22:
@rogerhub, I appreciate the concern but I would like to leave @ithinker1991 able to send PRs and file issues not just on grpc-gateway, but in the rest of the grpc-ecosystem projects. Ultimately, these PRs (confusing and strange though they may be) are a sign that they are interested in the trajectory of the project and therefore have a vested interest in its welfare. I would, however, understand if you're annoyed by these emails. You can block them personally which will prevent you getting PR notification emails which seems like a fair middle ground. @ithinker1991 if you would like to talk I'm all ears and eager to hear what it is you're trying to do. I'm going to close this, but feel free to reopen if you have some constructive steps forward.

Thanks!

rogerhub:
Ah, I wasn’t aware that was an option. Thanks for the explanation, and keep up the good work!

PR #686, opened 2018-06-28
Add openapiv2_json_name field option to specify non alphanumeric characters in JSON property name

co3k:
proto3 provides json_name option by default (see also https://developers.google.com/protocol-buffers/docs/proto3#json)
but we cannot use this option because it converts snake_cased field name to camelCased one. It means we should have
our own json_name field option for this purpose to keep backward compatibility.


co3k:
Oh, I didn't noticed #375 and #681.

If we don't need to keep compatibility of existing generated swagger definitions, we can choose #681 approach instead of this pull request.

johanbrandhorst:
This is being considered in #546

PR #692, opened 2018-07-03
Populate swagger method parameter description from message comments

co3k:


johanbrandhorst:
Hi @co3k, and thanks for your contribution! The title of this PR (and commit message) confuses me, could you reword it? I don't understand what this does just from reading the title.

co3k:
@johanbrandhorst 
Oh, I'm sorry for confusing you. How about the following?

```diff
- Reflect descriptions of message to Swagger parameter representation automatically
+ Reflect comments of proto message as description of Swagger's Parameter Object representation implicitly
```

johanbrandhorst:
Let me see if I understand correctly. Would this also be a correct description?

`Populate swagger method parameter description from message comments`

If so, I think that is clearer _yet_ than your update ;).

co3k:
@johanbrandhorst Thank you for your good suggestion. I've updated the title of this pull request and the commit message.

protoc-gen-swagger/genswagger/template.go
```diff
@@ -567,11 +574,15 @@ func renderServices(services []*descriptor.Service, paths swaggerPathsObject, re
...
+							desc = schema.Description
+						} else {
+							desc = fieldProtoComments(reg, lastField.Target.Message, lastField.Target)
+						}
 					}
 
-					desc := ""
 					if meth.GetClientStreaming() {
-						desc = "(streaming inputs)"
+						desc = desc + "(streaming inputs)"
```
johanbrandhorst: This can just be `desc += " (streaming inputs)"` right? (Also notice the space before the first bracket)

co3k: Thanks! Fixed.

PR #693, opened 2018-07-03
runtime: add support for time types in query parameters

johanbrandhorst:
Adds support for parsing the google.protobuf.Duration as well as native *time.Time and *time.Duration types in url query parameters.

Helps #400. Fixes https://github.com/gogo/protobuf/issues/420.

### DIFF
--- a/runtime/query.go
+++ b/runtime/query.go
@@ -221,6 +221,23 @@ func populateField(f reflect.Value, value string, props *proto.Properties) error
 			f.Field(0).SetInt(int64(t.Unix()))
 			f.Field(1).SetInt(int64(t.Nanosecond()))
 			return nil
+		case "Duration":
+			if value == "null" {
+				f.Field(0).SetInt(0)
+				f.Field(1).SetInt(0)
+				return nil
+			}
+			d, err := time.ParseDuration(value)
+			if err != nil {
+				return fmt.Errorf("bad Duration: %v", err)
+			}
+
+			ns := d.Nanoseconds()
+			s := ns / 1e9
+			ns %= 1e9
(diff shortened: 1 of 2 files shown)


runtime/query.go
```diff
@@ -284,14 +301,32 @@ func populateField(f reflect.Value, value string, props *proto.Properties) error
...
+		*t = pt
+		return nil
+	case *time.Duration:
+		d, err := time.ParseDuration(value)
+		if err != nil {
+			return fmt.Errorf("bad Duration: %v", err)
+		}
+		*t = d
+		return nil
+	}
```
johanbrandhorst: This part may be controversial, but I think considering this does not add any explicit dependencies on any gogoproto packages, I argue that this should be acceptable.

runtime/query_test.go
```diff
@@ -338,6 +347,51 @@ func TestPopulateParameters(t *testing.T) {
...
+
+		if err != nil {
+			t.Errorf("runtime.PopulateQueryParameters(msg, %v, utilities.NewDoubleArray(nil)) failed with %v; want success", spec.values, err)
+			continue
+		}
+		if got, want := msg, spec.want; !proto.Equal(got, want) {
+			t.Errorf("runtime.PopulateQueryParameters(msg, %v, utilities.NewDoubleArray(nil)) = %v; want %v", spec.values, got, want)
+		}
+	}
+}
```
johanbrandhorst: This needs to be a separate test because `golang/protobuf/proto.Clone` panics when using *time.Time in the proto message.

johanbrandhorst:
@achew22 @tmc any objections to this?

Issue #689, opened 2018-07-02
Vendor dependencies


jsha:
For the Trillian project (https://github.com/google/trillian/pull/1220), I am trying to vendor all dependencies. Right now, Trillian depends on a specific version of `protoc-gen-grpc-gateway`. However, even if we pull that specific version, we don't currently have a good guarantee that we are getting a consistent set of dependencies pulled into `protoc-gen-grpc-gateway`. Would you consider vendoring this project's dependencies using `dep`? Vendoring represents the dependency management plan that the Go team is moving forwards with: https://blog.golang.org/versioning-proposal. Go's end goal won't be exactly like `dep` but will be similar, and the Go team has committed to a smooth transition.

tmc:
I think going the Gopkg route as a temporary measure makes sense. @johanbrandhorst can you take that on?

johanbrandhorst:
@tmc I'm on it

PR #695, opened 2018-07-07
protoc-gen-swagger: support all well-known wrapper types

jriecken:
There were a few well-known wrapper types missing from the `wkSchemas` map, in specific: `UInt32Value`, `UInt64Value` and `BytesValue`. This change handles them (and maps them to the same swagger types as the non-wrapped values)

This also fixes the mapping of `Int64Value`. The `Int64Value` handling maps the value to a swagger `integer`. The documentation for `Int64Value` indicates that it should be mapped to a JSON `string` (also the mapping for normal `int64` in `protoc-gen-swagger` maps it to a `string`, so this was inconsistent.)


jriecken:
I signed it!

jriecken:
It looks like this was caused by an upstream bugfix in `github.com/golang/protobuf/jsonpb`  that was made yesterday: https://github.com/golang/protobuf/pull/645

The JSON generated by that `marshal_jsonpb` test before that fix looks like:

```json
{"singleNested":{},"uuid":"6EC2446F-7E89-4127-B3E6-5C05E6BECBA7","nested":[{"name":"foo","amount":12345}],"uint64Value":"18446744073709551615"
,"repeatedStringValue":[],"oneofString":"bar","mapValue":{"a":1,"b":0},"mappedStringValue":{},"mappedNestedValue":{},"timestampValue":"1970-01-01T00:00:00Z","repeatedEnumValue":[
]}
```

and after:

```json
{"singleNested":{},"uuid":"6EC2446F-7E89-4127-B3E6-5C05E6BECBA7","nested":[{"name":"foo","amount":12345}],"uint64Value":"18446744073709551615"
,"repeatedStringValue":[],"oneofString":"bar","mapValue":{"a":"ONE","b":"ZERO"},"mappedStringValue":{},"mappedNestedValue":{},"timestampValue":"1970-01-01T00:00:00Z","repeatedEnu
mValue":[]}
```

Note how the `mapValue` map's values changed from the numeric representation of the enum (`"mapValue":{"a":1,"b":0}`) to the string one (`"mapValue":{"a":"ONE","b":"ZERO"})`. This makes sense as the test is setting `EnumsAsInts` to `0` (i.e. output them a strings).

I'll update the test to match the fixed behavior from `jsonpb`

jriecken:
One of the browser tests was also assuming integer enum values - fixed it too

johanbrandhorst:
Looks like we still need some regeneration of files.

jriecken:
One of the travis runs failed because protoc generated a file descriptor `var` that was slightly different  (`fileDescriptor_wrappers_5850bd48b0187754` vs `fileDescriptor_wrappers_0cd12813574dca64`) and the git diff failed. The other 5 non-bazel test runs did not have this same issue.

The Bazel version of the build also failed (it must somehow be pulling in a different version of `jsonpb` as it output the old numeric enum values?)

I feel like I'm very quickly going down a rabbit hole in `grpc-gateway`'s build process...

johanbrandhorst:
I think the first failure should be fixed with a regeneration of the protofiles (I guess it's sha-dependent?). I admit I'm mostly ignorant of the build process. @achew22 @tmc could you please help?

jriecken:
I was using the latest version of `protoc` instead of 3.3 - regenerating with 3.3 created the same `var` name.

jriecken:
I did a little more digging and the Bazel Go rules pin the `golang/protobuf` dependency to this version: https://github.com/golang/protobuf/releases/tag/v1.1.0 (see https://github.com/bazelbuild/rules_go/blob/7e07b18263c08c8eca31207e1f5b28aa71539a72/go/private/repositories.bzl#L75)

The non-Bazel build uses `go get` (i..e pulling from `master` of `golang/protobuf`, which has the change made yesterday that caused the `marshal_jsonpb` test to break in the first place)

johanbrandhorst:
So it seems we have a mismatch of versions between that pinned by Bazel and that fetched in the normal test runs. Presumably this means we cannot get a successful build unless we change one of the two. I'm not comfortable with making the call here but I would assume we'd be more keen to pin versions to _releases_, so assuming that, how do we pin the non-Bazel tests? This might well not be fixable until something like https://github.com/grpc-ecosystem/grpc-gateway/issues/689 is done.

johanbrandhorst:
https://github.com/grpc-ecosystem/grpc-gateway/pull/696 should hopefully fix these dependency inconsistencies. Then this PR should be easy to merge.

jriecken:
Done

johanbrandhorst:
Great, this LGTM!

mayank-dixit:
@johanbrandhorst @jriecken 
With this change, can this be done:

```
service Greeter {
    option (google.api.http) = {
      get: "/say/{strVal}"
    };
  }
}

message HelloRequest {
  google.protobuf.StringValue strVal = 5;
}

```


Right now when I compile using protoc:
```
  -I$GOPATH/src \
  helloworld/helloworld.proto
```


I get the error:
`--grpc-gateway_out: aggregate type TYPE_MESSAGE in parameter of Greeter.SayHello: strVal`

johanbrandhorst:
That should absolutely work, I can't see why that is erroring. Could you raise an issue please?

mayank-dixit:
> That should absolutely work, I can't see why that is erroring. Could you raise an issue please?

Created: https://github.com/grpc-ecosystem/grpc-gateway/issues/808

PR #696, opened 2018-07-08
Add explicit dependency versions

### LINKED ISSUE #689: Vendor dependencies

johanbrandhorst:
Version constraints are copied from existing Bazel rules. In the future, these version upgrades must be performed atomically.

Fixes #689


johanbrandhorst:
The build errors imply files have been generated with `protoc-gen-go` 1.1, but Bazel dependencies are using `golang/protobuf` v1.0.0: https://github.com/grpc-ecosystem/grpc-gateway/pull/695#issuecomment-403295395.

I think we need to update the `rules_go` version we're using to something more recent, and at the same time update the locked dependency versions in Gopkg.toml.

johanbrandhorst:
Bump. I think we need to get this in. @achew22 @tmc @yugui

Issue #702, opened 2018-07-15
The swagger plugin couldn’t distinguish two rpcs if we use the resource name design style.


ch3rub1m:
The swagger plugin couldn’t distinguish two rpcs designed according to the Google API Design Guide [Resource Names](https://cloud.google.com/apis/design/resource_names) as below:
```
  rpc GetService(GetServiceRequest) returns (Service) {
    option (google.api.http) = {
      get: "/v1/{name=service_categories/*/services/*}"
    };
    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
      description: "Get a service.The **name** must have the format of `service_categories/*/services/*`.";
      tags: "Service";
    };
  }
```
```
  rpc GetUser(GetUserRequest) returns (User) {
    option (google.api.http) = {
      get: "/v1/{name=users/*}"
    };
    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
      description: "Get a user.The **name** must have the format of `users/*`.";
      tags: "User";
    };
  }
```
The two generated results from the swagger plugin both use the `/v1/{name}` as the key.
There is no problem if we use `ID` as the user input, but according to the Google API Design Guide the `name` is a better choice.
Is there any solution about this issue?

achew22:
@ch3rub1m, thanks for the detailed and interesting bug report. You have definitely run into an issue with our Swagger generation.

Reading through the linked documentation, I think you're interpreting things in a valid way, but I'm not sure how we would map it in Swagger/OpenAPI land. The concept works in the gateway because it can parse the whole string and determine the match, but in Swagger it wouldn't be able do to that since name is just a string to the type system. I would be *very* interested at hearing alternative ways to approach this, but as of right now the only solution I see is to call the field something other than `name` for the 2ndary match. You could also deviate from their naming scheme and decompose the URL into the constituent tuples with something like:

```
  rpc GetService(GetServiceRequest) returns (Service) {
    option (google.api.http) = {
      get: "/v1/service_categories/{category}/services/{id}"
    };
    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
      description: "Get a service.The **name** must have the format of `service_categories/*/services/*`.";
      tags: "Service";
    };
  }
  rpc GetUser(GetUserRequest) returns (User) {
    option (google.api.http) = {
      get: "/v1/users/{name}"
    };
[...]

PR #704, opened 2018-07-20
add support for resource name in swagger plugin (#702)

ch3rub1m:
Fix this issue: The swagger plugin couldn’t distinguish two rpcs if we use the resource name design style. (#702)

### DIFF
--- a/protoc-gen-swagger/genswagger/template.go
+++ b/protoc-gen-swagger/genswagger/template.go
@@ -525,7 +525,7 @@ func templateToSwaggerPath(path string) string {
 			}
 			// Pop from the stack
 			depth--
-			buffer += "}"
+			buffer += string(char)
 		case '/':
 			if depth == 0 {
 				parts = append(parts, buffer)
@@ -534,6 +534,7 @@ func templateToSwaggerPath(path string) string {
 				// section.
 				continue
 			}
+			buffer += string(char)
 		default:
 			buffer += string(char)
 			break
@@ -548,10 +549,22 @@ func templateToSwaggerPath(path string) string {
 	// memory.
 	re := regexp.MustCompile("{([a-zA-Z][a-zA-Z0-9_.]*).*}")
 	for index, part := range parts {
+		// If part is a resource name such as "parent", "name", "user.name", the format info must be retained.
+		prefix := re.ReplaceAllString(part, "$1")
+		if isResourceName(prefix) {
+			continue
+		}
 		parts[index] = re.ReplaceAllString(part, "{$1}")
 	}
 
 	return strings.Join(parts, "/")
+}
+
+func isResourceName(prefix string) bool {
+	words := strings.Split(prefix, ".")
+	l := len(words)
+	field := words[l-1]
+	return field == "parent" || field == "name"
 }
 
 func renderServices(services []*descriptor.Service, paths swaggerPathsObject, reg *descriptor.Registry, requestResponseRefs, customRefs refMap) error {
--- a/protoc-gen-swagger/genswagger/template_test.go
+++ b/protoc-gen-swagger/genswagger/template_test.go
@@ -697,6 +697,11 @@ func TestTemplateToSwaggerPath(t *testing.T) {
 		{"/{test=prefix/that/has/multiple/parts/to/it/*}", "/{test}"},
 		{"/{test1}/{test2}", "/{test1}/{test2}"},
 		{"/{test1}/{test2}/", "/{test1}/{test2}/"},
+		{"/{name=prefix/*}", "/{name=prefix/*}"},
+		{"/{name=prefix1/*/prefix2/*}", "/{name=prefix1/*/prefix2/*}"},
+		{"/{user.name=prefix/*}", "/{user.name=prefix/*}"},
+		{"/{user.name=prefix1/*/prefix2/*}", "/{user.name=prefix1/*/prefix2/*}"},
+		{"/{parent=prefix/*}/children", "/{parent=prefix/*}/children"},
 	}
 
 	for _, data := range tests {


ch3rub1m:
I signed it!

johanbrandhorst:
Reopening on the request of @ch3rub1m

johanbrandhorst:
@ch3rub1m could you push your changes again please?

ch3rub1m:
@johanbrandhorst @achew22 I already pushed. But it built failed again.

achew22:
Yeah, the problem is that some library we depend on for serializing changed their default for serializing enums from the numeric representation to the string representation. A reasonable change to make, but we need to explicitly set that in whatever library that is. Sorry I just haven't had time to dig into it

ch3rub1m:
@achew22 Ok, I got it, thanks.

achew22:
Could you rebase off of master? We have since fixed the CI issue

achew22:
Also, I was thinking about this. Would you be willing to (feel free to push back) add some new usage in the everything example? I would love it if we could do an e2e too

ch3rub1m:
@achew22 It still not work with GO: 1.10.x. Could you help to check it please?

ch3rub1m:
@achew22 Hi, I would like to add more usage, but it seems like it's not so suitable to add them directly to the everything example. May I create a new example? By the way, please take a look at the CI issue when you are not that busy.

achew22:
Yes, please. Feel free to add as many examples as you think are necessary

jon-whit:
What's the status on this PR? I have the same need.

I have a field in a protobuf message called "parent". It looks like this:
```
  message ListRolesRequest {
    // The resource name of the parent resource in one of the following formats:
    // `` (empty string) -- this refers to curated roles.
    // `organizations/{organization_id}`
    // `tenants/{tenant_id}`
    string parent = 1;

```

The parent can take the form "organizations/<org_id>" or "tenants/<tenant_id>" or it can be empty. I'd like the swagger to match all three. For example:

GET `/v1/organizations/<org_id>/roles`
GET `/v1/tenants/<tenant_id>/roles`
GET `/v1/roles`

jon-whit:
@ch3rub1m Are you saying I can already achieve what I'm looking for? If so, what would the protobuf HTTP options look like for that? That's what I'm struggling with.

If not, are you saying that once this PR is merged the example above will be supported? Sorry for the confusion.

ch3rub1m:
Yes, you could already achieve what you are looking for in current version! But there are some bug in generated swagger doc. And I am trying to fix this. But the bug just influence the doc so the code still work. You should write as follow to achieve what you want:
``` 
message ListRolesRequest {
    // The resource name of the parent resource in one of the following formats:
    // `` (empty string) -- this refers to curated roles.
    // `organizations/{organization_id}`
    // `tenants/{tenant_id}`
    string parent = 1;
    ...
  }

  rpc ListRoles(ListRolesRequest) returns (ListRolesResponse) {
      //Maps to HTTP GET.
      option (google.api.http) = {
        get: "/auth/iam/roles"
        additional_bindings {
          get: "/auth/iam/{parent=*/*}/roles"
        }
      };
  }
```

jon-whit:
@ch3rub1m That doesn't match the three mentioned above though.. It will only match

```
GET /auth/iam/organizations/<org_id>/roles
GET /auth/iam/tenants/<tenant_id>/roles
```

I'd like a match that matches all three of the following:

```
GET /auth/iam/roles
GET /auth/iam/organizations/<org_id>/roles
GET /auth/iam/tenants/<tenant_id>/roles
```

ch3rub1m:
Hi, @achew22. I tried to write some usage in everything example but it failed in travis CI again. The bazel really confused me and I don't know how to generate *.pb.go and swagger file by bazel. So I roll back the example commit and it works again. Could you approve this PR first please? I have done many tests in my own project and it seems the PR is ready to be merged. If you wants more usage in example I will open another PR.

johanbrandhorst:
This LGTM, thanks for your contribution!

tlyng:
I'm unable to generate valid swagger output after this merge. I've tried two validators / parsers and they both complain about the same thing.

```
Semantic error at paths./api/v1/{name=shelves/*/books/*}
Declared path parameter "name=shelves/*/books/*" needs to be defined as a path parameter at either the path or operation level
```

Take this example protobuf specification:

```protobuf
syntax = "proto3";

import "google/protobuf/empty.proto";

service LibraryService {
    
    // Shelf management
    rpc GetShelf(GetShelfRequest) returns (Shelf) {
        option (google.api.http) = {
            get: "/pai/v1/{name=shelves/*}"
        };
    };

    rpc CreateShelf(CreateShelfRequest) returns (Shelf) {
        option (google.api.http) = {
            post: "/api/v1/shelves"
            body: "shelf"
        };
    };

    rpc DeleteShelf(DeleteShelfRequest) returns (google.protobuf.Empty) {
        option (google.api.http) = {
            delete: "/api/v1/{name=shelves/*}"
        };
    };

    rpc ListShelfs(ListShelfRequest) returns (ListShelfResponse) {
        option (google.api.http) = {
            get: "/api/v1/shelves"
        };
    };

    // Book management
    rpc GetBook(GetBookRequest) returns (Book) {
        option (google.api.http) = {
            get: "/api/v1/{name=shelves/*/books/*}"
        };
    };

    rpc CreateBook(CreateBookRequest) returns (Book) {
        option (google.api.http) = {
            post: "/api/v1/{parent=shelves/*}/books"
            body: "book"
        };
    };

    rpc DeleteBook(DeleteBookRequest) returns (google.protobuf.Empty) {
        option (google.api.http) = {
            delete: "/api/v1/{name=shelves/*/books/*}"
        };
    };

    rpc ListBooks(ListBookRequest) returns (ListBookResponse) {
        option (google.api.http) = {
            get: "/api/v1/{parent=shelves/*}/books"
        };
    };
}

message Shelf {
    // Resource name of the shelf. It must have the format of "shelves/*".
    // For example: "shelves/shelf1"
    string name = 1;
}

message Book {
    // Resource name of the book. It must have the format of "shelves/*/books/*".
    // For example: "shelves/shelf1/books/book2"
    string name = 1;
    string isbn = 2;
    string author = 3;
}

message GetShelfRequest {
    // Resource name of the shelf.
    // For example: "shelves/shelf1"
    string name = 1;
}

message CreateShelfRequest {
    // Resource name of the shelf.
    // For example: "shelves/shelf1"
    Shelf shelf = 1;
}

message ListShelfRequest {
}

message ListShelfResponse {
    repeated Shelf shelves = 1;
}

message DeleteShelfRequest {
    // Resource name of the shelf.
    // For example: "shelves/shelf1".
    string name = 1;
}

message GetBookRequest {
    // Resource name of the book.
    // For example: "shelves/shelf1/books/book1"
    string name = 1;
}

message CreateBookRequest {
    // Resource name of the parent resource where to create the book.
    // For example: "shelves/shelf1".
    string parent = 1;
    Book book = 2;
}

message DeleteBookRequest {
    // Resource name of the book to be deleted.
    // For example: "shelves/shelf1/books/book1"
    string name = 1;
}

message ListBookRequest {
    // Resource name of the parent resource.
    // For example: "shelves/shelf1"
    string parent = 1;
}

message ListBookResponse {
    repeated Book books = 1;
}

```

when generating the swagger output for this the result contain the following problematic statements:

```
...
 "/api/v1/{name=shelves/*/books/*}": {
      "get": {
        "summary": "Book management",
        "operationId": "GetBook",
        "responses": {
          "200": {
            "schema": {
              "$ref": "#/definitions/libraryserviceBook"
            }
          }
        },
...
"/api/v1/{name=shelves/*}": {
      "delete": {
        "operationId": "DeleteShelf",
        "responses": {
          "200": {
            "schema": {
              "properties": {}
            }
          }
        },
...
```

`{name=shelves/*}`and similar is not defined in parameters, therefor the validators fail to validate it. Previously this got rendered to `/api/v1/{name}`, which was defined.

johanbrandhorst:
@ch3rub1m Any ideas for this error? I was looking through our examples but couldn't find one with a name definition like this. I'm tempted to revert this MR as it has broken some users.

johanbrandhorst:
I'll give @ch3rub1m a week to respond, there might be a workaround.

ch3rub1m:
@johanbrandhorst I just knew this problem. I'll try to find a trade-off to solve it.

johanbrandhorst:
Great, thanks for your response. Since this breaks existing users I will roll back this change and reopen the original issue.

ch3rub1m:
@tlyng Hi, I ran your examples in my environment, and I found that there are some errors was complained by validator but it works well for doc generation (by https://github.com/Redocly/redoc).

Could you tell me your real use case? Actually in your use case, if this PR was reverted, you couldn't both get apis `/api/v1/{name=shelves/*/books/*}"` and `/api/v1/{name=shelves/*}` in your swagger file. Because this PR is for fixing that the multiple apis using `resource name style` being translated to same swagger json key, in your example, which one is `/api/v1/name`, so just one of them will be kept.

@johanbrandhorst For the above reasons, I think this PR shouldn't be reverted, because it might fix the warnings of validator and generate a valid swagger but more logical errors back. We should consider another workaround to mitigate the problems of validator instead of reverting this PR.

johanbrandhorst:
As I understand it, it's not just a matter of "warnings in the validator", but not being able to use generators or web UIs that rely on valid swagger files. I think that is a bigger problem right now, although of course I would like to see both issues fixed. I am proposing reverting this PR until we can get a new solution to the duplicate resource generation that still produces valid swagger files.

ch3rub1m:
@johanbrandhorst Understood, how about using encoded slash `%2F` insead of `/` and use whole string `name=users%2F*` as the key of api, like below:

```
    "/v1/{name=users%2F*}": {
        "parameters": [
          {
            "name": "name=users%2F*",
            "in": "path",
            "required": true,
            "type": "string",
            "default": "users"
          }
        ],
...
```
It's a valid swagger and solve the duplicate resource generation even if it doesn't look good.

johanbrandhorst:
I'm happy with that solution, if you would like to draft a competing PR to #1000 I'd be happy to merge that if it fixes @tlyng's problem.

ch3rub1m:
I am working on it. Thank you.

tlyng:
Sounds good enough for me! I'm not personally using swagger documentation for consuming API's, it's a request I've gotten from other developers. Personally I use protobuf , grpc and grpc-web. Providing swagger/openapi reduces the brainwork required by people already used to operate in such environments.

Issue #446, opened 2017-08-21
recursive inputs for GETs with protoc-gen-swagger stalls program


heedson:
When a message used for input into a GET request (with annotations) contains a recursive structure, the swagger generation will lock up and get stuck. Here is a simple (albeit useless) snippet of code that can produce the effect:

```
syntax = "proto3";

// Example
//
// Example is a super important central data store used to... store data.
package example;

import "google/api/annotations.proto";

service ExampleStore {
    rpc GetData(Foo) returns (Bar) {
        option (google.api.http) = {
            get: "/api/v1/example/data"
        };
    }
}

message RecursiveMessage {
    RecursiveMessage rec_msg = 1;
}

message Foo {
    RecursiveMessage a_message_use_case = 1;
}

message Bar {
    int64 fizz = 1;
}
```

The command I was running was:
```
protoc ./example/example.proto --swagger_out=logtostderr=true,Mgoogle/protobuf/timestamp.proto=github.com/gogo/protobuf/types,Mgoogle/protobuf/empty.proto=github.com/gogo/protobuf/types:./ -I/usr/local/include -I./ -I./vendor/github.com/googleapis/googleapis/ -I./vendor/ -I/home/heedson/go/src
```

Interestingly, when `rec_msg` is a `repeated RecursiveMessage` it will once again work.
It will also work if the `option (google...` is changed to a POST.

I'm unsure whether the input to a GET is allowed to be recursive or if it's supposed to graceful exit with an error but irrespective of this, the current implementation does neither.

I believe the problem is at this line https://github.com/grpc-ecosystem/grpc-gateway/blob/master/protoc-gen-swagger/genswagger/template.go#L112 .

achew22:
That's a really interesting example case. Can you describe to me the how you would expect to be calling this in a get?

Also, just to confirm, protoc is hanging, not the swagger generator?

PR #706, opened 2018-07-23
protoc-gen-swagger primitive types in json_schema.type rendering policy

askurydzin:
This PR (preliminary) is an approach to bring a meaning into a json_schema type field definition. The rationale for it is following: if we don't want to expose some internal structure of the message that stands behind some field, we should not provide these details with swagger schema. Also it can be used to solve some corner cases like recursive message definitions via oneof where message itself is only a string on API side and somewhat tree-like on gRPC side. I

So the main features included:

  * protoc-gen-swagger will not render message that has primitive type or references a primitive type (STRING, INT, BOOLEAN, ...)

  * however, it will render message as a definition iff it's included as Request or Response in rpc.

  * example, that include MimicObjectResponse, MimicObjectRequest that are rendered because they are used as response and request messages accordingly, and MimicObjectHidden that is rendered as a string parameter with appropriate Marshal/UnmarshalJSONPB methods to work with it.

```
$ curl -X POST  http://localhost:8080/v1/example/a_bit_of_everything -d '{"mimic_string": "gggggg/  yzzzz/VALUE_ONE"}' | jq

{
  "uuid": "8a90c65f9ff27636c62bde644a7db0aff6f7de715d90b401",
  "mimic_string": "gggggg/yzzzz/VALUE_ONE"
}
```


askurydzin:
#446

johanbrandhorst:
Thanks for this PR @askurydzin, this is really cool and clever! However, is this a breaking change? Should we consider this for v2?

askurydzin:
Thanks, @johanbrandhorst. I would not consider this as a breaking change, as it employs nonsense combination of options (type "string" for message) that was unlikely used by anyone.

johanbrandhorst:
@askurydzin Terribly sorry, looks like we need another rebase and regenerating of the files :(. I'll get it merged as soon as CI passes again.

johanbrandhorst:

PR #708, opened 2018-07-24
Add test case and proposed fix for path component with trailing colon (and string)

### LINKED ISSUE #224: 404s using colons in the middle of the last path segment
I have the following endpoint and user_ids of the format "user:<number>"

```
  rpc GetUser(GetUserRequest) returns (GetUserResponse) {
    option (google.api.http) = {
      get: "/v0/users/{user_id}"
    };
  };
```

I'm getting 404 status code and it doesn't hit my GetUser handler when I call this endpoint with user_ids of the format above. However, if I call the endpoint with a user_id that doesn't have a colon it (e.g. "user123") it does hit my GetUser handler. Url encoding the colon doesn't work either.

What's even more interesting is if I add another colon to the end of the path, grpc-gateway parses the user_id into the correct format that I want. ("/v0/users/user:123:" -> user_id = "user:123"). Given this data point, I believe grpc-gateway is treating colons differently in the last path segment. 

For example here: https://github.com/grpc-ecosystem/grpc-gateway/blob/master/protoc-gen-grpc-gateway/httprule/parse.go#L86 it strips off the last colon in the segment. Though this looks like the rule for parsing the endpoint path definition and not what is being used to parse incoming requests.

jfhamlin:
If a pattern has no verb, match with the verb argument appended to last component.

As I understand the implementation, parsing out a "verb" from the input path is a convenience for matching. Whether the colon and string are _actually_ a verb depends on the matching HTTP rule. This change gives a verb-less pattern a chance to match a path by assuming the colon+string suffix are a part of the final component.

There is a backwards compatibility issue here: this can change the behavior of existing rule sets. For example:
```
    option(google.api.http) = {
      get: "/users/{user_id}"
    };
    option(google.api.http) = {
      get: "/users/{user_id}:averb"
    };
```
Before this change, only the second rule matched a request for `/users/123:averb`. After this change, both rules match (correctly, by my understanding of https://github.com/googleapis/googleapis/blob/master/google/api/http.proto). Note that the spec describes a deterministic way of breaking the tie: `// **NOTE:** All service configuration rules follow "last one wins" order.` grpc-ecosystem/grpc-gateway's implementation looks buggy on this score, since it accepts the first matching handler.

Fixes #224

### DIFF
--- a/runtime/mux_test.go
+++ b/runtime/mux_test.go
@@ -176,6 +176,38 @@ func TestMuxServeHTTP(t *testing.T) {
 			respStatus:  http.StatusOK,
 			respContent: "POST /foo:bar",
 		},
+		{
+			patterns: []stubPattern{
+				{
+					method: "GET",
+					ops:    []int{int(utilities.OpLitPush), 0, int(utilities.OpPush), 0, int(utilities.OpConcatN), 1, int(utilities.OpCapture), 1},
+					pool:   []string{"foo", "id"},
+				},
+			},
+			reqMethod: "GET",
+			reqPath:   "/foo/bar",
+			headers: map[string]string{
+				"Content-Type": "application/json",
+			},
+			respStatus:  http.StatusOK,
+			respContent: "GET /foo/{id=*}",
+		},
+		{
+			patterns: []stubPattern{
+				{
+					method: "GET",
+					ops:    []int{int(utilities.OpLitPush), 0, int(utilities.OpPush), 0, int(utilities.OpConcatN), 1, int(utilities.OpCapture), 1},
+					pool:   []string{"foo", "id"},
+				},
+			},
+			reqMethod: "GET",
+			reqPath:   "/foo/bar:123",
+			headers: map[string]string{
+				"Content-Type": "application/json",
+			},
+			respStatus:  http.StatusOK,
+			respContent: "GET /foo/{id=*}",
+		},
 	} {
 		mux := runtime.NewServeMux()
 		for _, p := range spec.patterns {
--- a/runtime/mux.go
+++ b/runtime/mux.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 
 	"context"
+
 	"github.com/golang/protobuf/proto"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/metadata"
--- a/runtime/pattern.go
+++ b/runtime/pattern.go
@@ -144,7 +144,16 @@ func MustPattern(p Pattern, err error) Pattern {
 // If otherwise, the function returns an error.
 func (p Pattern) Match(components []string, verb string) (map[string]string, error) {
 	if p.verb != verb {
-		return nil, ErrNotMatch
+		if p.verb != "" {
+			return nil, ErrNotMatch
+		}
+		if len(components) == 0 {
+			components = []string{":" + verb}
+		} else {
+			components = append([]string{}, components...)
+			components[len(components)-1] += ":" + verb
+		}
+		verb = ""
 	}
 
 	var pos int


jfhamlin:
fyi @yugui @tmc 

> After this change, both rules match (correctly, by my understanding of https://github.com/googleapis/googleapis/blob/master/google/api/http.proto).

I'll provide a little more color on why I think this is so. Here's the relevant section of http.proto:
```
// The syntax of the path template is as follows:
//
```

This is the syntax of __templates__, not the __paths__ templates are to match. It's actually a template that defines the syntax of the input paths it matches. If a template doesn't include a verb, then any trailing colon and string in an input path must be a part of the final segment.

jfhamlin:
bump on this (and cc @johanbrandhorst since i've seen you chime in on recent PRs 😃)

runtime/mux_test.go
```diff
@@ -176,6 +176,38 @@ func TestMuxServeHTTP(t *testing.T) {
...
+		{
+			patterns: []stubPattern{
+				{
+					method: "GET",
+					ops:    []int{int(utilities.OpLitPush), 0, int(utilities.OpPush), 0, int(utilities.OpConcatN), 1, int(utilities.OpCapture), 1},
+					pool:   []string{"foo", "id"},
+				},
+			},
+			reqMethod: "GET",
+			reqPath:   "/foo/bar:123",
```
achew22: Is the verb in this case "123"? What does it mean to have a verb of 123?

jfhamlin: > Is the verb in this case "123"?

My reading of the HttpRule spec makes me interpret it like this: this path matches the path template in the stub pattern, which has no verb, so "123" isn't a verb (I try to make a case for this reading in https://github.com/grpc-ecosystem/grpc-gateway/pull/708#issuecomment-407520891). Instead, "bar:123" is the captured value of the `id` variable.

To flesh this out further:
- An HTTP rule path template describes a set of paths, basically by a regular expression
- If a request path is matched by an RPC's HTTP rule (regular expression), then it is handled by that RPC, with fields mapped according to the rule
- It follows that if an HTTP rule path template doesn't end with a verb, then it expects no verb, and the trailing path component will be scanned the same way as any other segment or variable

Here's another way to think about it: an alternative implementation of these patterns in grpc-gateway could have been to build up golang `regexp.Regexp`s directly from the HTTP rules. Then when a request comes in, iterate through the regexps and use the last one that matches the path. Example mappings from rule to regular expression (`[^/]*` probably isn't precisely correct, it's just an approximation of ["The syntax `*` matches a single path segment"](https://github.com/googleapis/googleapis/blob/master/google/api/http.proto#L234)).

```
get: "/foo/{id}" => "/foo/(?P<id>[^/]*)"
get: "/foo/{id}:myverb" => "/foo/(?P<id>[^/]*):myverb"
```

jfhamlin: @achew22, does this make sense?

achew22: Ah, I see. You are trying to describe the positive case, and not the negative case as I had interpreted it. That seems entirely reasonable. Merging.

Thanks so much for fighting through my confusion and for contributing to the project as a whole!

kassiansun:
Then, how to gain the old behavior of `{id}:verb`? Breaking the old behavior without any solution?

kassiansun:
ping @jfhamlin @achew22

kassiansun:
Who can tell me what's the meaning of syntax `{id}:verb` after this patch, if it matches the whole segment to `id`?

kassiansun:
Actually it matches to the `{id}` route, which has no trailing `:verb`.

And in #224 , I don't think URN is valid inside URL, they are exclusive.

jfhamlin:
> Who can tell me what's the meaning of syntax {id}:verb after this patch, if it matches the whole segment to id?

Hi @kassiansun. It would be useful in these examples to not only have the path pattern but also some example paths and expected variable binding. For example:

pattern: `/{id}:verb`
path: `/foo:verb`  match binding: `id = foo`
path: `/foo` no match

pattern: `/{id}`
path: `/foo:verb`  match binding: `id = foo:verb`
path: `/foo` match binding: `id = foo`

Does that make sense?

kassiansun:
No, the real behavior is `/{id}:verb` matches `/foo:verb` with `id=foo:verb`. Have you tested this patch?

kassiansun:
#224 makes no sense, users should not use `:` in url as `grpc-gateway` use `:` for verb literals. If they really want to use, use url queries, don't mess with path variable.

PR #711, opened 2018-07-30
Make support paths option

izumin5210:
Make `protoc-gen-grpc-gateway` support `paths` option such like  https://github.com/golang/protobuf/issues/515

### DIFF
--- a/protoc-gen-grpc-gateway/gengateway/generator.go
+++ b/protoc-gen-grpc-gateway/gengateway/generator.go
@@ -19,15 +19,23 @@ var (
 	errNoTargetService = errors.New("no target service defined in the file")
 )
 
+type pathType int
+
+const (
+	pathTypeImport pathType = iota
+	pathTypeSourceRelative
+)
+
 type generator struct {
 	reg                *descriptor.Registry
 	baseImports        []descriptor.GoPackage
 	useRequestContext  bool
 	registerFuncSuffix string
+	pathType           pathType
 }
 
 // New returns a new generator which generates grpc gateway files.
-func New(reg *descriptor.Registry, useRequestContext bool, registerFuncSuffix string) gen.Generator {
+func New(reg *descriptor.Registry, useRequestContext bool, registerFuncSuffix, pathTypeString string) gen.Generator {
 	var imports []descriptor.GoPackage
 	for _, pkgpath := range []string{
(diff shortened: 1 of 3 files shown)


izumin5210:
@googlebot I signed it!

izumin5210:
@johanbrandhorst I fixed the message! 😄

johanbrandhorst:
@achew22 @tmc what's going on with CI here?

achew22:
@johanbrandhorst Very strange. We used to have this kind of problem all the time. It was from proto-go being upgraded and changing the output format. When there are no env variables set, we compare the compiled versions of the `.pb.go` to the version that is checked in for equality (golden testing). 

However, with the submission of #696, I wouldn't expect that to happen any more. I don't know a lot about dep unfortunately. I was hoping we could just wait till vgo, but I'm glad we did this. Looking at [this line](https://github.com/johanbrandhorst/grpc-gateway/blob/7d1b84fe44c5b6600819d3963f8befa0f85c540b/.travis.yml#L18) is it possible that we are overriding our go dep and explicitly installing the HEAD version of proto-go?

The other two failures appear to just be standard tooling errors. If we paid for Travis I bet it would be more stable, but the free version is kind of flaky. I re-triggered them and they passed.

achew22:
@johanbrandhorst, I wasn't meaning to imply that line specifically, but is there something in that file you would expect is updating our dependencies to head instead of the vendored versions?

achew22:
I fixed it by manually applying the patch. Could you please rebase this PR to see how Travis feels about it now? Thanks!

izumin5210:
I rebased!

But CI failed again 😢 
Could you rerun them?

> ```
> 15.27s$ test "${USE_BAZEL}" = true || go get github.com/golang/lint/golint
> package golang.org/x/lint: unrecognized import path "golang.org/x/lint" (https fetch: Get https://golang.org/x/lint?go-get=1: net/http: TLS handshake timeout)
> The command "test "${USE_BAZEL}" = true || go get github.com/golang/lint/golint" failed and exited with 1 during .
> ```

https://travis-ci.org/grpc-ecosystem/grpc-gateway/jobs/411080943#L557-L561

johanbrandhorst:
Looks like it's all working again, thanks for your contribution!

Issue #707, opened 2018-07-24
FR: Promote a field in the returned JSON message to a top-level returned value


ivucica:
Often pre-existing APIs return a list as a top-level JSON entity. Sometimes, the returned value may be a string or a number.

Example of an API that returns an array is Mastodon's [Getting who an account is following](https://github.com/tootsuite/documentation/blob/master/Using-the-API/API.md#getting-who-account-is-following) call. I've run into other APIs with the same issue.

While I believe the default behavior should be as-is, It would be useful to allow promoting a field of a message to be the actual value of a message.

There are multiple ways to do it; we could add an option to a particular RPC, and we could add a grpc-gateway-specific option to a message. I'm leaning towards attaching it to a message as that allows doing this multiple times.

Proposing the following:

```
service Accounts {
  rpc GetAccountFollowing (GetAccountFollowingRequest) returns (AccountArray) {
    option (google.api.http) = {
      get: "/accounts/{id}"
    };
  }
  message AccountArray {
    repeated Account account = 1 [(grpc.gateway.options).unwrap_field = true]; // or promote_field?
  }

  message Account {
    string id = 1;
    string username = 2;
    string display_name = 3;
  }
```

This should replace the following default behavior:

```
{
  "account": [
    {
      "id": "1",
      "username": "test",
      "display_name": "Esther T."
    },
    {
      "id": "2",
      "username": "test2",
      "display_name": "Esther T."
    }
  ]
}
```

with the following:

```
[
  {
    "id": "1",
    "username": "test",
    "display_name": "Esther T."
  },
  {
    "id": "2",
    "username": "test2",
    "display_name": "Esther T."
  }
]
```

Implementing the filter in the gRPC-Gateway serving binary is reasonably trivial. I've taken `net/http/httptest`, recorded the output, unmarshalled the JSON into a `map[string]interface{}`, taken the `account` key, encoded it into JSON and then output it as the response. It's tolerable.

But it is a hack, and it doesn't affect the `protoc-gen-swagger`-generated definition.

Since APIs that return arrays as a top level entity, it would be excellent if we could agree on an `option`-based design that would be respected by both `protoc-gen-grpc-gateway` and by `protoc-gen-swagger`.

Is this already possible? Did I miss something? Did I miss an open issue tracking this feature request? Thoughts on where the `option` would belong -- on the `rpc` or on the `message`?

achew22:
[...]

PR #712, opened 2018-07-30
Support HttpRule with field response

doroginin:
This PR adds support optional `response_body` field for google.api.HttpRule. Also see PR: https://github.com/google/go-genproto/pull/87 and https://github.com/googleapis/googleapis/pull/512

This version of code can work with current `HttpRule` implementation and with forked implementation with new field. This implementation does not break existing code.

Also `allow_repeated_fields_in_body` flag added to make it possible using repeated fields in `body` and `response_body`. This flag can be used for slices in request and response.

Also flag `json_name_in_swgdef` added to make it possible to use Field.GetJsonName() instead of Field.GetName() for generating swagger definitions. See: https://github.com/grpc-ecosystem/grpc-gateway/pull/681 for more info.


johanbrandhorst:
Ha, this seems to be exactly what @ivucica wanted in #707. Looking at https://github.com/googleapis/googleapis/blob/201d7be7f9da925df93bc52f8108963284f61aed/google/api/http.proto, this doesn't seem to be an official part of the `http.proto` file? Is this in the process of being merged? I don't think there's any chance of us supporting something that isn't part of the official spec.

ivucica:
I have not yet received upstream response for #707.

I am opposed to merging this particular PR, as it may _(cough cough)_ collide with upstream fields which are not publicly visible.

I am also opposed to forking `google.api.HttpRule`; if we want to introduce something like this, it should be a new option namespaced to gRPC-Gateway. (If upstream decides to support the equivalent field, we can switch at that point, but it should be close to what upstream supports.)

In that light, I suggest careful examination of what's currently present in [google.api.HttpRule documentation](https://cloud.google.com/service-management/reference/rpc/google.api#google.api.HttpRule), particularly look at `response_body`. I'd suggest comparing the comment that @doroginin put into [his fork of http.proto](https://github.com/doroginin/googleapis/blob/5539ecc567a5ddc1ee99c309f4c3530deeabcca1/google/api/http.proto#L302) which bears a resemblance to the documentation of certain fields missing from the public .proto file.

ivucica:
This is improved, but if you do receive a response on https://github.com/googleapis/googleapis/pull/512, I suspect it will be pushback.

protoc-gen-grpc-gateway/main.go
```diff
@@ -22,12 +22,14 @@ import (
...
+	importPrefix              = flag.String("import_prefix", "", "prefix to be added to go package paths for imported proto files")
+	importPath                = flag.String("import_path", "", "used as the package if no input files declare go_package. If it contains slashes, everything up to the rightmost slash is ignored.")
+	registerFuncSuffix        = flag.String("register_func_suffix", "Handler", "used to construct names of generated Register*<Suffix> methods.")
+	useRequestContext         = flag.Bool("request_context", true, "determine whether to use http.Request's context or not")
+	allowDeleteBody           = flag.Bool("allow_delete_body", false, "unless set, HTTP DELETE methods may not have a body")
+	grpcAPIConfiguration      = flag.String("grpc_api_configuration", "", "path to gRPC API Configuration in YAML format")
+	allowRepeatedFieldsInBody = flag.Bool("allow_repeated_fields_in_body", false, "allows to use repeated field in `body` field of `google.api.http` annotation")
+	useJSONNameInSwaggerDef   = flag.Bool("json_name_in_swgdef", false, "if it sets Field.GetJsonName() will be used for generating swagger definitions, otherwise Field.GetName() will be used")
```
ivucica: Why is this flag only affecting the swagger definition? Won't this break the api generated with `protoc-gen-grpc-gateway`?

If this affects only `protoc-gen-swagger`, why is this a flag in `protoc-gen-grpc-gateway`?

doroginin: You're right, this flag should be in protoc-gen-swagger, i'll move it.

doroginin:
@ivucica could you help with this failed job, please: https://travis-ci.org/grpc-ecosystem/grpc-gateway/jobs/410260132 ? What's wrong here?

protoc-gen-grpc-gateway/main.go
doroginin: >2. Please consider the very, very probable outcome where you won't be able to merge the upstream google.api.HttpRule change -- can you reserve a new set of proto options in the option registry for grpc-gateway (in addition to the existing swagger-specific ones) and implement this using that?

@ivucica do you mean implement option like this:
```
message AccountArray {
    repeated Account account = 1 [(grpc.gateway.options).unwrap_field = true]; // or promote_field?
}
```

I think it's less obvious solution than `response_body` field, and in this case you can't add additional bindings with different requests/responses mapping for example, how do you implement this with you solution?

```
service Strings {
    rpc ToUpper (String) returns (String) {
        option (google.api.http) = {
            post: "/strings/to_upper"
            body: "str"
            response_body: "str"
            additional_bindings: {
                post: "/strings/to_upper/v2",
                body: "*",
                response_body: "*",
            }
        };
    }
}

message String {
    repeated string str = 1;
}
```

ivucica: That was my original proposal, but I am fine with attaching it to `rpc` instead of `message`. We only need to move it to a proto field we actually control.

It is correct that you cannot add a different `response_body` with what I am proposing:

```
service Strings {
    rpc ToUpper (String) returns (String) {
        option (google.api.http) = {
            post: "/strings/to_upper"
            body: "str"
            additional_bindings: {
                post: "/strings/to_upper/v2",
                body: "*",
            }
        };
        option (grpc.gateway.rpc_options) = {
            response_body: "str"
        };
    }
}
```

That is indeed unfortunate, but I see `additional_bindings` as a bonus, and for the most part, I just use a 1:1 mapping.

Anything I can think of is ugly, including this possibly tolerable addition (which is not incompatible with what's above):

    rpc ToUpper (String) returns (String) {
        option (google.api.http) = {
            post: "/strings/to_upper"
            body: "str"
            additional_bindings: {
                post: "/strings/to_upper/v2",
                body: "*",
            }
        };
        option (grpc.gateway.rpc_options) = {
            configure_rule: {
                rule: { post: "/strings/to_upper" }
                config: { response_body: "str" }
            }
            configure_rule: {
                rule: { post: "/strings/to_upper/v2" }
                config: { response_body: "*" }
            }
        };
    }


To get what you want to happen, we _do_ need to get upstream to release the `response_body`, which may be difficult.

Once again: I am nearly certain your `.HttpRule` PR will not be accepted, due to insights that I cannot detail here without approval. So, once again: you should explore alternative designs unless upstream decides to provide the field.

I strongly suspect `grpc-gateway` will not move to a forked `google.api.HttpRule` either (especially one that may long-term conflict with upstream developments).

ivucica:
Some ongoing CI issues: https://github.com/grpc-ecosystem/grpc-gateway/pull/711#issuecomment-408860071

I would not be worried and, were this PR otherwise mergable, I would not object to merging it.

protoc-gen-grpc-gateway/main.go
ivucica: I propose we take design discussions into #707 instead of this particular PR?

wora:
Please document why this feature is needed. We intentionally keep gRPC transcoding extremely simple. The goal for gRPC transcoding is to build REST API from RPC API. We don't intend to be feature rich as OpenAPI. We generally don't need feature to build great and useful APIs. The less features, the better performance and quality.

ivucica:
@wora See #707 for my own feature request. As I said there, some *preexisting* APIs, such as Mastodon's, cannot currently be described with `google.api.HttpRule` as they return a JSON list as a top level object. With this trivial change, they could be.

Without this change, when attempting to describe Mastodon's existing API, one would

1. have to postprocess the JSON before returning it to the user,
1. and also have to postprocess output of `protoc-gen-swagger` so it documents returning a list instead of a message

While that's somewhat tolerably doable, it's way hackier than just allowing users to specify which field should be the top level JSON entity.

I do believe designing APIs like so is a bad design, but that's better to be strongly discouraged in documentation and in the [API Design Guide](https://cloud.google.com/apis/design/).

@doroginin might have another use.

doroginin:
> @doroginin might have another use.

We use `protoc-gen-swagger` in https://github.com/utrack/clay project (in this project we generate http handlers from proto file and http client for them), also we generate `swagger.json` via protoc-gen-swagger.
In this pull request: https://github.com/utrack/clay/pull/39 i added test which uses the feature from this pull request.
Here is the example of usage:
https://github.com/doroginin/clay/tree/871583110ffcbf9093c8aa429da1433557a4790e/integration/binding_with_body_and_response

wora:
I am publishing the change from upstream. It should land in a few days.

protoc-gen-grpc-gateway/descriptor/registry.go
```diff
@@ -46,6 +46,13 @@ type Registry struct {
 
 	// mergeFileName target swagger file name after merge
 	mergeFileName string
+
+	// allowRepeatedFieldsInBody permits repeated field in body field path of `google.api.http` annotation option
+	allowRepeatedFieldsInBody bool
+
+	// useJSONNameInSwaggerDef if it true Field.GetJsonName() will be used for generating swagger definitions,
+	// otherwise Field.GetName() will be used
+	useJSONNameInSwaggerDef bool
```
ivucica: Please rename this into `useJSONNamesForFields` and update the comment appropriately :-)

Separately, can you then make `protoc-gen-grpc-gateway` use the JSON names as well if the flag is set? If not, can you please add a TODO here for someone to take a stab at it. It would be long-term important that `-swagger` defines the same API that `-grpc-gateway` exposes.

protoc-gen-grpc-gateway/descriptor/registry.go
```diff
@@ -318,6 +325,22 @@ func (r *Registry) SetMergeFileName(mergeFileName string) {
...
+func (r *Registry) SetAllowRepeatedFieldsInBody(allow bool) {
+	r.allowRepeatedFieldsInBody = allow
+}
+
+// SetUseJSONNameInSwaggerDef sets useJSONNameInSwaggerDef
+func (r *Registry) SetUseJSONNameInSwaggerDef(use bool) {
+	r.useJSONNameInSwaggerDef = use
+}
+
+// GetUseJSONNameInSwaggerDef whether generation one swagger file out of multiple protos
```
ivucica: This comment is incorrect and nonsensical*, could you please fix it? :)

* Its grammar is nonsensical even in `IsAllowMerge`, but here the comment itself is outright wrong.

protoc-gen-grpc-gateway/descriptor/registry.go
```diff
@@ -318,6 +325,22 @@ func (r *Registry) SetMergeFileName(mergeFileName string) {
 	r.mergeFileName = mergeFileName
 }
 
+// SetAllowRepeatedFieldsInBody controls whether repeated field can be used in body field path
```
ivucica: Given you are providing a getter for `useJSONNameInSwaggerDef`, did you mean to provide it for `allowRepeatedFieldsInBody`?

protoc-gen-grpc-gateway/descriptor/types.go
```diff
@@ -227,10 +234,10 @@ func (p Parameter) ConvertFuncExpr() (string, error) {
 	return conv, nil
 }
 
-// Body describes a http requtest body to be sent to the method.
+// Body describes a http request(response) body to be sent to the method(client).
```
ivucica: How about the bash syntax:

```
// Body describes a http {request,response} body to be sent to the {method,client}.
```

or the regex syntax:

```
// Body describes a http (request|response) body to be sent to the (method|client).
```

Maybe expand the comment a bit, and mention this is used in `body` and `response_body` options in `google.api.HttpRule`?

protoc-gen-swagger/genswagger/template.go
```diff
@@ -625,6 +631,23 @@ func renderServices(services []*descriptor.Service, paths swaggerPathsObject, re
 
 				methProtoPath := protoPathIndex(reflect.TypeOf((*pbdescriptor.ServiceDescriptorProto)(nil)), "Method")
 				desc := ""
+				var responseSchema swaggerSchemaObject
+
+				if b.Response == nil || len(b.Response.FieldPath) == 0 {
```
ivucica: Add a comment that this is resolving the value of `response_body` in the `google.api.HttpRule`

doroginin:
@ivucica 
It's success! `http.proto` was updated in main repo: https://github.com/googleapis/googleapis/blob/master/google/api/http.proto#L303 :+1: 
We just need to wait when this PR will be merged: https://github.com/google/go-genproto/pull/87

doroginin:
Now i'm preparing some tests for changes in this PR

ivucica:
I previously asked:

> Separately, can you then make `protoc-gen-grpc-gateway` use the JSON names as well if the flag is set? If not, can you please add a TODO here for someone to take a stab at it. It would be long-term important that `-swagger` defines the same API that `-grpc-gateway` exposes.

If you don't plan on updating protoc-gen-grpc-gateway to respect JSON names flag, could you at least add a TODO for someone else to do it? Thanks

doroginin:
> Separately, can you then make protoc-gen-grpc-gateway use the JSON names as well if the flag is set? If not, can you please add a TODO here for someone to take a stab at it. It would be long-term important that -swagger defines the same API that -grpc-gateway exposes.

I think it's not needed. `grpc-gateway` uses json names or original name in request/response according with used marshaler which will created by user. So user can create marshaler as he wants and set corresponding option in `protoc-gen-swagger` for generating definition.

I added more comments about it for new flag.

protoc-gen-grpc-gateway/descriptor/registry.go
```diff
@@ -50,8 +50,12 @@ type Registry struct {
...
 	allowRepeatedFieldsInBody bool
 
-	// useJSONNamesForFields if it true Field.GetJsonName() will be used for generating swagger definitions,
-	// otherwise Field.GetName() will be used
+	// useJSONNamesForFields if true json tag name is used for generating fields in swagger definitions,
+	// otherwise original proto name is used. It's helpful for synchronization swagger definition with grpc-gateway
+	// response, if it uses json tags for marshaling.
+	// Pay attention. By default grpc-gataway uses original proto name in responses. If you would like
+	// to use names from json tags change marshaller for gateway with runtime.ServeMuxOption
+	// runtime.WithMarshalerOption(runtime.MIMEWildcard, &runtime.JSONPb{OrigName: false})
```
johanbrandhorst: While this is helpful, I don't like the tone of this. How about:
```
// otherwise the original proto name is used. It's helpful for synchronizing the swagger definition with grpc-gateway
// response, if it uses json tags for marshaling.
```

doroginin:
@ivucica can you help to fix bazel test: https://travis-ci.org/grpc-ecosystem/grpc-gateway/jobs/414555020 ? I have never work with this stuff. We need to update version of google.golang.org/genproto to 383e8b2c3b9e36c4076b235b32537292176bae20 how can i do it?

Gopkg.toml
```diff
@@ -51,7 +51,7 @@ required = [
 [[constraint]]
   # Also defined in bazelbuild/rules_go
   # https://github.com/bazelbuild/rules_go/blob/436452edc29a2f1e0edc22d180fbb57c27e6d0af/go/private/repositories.bzl#L131
-  revision = "86e600f69ee4704c6efbf6a2a40a5c10700e76c2"
+  revision = "383e8b2c3b9e36c4076b235b32537292176bae20"
```
johanbrandhorst: Will this change require a corresponding change to `bazelbuild/rules_go`?

johanbrandhorst: I see you already managed to confirm we will need this change.

johanbrandhorst:
@doroginin maybe https://github.com/grpc-ecosystem/grpc-gateway/blob/master/repositories.bzl#L31?

doroginin:
@johanbrandhorst 
we need at least this version of googleapis: https://github.com/googleapis/googleapis/commit/3544ab16c3342d790b00764251e348705991ea4b (here is `http.proto`)
and this version of google.golang.org/genproto: https://github.com/google/go-genproto/commit/383e8b2c3b9e36c4076b235b32537292176bae20 (here is `http.pb.go`)

protoc-gen-grpc-gateway/descriptor/registry.go
doroginin: done!

doroginin:
Here i changed `repositories.bzl`: https://github.com/doroginin/grpc-gateway/commit/f4300d0389f5ce363344ff405903d86e634f0283
But it is still not working: https://travis-ci.org/doroginin/grpc-gateway/jobs/414800225

johanbrandhorst:
I'm not personally super familiar with Bazel so I don't know about that error, maybe @achew22 or @yugui can see what's left?

Gopkg.toml
johanbrandhorst: Can you update this comment as well since this is no longer correct?

johanbrandhorst:
Please squash the commits into a single commit with a nice descriptive commit message as well, in preparation for merge.

doroginin:
@johanbrandhorst squashed!

doroginin:
@johanbrandhorst, I think we need to update https://github.com/bazelbuild/rules_go/blob/c73c25f2ead2b6d2cf772878e88a650f07a3de7c/go/private/repositories.bzl#L141 from daca94659cb50e9f37c1b834680f2e46358f10b0 to 383e8b2c3b9e36c4076b235b32537292176bae20
Who can help with it?

johanbrandhorst:
@doroginin I think we just have to open an issue or submit a PR against https://github.com/bazelbuild/rules_go to have the version upgraded.

doroginin:
@johanbrandhorst @ivucica, I fixed bazel tests. Now it's green! Could you merge this PR, if you don't have any other requests?

Gopkg.toml
johanbrandhorst: This comment still needs updating

johanbrandhorst:
@doroginin See my comment on the Gopkg.toml changes. Can you also make a reference in the WORKSPACE file to Gopkg.toml?

Gopkg.toml
doroginin: Done!

WORKSPACE
```diff
@@ -12,17 +12,16 @@ http_archive(
...
-load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies")
-
-gazelle_dependencies()
-
-load("@io_bazel_rules_go//go:def.bzl", "go_repository")
+go_repository(
+    name = "org_golang_google_genproto",
+    commit = "383e8b2c3b9e36c4076b235b32537292176bae20",
+    importpath = "google.golang.org/genproto",
+)
```
johanbrandhorst: Sorry, could you reference `Gopkg.toml` here as well? These two must be kept in sync, obviously.

johanbrandhorst: Really these should all reference Gopkg.toml :thinking:

doroginin: Done!

johanbrandhorst:
This LGTM, but I'll leave it for @ivucica to merge as he's more involved than I with this. Thanks for your contribution @doroginin!

doroginin:
Guys, any update here?
@ivucica we need your approve!

johanbrandhorst:
@doroginin Thanks for your patience, I've reached out to Ivan, we'll try and get this merged as soon as possible :).

ivucica:
I'd like that too; I reopened the internal bug.

bcdurden:
Can we roll backwards? This just broke all of my builds that depend on this project.

johanbrandhorst:
@bcdurden Fair enough, we'll sort this asynchronously.

bcdurden:
Much appreciated!  I should have read upwards a bit before I commented anyhow :100:

ivucica:
@bcdurden Upstream has *already* merged all the changes! Can you verify that your projects are no longer broken and that I can drop PR #730?

bcdurden:
All good! Thanks for the quick turn-around!

doroginin:
Thank you for merging this, guys! Is there any plan to set tag for this version?

doroginin:
And let's restart the build on master branch, cause now it's errored!

johanbrandhorst:
@doroginin See https://github.com/grpc-ecosystem/grpc-gateway/issues/733

PR #687, opened 2018-06-28
protoc-gen-swagger: support JSON Schema Validation properties and add openapiv2_field option

co3k:
This change supports JSON Schema Validation properties.
It allows you to specify some validation rules in `.proto` file like this:

```proto
message GeoCoordinate {
  option (grpc.gateway.protoc_gen_swagger.options.openapiv2_schema) = {
    json_schema: {
      required: ["latitude", "longitude"]
    }
  };

  string latitude = 1;
  string longitude = 2;
}
```

And also this supports openapiv2_field field option.
It allows you to specifiy additional JSON schema setting like this:

```proto
string latitude = 1 [(grpc.gateway.protoc_gen_swagger.options.openapiv2_field) = {pattern: "^[\-]?[0-9]{0,2}\.[0-9]+$"}];
string longitude = 2 [(grpc.gateway.protoc_gen_swagger.options.openapiv2_field) = {pattern: "^[\-]?[0-9]{0,3}\.[0-9]+$"}];
```

The both of `grpc.gateway.protoc_gen_swagger.options.openapiv2_schema` and `grpc.gateway.protoc_gen_swagger.options.openapiv2_field` now supports the following validation properties.

* multiple_of
* maximum
* exclusive_maximum
* minimum
* exclusive_minimum
* max_length
* min_length
* pattern
* max_items
* min_items
* unique_items
* max_properties
* min_properties
* required

Please see also [JSON Schema Validation spec](https://tools.ietf.org/html/draft-fge-json-schema-validation-00) to understand about these properties.

PR #716, opened 2018-08-02
support to add list of required fields to objects generated in swagger.json

srikanth6k:
In .proto file for each message we can add openapiv2_schema.json_schema option to generate some metadata about the message in swagger.json. As of now only title and description are supported. This change enables them to generate a list of required fields.


johanbrandhorst:
Looks to me like support for this was merged with https://github.com/grpc-ecosystem/grpc-gateway/pull/687

PR #718, opened 2018-08-02
Add message field to the error message emitted by grpc-gateway

ffredsh:
This is to make it more consistent with the Status proto message in order to make the error more consistent between grpc and json.

We currently use grpc-gateway for local developement and ESP in production and having to deal with the message in two different fields is a bit annoying. This makes the behaviour of grpc-gateway a bit more consistent with ESP.

### DIFF
--- a/runtime/errors.go
+++ b/runtime/errors.go
@@ -66,8 +66,9 @@ var (
 
 type errorBody struct {
 	Error   string     `protobuf:"bytes,1,name=error" json:"error"`
-	// This is to make the error more compatible with users that expect errors to be Status objects. It should be
-	// the exact same message as the Error field.
+	// This is to make the error more compatible with users that expect errors to be Status objects:
+	// https://github.com/grpc/grpc/blob/master/src/proto/grpc/status/status.proto
+	// It should be the exact same message as the Error field.
 	Message string     `protobuf:"bytes,1,name=message" json:"message"`
 	Code    int32      `protobuf:"varint,2,name=code" json:"code"`
 	Details []*any.Any `protobuf:"bytes,3,rep,name=details" json:"details,omitempty"`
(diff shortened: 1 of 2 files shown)


runtime/errors.go
```diff
@@ -66,6 +66,9 @@ var (
 
 type errorBody struct {
 	Error   string     `protobuf:"bytes,1,name=error" json:"error"`
+	// This is to make the error more compatible with users that expect errors to be Status objects. It should be
```
achew22: Could you add a link to https://github.com/grpc/grpc/blob/master/src/proto/grpc/status/status.proto?

ffredsh: done!

ti:
this is really ugly and ？ how to fix "error" and "message" repeat problem ?
before this code:
```json
{
    "error": "something error",
    "code": 3
}
```
after:

```json
{
    "error": "something error",
    "message": "something error",
    "code": 3
}
```

should web restore the code?

ti:
you can just use to fix it before:
```
type Error struct {
	Message string   `json:"error,omitempty"`
```

@ffredsh

johanbrandhorst:
I agree, but unfortunately we have to keep the old `"error"` field for backwards compatibility purposes. We'll want to remove it come 2.0.

ti:
@johanbrandhorst web should not remove the "error" field in http response json, the "error" field is defined for compatible for some rfc documents, 
For exp:
https://tools.ietf.org/html/rfc6749#page-45

more:
the "error" means some  "enmu text" for the front apps, but he "message" just means some error details, if you want you should and the "message" to the 2.0

@ffredsh

ti:
This change if really affect our current restful document for open apis 😢

ti:
@ffredsh

PR #719, opened 2018-08-07
Fix naming convention of JSON Schema didn't matched with the spec

co3k:
e.g. min_length should be converted as minLength in Swagger spec, but it wasn't.

please see also the [schemaObject definition of OpenAPI v2 spec](https://github.com/OAI/OpenAPI-Specification/blob/3.0.0/versions/2.0.md#schemaObject)

This bug is introduced by #687 changes.


johanbrandhorst:
Doh!

johanbrandhorst:
Could you just include a link to the spec in the commit message and description as well? Good for future reference.

ivucica:
Oops. I did not realize @johanbrandhorst requested a change. I only saw the change was approved. *sigh*

PR #724, opened 2018-08-15
Lock versions to tags where possible

johanbrandhorst:
Using an exact revisions in `Gopkg.toml` prevents `dep` from detecting equivalence between constraints in the case where one dependency depends on, for example, golang/protobuf v1.1.0 and another depends on the revision that is tagged with that version. Using tagged versions should help with this.


johanbrandhorst:
This change does not introduce any version changes, it just relaxes the versions that downstream dependants can use to `1.x` and `1.x` for `golang/protobuf` and `grpc`.

Issue #322, opened 2017-03-06
Support Enum Types in Path


johanbrandhorst:
Having merged Enum support into the query parameters, would it be possible to now support Enum Types in the path as well? Currently this appears to be limited by https://github.com/grpc-ecosystem/grpc-gateway/blob/fdcb7eab2d6ed4919e83ddcef4f199c161611e1f/protoc-gen-grpc-gateway/descriptor/types.go#L293 not being implemented for Enum types.

patrickwmcgee:
Hello wonderful `grpc-gateway` maintainers! I was wondering what the state of this issue is. Currently the `Readme.md` has the line
 > Enum fields in path parameter (including repeated enum fields). 

This made me think I could use `Enum` fields in the request path via options like
```
rpc Echo(MyMessage) returns (SimpleMessage) {
	option (google.api.http) = {
		post: "/v1/example/echo/{my_enum}"
	};
}
```
Where `my_enum` is an `Enum` that is used in `MyMessage`.

If this is not the case, perhaps the `Readme.md` could be updated?

johanbrandhorst:
The wording might not be great, enums are supported in _query_ parameters of the URL, not the path.

PR #738, opened 2018-08-28
Add support for enum path parameters

maros7:
This is related to #322. I've taken a stab at implementing enum type path parameter support.

I have updated protoc-gen-swagger to render enum path parameters correctly. However, since swagger-codegen-cli 2.2.2 doesn't support generating go enums (it was added in a later version, see https://github.com/swagger-api/swagger-codegen/pull/5635) I also had to add some helper functions to make the tests pass correctly.

protoc-gen-grpc-gateway was of course updated to handle enum path parameters. The runtime enum convert function will return an int32 that will be type casted to an enum, which is generated via the update to the template.

Any feedback on my take on this is appreciated.


examples/proto/examplepb/a_bit_of_everything.proto
```diff
@@ -226,9 +229,8 @@ service ABitOfEverythingService {
 	//
 	// This API creates a new ABitOfEverything
 	rpc Create(ABitOfEverything) returns (ABitOfEverything) {
-		// TODO add enum_value
 		option (google.api.http) = {
-			post: "/v1/example/a_bit_of_everything/{float_value}/{double_value}/{int64_value}/separator/{uint64_value}/{int32_value}/{fixed64_value}/{fixed32_value}/{bool_value}/{string_value=strprefix/*}/{uint32_value}/{sfixed32_value}/{sfixed64_value}/{sint32_value}/{sint64_value}/{nonConventionalNameValue}"
+			post: "/v1/example/a_bit_of_everything/{float_value}/{double_value}/{int64_value}/separator/{uint64_value}/{int32_value}/{fixed64_value}/{fixed32_value}/{bool_value}/{string_value=strprefix/*}/{uint32_value}/{sfixed32_value}/{sfixed64_value}/{sint32_value}/{sint64_value}/{nonConventionalNameValue}/{enum_value}/{path_enum_value}/{nested_path_enum_value}"
```
johanbrandhorst: :100:

protoc-gen-swagger/genswagger/template.go
```diff
@@ -594,8 +595,13 @@ func renderServices(services []*descriptor.Service, paths swaggerPathsObject, re
 							return fmt.Errorf("only primitive and well-known types are allowed in path parameters")
 						}
 					case pbdescriptor.FieldDescriptorProto_TYPE_ENUM:
-						paramType = fullyQualifiedNameToSwaggerName(parameter.Target.GetTypeName(), reg)
+						paramType = "string"
```
johanbrandhorst: It seems the handler below handles both the numerical and string representations of an enum, should we really hardcode this to string?

runtime/convert.go
```diff
@@ -85,3 +86,23 @@ func Duration(val string) (*duration.Duration, error) {
 	err := jsonpb.UnmarshalString(val, r)
 	return r, err
 }
+
+// Enum converts the given string into an int32 that should be type casted into the
+// correct enum proto type.
+func Enum(val string, enumValMap map[string]int32) (int32, error) {
+	e, ok := enumValMap[val]
```
johanbrandhorst: This lookup means we define the correct enum values as the strings generated by the go Protobuf implementation. Is this appropriate? It'll be things like `MyEnum_FIRST_VALUE` etc. Is there some way we can use the values from the protofile directly (i.e. `FIRST_VALUE`). I like having the protofile content as the source of truth.

protoc-gen-swagger/genswagger/template.go
maros7: I agree. I chose to use the same logic as already existed for enums in query params, see https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/query.go#L339-L356. Do you know the rationale of supporting numerical -> string? From my perspective string support is sufficient.

runtime/convert.go
maros7: Not sure I follow. The convert Enum function will use the name->index map that is generated by the proto compiler, see e.g. [grpc-gateway/examples/proto/examplepb/a_bit_of_everything.pb.go]( https://github.com/maros7/grpc-gateway/blob/enum-path-parameter/examples/proto/examplepb/a_bit_of_everything.pb.go#L48-L51). That map is provided to the convert Enum function when conversion from path parameters to protobuf is done, see e.g. [grpc-gateway/examples/proto/examplepb/a_bit_of_everything.pb.gw.go](https://github.com/maros7/grpc-gateway/blob/enum-path-parameter/examples/proto/examplepb/a_bit_of_everything.pb.gw.go#L216-L253). So not sure exactly what you mean.

johanbrandhorst: Ah, my bad, the values aren't prefixed. Carry on!

johanbrandhorst:
This LGTM, @achew22, @ivuvica want to have a look before I merge?

johanbrandhorst:
Thanks for your contribution!

PR #757, opened 2018-09-23
Update README.md

wora:
Point users to the proto definition instead of Service Management API documentation. Latter is generated from former, and is more likely to be out of date.

Rephrase "custom options" to "google.api.http" annotation to be more straight forward.


achew22:
@wora, thank you so much for your contribution! These documentation fixes are some of my favorite kinds of changes. It's so hard to look at some text you know very well and determine what could be improved about it. Thanks for the extra set of eyes!

PR #761, opened 2018-10-01
Revert #708 since it breaks backwards compatibility

### LINKED ISSUE #760: URLs using verb no longer work after upgrading to v1.5.0
We tested out the latest 1.5.0 release and noticed it broke our URLs that use colons with a custom verb, like described here: https://cloud.google.com/apis/design/custom_methods

In our case we have a both a "Get" method and a custom method sharing parts of the URL, like this:

```
  rpc GetDevice(GetDeviceRequest) returns (Device) {
    option (google.api.http) = { get: "/v2/{name=projects/*/devices/*}" };
  }
  rpc StreamDevices(StreamDevicesRequest) returns (stream StreamDevicesResponse) {
      option (google.api.http) = { get: "/v2/{name=projects/*/devices/*}:stream" };
   }
```

When upgrading from v1.4.1 to v1.5.0 requests no longer seem to reach the `StreamDevices` method.

By extending the `TestMuxServeHTTP` test with the following test-case, I'm able to trigger the same issue:

```
{
			patterns: []stubPattern{
				{
					method: "GET",
					ops:    []int{int(utilities.OpLitPush), 0, int(utilities.OpPush), 0, int(utilities.OpConcatN), 1, int(utilities.OpCapture), 1},
					pool:   []string{"foo", "id"},
				},
				{
					method: "GET",
					ops:    []int{int(utilities.OpLitPush), 0, int(utilities.OpPush), 0, int(utilities.OpConcatN), 1, int(utilities.OpCapture), 1},
					pool:   []string{"foo", "id"},
					verb:   "verb",
				},
			},
			reqMethod: "GET",
			reqPath:   "/foo/bar:verb",
			headers: map[string]string{
				"Content-Type": "application/json",
			},
			respStatus:  http.StatusOK,
			respContent: "GET /foo/{id=*}:verb",
		},
```

I would expect the test to reach the `/foo/{id=*}:verb` pattern, or in our case reach the `StreamDevices` method as in the previous releases.

It seems to be a side-effect of fixing #224.

johanbrandhorst:
The solution for #224 turned out to break backwards compatibility, so we're going to have to find another
solution for users who desire this behaviour.

Also adds test cases from #760.

Fixes #760.

### DIFF
--- a/runtime/mux.go
+++ b/runtime/mux.go
@@ -1,10 +1,10 @@
 package runtime
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 	"strings"
-	"context"
 
 	"github.com/golang/protobuf/proto"
 	"google.golang.org/grpc/codes"
--- a/runtime/mux_test.go
+++ b/runtime/mux_test.go
@@ -183,30 +183,20 @@ func TestMuxServeHTTP(t *testing.T) {
 					ops:    []int{int(utilities.OpLitPush), 0, int(utilities.OpPush), 0, int(utilities.OpConcatN), 1, int(utilities.OpCapture), 1},
 					pool:   []string{"foo", "id"},
 				},
-			},
-			reqMethod: "GET",
-			reqPath:   "/foo/bar",
-			headers: map[string]string{
-				"Content-Type": "application/json",
-			},
-			respStatus:  http.StatusOK,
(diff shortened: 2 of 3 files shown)


achew22:
Down with the CI overlords!

Issue #766, opened 2018-10-02
proto-gen-swagger: provide default description for HTTP 200 responses


johnchildren:
In the swagger specification it says the following about description fields:

Field Name | Type | Description
-- | -- | --
description | string | REQUIRED. A short description of the response. CommonMark syntax MAY be used for rich text representation.

https://swagger.io/specification/#responseObject

Currently proto-gen-swagger will generate an empty string for this field on a 200 response, but I am unsure if this is enough to meet the specification. This has some ramifications as at least one of the generators in `swagger-codegen` or `openapi-codegen`, `rust-server` will not compile if the description is set to an empty string without manual intervention.

It seems to be this line which is responsible, so it may just be a matter of setting a default value, though I am not sure what this value should actually be.

https://github.com/grpc-ecosystem/grpc-gateway/blob/master/protoc-gen-swagger/genswagger/template.go#L697

johanbrandhorst:
Thanks for submitting this issue @johnchildren! This does indeed seem to be in breach of the spec. What can I do to help you get a pull request in to fix this? I think it should be okay to introduce this backwards-incompatibility since any comments or explicit description will override it. How about just `A successful response`?

johnchildren:
Have created a pull request for the change, but I am unsure how best to provide automated tests. I also interested to see how the default value looks when used with streaming responses.

johanbrandhorst:
Closed with #767

PR #767, opened 2018-10-02
Add default value for swagger 200 response

johnchildren:
Re: https://github.com/grpc-ecosystem/grpc-gateway/issues/766

This commit adds a default value to the description field of 200
responses. This is needed to meet the swagger specification requirement
of a short description for all response objects.

### DIFF
--- a/protoc-gen-swagger/genswagger/template.go
+++ b/protoc-gen-swagger/genswagger/template.go
@@ -694,7 +694,7 @@ func renderServices(services []*descriptor.Service, paths swaggerPathsObject, re
 				}
 
 				methProtoPath := protoPathIndex(reflect.TypeOf((*pbdescriptor.ServiceDescriptorProto)(nil)), "Method")
-				desc := ""
+				desc := "A successful response."
 				var responseSchema swaggerSchemaObject
 
 				if b.ResponseBody == nil || len(b.ResponseBody.FieldPath) == 0 {


johnchildren:
I couldn't see an obvious place to add this in the tests, but they seemed to pass locally regardless of the change. It may be worth also updating the example swagger files as they currently have empty descriptions.

johanbrandhorst:
@ivucica would like your opinion on this if possible.

johanbrandhorst:
You needn't really bother with tests, it should be good enough that the generated files now have a description set somewhere where it wasn't before.

ivucica:
Regarding justification: I'd argue the spec only requires a string to be set, and not that it's a non-empty string. See [relevant section in spec](https://github.com/OAI/OpenAPI-Specification/blob/fb059ca461bd17b10a9e3e59879f04485886d356/versions/2.0.md#fixed-fields-10); while it says it's required, it doesn't say what the string should look like. 

(For what it's worth, the spec also actually doesn't say what it means by the word "required" in this context; the quoted RFCs use it slightly differently.)

editor.swagger.io does not complain about a `description` set to an empty string.

But, having said this, I believe the change to be an improvement. LGTM.

On a separate note: line 717 will make the resulting description look ugly, but that's already the case if people specify a documentation comment. I'd remove that case, but that's something to be done in a separate PR.

Issue #254, opened 2016-11-07
Update swagger-codegen to latest


tmc:
We're on 2.1.6 while 2.2.0 is latest: https://github.com/swagger-api/swagger-codegen/releases

achew22:
@tmc, Do we need to do more to upgrade than change the [hardcoded compatibility string](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/protoc-gen-swagger/genswagger/template.go#L465)? I didn't see anything in there listed as a breaking change

tmc:
@achew22 afaict it dumps out code with different package names into the same directory.

tmc:
I pushed https://travis-ci.org/tmc/grpc-gateway/builds/177507616 to demonstrate

tmc:
with 2.2.0 or 2.2.1 I get:

```
go test -race github.com/grpc-ecosystem/grpc-gateway/...
can't load package: package github.com/grpc-ecosystem/grpc-gateway/examples/clients: found packages abe (a_bit_of_everything_nested.go) and echo (echo_service_api.go) in /Users/tmc/go/src/github.com/grpc-ecosystem/grpc-gateway/examples/clients
make: *** [test] Error 1``

Thoughts on removing  or disabling the swagger-codegen calls? It's a bit tangential and it will be a contribution barrier.

tmc:
looking more closely this appears to be the `--additional-properties packageName=echo` parameter to swagger-codegen not affecting the output directory anymore.

achew22:
It feels like a test that is testing something outside the control of this project. It does, however, add a nice e2e test to the repo. Honestly I don't even remember this commit going by which explains my total lack of understanding in my earlier comment.

Let's ask ourselves the question "What is this testing and can it be tested in another way" specifically without bringing in a dependency on swagger-codegen

tmc:
If were were actually exercising the `swagger-codegen`-generated code in test I'd be more inclined to keep it around. Right now it's basically testing well-formedness of the generated swagger output.

tmc:
@yugui thoughts here?

dmitris:
the current version of swagger-codegen (that you get with `brew install` for example) is `2.3.1` - CONTRIBUTING.md says "use swagger-codegen 2.2.2, not newer versions".  @johanbrandhorst  - would it be possible to make grpc-gateway to work with 2.3.1?  Thanks.

johanbrandhorst:
[...]

PR #770, opened 2018-10-08
replace github.com/go-resty/resty with gopkg.in/resty.v1

dmitris:
@johanbrandhorst - as discussed on Slack, did not update generated files


Gopkg.toml
```diff
@@ -16,7 +16,7 @@ required = [
 [[constraint]]
   # Also defined in WORKSPACE
   revision = "f8815663de1e64d57cdd4ee9e2b2fa96977a030e"
-  name = "github.com/go-resty/resty"
+  name = "gopkg.in/resty.v1"
```
johanbrandhorst: Does this still work? I think since our generated files use the other import we still need to include both. I'm not sure this is a complete fix. The correct way to fix this would be to correct the generated files as well. Sorry I know that means it will be harder to get done.

johanbrandhorst:
I don't think we can merge this as is - `dep` will still add the old dependency to the lockfile since some of the checked in files import it. We need this to be solved in the generator first (or run a post-generate script to replace the import ourselves...). Have you raised an issue with swagger-codegen?

dmitris:
related (for reference) - https://github.com/go-resty/resty/issues/178

also https://github.com/go-resty/resty/blob/master/go.mod#L1 contains `module gopkg.in/resty.v1`

johanbrandhorst:
Did you regenerate these files or just search replace?

dmitris:
@johanbrandhorst - search-replaced in the branch for now

opened https://github.com/swagger-api/swagger-codegen/issues/8798 - please add the details as needed.  

It would be great if we could move to using the latest version of swagger-codegen (2.3.1) as mentioned in #254.  Currently when I try to generate code with swagger-codegen 2.3.1, I'm getting invalid Go generated:
```
./examplepb_a_bit_of_everything.go:48:13: undefined: ExamplepbNumericEnum
./examplepb_a_bit_of_everything.go:50:17: undefined: PathenumPathEnum
./examplepb_a_bit_of_everything.go:68:22: undefined: ExamplepbNumericEnum
./examplepb_a_bit_of_everything.go:78:22: undefined: ExamplepbNumericEnum
```

If I understand it correctly, `a_bit_of_everything.swagger.json has definitions for the lower-case fields:
```
    "examplepbNumericEnum": {
      "type": "string",
      "enum": [
        "ZERO",
        "ONE"
      ],
      "default": "ZERO",
      "description": "NumericEnum is one or zero.\n\n - ZERO: ZERO means 0\n - ONE: ONE means 1"
    },
    "pathenumPathEnum": {
      "type": "string",
      "enum": [
        "ABC",
        "DEF"
      ],
      "default": "ABC"
    },
```
which becomes in `examplepb_numeric_enum.go`:
```
// ExamplepbNumericEnum : NumericEnum is one or zero.   - ZERO: ZERO means 0  - ONE: ONE means 1
type examplepbNumericEnum string
```
and contradicts usage as in the following enum: 
```
// ExamplepbNumericEnum : NumericEnum is one or zero.   - ZERO: ZERO means 0  - ONE: ONE means 1
type examplepbNumericEnum string

// List of examplepbNumericEnum
const (
	ZERO ExamplepbNumericEnum = "ZERO"
	ONE  ExamplepbNumericEnum = "ONE"
)
```

johanbrandhorst:
I'm gonna close this, lets resume the discussion in #254

PR #773, opened 2018-10-10
Propagate traced requests from gateway to server

ZachEddy:
Currently, there isn't any out-of-the-box support for tracing requests across the gateway to server boundary. This makes it difficult to fully trace a request from end-to-end (e.g. start tracing at NGINX gateway, then propagate trace to the gRPC gateway, and finally to the gRPC server). There is currently a documented approach [here](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/docs/_docs/customizingyourgateway.md#opentracing-support), but this has to be duplicated from gateway to gateway and it requires extra legwork from developers.

The `WithTracer` option is intended to fix this. When this option is used, the gRPC gateway will pass the trace forward to the gRPC server.

I'm not completely sure this is the right approach. It requires the gateway to be initiated with the client-side, so that's something to consider.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,21 +2,28 @@
 
 
 [[projects]]
+  digest = "1:2cd7915ab26ede7d95b8749e6b1f933f1c6d5398030684e6505940a10f31cfda"
   name = "github.com/ghodss/yaml"
   packages = ["."]
+  pruneopts = "UT"
   revision = "0ca9ea5df5451ffdf184b4428c902747c2c11cd7"
 
 [[projects]]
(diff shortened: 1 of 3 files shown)


johanbrandhorst:
Hi @ZachEddy, thanks for your contribution! This is a great step in the right direction, but I think we're coupling ourselves a little unnecessarily to OpenTracing here. If we want to add an option to the gateway I think we want to expose it via an implicit interface that will work with OpenTracing and other providers (ZipKin, OpenCensus, whatever else there is). Do you have any thoughts on how we can accomplish that?

johanbrandhorst:

johanbrandhorst:
https://grpc-ecosystem.github.io/grpc-gateway/docs/customizingyourgateway.html#opentracing-support has a section on tracing with the gateway. Closing this as stale.

PR #776, opened 2018-10-11
remove message field from the error message emitted by grpc-gateway

ti:
this is just recovery of the code, the "error" filed is more meaning for than the message

more info: https://github.com/grpc-ecosystem/grpc-gateway/pull/718

### DIFF
--- a/runtime/errors.go
+++ b/runtime/errors.go
@@ -66,10 +66,6 @@ var (
 
 type errorBody struct {
 	Error   string     `protobuf:"bytes,1,name=error" json:"error"`
-	// This is to make the error more compatible with users that expect errors to be Status objects:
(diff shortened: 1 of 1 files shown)


johanbrandhorst:
Hi @ti, while I appreciate that the duplication is ugly, I don't think removing the message that we added there for consistency with the googleapis.Status type is the right way to go. We cannot accept this right now for backwards compatbility reasons, but we are also more likely to just remove `error` with version 2.0.

ti:
@ffredsh we should recovery of the code.

ti:
@johanbrandhorst  https://github.com/grpc-ecosystem/grpc-gateway/pull/718

johanbrandhorst:
What do you mean? We added that for the reasons listed in the pull request.

PR #783, opened 2018-10-17
fix #782 Cannot return HTTP header using "Grpc-Metadata-" prefix

### LINKED ISSUE #782: Cannot return HTTP header using "Grpc-Metadata-" prefix
I have been using the following line of code to return HTTP headers in the server response:

```
import (
    "google.golang.org/grpc"
    "google.golang.org/grpc/metadata"
)

grpc.SetHeader(ctx, metadata.Pairs("Grpc-Metadata-Some-Key", someValue))
```

This was working OK using the `runtime.DefaultHeaderMatcher` when starting the server, but after updating this no longer works.

I went over the `google.golang.org/grpc/metadata` package and it seems that every key set in the metadata is transformed to lower case, but the `runtime.DefaultHeaderMatcher` function compares the transformed metadata key with the prefix "Grpc-Metadata-", so it never matches and the header is never returned.

I think [this](https://github.com/grpc-ecosystem/grpc-gateway/commit/8976e602c518589cb6d40aca52a7dd8aef83706a#diff-85c4fe88fcf7047845e0e64e2f4a0722L54) broke it.

Am I doing anything wrong? How should I set the metadata so it is returned as an HTTP header?

joelurraco:


joelurraco:
I signed it

johanbrandhorst:
@joelurraco it seems you've committed with an email/username that is different from the one you're using to open this PR.

joelurraco:
@johanbrandhorst yes, but I've just signed the CLA with both accounts

johanbrandhorst:
Not much I can do about the Google bot I'm afraid - would you mind force pushing a new commit or opening a new PR with the other account?

johanbrandhorst:
Superceded by #784

PR #784, opened 2018-10-17
fix #782 Cannot return HTTP header using "Grpc-Metadata-" prefix

### LINKED ISSUE #728: Client grabbing errors
### This is not technically a bug with grpc-gateway, but more of a request for information/help for usage.

**Steps you follow to reproduce the error:**

1. Setup and run a grpc-gateway with no backend service running (so that all requests fail).
2. Send a legitimate REST call using a python `requests` client or php `guzzle` client.
3. Send the same command using curl.
4. Compare the error responses. (Curl command gives use information.)

**What did you expect to happen instead:**

I expected that the clients would pick-up the error message that curl did.
In the aforementioned situation, Curl returns useful information such as:
`{"error":"transport is closing","message":"transport is closing","code":14}`
or 
`{"error":"all SubConns are in TransientFailure, latest connection error: connection error: desc = \"transport: Error while dialing dial tcp 123.45.678.910:12345: connect: operation timed out\"","code":14}`

But the client will only return `503 Service Unavailable`

**What's your theory on why it isn't working:**

I suspect that the errors are chunked or in the trailer, but I am unsure how to capture them using a client. I am hoping somebody could give me some insight towards a solution. Thanks!

### LINKED ISSUE #782: Cannot return HTTP header using "Grpc-Metadata-" prefix

joelclouddistrict:
Fixes #728, supercedes #783

### DIFF
--- a/runtime/mux.go
+++ b/runtime/mux.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"net/textproto"
 	"strings"
 
 	"github.com/golang/protobuf/proto"
@@ -50,6 +51,7 @@ type HeaderMatcherFunc func(string) (string, bool)
 // keys (as specified by the IANA) to gRPC context with grpcgateway- prefix. HTTP headers that start with
 // 'Grpc-Metadata-' are mapped to gRPC metadata after removing prefix 'Grpc-Metadata-'.
 func DefaultHeaderMatcher(key string) (string, bool) {
+	key = textproto.CanonicalMIMEHeaderKey(key)
 	if isPermanentHTTPHeader(key) {
 		return MetadataPrefix + key, true
 	} else if strings.HasPrefix(key, MetadataHeaderPrefix) {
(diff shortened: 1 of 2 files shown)


joelclouddistrict:
Can this be merged?

johanbrandhorst:
Yes, great, thanks for your contribution!

PR #739, opened 2018-08-29
Fixed operation id support

robbert229:
Currently there is an issue with the grpc-gen-swagger where it does not respect the operation_id that can be given in the proto file.

This is an example of what it may look like to fix the issue.

PR #786, opened 2018-10-19
Make operationID more unique using svc name + call name

mestrade:
Following the discussion on the issue #739 
The OperationID could be more unique by default.
Ie. when a .proto is containing multiple CRUD services, operationID is duplicated in each service.

### DIFF
--- a/protoc-gen-swagger/genswagger/template.go
+++ b/protoc-gen-swagger/genswagger/template.go
@@ -727,10 +727,10 @@ func renderServices(services []*descriptor.Service, paths swaggerPathsObject, re
 					},
 				}
 				if bIdx == 0 {
-					operationObject.OperationID = fmt.Sprintf("%s", meth.GetName())
+					operationObject.OperationID = fmt.Sprintf("%s.%s", svc.GetName(), meth.GetName())
 				} else {
 					// OperationID must be unique in an OpenAPI v2 definition.
-					operationObject.OperationID = fmt.Sprintf("%s%d", meth.GetName(), bIdx+1)
+					operationObject.OperationID = fmt.Sprintf("%s.%s%d", svc.GetName(), meth.GetName(), bIdx+1)
 				}
 
 				// Fill reference map with referenced request messages
--- a/examples/clients/abe/a_bit_of_everything_service_api.go
+++ b/examples/clients/abe/a_bit_of_everything_service_api.go
@@ -62,7 +62,7 @@ func NewABitOfEverythingServiceApiWithBasePath(basePath string) *ABitOfEverythin
  * @param nestedPathEnumValue 
  * @return *ExamplepbABitOfEverything
  */
-func (a ABitOfEverythingServiceApi) Create(floatValue float32, doubleValue float64, int64Value string, uint64Value string, int32Value int32, fixed64Value string, fixed32Value int64, boolValue bool, stringValue string, uint32Value int64, sfixed32Value int32, sfixed64Value string, sint32Value int32, sint64Value string, nonConventionalNameValue string, enumValue string, pathEnumValue string, nestedPathEnumValue string) (*ExamplepbABitOfEverything, *APIResponse, error) {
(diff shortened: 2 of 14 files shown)


protoc-gen-swagger/genswagger/template.go
```diff
@@ -694,7 +694,7 @@ func renderServices(services []*descriptor.Service, paths swaggerPathsObject, re
 				}
 
 				methProtoPath := protoPathIndex(reflect.TypeOf((*pbdescriptor.ServiceDescriptorProto)(nil)), "Method")
-				desc := "A successful response."
+				desc := ""
```
johanbrandhorst:
```suggestion
				desc := "A successful response."
```

protoc-gen-swagger/genswagger/template.go
```diff
@@ -727,10 +727,10 @@ func renderServices(services []*descriptor.Service, paths swaggerPathsObject, re
 					},
 				}
 				if bIdx == 0 {
-					operationObject.OperationID = fmt.Sprintf("%s", meth.GetName())
+					operationObject.OperationID = fmt.Sprintf("%s%s", svc.GetName(), meth.GetName())
```
johanbrandhorst: How about
```suggestion
					operationObject.OperationID = fmt.Sprintf("%s.%s", svc.GetName(), meth.GetName())
```

protoc-gen-swagger/genswagger/template.go
```diff
@@ -727,10 +727,10 @@ func renderServices(services []*descriptor.Service, paths swaggerPathsObject, re
...
 				} else {
 					// OperationID must be unique in an OpenAPI v2 definition.
-					operationObject.OperationID = fmt.Sprintf("%s%d", meth.GetName(), bIdx+1)
+					operationObject.OperationID = fmt.Sprintf("%s%s%d", svc.GetName(), meth.GetName(), bIdx+1)
```
johanbrandhorst:
```suggestion
					operationObject.OperationID = fmt.Sprintf("%s.%s%d", svc.GetName(), meth.GetName(), bIdx+1)
```

johanbrandhorst:
You'll need to regenerate the files as well before we can merge this change, please see https://github.com/grpc-ecosystem/grpc-gateway/blob/master/CONTRIBUTING.md#i-want-to-regenerate-the-files-after-making-changes

johanbrandhorst:
Interesting, the CLA bot doesn't like us sharing commits.

johanbrandhorst:
OK @mestrade can you please squash the commits into on commit made by you and force push it to this branch and then we can get it in. Remember you need to regenerate as well.

johanbrandhorst:
Not sure about the build error, but on second thought, could you add the changes made in https://github.com/grpc-ecosystem/grpc-gateway/pull/739/files as well? So users can manually overwrite the operationID? This part:

```go
if opts.OperationId != "" {
	operationObject.OperationID = opts.OperationId
	if bIdx > 0 {
		operationObject.OperationID = fmt.Sprintf("%s%d", opts.OperationId, bIdx + 1)
	}
}
```

johanbrandhorst:
I think the tests are failing because the generated types have changed so we need to also change the integration tests: https://github.com/grpc-ecosystem/grpc-gateway/blob/master/examples/integration/integration_test.go

johanbrandhorst:
I'm also not sure if we consider this to be a backwards compatibility breaking change since we're changing the swagger output? Maybe we should do it and see if someone complains :grimacing:.

mestrade:
Agree that it may break something, but for now, if you have two rpc call with the same name in two different services, the output swagger is not openapi compliant.

trusch:
@mestrade @johanbrandhorst Are there any updates on this issue?

johanbrandhorst:
There's another PR open for this which adds the package name to deduplicate methods, see #

mestrade:
Hello again =) 

I will try to come back again on this in the next week. Just upgraded to the last version and it seems not fixed. Hopefully i will be able to find a solution on that as all generated swagger are not compliant to openapi.

Btw, there is a tool here to test if it's compliant or not regarding the spec (openapi 2.0 only for now).
I will try to write a blog post soon on how to have a good score by using annotations and options in order to define your .proto correctly to generate a good swagger !

johanbrandhorst:
That's great, maybe something we could add to the docs as well?

PR #793, opened 2018-10-30
ci: add job for building binaries for releases

### LINKED ISSUE #790: Proposal: pre-compiled per-platform release binaries
Currently it can be tricky for teams to ensure they are using consistent versions of protoc-gen-grpc-gateway (and protoc-gen-swagger). For example, the installation instructions on the repo suggest `go get -u github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway` - which recently led to out-of-sync generated code when new folks joined our team.

Version management is particularly important here because these tools are typically part of build processes which are meant to be reproducible.

It would be great to have pre-compiled, per-platform release binaries, similar to protoc https://github.com/protocolbuffers/protobuf/releases or dep https://github.com/golang/dep/releases

A set of release binaries would also make grpc-gateway easy to integrate with version-management tooling such as https://github.com/asdf-vm/asdf

(I'm happy to help with this, but attaching binaries to releases is something you can't do in a PR)

johanbrandhorst:
This will trigger a build of a linux, darwin and windows
build of protoc-gen-grpc-gateway and protoc-gen-swagger
on every new tag matching the semver regex.

Fixes #790

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: "jfbrandhorst/grpc-gateway-build-env"
+      - image: jfbrandhorst/grpc-gateway-build-env
     working_directory: /go/src/github.com/grpc-ecosystem/grpc-gateway
     steps:
       - checkout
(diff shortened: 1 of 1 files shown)


johanbrandhorst:
Now I haven't the foggiest idea how we're going to be able to test this unfortunately - any ideas?

johanbrandhorst:
It's based on https://circleci.com/blog/publishing-to-github-releases-via-circleci/

johanbrandhorst:
I didn't expect the release builds to run, perhaps we should only run those on tags as well?

johanbrandhorst:
Disabled the release builds for non-tags

johanbrandhorst:
So I think the only way to really test this is to make a new release :tada:

achew22:
LGTM. I'm not merging because there is a chance you want to watch the next run. Feel free to merge at will

PR #796, opened 2018-11-06
protoc-gen-swagger: Fix formatting of license field definition.

ivucica:


johanbrandhorst:
Looks like we need to regenerate the protofile as well - not sure where the hash comes from but there it is. https://github.com/grpc-ecosystem/grpc-gateway/blob/master/CONTRIBUTING.md#i-want-to-regenerate-the-files-after-making-changes

ivucica:
It's so very strange that touching just the whitespace causes this.

I can't do this straight away unfortunately.

johanbrandhorst:
I can force push a change, leave it to me.

ivucica:
If you feel like it, sure, but this is really a nit anyway :-)

johanbrandhorst:
Files regenerated.

PR #797, opened 2018-11-07
Run buildifer on Bazel files

drigz:
buildifier autoformats the files to give a consistent style. This makes
edits easier for contributors that have their editor set to
autoformat-on-save.


johanbrandhorst:
If we're doing this, I'd also be tempted to add a formatter step to CI. Would you be interested in adding a CI step to this PR that runs the buildifier and reports transgressions?

drigz:
@johanbrandhorst I'm happy to try. Would you like me to do it in this or another PR?

johanbrandhorst:
This PR please.

achew22:
[Gazelle](https://github.com/bazelbuild/bazel-gazelle) is a good solution to this problem in the long run. It's a simple binary that is able to regenerate all the BUILD files as they should be written including doing these kinds of formattings.

The maintainer of Gazelle, is presently working on expanding his binary so that you can add language plugins. Once this happens I am going to write a language plugin for grpc-gateway that auto-generates these BUILD files. When that happens I will make it a part of the presubmit process.

I'm going to merge these because my hopes/dreams shouldn't slow other ppl down.

@drigz, thanks so much for contributing! If you know about Bazel and would be interested in contributing to the project, there is a lot more neat stuff we could do!

drigz:
@achew22 I've compared gazelle v0.10.1 vs buildifier (build from oct 24) and found the following:

- buildifier can reformat WORKSPACE but gazelle doesn't (by default at least)
- buildifier can reformat *.bzl but gazelle doesn't (by default at least)
- gazelle (unsurprisingly) wants to make a few changes to the go rules (see :gazelle_diff output [from last run](https://circleci.com/gh/grpc-ecosystem/grpc-gateway/272))
- neither sets their exit code based on differences ([gazelle issue](https://github.com/bazelbuild/bazel-gazelle/issues/15))

Do you think the CI check should fail when Gazelle (/buildifier) would make a change?

drigz:
I'm assuming that, for now, running both buildifier and gazelle is useful, so I've added it to CI steps.

@achew22 I followed your lead in preserving `--batch`, although we could probably make the presubmit faster by dropping `--batch` and combining `:bazel_build` and `:bazel_test` (since `test //...` implies `build //...`).

johanbrandhorst:
Please revise the CI tests to make sense - I only reproduced the existing travis tests to the best of my ability when creating the circle CI configuration.

johanbrandhorst:
LGTM, but I'll let @achew22 have the final say.

PR #798, opened 2018-11-07
Use newer Bazel repo rules

drigz:
This uses the new go_repository and http_archive rules. Using the former
avoids a warning:

DEPRECATED: the rules_go version of go_repository is deprecated. Migrate
to the new version in @bazel_gazelle//def.bzl instead.

The latter will avoid breakage when the native http_archive rule is
disabled by default in a future version of Bazel.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,5 +1,7 @@
 workspace(name = "grpc_ecosystem_grpc_gateway")
 
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
 http_archive(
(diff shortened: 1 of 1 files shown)


johanbrandhorst:
Thanks for the contribution! Do you want to merge this before or after #797?

drigz:
@johanbrandhorst Let's do it before if you don't have any comments here. They shouldn't conflict. (famous last words)

johanbrandhorst:
I'll leave this to @achew22 now that he's shown signs of interest, but no objections on ordering to me.

PR #802, opened 2018-11-09
Update rules_go and gazelle

drigz:
This requires/allows the following changes:
- rerun Gazelle to regenerate the build files
- regenerate examples with the newer protoc
- fix tests to correctly handle enums inside maps (golang/protobuf#645)
- remove `repositories.bzl` (replaced by @go_googleapis)

Also, enforce up-to-date build files in the CI check.


WORKSPACE
```diff
@@ -2,16 +2,24 @@ workspace(name = "grpc_ecosystem_grpc_gateway")
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
+# master as of 2018-11-09. This should be updated when a new release is tagged.
```
johanbrandhorst: I don't have a great understanding of this, but does this update any transative dependencies? We need to keep this file in sync with `Gopkg.toml`.

runtime/marshal_jsonpb_test.go
```diff
@@ -270,8 +270,8 @@ func TestJSONPbEncoder(t *testing.T) {
 				if strings.ContainsAny(json, " \t\r\n") {
 					t.Errorf("strings.ContainsAny(%q, %q) = true; want false", json, " \t\r\n")
 				}
-				if strings.Contains(json, "ONE") {
-					t.Errorf(`strings.Contains(%q, "ONE") = true; want false`, json)
+				if !strings.Contains(json, "ONE") {
```
johanbrandhorst: Is this a change in behaviour?

WORKSPACE
drigz: It does, thanks for pointing that out. I've updated it and rerun `dep ensure`, is there anything else to do?

johanbrandhorst: I guess we still want to make sure the tests pass and everything with the new versions, but I assume Bazel will do that for us as well? Just commit it and we shall see :).

runtime/marshal_jsonpb_test.go
drigz: I think so - I assumed it was fixing a bug since the test seemed odd to me before.

In master, the behavior is different in two tests:

`TestJSONPbMarshal` tests:
- if `EnumsAsInts == false`, assert `!strings.Contains(json, "ONE")`
- if `EnumsAsInts == true`, assert `strings.Contains(json, "ONE")`

`TestJSONPbEncoder` tests:
- if `EnumsAsInts == false`, assert `!strings.Contains(json, "ONE")`
- if `EnumsAsInts == true`, assert `!strings.Contains(json, "ONE")`

The change is in the last case, but I'm not 100% sure why this changed or what the behavior ought to be. Do you have any insight into this area?

drigz: The node_tests make this look more like a bug - there's a `TODO(yugui) Support enum by name` before `map_value: { a: 1, b: 2 }`. If you change the digits to strings, you get `unknown value \"TWO\" for enum grpc.gateway.examples.examplepb.NumericEnum`. So it seems that not supporting strings is almost intentional...

Maybe we need to change the configuration such that `jsonpb` will continue to have the old behavior.

@achew22, do you know anything about this code?

drigz: FYI: I suspect this PR is the cause: https://github.com/golang/protobuf/pull/645

It looks like the grpc-gateway tests rely on behavior that golang/protobuf considered broken (enum values in maps).

johanbrandhorst: Great digging, I think we should obviously fix the test then.

johanbrandhorst:
What's next to do here? I've lost sight a bit. Do you need anything from us?

johanbrandhorst:
If the behavior has been incorrect I think we can safely correct it without worrying about backwards compatibility. Is that simply a matter of copying the fix from the protobuf repo?

drigz:
I wasn't able to easily fix `node_test`, as changing ints to strings introduced another error. Additionally, I need to regenerate the examples, which I haven't worked out how to do yet.

johanbrandhorst:
Regenerating everything should be covered in https://github.com/grpc-ecosystem/grpc-gateway/blob/master/CONTRIBUTING.md#i-want-to-regenerate-the-files-after-making-changes.

If fixing the tests is a pain, lets raise an issue and solve it in a separate PR. Is it necessarily part of this cleanup? Don't want to waste your time with red tape.

johanbrandhorst:
Right I see what happened, we updated our dependency of golang/protobuf, and all of a sudden incorrect behaviour that we depended on was fixed, and now our tests fail. It seems we have to fix it in this PR then. I'm not too familiar with this code unfortunately, so I think we just have to do a little digging to get to the bottom of it. Changing the test is correct, so any part of our json marshaller that is doing the wrong thing is what needs to be fixed now.

Is that a correct summary?

johanbrandhorst:
Node tests are failing with:

```
Expected $.map_value.a = 'ONE' to equal 1.
```

I wonder if it's just a matter of fixing these tests too?

johanbrandhorst:
:crossed_fingers:

johanbrandhorst:
This has become bigger than initially anticipated I think, but lets wait for @achew22 to have a chance to review it as well before merging. Huge thanks for getting the ball rolling on this though @drigz :).

johanbrandhorst:
Node tests are looking good!

drigz:
Yep, looks like I gave up too quickly. Thanks for the encouragement :)

WORKSPACE
```diff
@@ -21,12 +29,10 @@ http_archive(
     url = "https://github.com/bazelbuild/buildtools/archive/e90e7cc6ef3e6d08d4ca8a982935c3eed638e058.tar.gz",
 )
 
-load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies")
+load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies", "go_repository")
 
 gazelle_dependencies()
 
-load("@bazel_gazelle//:def.bzl", "go_repository")
-load("//:repositories.bzl", "repositories")
```
ivucica: Not objecting, just trying to learn: Why remove repositories.bzl? Wouldn't it be better to allow third parties to import a repositories.bzl and transitively depend?

WORKSPACE
```diff
@@ -21,12 +29,10 @@ http_archive(
     url = "https://github.com/bazelbuild/buildtools/archive/e90e7cc6ef3e6d08d4ca8a982935c3eed638e058.tar.gz",
 )
 
-load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies")
+load("@bazel_gazelle//:deps.bzl", "gazelle_dependencies", "go_repository")
 
 gazelle_dependencies()
```
ivucica: Nob objecting, just trying to learn: Maybe I'm not tracking developments in Bazel world closely enough, but if gazelle_dependencies() is no longer imported, where is this defined? Why does this work?

drigz: gazelle_dependencies is still loaded from bazel_gazelle//:deps.bzl, along with go_repository. The only change here is that two load() statements have been consolidated into one.

WORKSPACE
drigz: The repository defined by repositories.bzl is obsolete, having been replaced by @go_googleapis (created by go_rules_dependencies()). Since repositories.bzl is unable to call go_rules_dependencies() (https://github.com/bazelbuild/bazel/issues/1550) there's nothing left for it to do.

achew22:
> This has become bigger than initially anticipated I think, but lets wait for @achew22 to have a chance to review it as well before merging. Huge thanks for getting the ball rolling on this though @drigz :).

I think we should merge this as is. It will probably break people but upstream changed on us and we don't have much of an option. If this behavior is required by integrators there is a parameter they can set.

Gopkg.toml
```diff
@@ -37,30 +37,30 @@ required = [
 
 [[constraint]]
   # Also defined in bazelbuild/rules_go
-  # https://github.com/bazelbuild/rules_go/blob/436452edc29a2f1e0edc22d180fbb57c27e6d0af/go/private/repositories.bzl#L75
-  version = "1.1.0"
+  # https://github.com/bazelbuild/rules_go/blob/109c520465fcb418f2c4be967f3744d959ad66d3/go/private/repositories.bzl#L52
+  revision = "aa810b61a9c79d51363740d207bb46cf8e620ed5"
```
javasgl: `master: Could not introduce github.com/grpc-ecosystem/grpc-gateway@master, as it has a dependency on github.com/golang/protobuf with constraint aa810b61a9c79d51363740d207bb46cf8e620ed5, which has no overlap with existing constraint master from github.com/grpc-ecosystem/go-grpc-middleware@master`

broken!

drigz: I'm not a `go dep` but I don't think this change introduced the problem, as even before the version was constrained and would have conflicted with `go-grpc-middleware`'s constraint. @javasgl Maybe you need to use an override to resolve the difference?

johanbrandhorst: Yeah I've explained the situation in #829

PR #803, opened 2018-11-12
add allow_get_body option to allow send body with GET request

ChenLingPeng:
I have meet a case that every request has a requestId in request body, and the response should reply with that body. but it seems not support with method GET. This PR enable this

Signed-off-by: forrestchen <forrestchen@tencent.com>

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -36,6 +36,7 @@
     "ptypes/wrappers"
   ]
   revision = "b4deda0973fb4c70b50d226b1af49f3da59f5265"
+  version = "v1.1.0"
 
 [[projects]]
   name = "github.com/rogpeppe/fastuuid"
@@ -115,6 +116,7 @@
     "transport"
   ]
   revision = "d11072e7ca9811b1100b80ca0269ac831f06d024"
+  version = "v1.11.3"
 
 [[projects]]
   name = "gopkg.in/yaml.v2"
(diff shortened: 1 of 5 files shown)


ChenLingPeng:
I signed it!

ChenLingPeng:
Since most http framework support GET with body, I think there's also some scenarios user would like to use with it. It is better if we support this.

johanbrandhorst:
What for supports GET with a body?

ChenLingPeng:
I have an api spec which defined that every reqeust(include GET/DELETE) should have a requestId in request body and the response body will carry back this requestId.

johanbrandhorst:
Sorry, what frameworks support GET with a body? I have always seen it referenced to as a violation of the spec. Your design should account for this and take the request id as a query parameter. I don't think you can assume any HTTP servers will support GET with a body.

ChenLingPeng:
we also use [go-restful](https://github.com/emicklei/go-restful) as our router and it support read from GET request body

ChenLingPeng:
How could I send a header in grpc-gateway?
i.e
I send a request with header 'X-Request-Id', I want to send it back. 
I have tried this:
```
runtime.NewServeMux(util.DefaultJsonMarshal,
			//runtime.WithProtoErrorHandler(proto_err.ProtoError),
			runtime.WithForwardResponseOption(mycode.Forward),
			runtime.WithIncomingHeaderMatcher(mycode.HeaderMatcherFunc),
			runtime.WithOutgoingHeaderMatcher(mycode.OutgoingHeaderMatcherFunc),
			runtime.WithMetadata(mycode.CookiesProvider),),
```
but seems it doesn't work.

ChenLingPeng:
May I ask how to send back header in grcp-gateway?
i.e 
I send a request with header 'X-Request-Id', I want get it from response
I tried set some function when `NewServeMux`, but it doesn't work

johanbrandhorst:
It's definitely possible to set a header response in the gRPC server and have it sent out, but you need to share more of your code for us to see where the error is. Maybe you could join the slack channel #grpc-gateway on gophers Slack and we can help you in there?

johanbrandhorst:
Closing as it was agreed out-of-band that this wasn't necessary.

PR #804, opened 2018-11-12
fix body with "*" bug

ChenLingPeng:
[this code](https://github.com/googleapis/googleapis/blob/master/google/api/http.proto#L149) says that when body is set to `"*"` in a rpc service, all the query param will map to the request body, but the truth is not. when body is `"*"`, `b.Body != nil && len(b.Body.FieldPath)` will be true and we should continue the method

Signed-off-by: forrestchen <forrestchen@tencent.com>

### DIFF
--- a/examples/proto/examplepb/a_bit_of_everything.pb.gw.go
+++ b/examples/proto/examplepb/a_bit_of_everything.pb.gw.go
@@ -261,11 +261,19 @@ func request_ABitOfEverythingService_Create_0(ctx context.Context, marshaler run
 
 }
 
+var (
+	filter_ABitOfEverythingService_CreateBody_0 = &utilities.DoubleArray{Encoding: map[string]int{"": 0}, Base: []int{1, 1, 0}, Check: []int{0, 1, 2}}
+)
+
 func request_ABitOfEverythingService_CreateBody_0(ctx context.Context, marshaler runtime.Marshaler, client ABitOfEverythingServiceClient, req *http.Request, pathParams map[string]string) (proto.Message, runtime.ServerMetadata, error) {
 	var protoReq ABitOfEverything
 	var metadata runtime.ServerMetadata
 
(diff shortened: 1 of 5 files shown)


johanbrandhorst:
I'm confused, are you saying that parameters provided as query parameters should be parsed when `body: "*"` is set? I'm not sure it's a correct interpretation of the rule. My interpretation is that any parameters not part of the path should be in the request body. Do you have a practical example of this instead of a hypothetical interpretation?

johanbrandhorst:
Leaving aside the point about having a body in a GET, what does your http.proto annotations look like for this RPC?

ChenLingPeng:
```
message TsfListClusterRequest {
    string requestId = 1;
    int32 offset = 2;
    int32 limit = 3;
}

rpc ListClusters(TsfListClusterRequest) returns (TsfListClusterResponse) {
    option (google.api.http)  = {
        get: "/tsf/v1.0/clusters"
        body: "*"
    };
}
```

johanbrandhorst:
So judging by your annotations I think the wording in http.proto implies that the parameters should go in the body. You can't expect it to read it from query parameters and the body when you use `"*"`. Have you tried just using `body: "request_id"`?

Also; you should really use `snake_case` for proto file field names. https://developers.google.com/protocol-buffers/docs/style

ChenLingPeng:
when body is set to "request_id", the generate code is `marshaler.NewDecoder(req.Body).Decode(&protoReq.RequestId)` and will return an unmarshal error.

johanbrandhorst:
What does that error look like?

ChenLingPeng:
`{"error":"json: cannot unmarshal object into Go value of type string","code":3}`

johanbrandhorst:
I'm wondering if that's a bug... it should be possible to have the specific parameter parsed from the body. Can you see why this is broken like this? You can't use * and expect query parameters to work.

johanbrandhorst:
I think the premise of this PR is a flawed interpretation of `http.proto`, as I discussed in #234. Closing since the current behaviour is correct.

PR #671, opened 2018-06-18
add patch behavior

dmacthedestroyer:
this addresses #379 

If a binding is mapped to `patch` and the request message has exactly one `FieldMask` message in it, additional code is rendered for the gateway handler that will populate the `FieldMask` based on the request body.

This handles two scenarios:
*The FieldMask is hidden from the REST request* (as in the `UpdateV2` example). In this case, the FieldMask is updated from the request body and set in the gRPC request message.

*The FieldMask is exposed to the REST request* (as in the `PatchWithFieldMaskInBody` example). For this case, a check is made as to whether the FieldMask is nil/empty prior to populating with the request body. If it's not nil, then it converts the FieldMask paths from the REST (snake_case) to gRPC (PascalCase) format. Otherwise, it acts like the previous case.

Additional comments of interest are inline on this PR.

Issue #379, opened 2017-05-05
REST PUT request only passes fields to update


hangll:
I'm trying to export REST APIs and the problem is that in my PUT request, client usually tries to pass only fields to update, but the generated proto will set all the other fields to its default value. I have no way to tell which fields to update.

Of course I can restrict client to always pass the full object, but is there better way to do this?

TamalSaha:
@hangll , on your server side, you could check for fields that have values set (aka, non-default value) and then only update those. That is what we do for our service.

hangll:
@TamalSaha that is hard. How can you tell if a bool value is updated to false or is not set?

TamalSaha:
@hangll you could define your boolean variables in a way that settings `true` will require update. Or, you could use a tr-state variable.

evolsnow:
[FieldMask](https://developers.google.com/protocol-buffers/docs/reference/csharp/class/google/protobuf/well-known-types/field-mask) may be helpful, and it would be wonderful if `grpc-gateway` could generate the mask from REST API.

achew22:
The semantics for `PUT` are that you will include the [entire contents of the file](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PUT). If this were going to be supported it is would have to be under the verb [PATCH](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PATCH)

adematte:
Hello, we have our own implementation to map REST -> GRPC in nodejs and are evaluating if we could use/support this project instead. One of the things we are doing is populating the fieldMask automatically based on the json request.  Using a fieldmask is the de facto GRPC approach for "patching" and I can see real value in it being handled by the api gateway and not having it being generated by the REST clients.

Is this something that would make sense for this projet core features, or is this something that should be handled via a custom middleware of some kind?

achew22:
@adematte I think that is the way to go. An auto populated fieldMask is, I think, the way to do this. That said, I'm not sure what the best way to go about doing that would be. Do you have a magically named fieldmask? Is there a part of the spec that allows for that that I've missed?

wora:
I worked on proto3, gRPC, and Google API Design Guide. This is a well known feature request. I would recommend gRPC Gateway implement this feature.

[...]

PR #806, opened 2018-11-13
Updated Patch behavior

razamiDev:
This PR is a cleaner version of PR #671 
This was a collaborative effort by @razamiDev and @dmacthedestroyer 

this addresses #379
If a binding is mapped to patch and the request message has exactly one FieldMask message in it, additional code is rendered for the gateway handler that will populate the FieldMask based on the request body.
This handles two scenarios: The FieldMask is hidden from the REST request (as in the UpdateV2 example). In this case, the FieldMask is updated from the request body and set in the gRPC request message.
The FieldMask is exposed to the REST request (as in the PatchWithFieldMaskInBody example). For this case, a check is made as to whether the FieldMask is nil/empty prior to populating with the request body. If it's not nil, then it converts the FieldMask paths from the REST (snake_case) to gRPC (PascalCase) format. Otherwise, it acts like the previous case.
Additional comments of interest are inline on this PR.


johanbrandhorst:
@razamiDev Thanks a lot for picking this up. It's very hotly anticipated feature among the maintainers :). Looks like we've got some bazel failures, something about the field_mask.proto file not being found. @achew22 @drigz maybe you can help here? I think it's a new dependency for the test files.

drigz:
@razamiDev Could you [install Bazel](https://docs.bazel.build/versions/master/install.html) and run `bazel run :gazelle_fix`?

@johanbrandhorst I'll try to make this clearer in the CI logs (#807, in progress). Do you think we need an invocation that works with docker like in CONTRIBUTING.md, or is it OK to ask contributors to install Bazel?

johanbrandhorst:
@drigz I definitely would like to see a docker-based solution, and I think we should be able to put together a oneliner with the docker image we're using in circleCI.

drigz:
@johanbrandhorst We need to use a different image to get Bazel - CircleCI uses l.gcr.io/google/bazel - so we'll need two `docker run` commands. The simplest I could come up with is:

```
    l.gcr.io/google/bazel -c 'bazel run :gazelle_fix; bazel run :buildifier'
```

@razamiDev If you don't have Bazel installed, perhaps you could try that command and let me know if it works for you?

johanbrandhorst:
Nice job @drigz, lets see if we can get that into the contribution guidelines with #807?

razamiDev:
I appreciate the help as I am new to bazel. @johanbrandhorst I installed bazel and ran `bazel run :gazelle_fix` and it didn't give me any errors it said Build completed successfully.

@drigz  I also ran the docker command and it said Build completed successfully. 
I added all the new BUILD.bazel files and see that it still failed... giving an error not finding`fastuuid`

drigz:
@razamiDev Thanks for following up! It looks like gazelle has rewritten the dependencies to point an `vendor`, which isn't what we want. Could you rebase on top of master to pull in #807? That tells gazelle to ignore `vendor`.

razamiDev:
@johanbrandhorst I am not sure why the googlebot failed... I am okay with my work being contributed to this project.

johanbrandhorst:
So it looks like your merged master into this branch instead of rebasing on top of master. This meant this branch now contains commits that weren't made by you, and it's confusing the Google bot. I can't overrule the Google bot, do the easiest thing for you is to try and rebase on master again (not merge!). If you can't do that, the next easiest is probably to open a new PR in a new branch.

drigz:
@razamiDev I think the CLA bot is unhappy because some commits from master have made it into the pull request:

image

You may be able to fix it by running:

```
git checkout -b feature/patch2-rebased
git cherry-pick 398e4356300c19959d4cc65668975b2254b06b51
bazel run :gazelle_fix
git commit -am "Regenerate BUILD files"
```

As @johanbrandhorst you may need to create a new pull request so the CLA bot can run again.

razamiDev:
@johanbrandhorst @drigz thank you both for the help! 
I have created a new PR (#812) and this can be closed.

PR #807, opened 2018-11-14
Make Bazel CI failures clearer

drigz:
- don't try to run tests if the BUILD files are out-of-date
- run buildifier even if the tests fail, as it may still be useful
- use a parameter to hold Bazel config to simplify config.yml
- use default output_base (hopefully it works, this is because it's hard
  to have nested envvars with CircleCL)
- ignore the vendor/ directory when running Gazelle
  (otherwise it will fail if you've used `dep` to create vendor/)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,23 +50,32 @@ jobs:
       - run: dep ensure --vendor-only
       - run: go get golang.org/x/lint/golint
       - run: make lint
-  bazel_lint:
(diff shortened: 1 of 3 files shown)


drigz:
Don't mind me, just trying to get the CI config valid.

drigz:
@achew22 Could you review this? Hopefully it'll help avoid confusion like in #806.

Issue #808, opened 2018-11-14
Unable to add protobuf wrappers in url template option


mayank-dixit:
## Pre
1. Using string value from: https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/wrappers.proto
2. protoc version: 3.6.1

## Steps
My proto file looks like this:
>[helloworld.proto](https://github.com/mayank-dixit/examples/blob/master/grpc-gateway/protobuf-wrappers-in-url-template/helloworld.proto)
```
syntax = "proto3";

package helloworld;

import "google/api/annotations.proto";
import "google/protobuf/wrappers.proto";

service Greeter {
  rpc SayHello (HelloRequest) returns (HelloReply) {
    option (google.api.http) = {
      get: "/say/{strVal}"
    };
  }
}

message HelloRequest {
  google.protobuf.StringValue strVal = 5;
}

message HelloReply {
  string message = 1;
}

```

And then I try compile protos:
```
protoc -I/usr/local/include -I. \
  -I$GOPATH/src \
  -I$GOPATH/src/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis \
  --go_out=plugins=grpc:. \
  helloworld/helloworld.proto
```
Works fine. And then I try to generate reverse proxy using:
```
protoc -I/usr/local/include -I. \
  -I$GOPATH/src \
  -I$GOPATH/src/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis \
  --grpc-gateway_out=logtostderr=true:. \
  helloworld/helloworld.proto
```

### expected:
Successful compilation

### result: 
```
--grpc-gateway_out: aggregate type TYPE_MESSAGE in parameter of Greeter.SayHello: strVal
```

johanbrandhorst:
I've no idea why this is erroring, have you had a chance to do some debugging?

mayank-dixit:
Yes, I tried to debug:
Error is thrown from:
https://github.com/grpc-ecosystem/grpc-gateway/blob/master/protoc-gen-grpc-gateway/descriptor/services.go#L221
Where it doesn't find `*target.TypeName`(_.google.protobuf.StringValue_) to be a well known type.

Recognised well-known types are added here:
https://github.com/grpc-ecosystem/grpc-gateway/blob/master/protoc-gen-grpc-gateway/descriptor/types.go#L453

johanbrandhorst:
[...]

PR #809, opened 2018-11-14
Adding support for more well known types in descriptor

mayankcpdixit:
This change is to support more wellknown types in "protoc-gen-grpc-gateway". It solves the following issue:
Ref: https://github.com/grpc-ecosystem/grpc-gateway/issues/808

### DIFF
--- a/protoc-gen-grpc-gateway/descriptor/types.go
+++ b/protoc-gen-grpc-gateway/descriptor/types.go
@@ -451,7 +451,16 @@ var (
 	}
 
 	wellKnownTypeConv = map[string]string{
-		".google.protobuf.Timestamp": "runtime.Timestamp",
-		".google.protobuf.Duration":  "runtime.Duration",
+		".google.protobuf.Timestamp":   "runtime.Timestamp",
+		".google.protobuf.Duration":    "runtime.Duration",
+		".google.protobuf.StringValue": "runtime.StringValue",
+		".google.protobuf.BytesValue":  "runtime.BytesValue",
+		".google.protobuf.Int32Value":  "runtime.Int32Value",
+		".google.protobuf.UInt32Value": "runtime.UInt32Value",
+		".google.protobuf.Int64Value":  "runtime.Int64Value",
+		".google.protobuf.UInt64Value": "runtime.UInt64Value",
+		".google.protobuf.FloatValue":  "runtime.FloatValue",
+		".google.protobuf.DoubleValue": "runtime.DoubleValue",
+		".google.protobuf.BoolValue":   "runtime.BoolValue",
 	}
 )


mayankcpdixit:
I signed it!

protoc-gen-grpc-gateway/descriptor/types.go
```diff
@@ -451,7 +451,16 @@ var (
 	}
 
 	wellKnownTypeConv = map[string]string{
-		".google.protobuf.Timestamp": "runtime.Timestamp",
-		".google.protobuf.Duration":  "runtime.Duration",
+		".google.protobuf.Timestamp":   "runtime.Timestamp",
+		".google.protobuf.Duration":    "runtime.Duration",
+		".google.protobuf.StringValue": "runtime.StringValue",
```
mayank-dixit: @achew22 @johanbrandhorst These values don't really exist in [runtime](https://github.com/grpc-ecosystem/grpc-gateway/tree/master/runtime) package. 

So this change only helps in generating reverse proxy (gw.go code) files. It's not sufficient to support "adding protobuf wrappers in url template option" (https://github.com/grpc-ecosystem/grpc-gateway/issues/808).

protoc-gen-grpc-gateway/descriptor/types.go
```diff
@@ -451,7 +451,16 @@ var (
 	}
 
 	wellKnownTypeConv = map[string]string{
-		".google.protobuf.Timestamp": "runtime.Timestamp",
-		".google.protobuf.Duration":  "runtime.Duration",
+		".google.protobuf.Timestamp":   "runtime.Timestamp",
```
mayankcpdixit: @achew22 @johanbrandhorst These values don't really exist in [runtime](https://github.com/grpc-ecosystem/grpc-gateway/tree/master/runtime) package. 

So this change only helps in generating reverse proxy (gw.go code) files. It's not sufficient to support "adding protobuf wrappers in url template option" (https://github.com/grpc-ecosystem/grpc-gateway/issues/808).

johanbrandhorst: I will revert this MR until we have a proper fix in place then.

johanbrandhorst: Wait, what do you mean? These functions do exist in the runtime:

https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/convert.go#L16

mayankcpdixit: StringValue, BytesValue, Int32Value... etc don't.

mayankcpdixit: It'll work.But I'm afraid it'll defeat the purpose of using `StringValue` in the first place. It'll just look like we're providing support for `StringValue` but internally we'll just be giving them `String`.
Correct me if I'm wrong.

johanbrandhorst: We intentionally treat `google.protobuf.StringValue` as a string. That's a feature!

mayankcpdixit: Oh! If that's the case I'll check if I can make it work.

johanbrandhorst: I'm going to revert this since it's broken. Please submit a separate PR.

mayankcpdixit: Yes makes sense.

Issue #438, opened 2017-07-18
any way to let int64 marshal to int not string?


jinleileiking:
https://github.com/grpc-ecosystem/grpc-gateway/issues/296

But I really need this feature.

achew22:
You can also use int32 if you want your data to be represented as an int still.

jinleileiking:
@achew22 my data is bigger than int32. Can int32 work?

achew22:
If your data is bigger than an int32 then probably not but it depends on your usecase. Talk around with your team and see if there may be another strategy for attacking the problem (or use a float)

jinleileiking:
I just want to set a unixtime(epoch) which is a int64

achew22:
If you want to store a time, you should use the well known type Timestamp. This will correctly get marshalled to a Date object if you're using a swagger generated client.

https://developers.google.com/protocol-buffers/docs/reference/csharp/class/google/protobuf/well-known-types/timestamp

Raviaryacsc:
What is the reasoning behind marshaling the int64, uint64 to string instead of number.  unit32, float, doubles are marshaled as numbers. Is not it a bug which should be fixed?
Suggestion of using timestamp for epoch is not going to work.

achew22:
To partially quote an earlier issue [comment](https://github.com/grpc-ecosystem/grpc-gateway/issues/219#issuecomment-251250029):

>gRPC gateway follows the serialization rules of proto3 and the proto3 spec says that int64, fixed64, and uint64 are all strings.
>
>This is the case because of IEEE 754 Double precision which JS implements using a 52 bit mantissa (1 for sign, and 11 for exponent). This means the largest number that JS can represent without losing precision is 2^52. Try it out:
>
> ```
> (Math.pow(2,53) + 0) == (Math.pow(2,53) + 1)
>true
>```
>
>So, rather than pretending that Javascript can handle it (and lose data) proto3 prefers to have the client interact with strings. Fewer footguns == better libraries.
>
>If you want to interact with int64 as numbers, there are a bunch of JS bigint libraries out there that you can use to parse the string, perform whatever operations you want and get back a string. That would probably be a better way to handle this.

That is why I am suggesting using timestamp for epoch, since it was designed to work around the issue of the 52 bit mantissa.

Raviaryacsc:
Proto3 to JSON Mapping:
int64, uint64    --->  String
float, double    ----> number.
[...]

PR #810, opened 2018-11-15
'type' should be 'integer' for int64

koron:
According to OpenAPI 2.0 spec, `"type"` should be `"integer"` for `int64`, not `"string"`
cf: <https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md#data-types>

so current protoc-gen-swagger generates invalid swagger file when `int64` is used.

This PR fix this problem, just replace `"string"` by `"integer"`.

### DIFF
--- a/examples/proto/examplepb/wrappers.swagger.json
+++ b/examples/proto/examplepb/wrappers.swagger.json
@@ -54,7 +54,7 @@
           "format": "int32"
         },
         "int64_value": {
-          "type": "string",
+          "type": "integer",
           "format": "int64"
         },
         "float_value": {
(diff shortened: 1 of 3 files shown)


koron:
I signed it!

johanbrandhorst:
This has been discussed before in https://github.com/grpc-ecosystem/grpc-gateway/issues/438, I don't think we will be changing this as it's fundamentally wrong when applied to JSON. See https://github.com/OAI/OpenAPI-Specification/issues/704 for more discussion on this, but as the gRPC-Gateway works with JSON for swagger, we will not be changing this.

koron:
I see very well. Thank you for explaing it. 🙇

PR #811, opened 2018-11-15
Correct wellKnownTypeConv function references

### LINKED ISSUE #808: Unable to add protobuf wrappers in url template option

johanbrandhorst:
Correct wellKnownTypeConv function references.

Fixes #808

### DIFF
--- a/protoc-gen-grpc-gateway/descriptor/types.go
+++ b/protoc-gen-grpc-gateway/descriptor/types.go
@@ -453,14 +453,14 @@ var (
 	wellKnownTypeConv = map[string]string{
 		".google.protobuf.Timestamp":   "runtime.Timestamp",
 		".google.protobuf.Duration":    "runtime.Duration",
-		".google.protobuf.StringValue": "runtime.StringValue",
-		".google.protobuf.BytesValue":  "runtime.BytesValue",
-		".google.protobuf.Int32Value":  "runtime.Int32Value",
-		".google.protobuf.UInt32Value": "runtime.UInt32Value",
-		".google.protobuf.Int64Value":  "runtime.Int64Value",
-		".google.protobuf.UInt64Value": "runtime.UInt64Value",
-		".google.protobuf.FloatValue":  "runtime.FloatValue",
-		".google.protobuf.DoubleValue": "runtime.DoubleValue",
-		".google.protobuf.BoolValue":   "runtime.BoolValue",
+		".google.protobuf.StringValue": "runtime.String",
+		".google.protobuf.BytesValue":  "runtime.Bytes",
+		".google.protobuf.Int32Value":  "runtime.Int32",
+		".google.protobuf.UInt32Value": "runtime.UInt32",
+		".google.protobuf.Int64Value":  "runtime.Int64",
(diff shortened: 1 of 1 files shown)


protoc-gen-grpc-gateway/descriptor/types.go
```diff
@@ -453,14 +453,14 @@ var (
...
-		".google.protobuf.StringValue": "runtime.StringValue",
-		".google.protobuf.BytesValue":  "runtime.BytesValue",
-		".google.protobuf.Int32Value":  "runtime.Int32Value",
-		".google.protobuf.UInt32Value": "runtime.UInt32Value",
-		".google.protobuf.Int64Value":  "runtime.Int64Value",
-		".google.protobuf.UInt64Value": "runtime.UInt64Value",
-		".google.protobuf.FloatValue":  "runtime.FloatValue",
-		".google.protobuf.DoubleValue": "runtime.DoubleValue",
-		".google.protobuf.BoolValue":   "runtime.BoolValue",
+		".google.protobuf.StringValue": "runtime.String",
```
mayankcpdixit: @johanbrandhorst seems it's still broken.

I try doing `go install` where my `*.pb.gw.go` files are.
Throws: `cannot assign string to protoReq.Domain (type *wrappers.StringValue) in multiple assignment`

mayankcpdixit: One way I could find is to update our `runtime/convert.go` to support `StringValue`.
```                                                                                                                                       
func StringValue(val string) (*wrappers.StringValue, error) {                                                                                
        return &wrappers.StringValue{Value: val}, nil                                                                                        
}
```

and keep the map here in `types.go` as:
```
".google.protobuf.BoolValue":   "runtime.BoolValue"
```

johanbrandhorst:
Sigh it's clear to me that we need a little more time to work this out. I agree that your suggestion makes sense, can you submit a new PR? We need to add some examples in this area at the same time. I will revert my change to fix master.

mayankcpdixit:
Sure.

PR #812, opened 2018-11-15
Feature/patch2 rebased

razamiDev:
This PR is a cleaner version of PR #806 
This was a collaborative effort by @razamiDev and @dmacthedestroyer

this addresses #379
If a binding is mapped to patch and the request message has exactly one FieldMask message in it, additional code is rendered for the gateway handler that will populate the FieldMask based on the request body.
This handles two scenarios: The FieldMask is hidden from the REST request (as in the UpdateV2 example). In this case, the FieldMask is updated from the request body and set in the gRPC request message.
The FieldMask is exposed to the REST request (as in the PatchWithFieldMaskInBody example). For this case, a check is made as to whether the FieldMask is nil/empty prior to populating with the request body. If it's not nil, then it converts the FieldMask paths from the REST (snake_case) to gRPC (PascalCase) format. Otherwise, it acts like the previous case.
Additional comments of interest are inline on this PR.


johanbrandhorst:
Hi @razamiDev, thanks for putting this together. I just had one quick idea - do you think you could put together a couple of little examples of requests this translates and compile it into a helpful page for the documentation (/docs)? I'd love for us to make great features like this more visible to the users. What do you think?

razamiDev:
@johanbrandhorst I added some documentation. Let me know if that is what you had in mind.

johanbrandhorst:
This is fantastic @razamiDev, thank you so much! I can't wait to merge this. Lets see what CI thinks :).

johanbrandhorst:
Actually, just a quick thought, maybe add some `cURL` examples of each case as well?

razamiDev:
@johanbrandhorst good idea! I added some examples for curl :)

johanbrandhorst:
Fantastic, let's merge this thing!

johanbrandhorst:
Congrats @razamiDev and @dmacthedestroyer on finally getting this in, we've all been very excited for this!

PR #813, opened 2018-11-16
Revert "Adding support for more well known types in descriptor"

johanbrandhorst:
Reverts grpc-ecosystem/grpc-gateway#809


birdayz:
hey @johanbrandhorst :) i'm working currently on related changes as well, and well on my local branch i have exactly these changes (adding the missing WKN-types to this map) as well..
can you elaborate why the revert, which problems occurred?

johanbrandhorst:
This was reverted because the function signatures weren't correct. We want to add this, but we need to add some tests and make sure it all plumbs together. Please feel free to submit a PR with the changes.

PR #816, opened 2018-11-17
Added support for more WKT

mayankcpdixit:
Adding support for more well known types.
Adding support for wrappers (_google.protbuf.wrappers_)
Adding Example for using StringValue protobuf wrapper in url template option

Should fix: https://github.com/grpc-ecosystem/grpc-gateway/issues/808


mayankcpdixit:
@johanbrandhorst Right. Removed those.

johanbrandhorst:
It looks like the generated files don't build, can you see if we can fix the signature of the new functions?

johanbrandhorst:
This is great, thanks! I will merge once the build has finished successfully :).

drigz:
@johanbrandhorst Created #817 to avoid noise on this PR.

johanbrandhorst:
That bazel fix command is really impressive, wow! I'll merge this once the build finishes :).

mayankcpdixit:
> That bazel fix command is really impressive, wow! I'll merge this once the build finishes :).

Agree. It finished smoothly though it took really long. I felt there should be easier way so I don't have to download heavy docker images. Maybe use multistage build. idk. just thinking out loud.

johanbrandhorst:

mayankcpdixit:
No problem @johanbrandhorst . I'm using it for one of my work item. So I had to make it work anyway.

PR #819, opened 2018-11-30
Add constant behavior to protoc-gen-swagger

burov:
This PR should cover the use case when we map path to a payload data, but it is not properly reflected in generated swagger schema.

Lets assume we have following in proto:

```
message Id {
  string app_name = 1;
  string resource_type = 2;
}
```

and following gateway mappings for some of the rpcs:

```
post: /{id.app_name=app}/{id.resource_type=foo}
...
post: /{id.app_name=app}/{id.resource_type=bar}
```

In swagger schema it will look like ` /{id.app_name}/{id.resource_type}` in both cases

This PR will make plugin to generate `/app/foo` and `/app/bar` respectively which seems to be more correct. At the same time it will generate `/{id.app_name}/{id.resource_type}` is there is no `=` assignments in proto.

Thanks

### DIFF
--- a/protoc-gen-swagger/genswagger/template_test.go
+++ b/protoc-gen-swagger/genswagger/template_test.go
@@ -693,8 +693,8 @@ func TestTemplateToSwaggerPath(t *testing.T) {
 	}{
 		{"/test", "/test"},
 		{"/{test}", "/{test}"},
-		{"/{test=prefix/*}", "/{test}"},
-		{"/{test=prefix/that/has/multiple/parts/to/it/*}", "/{test}"},
+		{"/{test=prefix/*}", "/prefix/*"},
+		{"/{test=prefix/that/has/multiple/parts/to/it/*}", "/prefix/that/has/multiple/parts/to/it/*"},
 		{"/{test1}/{test2}", "/{test1}/{test2}"},
 		{"/{test1}/{test2}/", "/{test1}/{test2}/"},
 		{"/{name=prefix/*}", "/{name=prefix/*}"},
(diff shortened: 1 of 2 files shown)


johanbrandhorst:
Hi @burov, thanks for your PR! Could you please reference the part of the `http.proto` spec that defines the meaning of the `=` operator in this case? I'm not familiar with this use.

johanbrandhorst:
Btw, the CLA bot is unhappy because your commit is using a different user than your github account seems to think. They must both match unfortunately.

burov:
@johanbrandhorst see https://github.com/grpc-ecosystem/grpc-gateway/blob/master/third_party/googleapis/google/api/http.proto#L230

protoc-gen-swagger/genswagger/template_test.go
```diff
@@ -693,8 +693,8 @@ func TestTemplateToSwaggerPath(t *testing.T) {
 	}{
 		{"/test", "/test"},
 		{"/{test}", "/{test}"},
-		{"/{test=prefix/*}", "/{test}"},
-		{"/{test=prefix/that/has/multiple/parts/to/it/*}", "/{test}"},
+		{"/{test=prefix/*}", "/prefix/*"},
```
johanbrandhorst: This isn't correct anymore, is it? I don't think we should leave any `*`in the path.

johanbrandhorst:
Is this really a correct interpretation? I thought segments were used to capture paths into variable names. I don't think this produces correct output with `/*` paths.

burov:
It more about another cases explained in description to my PR
post: /{id.app_name=app}/{id.resource_type=foo} on gateway level it translate to /app/foo but on swagger to /{id.app_name}/{id.resource_type} if you have to similar handlers with different {id.resource_type} in swagger schema conflict will happen

PR #820, opened 2018-11-30
Add constant behavior to protoc-gen-swagger Duplicate for #819

burov:
This PR should cover the use case when we map path to a payload data, but it is not properly reflected in generated swagger schema.

Lets assume we have following in proto:

message Id {
  string app_name = 1;
  string resource_type = 2;
}
and following gateway mappings for some of the rpcs:

post: /{id.app_name=app}/{id.resource_type=foo}
...
post: /{id.app_name=app}/{id.resource_type=bar}
In swagger schema it will look like /{id.app_name}/{id.resource_type} in both cases

This PR will make plugin to generate /app/foo and /app/bar respectively which seems to be more correct. At the same time it will generate /{id.app_name}/{id.resource_type} is there is no = assignments in proto.

Thanks

It duplicate #819  because

### DIFF
--- a/protoc-gen-swagger/genswagger/template.go
+++ b/protoc-gen-swagger/genswagger/template.go
@@ -560,6 +560,16 @@ func templateToSwaggerPath(path string) string {
 		if isResourceName(prefix) {
 			continue
 		}
+
+		if strings.Contains(part, "=") {
+			v := strings.Split(strings.TrimSuffix(strings.TrimPrefix(part, "{"), "}"), "=")
+			if len(v) == 2 {
+				part = v[1]
+			}
+			parts[index] = part
+			continue
+		}
+
 		parts[index] = re.ReplaceAllString(part, "{$1}")
 	}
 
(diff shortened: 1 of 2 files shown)


johanbrandhorst:
Can explain why you opened this please?

burov:
@johanbrandhorst I open PR #819 and i accidentally close it i reopen it because it seems like an issue in protoc-gen-swagger , we can create two separated api like /{id.application_name=app}/{id.resource_type=foo} and /{id.application_name=app}/{id.resource_type=bar} on gateway level it works fine but on swagger schema both api seems /{id.application_name}/{id.resource_type} and it prevent to conflict on protoc-gen-swagger, i think it seems like an issue because if it possible on gateway level and work fine it must be possible in swagger schema and work fine too, Correct me if it wrong, Thanks

johanbrandhorst:
If I understand you right, you're saying that if you have two methods defined with similar paths but different segments, the gateway generator handles it fine, but the swagger generator does not? If that's true, then I agree there seems to be a bug, but I'm not sure the current fix is appropriate. Can you start by adding a failing test case instead? Assume that the existing examples are correct, and changing them will break the generator. Can you do that?

burov:
@johanbrandhorst Yes, you understand me, ok thanks,  i will do it

PR #821, opened 2018-12-05
Add filegroup for options proto files

kellycampbell:
This allows other projects to generate proto code for other languages, e.g. py_proto_library

### DIFF
--- a/protoc-gen-swagger/options/BUILD.bazel
+++ b/protoc-gen-swagger/options/BUILD.bazel
@@ -3,6 +3,20 @@ load("@io_bazel_rules_go//proto:def.bzl", "go_proto_library")
 
 package(default_visibility = ["//visibility:public"])
 
+filegroup(
+    name = "options_proto_files",
+    srcs = [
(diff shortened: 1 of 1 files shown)


achew22:
I was thinking about this, and as a curiosity, what are the circumstances where you need a `py_proto_library` for this? I expect these protos to really only be consumed by the swagger code generator

kellycampbell:
We have python gRPC backends which implement the services defined with rest annotations and documented using these swagger annotations.

It only needs them because they're imported and referenced by our services .profo file. If I run protoc for python it errors out if it can't find these options protos. Nothing in our python code actually needs the annotations.

Ideally, the py_proto_library rule should be able to use the proto provider from the proto_library to find the files, but currently the one we're using doesn't. (https://github.com/pubref/rules_protobuf/tree/master/python)

achew22:
That's an unfortunate outcome but it makes sense that proto works that way. I meant to merge earlier, but forgot to click the button. Sorry about that

PR #822, opened 2018-12-07
change ForwardResponseMessage to variable

ZhiqinYang:
sometimes user need custom  handler ForwardResponseMessage, so change this is a variable like HttpErr

### DIFF
--- a/runtime/handler.go
+++ b/runtime/handler.go
@@ -7,6 +7,7 @@ import (
 	"net/textproto"
 
 	"context"
+
 	"github.com/golang/protobuf/proto"
 	"github.com/golang/protobuf/ptypes/any"
 	"github.com/grpc-ecosystem/grpc-gateway/runtime/internal"
@@ -14,6 +15,8 @@ import (
 	"google.golang.org/grpc/grpclog"
 	"google.golang.org/grpc/status"
 )
+
+var ForwardResponseMessage = DefaultForwardResponseMessage
 
 // ForwardResponseStream forwards the stream from gRPC server to REST client.
 func ForwardResponseStream(ctx context.Context, mux *ServeMux, marshaler Marshaler, w http.ResponseWriter, req *http.Request, recv func() (proto.Message, error), opts ...func(context.Context, http.ResponseWriter, proto.Message) error) {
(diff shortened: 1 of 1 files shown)


johanbrandhorst:
Is there anything in particular you want to do that isn't already covered by using https://github.com/grpc-ecosystem/grpc-gateway/blob/7ad920cb70af14a60b14add071a4605a0655ea6e/runtime/mux.go#L41?

ZhiqinYang:
@johanbrandhorst I think maybe not, I just wanna to handle response with myself, it not stop when called WithForwardResponseOption, it will print more data which I don't need!

ZhiqinYang:
@johanbrandhorst I think to use this is enough, I can use HTTP middleware wrap response!

ZhiqinYang:
@johanbrandhorst  I think this maybe need , I handle it like this 
```
                 gwruntime.WithForwardResponseOption(func(ctx context.Context, w http.ResponseWriter, pb proto.Message) error {
		var buf = pools.Get().(*bytes.Buffer)
		buf.Reset()
		if err := marshaler.Marshal(buf, pb); err != nil {
			pools.Put(buf)
			return err
		}

		bts, err := json.Marshal(struct {
			Code int             `json:"code"`
			Msg  string          `json:"msg"`
			Data json.RawMessage `json:"data,omitempty"`
		}{
			Code: 200,
			Msg:  "Success",
			Data: buf.Bytes(),
		})

		// 回收数据 一定要放在marshal后面
		pools.Put(buf)
		if err != nil {
			return err
		}

		_, err = w.Write(bts)
		if err == nil {
			// 使用完后不允许再往body 填写数据
			((w.(*echo.Response).Writer).(*lm.CustomResponseWriter)).Complete()
		}
		return err
	})




// Complete if set , can not write into body
func (w *CustomResponseWriter) Complete() {
	w.done = true
}



func (w *CustomResponseWriter) Write(bts []byte) (int, error) {
	if w.done {
		return 0, nil
	}
	return w.w.Write(bts)
}

func (w *CustomResponseWriter) WriteHeader(statusCode int) {
	w.w.WriteHeader(statusCode)
}
```
I hava to use CustomResponseWriter to avoid forwardresponse write data to responsewriter !

johanbrandhorst:
Yeah so if you need complete control over what gets written to the ResponseWriter, you will probably need this change. Are you just trying to avoid the runtime from writing to the ResponseWriter after you ForwardResponse method has been called?

ZhiqinYang:
@johanbrandhorst  Yes, it is!

johanbrandhorst:
What messages are written by the runtime after the forward response method has been called?

PR #826, opened 2018-12-07
Write version to intermediate file for release publish

johanbrandhorst:
This was failing because the git repo is not accessible when running the `ghr` command.


johanbrandhorst:
Think I've finally got this done T_T

Issue #746, opened 2018-09-05
Comments of rpc method gets copied if multiple services are present in a proto file.


princejha95:
I have a proto file which consists of multiple services. When I generate the swagger json file from it,
the comment which are mentioned for the **first service** gets copied down to the **second service** in the json file.

Here is the proto file:

```protobuf
syntax = "proto3";

import "google/api/annotations.proto";

service SampleService1{
    // Comment 1
    rpc Method (Empty) returns (Empty){
        option (google.api.http) = {
            get : "/api/method1"
        };
    }
}

service SampleService2{
    // Comment 2
    rpc Method (Empty) returns (Empty){
        option (google.api.http) = {
            get : "/api/method2"
        };
    }
}

message Empty {}
```

Here is the generated swagger json file:

```json
{
    "paths": {
        "/api/method1": {
            "get": {
                "operationId": "Method",
                "responses": {
                    "200": {
                        "description": "",
                        "schema": {
                            "$ref": "#/definitions/Empty"
                        }
                    },
                    "404": {
                        "description": "Not Found",
                        "schema": {}
                    }
                },
                "tags": [
                    "SampleService1"
                ],
                "summary": "Comment 1"
            }
        },
        "/api/method2": {
            "get": {
                "operationId": "Method",
                "responses": {
                    "200": {
                        "description": "",
                        "schema": {
                            "$ref": "#/definitions/Empty"
                        }
                    },
                    "404": {
                        "description": "Not Found",
                        "schema": {}
                    }
                },
                "tags": [
                    "SampleService2"
                ],
                "summary": "Comment 1"
            }
        }
    },
    "schemes": [
        "http",
        "https"
    ],
    "tags": [
        {
            "name": "SampleService1",
            "description": "Description"
        },
        {
            "name": "SampleService2",
            "description": "Description"
        }
    ],
    "basePath": "/",
    "host": "localhost:3000",
    "definitions": {
        "Empty": {
            "type": "object"
        }
    },
    "swagger": "2.0"
}
```

Can anyone help in fixing this issue ?

I don't want the comments to get copied. The comments should be present as it is in json file for their respective rpc methods under their respective services.

johanbrandhorst:
Thanks for raising this issue @princejha95. I'm guessing there's either some variable being overwritten incorrectly or an index not being used here. I predict that if we add a third service with the same method name it will also give that the comment from the first service.

[...]

PR #833, opened 2018-12-18
protoc-gen-swagger: check typeIndex when typeName is Method

hexfusion:
isProtoPathMatches() doesn't return false if svcIdx != methIdx when typeName == Method. The results in comments belonging to other services polluting swagger spec. We noticed this regression when using protoc 3.5.1+, although I still can't say why it started.

Fixes: https://github.com/grpc-ecosystem/grpc-gateway/issues/746

### DIFF
--- a/examples/clients/abe/camel_case_service_name_api.go
+++ b/examples/clients/abe/camel_case_service_name_api.go
@@ -37,8 +37,7 @@ func NewCamelCaseServiceNameApiWithBasePath(basePath string) *CamelCaseServiceNa
 }
 
 /**
- * Create a new ABitOfEverything
- * This API creates a new ABitOfEverything
+ * 
  *
  * @return *interface{}
  */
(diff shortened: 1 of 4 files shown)


johanbrandhorst:
Hi @hexfusion, thanks a bunch for the contribution! Any chance you could add some new test casee for this to prevent future regressions? Thanks!

johanbrandhorst:
It seems the generator changes caused a diff in one of the generated files. Is that expected? If so, you'll need to rerun the generator (see CONTRIBUTING.md).

hexfusion:
@johanbrandhorst looking at the source of the proto I believe the change is expected, will update.

johanbrandhorst:
Looks great, thanks!

johanbrandhorst:
A new release you mean? We don't currently have a set release schedule, I can see if we're waiting for anything in particular but I might just release 1.6.4 tomorrow.

hexfusion:
that would be fantastic, and greatly appreciated.

johanbrandhorst:
FYI @hexfusion I will be backing out these changes, it seems they caused #923, please see #924. Please feel free to open another PR reverting the revert where we can add some more tests to ensure we get the best of both worlds.

PR #836, opened 2018-12-26
Fix the url of gRPC timeouts on README.md

royeo:
The link http://www.grpc.io/docs/guides/wire.html is no longer valid, it should be https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md .


johanbrandhorst:
Hi @royeo, thanks for spotting this! Could you perhaps even point it at https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests to make it even more specific? Thanks!

royeo:
> Hi @royeo, thanks for spotting this! Could you perhaps even point it at https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests to make it even more specific? Thanks!

Ok, this is really better. I've already changed it.

johanbrandhorst:
Thanks for your contribution!

PR #842, opened 2019-01-08
Use the canonical import name for resty

dweitzman:
This is more important now that resty has a go.mod file

I'm having trouble getting "make test" to work for me, so I'm hoping the github checks will do a decent job of checking for compilation errors


dweitzman:
Hmm. Looks like resty actually changed its import path in https://github.com/go-resty/resty/commit/8fa06f264236504747b03e35acf013e80f6bd04d so this diff may only make sense when the dep is being upgraded. I'll close this for now and I think I can work around it be choosing an older version of resty in my own go.mod

Issue #755, opened 2018-09-12
Support go modules


kasuboski:
Steps you follow to reproduce the error:

1. Using go1.11 outside of go path
2. Run `go get -u github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway`

Returns
```
go: finding github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway latest

# github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway/descriptor

../../go/pkg/mod/github.com/grpc-ecosystem/grpc-gateway@v1.5.0/protoc-gen-grpc-gateway/descriptor/services.go:146:49: opts.ResponseBody undefined (type *annotations.HttpRule has no field or method ResponseBody)
-->
```

What did you expect to happen instead:
Successfully added to mod file

It seems like it's able to find the latest version 1.5.0, but then isn't up to date? The error that's output seems to be what was fixed with #731

achew22:
Hey @kasuboski, thanks so much for your interest! We've been talking about doing this migration but no one has the time to do it for a while. Sounds like you're blazing a new path forward in the golang world, would you have any interest in generating anew, or migrating our existing gopkg file? That might be the quickest way to get to this working if that's your first priority. Let me know.

kasuboski:
I looked into it today. It seems like the go modules don't like having the top level dependencies like is captured with the Gopkg.lock in the root. Let me know if I misunderstood how it's currently working.

I'm not going to continue looking into it for now, but might come back to it later.

kasuboski:
Yeah I didn't read it...I ran the conversion on the repo and got the module file well enough, but then had issues with the build, namely it couldn't resolve a package for the module.

Maybe it needs multiple modules for the different packages. I had tried that as well, but had issues with it still trying to download the source from github instead of using the higher level module.

My experience with it has only been with much simpler projects so I'm definitely interested to see how this works.

johanbrandhorst:
I don't think we should need multiple modules, we'll definitely try to have a single root module and only resort to multiple if we absolutely need to. What error did you get?

kasuboski:
The error I get is
```
can't load package: package github.com/grpc-ecosystem/grpc-gateway: unknown import path "github.com/grpc-ecosystem/grpc-gateway": cannot find module providing package github.com/grpc-ecosystem/grpc-gateway
```

go.mod at the root
```
[...]

PR #844, opened 2019-01-08
Better protbuf version pin

gen1us2k:


gen1us2k:
I signed it!

johanbrandhorst:
Hi @gen1us2k. This is unfortunately one of those _it's not that simple_ cases. We're very keen to keep the versions pinned between our Bazel dependencies and our `dep` dependencies the same, and this PR only updates the `dep` version. Indeed, the comment just above the line that you edited points to the other file that you'd need to edit. Can you see if the Bazel version can be updated in this fashion too?

The longer term solution to this is #755. If you are interested in implementing that, please do, but this PR as it is is incomplete.

johanbrandhorst:
Closing this due to inactivity, please open another one if you wish to continue this work.

PR #848, opened 2019-01-09
fill example attribute of swagger schema if provided for messages

srenatus:
use json.RawMessage for property & object examples

with json.RawMessage, users can supply complete json for swagger objects (i.e.
for a proto Message). If it's correct json, it will be added to the output.json
and rendered correctly by consuming applications (i.e. code generators or
swagger-ui)

Example, in proto:
```
option (grpc.gateway.protoc_gen_swagger.options.openapiv2_schema) = {
  example: {
    value:
    '{"someKey" : "someValue", "anotherKey" : 5}';
  };
};
```

Rendered JSON:
```
"example": {
  "someKey": "someValue",
  "anotherKey": 5
}
```

If we used string instead of json.RawMessage, the serializer would escape the
quotes and it would be serialized as a string. This is not desirable, because
the output should be in fact JSON key/value pairs.

Also:

* a_bit_of_everything.proto: add 'example' example

Co-authored-by: Johannes Brüderl <johannes.bruederl@gmail.com>
Signed-off-by: Stephan Renatus <srenatus@chef.io>

----
🌻 it's all @birdayz's fine work 🎈


srenatus:
OMG so I'll leave off the `Co-authored-by`... 🤕

johanbrandhorst:
We'll get there in the end :joy:

PR #849, opened 2019-01-09
test: "fill attributes of swagger schema if provided for messages"

srenatus:
Back-filling tests for #799:

* [X] a_bit_of_everything.proto: add 'example' example
* [X] add basic tests
* [ ] ~test nested messages~ maybe next time 😅


johanbrandhorst:
Right then; re: test, I think it's good that we've got an example added to `a_little_bit_of_everything.proto`, but I'm wondering if we can add a test in https://github.com/grpc-ecosystem/grpc-gateway/blob/master/protoc-gen-swagger/genswagger/template_test.go that tests that populating the example field makes it into the output? What do you think?

srenatus:
Yup, will do

johanbrandhorst:
Seems about as good as my understanding of these tests :grimacing:. Please try and add a new structure within which we can test the messages :). Also, maybe add some more examples in `a_little_bit_of_everything.proto`, to illustrate that it's verbatim JSON that is the format expected.

birdayz:
Thanks a lot for taking over! I didn't find the time to finish it. Also, my code was lacking one thing: setting the example value on field level, as the protobuf option doesn't make it possible (see my old comment in #799). Do you think you could work on that as well? Otherwise, you can only set the example for the WHOLE rpc, which is annoying i think.

achew22:
@birdayz I'm in strong agreement that your contribution shouldn't be lost. I will make an exception to the single commit/PR rule to ensure that you get credit for your hard work. Thanks so much for contributing to the project. Without people like you this couldn't work

srenatus:
:tada: That's perfect. I'll restore the commit history tomorrow :)

srenatus:
Alternatively, we could merge #799 as-is and I'll put up a test-only PR following that.

johanbrandhorst:
Yeah, let's merge the original and add tests asynchronously, that'd be the easiest solution.

johanbrandhorst:
This is great, thanks for taking the lead on this @srenatus. Looks like we've got a bazel failure. You should be able to fix it like so:
```
    l.gcr.io/google/bazel -c 'bazel run :gazelle_fix; bazel run :buildifier'
```

johanbrandhorst:
Great work on this, thanks! Increasing the coverage of that file by 5%!

Issue #851, opened 2019-01-17
Support swagger annotations for default and required fields


fahernandez:
Swagger annotation for default and required fields aren't suportted by the swagger generator tool. The idea is to support the annotations as:
  ```
  float float_value = 3 [(grpc.gateway.protoc_gen_swagger.options.openapiv2_field) = {description: "Float value field", default: "0.2", required: ['float_value']}];
  ```

That will generated the swagger json as:
  ```json
          {
            "name": "float_value",
            "description": "Float value field",
            "in": "query",
            "required": true,
            "type": "number",
            "format": "float",
            "default": "0.2"
          }
  ```

johanbrandhorst:
Closed by #852

PR #852, opened 2019-01-17
Adding support for default and required swagger annotation fields.

fahernandez:
Solves issue #851


examples/clients/abe/.travis.yml
```diff
@@ -0,0 +1,8 @@
+language: go
+
+install:
+  - go get -d -v .
+
+script:
+  - go build -v ./
+
```
johanbrandhorst: This file shouldn't have been generated :thinking:. Could you delete it please?

examples/clients/echo/.travis.yml
```diff
@@ -0,0 +1,8 @@
+language: go
+
+install:
+  - go get -d -v .
+
+script:
+  - go build -v ./
+
```
johanbrandhorst: Ditto these files

examples/clients/responsebody/.travis.yml
```diff
@@ -0,0 +1,8 @@
+language: go
+
+install:
+  - go get -d -v .
+
+script:
+  - go build -v ./
+
```
johanbrandhorst: .

examples/clients/unannotatedecho/.travis.yml
```diff
@@ -0,0 +1,8 @@
+language: go
+
+install:
+  - go get -d -v .
+
+script:
+  - go build -v ./
+
```
johanbrandhorst: .

examples/proto/examplepb/a_bit_of_everything.swagger.json
```diff
@@ -1452,7 +1456,12 @@
         },
         "float_value": {
           "type": "number",
-          "format": "float"
+          "format": "float",
+          "default": "0.2",
+          "description": "Float value field",
+          "required": [
+            "float_value"
+          ]
```
johanbrandhorst: Is this expected?

protoc-gen-swagger/genswagger/template_test.go
```diff
@@ -67,13 +67,15 @@ func TestMessageToQueryParameters(t *testing.T) {
...
 					Type:     "string",
+					Default:  "1",
 				},
 				swaggerParameterObject{
 					Name:     "b",
 					In:       "query",
 					Required: false,
 					Type:     "number",
 					Format:   "double",
+					Default:  "2",
```
johanbrandhorst: Should this not be the result in different inputs?

examples/proto/examplepb/a_bit_of_everything.proto
```diff
@@ -168,7 +168,7 @@ message ABitOfEverything {
 
 	string uuid = 1 [(grpc.gateway.protoc_gen_swagger.options.openapiv2_field) = {pattern: "[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}", min_length: 1}];
 	repeated Nested nested = 2;
-	float float_value = 3;
+	float float_value = 3 [(grpc.gateway.protoc_gen_swagger.options.openapiv2_field) = {description: "Float value field", default: "0.2", required: ['float_value']}];
```
johanbrandhorst: I'm a little confused about this taking an array of "required" parameters. What does it mean in this case?

examples/clients/abe/.travis.yml
fahernandez: ok

examples/clients/echo/.travis.yml
fahernandez: done

examples/clients/responsebody/.travis.yml
fahernandez: done

examples/clients/unannotatedecho/.travis.yml
fahernandez: done

johanbrandhorst:
It looks like this change breaks defaults when the value is unset as well - I think we need to make sure we maintain the current behavior where a default isn't specified.

examples/proto/examplepb/a_bit_of_everything.swagger.json
```diff
@@ -1418,16 +1422,14 @@
       "enum": [
         "GHI",
         "JKL"
-      ],
-      "default": "GHI"
```
johanbrandhorst: This change is what I mean. Could we ensure this doesn't go away?

johanbrandhorst:
See my last comment, enum fields had their default set to the first value previously.

examples/proto/examplepb/a_bit_of_everything.proto
johanbrandhorst: I assume it's this way because when the field is a _Message_ type you are able to specify several of them? Could we try to make an example where several are of them are specified?

examples/proto/examplepb/a_bit_of_everything.swagger.json
fahernandez: Oh I see , sorry, I will remove the code that breaks this part

examples/proto/examplepb/a_bit_of_everything.proto
fahernandez: Sure I can but inside openapiv2_field It won't have any effect because the field is translated to a boolean value (here)[https://github.com/grpc-ecosystem/grpc-gateway/pull/852/files#diff-d419038d63db0fdabc61f0d6cc6683aeR125]. openapiv2_field and openapiv2_schema use (swaggerSchemaObject)[https://github.com/grpc-ecosystem/grpc-gateway/blob/master/protoc-gen-swagger/genswagger/types.go#L200] as schema, that's why on both cases it has to be managed as array but only on openapiv2_schema has any sense to use it as a multiple value array. On  openapiv2_field is used as array due to the actual implementation but it will have more sense to used it as a boolean.

johanbrandhorst: I see, nevermind then. It feels a little bit strange though, how much work would it be to make it a boolean instead?

johanbrandhorst: Isn't there somewhere we can specify multiple `required` for a Message still? I understand the field uses an array for other reasons, but it looks like the swagger definition expects an array, and specifies an array. Can we find one such example?

johanbrandhorst:
Thanks for this contribution!

PR #853, opened 2019-01-17
Add flag 'allow_repeated_fields_in_body' to protoc-gen-swagger

abice:
PR #712 provided support for repeated fields in the body of a response to protoc-gen-gateway, but failed to add it to protoc-gen-swagger.

I simply copied the flag over and updated the registry, and it handles the swagger generation for the same scenario.

```proto
syntax = "proto3";

package v1;

service Service {
  rpc ListStrings(ListStringsRequest) returns (ListStringsResponse) {
    option (google.api.http) = {
      get : "/v1/strings"
      response_body : "values"
    };
  }
}

message ListStringsRequest {}
message ListStringsResponse { repeated string values = 1; }
```


abice:
I asked that question in the slack room, this change *does* change the output of the swagger file, but it requires a CLI parameter in order to trigger the change, and I was having trouble getting the bazel part to run with the added flag required.  I can go back and check to see what bazel stuff I need to change in order to pass in the flag, but I'm unfamiliar with bazel, and it might take a bit.

johanbrandhorst:
I see.. could you add a new file (e.g. the snippet in your PR) that we can generate with the relevant CLI flag and add it to the Makefile?

johanbrandhorst:
Ignore Bazel, I don't know if it's relevant in this case, the generation all goes through the Makefile AFAIK.

abice:
Well, it's less bazel, and more the circle step of running bazel... I had updated the response_body_service.proto to include a sample since it was related, but ran into problems... I'll add a new proto that just requires that flag.

johanbrandhorst:
What was the problems with that proto if I might ask? I hope the generation step is easy enough? A lot of effort has gone into making this process simpler so let me know if there's anything we can improve.

abice:
Well, my first attempt was to just change the circleci config to add the `SWAGGER_PLUGIN_FLAGS` and `GATEWAY_PLUGIN_FLAGS` to the `make examples` line like:
```make
GATEWAY_PLUGIN_FLAGS=allow_repeated_fields_in_body=true SWAGGER_PLUGIN_FLAGS=allow_repeated_fields_in_body=true make examples SWAGGER_CODEGEN="${SWAGGER_CODEGEN}" # Set in Docker image
```

and then generated everything like it says in the contributing.md

but when I ran the bazel job locally using `circleci local execute --job bazel`, it failed on
```
/go/src/github.com/grpc-ecosystem/grpc-gateway/examples/proto/examplepb/BUILD.bazel:39:1: Generating into bazel-out/k8-fastbuild/bin/examples/proto/examplepb/linux_amd64_race_stripped/examplepb_go_proto%/github.com/grpc-ecosystem/grpc-gateway/examples/proto/examplepb failed (Exit 1) go-protoc failed: error executing command bazel-out/host/bin/external/io_bazel_rules_go/go/tools/builders/linux_amd64_stripped/go-protoc -protoc bazel-out/host/bin/external/com_google_protobuf/protoc -importpath ... (remaining 86 argument(s) skipped)
```

Maybe I missed that it was looking at the BUILD.bazel file in the examplepb folder.  But I dont' know where I would add the extra flag in that file.

johanbrandhorst:
I can't help with that error I'm afraid (@achew22?) but perhaps it will be easier to add a new file :sweat_smile:.

achew22:
I think adding a new option to the compiler is the way to go.

```
    options = ["logtostderr=true"],
```

Becomes

```
options = [
    "your_new=option",
    "logtostderr=true",
]
```


https://github.com/grpc-ecosystem/grpc-gateway/blob/20e8cf9fb34a4c43782226b3e0dd42913a5c0a9b/protoc-gen-grpc-gateway/BUILD.bazel#L28

abice:
Awesome, thanks!  I'll add that and see where I get

PR #854, opened 2019-01-18
marshal_jsonpb: Added nil slice default value

abice:
I noticed that when using the repeated response body, that when an empty array is returned from the grpc server, it is translated to a nil object `([]*examplepb.RepeatedResponseBodyOut_Response)(nil)`, which does not implement the `proto.Message` interface, and therefore is marshalled locally in the `runtime.JSONPb` implementation as `null` even if `EmitDefaults` is set on the marshaller.

I updated the marshaller to check for the case mentioned above and to make sure it behaves according to the `EmitDefaults` option correctly.  Let me know if I missed anything on it.

Thanks!

### DIFF
--- a/runtime/marshal_jsonpb.go
+++ b/runtime/marshal_jsonpb.go
@@ -67,6 +67,12 @@ func (j *JSONPb) marshalNonProtoField(v interface{}) ([]byte, error) {
 		rv = rv.Elem()
 	}
 
+	if rv.Kind() == reflect.Slice {
+		if rv.IsNil() && j.EmitDefaults {
+			return []byte("[]"), nil
+		}
+	}
+
 	if rv.Kind() == reflect.Map {
 		m := make(map[string]*json.RawMessage)
 		for _, k := range rv.MapKeys() {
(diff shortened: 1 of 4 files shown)


runtime/marshal_jsonpb.go
```diff
@@ -67,6 +67,12 @@ func (j *JSONPb) marshalNonProtoField(v interface{}) ([]byte, error) {
 		rv = rv.Elem()
 	}
 
+	if rv.Kind() == reflect.Slice {
+		if rv.IsNil() && j.EmitDefaults {
+			return []byte("[]"), nil
+		}
```
johanbrandhorst: I suppose we could put this all in one big if instead to reduce indentation?

abice: That works for me, honestly wasn't sure if you wanted me to return `"null"` if j.EmitDefaults wasn't set.  I can make it all one big if.

👍

johanbrandhorst:
Thanks for the contribution!

PR #856, opened 2019-01-20
marshal_jsonpb: add check for slice sub types implementing proto.Message

abice:
I found that the slice of response bodies, even if they implement `proto.Message` are still being marshalled using the default json marshaller and ignoring the `EnumsAsInts` flag.  I've updated the implementation to handle this case, and pass off each individual object in the slice to the jsonpb marshaller.

Let me know if there's a better way of handling these slice types that I'm missing.

Thanks!


runtime/marshal_jsonpb.go
```diff
@@ -67,8 +72,38 @@ func (j *JSONPb) marshalNonProtoField(v interface{}) ([]byte, error) {
...
 	}
 
-	if rv.Kind() == reflect.Slice && rv.IsNil() && j.EmitDefaults {
-		return []byte("[]"), nil
+	if rv.Kind() == reflect.Slice {
+		if rv.IsNil() {
+			if j.EmitDefaults {
+				return []byte("[]"), nil
+			}
+			return []byte("null"), nil
```
johanbrandhorst: In this case we probably want to write nothing?

runtime/marshal_jsonpb_test.go
```diff
@@ -727,6 +727,37 @@ func TestJSONPbMarshalResponseBodies(t *testing.T) {
...
+					Type: examplepb.RepeatedResponseBodyOut_Response_B,
+				},
+			},
+			verifier: func(json string) {
+				expected := `[{"data":"","type":"UNKNOWN"},{"data":"abc","type":"B"}]`
+				if json != expected {
+					t.Errorf("json not equal (%q, %q)", json, expected)
+				}
+			},
+		},
```
johanbrandhorst: Could we add a test case for `EmitDefaults: false` and an empty/nil slice?

abice: Test case 5 (line 665) and 7 (line 684) are already testing those cases.
```
{
			input: ([]*examplepb.RepeatedResponseBodyOut_Response)(nil),
			verifier: func(json string) {
				expected := `null`
				if json != expected {
					t.Errorf("json not equal (%q, %q)", json, expected)
				}
			},
		},
```
and 
```
		{
			input: []*examplepb.RepeatedResponseBodyOut_Response{},
			verifier: func(json string) {
				expected := `[]`
				if json != expected {
					t.Errorf("json not equal (%q, %q)", json, expected)
				}
			},
		},
```

PR #857, opened 2019-01-23
Use github.com/golang/protobuf/descriptor ForMessage and fix CI from not rebasing

mechinn:
@johanbrandhorst this is the fix because I didn't rebase on master before you merged, also stumbled on https://godoc.org/github.com/golang/protobuf/descriptor as a replacement for the function I built to extract Descriptor

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -15,6 +15,11 @@ required = [
 
 [[constraint]]
   # Also defined in WORKSPACE
+  revision = "f8815663de1e64d57cdd4ee9e2b2fa96977a030e"
+  name = "github.com/go-resty/resty"
+
+[[constraint]]
+  # Also defined in WORKSPACE
   revision = "6724a57986aff9bff1a1770e9347036def7c89f6"
   name = "github.com/rogpeppe/fastuuid"
 
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -1,5 +1,4 @@
 required = [
-  "github.com/golang/protobuf/descriptor",
   "github.com/golang/protobuf/protoc-gen-go",
 ]
 
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -1,4 +1,5 @@
 required = [
+  "github.com/golang/protobuf/descriptor",
   "github.com/golang/protobuf/protoc-gen-go",
 ]
 
@@ -12,11 +13,6 @@ required = [
   # Also defined in WORKSPACE
   revision = "0ca9ea5df5451ffdf184b4428c902747c2c11cd7"
   name = "github.com/ghodss/yaml"
-
-[[constraint]]
-  # Also defined in WORKSPACE
-  revision = "f8815663de1e64d57cdd4ee9e2b2fa96977a030e"
-  name = "github.com/go-resty/resty"
 
 [[constraint]]
   # Also defined in WORKSPACE
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -23,9 +23,10 @@
   revision = "23def4e6c14b4da8ac2ed8007337bc5eb5007998"
 
 [[projects]]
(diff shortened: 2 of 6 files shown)


Gopkg.toml
```diff
@@ -1,4 +1,5 @@
 required = [
+  "github.com/golang/protobuf/descriptor",
```
johanbrandhorst: You shouldn't need to explicitly add this as we're importing it.

johanbrandhorst:
Hm, I figured our CI tests would have flagged that up :thinking:.

johanbrandhorst:
> There should be an unused function now to delete as well, right?
By this I mean the function where we set the sourceCodeInfo explicitly before.

Gopkg.toml
mechinn: you would think wouldn't you:

`"github.com/golang/protobuf/descriptor" is not imported by your project, and has been temporarily added to Gopkg.lock and vendor/.
If you run "dep ensure" again before actually importing it, it will disappear from Gopkg.lock and vendor/.`

johanbrandhorst: That doesn't sound right at all

mechinn: agreed, maybe it's because there are no go files in the root of the repo? dep is weird sometimes...

mechinn:
> > There should be an unused function now to delete as well, right?
> 
> By this I mean the function where we set the sourceCodeInfo explicitly before.

yeah good catch, removed

Gopkg.toml
johanbrandhorst: I don't think that should matter. I've never had to add a package explicitly like this. Let me try checking out this branch.

Gopkg.toml
```diff
@@ -13,11 +14,6 @@ required = [
   revision = "0ca9ea5df5451ffdf184b4428c902747c2c11cd7"
   name = "github.com/ghodss/yaml"
 
-[[constraint]]
-  # Also defined in WORKSPACE
-  revision = "f8815663de1e64d57cdd4ee9e2b2fa96977a030e"
-  name = "github.com/go-resty/resty"
-
```
johanbrandhorst: Where did this change come from?

Gopkg.toml
mechinn: huh you're right, if I get rid of the requires everything still works fine, I guess I just confused dep by doing `dep ensure -add github.com/golang/protobuf/descriptor`

Gopkg.toml
mechinn: me doing `dep ensure -add github.com/golang/protobuf/descriptor` for some reason removed it and I can't recreate why at this point, reverted that change

johanbrandhorst: `dep` works in mysterious ways

Gopkg.toml
```diff
@@ -13,9 +13,9 @@ required = [
   revision = "0ca9ea5df5451ffdf184b4428c902747c2c11cd7"
   name = "github.com/ghodss/yaml"
 
-[[constraint]]
-  # Also defined in WORKSPACE
-  revision = "f8815663de1e64d57cdd4ee9e2b2fa96977a030e"
+[[constraint]]	
+  # Also defined in WORKSPACE	
+  revision = "f8815663de1e64d57cdd4ee9e2b2fa96977a030e"
```
johanbrandhorst: Don't make me say it :joy: :

mechinn: baaaaaaahhhhhhhh

mechinn:
> Hm, I figured our CI tests would have flagged that up 🤔.

considering another change that was merged before my other one changed the ABitOfEverything message and we had never generated the stream.swagger.json before my change I would have been impressed if the CI picked up this issue before the merge into master

johanbrandhorst:
Thanks for this, I will confirm master builds successfully and release 1.7.0 :tada:

PR #859, opened 2019-01-24
add the package name to the tags field of each endpoint in protoc-gen-swagger

zwcn:
Currently, given a path and a http method, there is no way to identify which package it is generated from. For example, since the following swagger file generated by `protoc-gen-swagger` contains only the service name `dataService` and the rpc name `GetData`, we can't derive the _package_ name in the proto file that defines `dataService` and `GetData`. 

```json
{
  ...
  "paths": {
    "/api/v1/data": {
      "get": {
          "operationId": "GetData", // <--- the rpc name
          "tags": [
            "dataService" // <--- the service name
          ]
    }
  }
  ...
}
```

This pull request prepends the package name to the `service` name in the `tags` field. For example, if the proto file includes a line `package data;`, the tags field will be `["data.dataService"]`.  Doing this allows us to trace back the origin of the `GET /api/v1/data` http request.

### DIFF
--- a/protoc-gen-swagger/genswagger/template.go
+++ b/protoc-gen-swagger/genswagger/template.go
@@ -821,7 +821,7 @@ func renderServices(services []*descriptor.Service, paths swaggerPathsObject, re
 					responseSchema.Ref = strings.Replace(responseSchema.Ref, `#/definitions/`, `#/x-stream-definitions/`, 1)
 				}
 				operationObject := &swaggerOperationObject{
-					Tags:       []string{svc.GetName()},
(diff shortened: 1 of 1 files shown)


johanbrandhorst:
Hi @zwcn, sorry to see you close this PR, do you disagree with the CLA or is the another reason? It sounds like a good idea.

zwcn:
Hi @johanbrandhorst , I'd like to sign the CLA with another email address. I'm gonna reopen the PR once I figure it out how to do it.

PR #860, opened 2019-01-24
protoc-gen-swagger: add the package name to the tags field of each endpoint if the package name exists in the proto file

zwcn:
Currently, given a path and a http method, there is no way to identify which package it is generated from. For example, since the following swagger file generated by `protoc-gen-swagger` contains only the service name `dataService` and the rpc name `GetData`, we can't derive the package name in the proto file that defines `dataService` and `GetData`.

```json
{
  ...
  "paths": {
    "/api/v1/data": {
      "get": {
          "operationId": "GetData", // <--- the rpc name
          "tags": [
            "dataService" // <--- the service name
          ]
    }
  }
  ...
}
```

This pull request prepends the package name to the service name in the tags field. For example, if the proto file includes a line `package data;`, the tags field will be `["data.dataService"]`. Doing this allows us to trace back the origin of the `GET /api/v1/data` http request.


johanbrandhorst:
Hi again @zwcn. It looks like you need to regenerate the files after adding this fix. Please see the [contribution guide](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/CONTRIBUTING.md#i-want-to-regenerate-the-files-after-making-changes) for more information.

protoc-gen-swagger/genswagger/template.go
```diff
@@ -821,7 +821,7 @@ func renderServices(services []*descriptor.Service, paths swaggerPathsObject, re
 					responseSchema.Ref = strings.Replace(responseSchema.Ref, `#/definitions/`, `#/x-stream-definitions/`, 1)
 				}
 				operationObject := &swaggerOperationObject{
-					Tags:       []string{svc.GetName()},
+					Tags:       []string{fmt.Sprintf("%s.%s", svc.File.GetPackage(), svc.GetName())},
```
johanbrandhorst: What will this produce if there is no package name?

zwcn: if there's no package name, there will be only a dot prepended to the service name, like `.svcName`. It seem to me that's fine if we assume the tag always consists of two parts: the package name and the service name. We could use `strings.Split` to extract the two parts even if the optional `package` name is not defined in the proto file.

johanbrandhorst: I think we should only add the package name if it is set. Could we just add a:

```go
tag := svc.GetName()
if pkg := svc.File.GetPackage(); pkg != "" {
    tag = pkg + "." + tag
}
```
?

zwcn: Sure, that's a good idea. Will do.

johanbrandhorst:
Looks like we need to update some of the tests after this change:
```
examples/integration/client_test.go
spec.js
```

zwcn:
@johanbrandhorst , yes. As shown at https://github.com/grpc-ecosystem/grpc-gateway/pull/860/files#diff-eab1d67e52af78e98dd23e2b08fcc22aR25 , `swagger-codegen` generates go files based on the `tags` field of the input swagger file.

How do you feel about these long-name changes?

johanbrandhorst:
Haha yeah I noticed. I suppose it's not wrong but they're not pretty names. I wonder if we can remove the tag from the generated names with a generator option? Please take a look.

zwcn:
Hi @johanbrandhorst , 

Yes, the names aren't wrong. As shown at https://github.com/swagger-api/swagger-codegen/wiki/FAQ#how-do-tags-affect-the-generated-code, the `tags` filed is required for `swagger-codegen` to group APIs. So `swagger-codegen` did the right thing.

i've browsed the `swagger-codegen` docs. Unfortunately, I couldn't find any argument for `swagger-codegen` to ignore (or override) tags during code generation.

One possible solution would be to add a new boolean flag called `allow_package_in_tags` to `protoc-gen-swagger`. The package name is prepended only when it's `true`. This gives users more flexibility to have pretty names when they don't care about an API's origin at all.

How do you think?

johanbrandhorst:
I think we're gonna have to add a flag as you say as this could be considered backwards compatibility breaking as well. Could you update the PR with a flag please? See if you could add some tests or a new file generated with the new flag as well please.

zwcn:
@johanbrandhorst : Sounds like a plan. Will do.

protoc-gen-grpc-gateway/descriptor/registry.go
```diff
@@ -50,6 +50,10 @@ type Registry struct {
 	// allowRepeatedFieldsInBody permits repeated field in body field path of `google.api.http` annotation option
 	allowRepeatedFieldsInBody bool
 
+	// allowPackageNameInTags controls whether the package name defined in the `package` directive
+	// in the proto file can be prepended to the gRPC service name in the `Tags` field of every operation.
+	allowPackageNameInTags bool
```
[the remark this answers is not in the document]
zwcn: Yeah, `include_package_in_tags` is better. I've done the replacement everywhere.

johanbrandhorst:
Thanks a lot for your contribution!

PR #861, opened 2019-01-28
add protoc-gen-swagger parameter -zero_examples=true

therealmikelee:
* parameter will cause generation of swagger.json field example values w/ zero values.

### DIFF
--- a/protoc-gen-grpc-gateway/descriptor/registry.go
+++ b/protoc-gen-grpc-gateway/descriptor/registry.go
@@ -57,6 +57,9 @@ type Registry struct {
 	// otherwise the original proto name is used. It's helpful for synchronizing the swagger definition
 	// with grpc-gateway response, if it uses json tags for marshaling.
 	useJSONNamesForFields bool
+
+	// emit examples w/ zero values
+	emitZeroExamples bool
 }
 
 type repeatedFieldSeparator struct {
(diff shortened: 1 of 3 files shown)


therealmikelee:
> I signed it!

johanbrandhorst:
Hi @therealmikelee, thanks for this contribution! Any chance we could add a new proto file with some generic types that we could generate with this new option? It'll both serve as a test that we don't break it in the future and an example to our users about what this flag does.

What do you think?

(PS I LOVE your avatar. SQUIRREL!)

therealmikelee:
> Hi @therealmikelee, thanks for this contribution! Any chance we could add a new proto file with some generic types that we could generate with this new option? It'll both serve as a test that we don't break it in the future and an example to our users about what this flag does.
> 
> What do you think?
> 
> (PS I LOVE your avatar. SQUIRREL!)

Thanks, I will try to get this done as time permits.  I'm working on windows and I'm admittedly ignorant as to getting the Makefile running properly, but perhaps I can just manually generate the things I need.  I was thinking about using a_bit_of_everything.proto.

johanbrandhorst:
I don't think you should turn this flag on for `a_little_bit_of_everything.proto`, that would cause lots of changes and it serves as an example of what the files look like _without_ this new setting enabled.

As for generation, you should be able to edit the makefile and try things out by regenerating (see https://github.com/grpc-ecosystem/grpc-gateway/blob/master/CONTRIBUTING.md#i-want-to-regenerate-the-files-after-making-changes).

Issue #863, opened 2019-02-07
Repeated field documentation is overwritten by fields comments


fahernandez:
### Bug report:

When defining repeated fields the documentation of the field is overwritten by the fields comments. For example:

```
// Some field comments
repeated string names = 2 [(grpc.gateway.protoc_gen_swagger.options.openapiv2_field) =
    {
        description:
            "Some field description documentation."
           array: ["name1", "name2"];
    }];
```
After generating the swagger documentation the field comments takes precendence over the field documentation resulting in something like:

```
{
        "name": "names",
        "description": "Some field comments",
        "in": "query",
        "required": false,
        "type": "array",
        "items": {
             "type": "string"
         }
}
```

I think that the field documentation must take precedence over the field comments.

achew22:
This would be a really great starter project for getting involved!

This comes from the interaction of these two pieces of code.

https://github.com/grpc-ecosystem/grpc-gateway/blob/c3923b1034cd4a10cea4dfae173804b19f6f7fe0/protoc-gen-swagger/genswagger/template.go#L120-L124

https://github.com/grpc-ecosystem/grpc-gateway/blob/c3923b1034cd4a10cea4dfae173804b19f6f7fe0/protoc-gen-swagger/genswagger/template.go#L294-L296

A little bit of rejiggering in there + a new test case and I think you'd be done if you're interested.

fahernandez:
Sure, I will definitely try, thank you for taking me in the right direction!

johanbrandhorst:
Closed by #863

PR #865, opened 2019-02-11
863 adding swagger annotation support for enum and nested objects

fahernandez:


fahernandez:
Linked to [https://github.com/grpc-ecosystem/grpc-gateway/issues/863](issue) 836

examples/proto/examplepb/echo_service.swagger.json
```diff
@@ -1,7 +1,7 @@
 {
   "swagger": "2.0",
   "info": {
-    "title": "Echo Service",
+    "title": "examples/proto/examplepb/echo_service.proto",
```
achew22: This seems like an odd change. Why did this become the file path?

protoc-gen-swagger/genswagger/template.go
```diff
@@ -1214,26 +1221,32 @@ func updateSwaggerDataFromComments(swaggerObject interface{}, comment string) er
 
 	paragraphs := strings.Split(comment, "\n\n")
 
-	// If there is a summary (or summary-equivalent), use the first
+	// If there is a summary (or summary-equivalent) and its empty, use the first
```
achew22: its => it's

examples/proto/examplepb/echo_service.swagger.json
fahernandez: I will remove it but the strange thing is that is part of the autogenerated code so I don't know why it is changed.

fahernandez: Actually, I if remove that line, the build process fails https://circleci.com/gh/grpc-ecosystem/grpc-gateway/1982?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

protoc-gen-swagger/genswagger/template.go
```diff
@@ -408,31 +408,35 @@ func schemaOfField(f *descriptor.Field, reg *descriptor.Registry, refs refMap) s
...
-			updateSwaggerObjectFromJSONSchema(&ret, j)
-		}
-		return ret
 	}
+
+	if j, err := extractJSONSchemaFromFieldDescriptor(fd); err == nil {
+		updateSwaggerObjectFromJSONSchema(&ret, j)
+	}
+
+	return ret
```
johanbrandhorst: Is this change necessary? It changes the way we handle `object` and `array` aggregates for no obvious reason to me.

protoc-gen-swagger/genswagger/template.go
```diff
@@ -1189,7 +1196,7 @@ func applyTemplate(p param) (*swaggerObject, error) {
 //
 // If there is no 'Summary', the same behavior will be attempted on 'Title',
 // but only if the last character is not a period.
-func updateSwaggerDataFromComments(swaggerObject interface{}, comment string) error {
+func updateSwaggerDataFromComments(swaggerObject interface{}, comment string, typeName string) error {
```
johanbrandhorst: Instead of passing in the typename and adding if cases, can't we skip calling this function altogether if the summary isn't empty?

protoc-gen-swagger/genswagger/template.go
fahernandez: This change allows add field annotations to objects and arrays, before these changes, all field annotations added to object or arrays were ignored.

protoc-gen-swagger/genswagger/template.go
fahernandez: This methods handle specific behavior about setting title/info/description/summary. If we avoid calling this method, this logic needs to be handled on an upper level. The problem is that by default the package title is set to the file name protoc-gen-swagger/genswagger/template.go:949(I don't know why but I suppose we want to keep this behavior) so by changing the behavior of field options taking precedence over comments the title of all protos without package definition was overwritten by the file name so to handle this specific case, the **if** was added

johanbrandhorst: I see, thanks. Maybe we can change the parameter from a string to a boolean at least? `IsPackageObject`?

fahernandez: sure, I will change it

protoc-gen-swagger/genswagger/template.go
fahernandez: I will change it

examples/proto/examplepb/echo_service.swagger.json
fahernandez: The title of the documents was kept due to some changes when generating the swagger documentation.

achew22:
@fahernandez, thanks so much for finding a problem and reporting it but then going above and beyond to fix it for everyone! Really really great work. I look forward to doing it again soon.

PR #878, opened 2019-02-27
protoc-gen-swagger: return error when encoding swagger file

elliots:
Json marshalling errors were being ignored, and it was generating a blank string swagger file. 

This PR causes the error to be returned.

I ran into this (and it very well could be the only way this could happen) when using the json schema example option, which must be valid json.

For example:

```proto

message Request {
    option (grpc.gateway.protoc_gen_swagger.options.openapiv2_schema) = {
        example: { value: 'this is not json'}
    };

    bytes content = 90;
}
```

would silently fail.

Now, it fails like this:

```
failed to encode swagger for attachment/web/attachment.proto: json: error calling MarshalJSON for type json.RawMessage: invalid character 'h' in literal true (expecting 'r')
```
which isn't great, but is better.

I'm using this from inside https://github.com/utrack/clay/ but I think it would apply the same using it directly.

### DIFF
--- a/protoc-gen-swagger/genswagger/generator.go
+++ b/protoc-gen-swagger/genswagger/generator.go
@@ -67,11 +67,13 @@ func mergeTargetFile(targets []*wrapper, mergeFileName string) *wrapper {
 }
 
 // convert swagger file obj to plugin.CodeGeneratorResponse_File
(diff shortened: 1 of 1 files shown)


johanbrandhorst:
Thanks for your contribution!

PR #879, opened 2019-02-28
Fix parameter names when using JSON names.

brocaar:
When using `json_names_for_fields=true`, I noticed that this works fine for field names in the request body when generating Swagger definitions, but URL parameters remain `snake_case`.

I believe this should fix it. Looking forward to your feedback.

### DIFF
--- a/protoc-gen-swagger/genswagger/template.go
+++ b/protoc-gen-swagger/genswagger/template.go
@@ -143,9 +143,7 @@ func queryParams(message *descriptor.Message, field *descriptor.Field, prefix st
 
 		if reg.GetUseJSONNamesForFields() {
 			param.Name = prefix + field.GetJsonName()
-		}
-
-		if !reg.GetUseJSONNamesForFields() || param.Name == "" {
+		} else {
 			param.Name = prefix + field.GetName()
 		}
 
--- a/protoc-gen-swagger/genswagger/template.go
+++ b/protoc-gen-swagger/genswagger/template.go
@@ -143,7 +143,9 @@ func queryParams(message *descriptor.Message, field *descriptor.Field, prefix st
 
 		if reg.GetUseJSONNamesForFields() {
 			param.Name = prefix + field.GetJsonName()
-		} else {
+		}
+
+		if !reg.GetUseJSONNamesForFields() || param.Name == "" {
 			param.Name = prefix + field.GetName()
 		}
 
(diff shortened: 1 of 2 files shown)


brocaar:
@johanbrandhorst are there any tests that cover the `GetJsonName` functionality to which I can append this?

johanbrandhorst:
> @johanbrandhorst are there any tests that cover the `GetJsonName` functionality to which I can append this?

I'm not sure, but I think you should be able to copy and modify one of the tests in https://github.com/grpc-ecosystem/grpc-gateway/blob/master/protoc-gen-swagger/genswagger/template_test.go. Let me know if you need more assistance.

brocaar:
I've pushed a test which tests a field with and without `json_name`. I've also added a fallback to `GetName` when `GetJsonName` returns `""`. Without that fallback, https://github.com/grpc-ecosystem/grpc-gateway/pull/879/commits/413f6e236967789cfe6c196b2aa7965b6aa8525e#diff-20461b822a903046234d0fa5073ba2c4R236 would be `""`.

I believe the same must be added to:
https://github.com/grpc-ecosystem/grpc-gateway/blob/master/protoc-gen-swagger/genswagger/template.go#L310 ?

It was not clear to me how to add a testcase for that. Would you mind making that modification?

johanbrandhorst:
> I've pushed a test which tests a field with and without `json_name`. I've also added a fallback to `GetName` when `GetJsonName` returns `""`. Without that fallback, [413f6e2#diff-20461b822a903046234d0fa5073ba2c4R236](https://github.com/grpc-ecosystem/grpc-gateway/commit/413f6e236967789cfe6c196b2aa7965b6aa8525e#diff-20461b822a903046234d0fa5073ba2c4R236) would be `""`.
> 
> I believe the same must be added to:
> https://github.com/grpc-ecosystem/grpc-gateway/blob/master/protoc-gen-swagger/genswagger/template.go#L310 ?
> 
> It was not clear to me how to add a testcase for that. Would you mind making that modification?

I expect the reason that isn't required is because the JsonName is guaranteed by the protoc compiler. See https://github.com/protocolbuffers/protobuf/blob/f2ef7970fe704c4ec604d39fc1d70bec2c59e4a7/src/google/protobuf/descriptor.proto#L207. You can probably remove it from your test case and the code as well if you wish. Otherwise looks good!

brocaar:
Ah, I didn't know that. I've removed the fallback :-)

johanbrandhorst:
Great work, thanks for this!

brocaar:
Thanks for reviewing and your prompt responses!

Issue #877, opened 2019-02-25
Cleaner Swagger-generated schema names (omitting prefixes)


jgiles:
This issue resurrects some of the discussion in https://github.com/grpc-ecosystem/grpc-gateway/issues/525, and seeks a resolution to the problem of "ugly" prefixing without violating the principle of least surprise.

For the first time, we're planning to use grpc-gateway and its generated Swagger definitions to generate a public-facing API rather than an internal API. This means the "cosmetic" concern of prefixes we do not control on the generated Swagger schema names takes on new importance.

I've hacked together a bit of code that does "minimal" package prefixing as needed, but I think the conclusion in https://github.com/grpc-ecosystem/grpc-gateway/issues/525 is correct - more magic here is undesirable.

**Short Term**

In the short term, it would be great fix the simple common case where only messages from one proto package are rendered into the Swagger doc.

The current uniquification code here https://github.com/grpc-ecosystem/grpc-gateway/blob/9a7c952d3c7e6c0239ddb5d1d351e0fdb6c3c8d3/protoc-gen-swagger/genswagger/template.go#L542 is called on a bunch of fully-qualified proto names which do not actually get rendered into Swagger. In a build for my proto service with one endpoint, this is the list of names uniquified:
```
[...]

PR #881, opened 2019-02-28
Add fqn_for_swagger_name option

hypnoce:
Problem : 
Name generation can output unreadable concatenation. It is also dependent on the unicity, which can completely change the documentation by adding one message. It is a big side effect.

To have a more repeatable name generation as well as to ensure unicity and readability, generate swagger name as a dot-separated concatenation for proto fully qualified name.

Can help fix https://github.com/grpc-ecosystem/grpc-gateway/issues/877

### DIFF
--- a/protoc-gen-grpc-gateway/descriptor/registry.go
+++ b/protoc-gen-grpc-gateway/descriptor/registry.go
@@ -61,6 +61,12 @@ type Registry struct {
 	// otherwise the original proto name is used. It's helpful for synchronizing the swagger definition
 	// with grpc-gateway response, if it uses json tags for marshaling.
 	useJSONNamesForFields bool
+
+	// useFQNForSwaggerName if true swagger names will use the full qualified name (FQN) from proto definition,
+	// and generate a dot-separated swagger name concatenating all elements from the proto FQN.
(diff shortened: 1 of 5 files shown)


johanbrandhorst:
Nice work, complete with tests and all :heart_eyes:! I'm just wondering, have you explored what the standard way of namespacing things is in the OpenAPI/swagger world? It might be we might want to do a little research before merging this to see if there's another alternative representation that might be better. In any case, this is a backwards-compatible change so I have no real objections.

hypnoce:
Indeed, always trying to keep things backward compatible. 
Let me research things to see if we can have a more 'swagger' way of namespacing.

hypnoce:
Hi,

after some search, the only thing that is the closest to package/namespace if the [xml namespace/prefix](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md).  Also, the swagger code gen has a specific option for generated package : https://github.com/swagger-api/swagger-codegen/tree/master/modules/swagger-codegen-maven-plugin. I believe their is no good support for namespaces/package in openapi 2 (as well as 3)

Thanks !

johanbrandhorst:
SGTM, thanks for this!

PR #882, opened 2019-03-01
Read only support

hypnoce:
Add read only support.

This helps simplifying asymmetry in request/response using the same proto message definition
For instance, a proto can define some HATEOAS links that are read-only that should not be present in the request body of PUT or POST but present in the response body of GET for instance.

What do you think ?

Thanks


johanbrandhorst:
Looks like you need to rebase on master to start with.

johanbrandhorst:
Note: these are normally simply documented as `// Output Only` in idiomatic protobuf: https://cloud.google.com/apis/design/design_patterns#output_fields. Since swagger supports this field, maybe we could detect this comment and translate the properties to use this too?

hypnoce:
@johanbrandhorst PR updated with support for "Output only." comment on field.

johanbrandhorst:
The generate job is failing; did you run the generate job as described in CONTRIBUTING.md to regenerate the files?

protoc-gen-swagger/genswagger/template.go
```diff
@@ -1223,6 +1224,12 @@ func updateSwaggerDataFromComments(swaggerObject interface{}, comment string, is
 	// Figure out which properties to update.
 	summaryValue := infoObjectValue.FieldByName("Summary")
 	descriptionValue := infoObjectValue.FieldByName("Description")
+	readOnlyValue := infoObjectValue.FieldByName("ReadOnly")
+
+	if readOnlyValue.Kind() == reflect.Bool && readOnlyValue.CanSet() && strings.Contains(comment, "Output only.") {
```
johanbrandhorst: _Nice_ :sunglasses:

protoc-gen-swagger/options/openapiv2.proto
```diff
@@ -170,7 +170,7 @@ message JSONSchema {
   string description = 6;
   string default = 7;
   // field 8 is reserved for 'readOnly', which has an OpenAPI v2-specific meaning and is defined there.
```
johanbrandhorst: Lets remove this comment

hypnoce:
Do you think it's possible to release grpc-gateway at some point with the generated artifacts ?

protoc-gen-swagger/options/openapiv2.proto
hypnoce: done

johanbrandhorst:
We automatically build binaries for all releases. We're planning on making another release soon!

Issue #897, opened 2019-03-06
Error while defining enum comments


fahernandez:
### Steps to reproduce the issue

**Steps**
1.  Define an enum proto message with comments containing a **.**
```
// DeepEnum is one or zero.
enum DeepEnum {
    // FALSE is false.
    FALSE = 0;
   // TRUE is true.
   TRUE = 1;
}

// DeepEnum comment.<--This dot
DeepEnum ok = 3 [(grpc.gateway.protoc_gen_swagger.options.openapiv2_field) = {description: "DeepEnum description."}];
```
2. Generate the swagger documentation.

**Expected behavior**
1. The swagger code is generated without errors

**Actual behavior**
1. An error code is triggered.

**Hipotesis**
This error is triggered by an error on protoc-gen-swagger/genswagger/template.go:1268. The solution is to refactor the **if** behavior.

If you remove the dot from the comment it will generate the swagger code without errors.

johanbrandhorst:
What does the generated swagger look like? Would you like to contribute a fix for this?

fahernandez:
The swagger file is not generated due the error code. Sure I will fix it.

johanbrandhorst:
Fixed by #898

PR #898, opened 2019-03-06
897 fixing problem while generating swagger documentation for enum messages 

fahernandez:


fahernandez:
This pull request will fix issue https://github.com/grpc-ecosystem/grpc-gateway/issues/897

fahernandez:
It's a very strange behavior but if you look line protoc-gen-swagger/genswagger/template.go:1246 you will see a condition to avoid executing a particular flow if the comment has a dot at the end. So when the support for enum and objects annotation was [added](https://github.com/grpc-ecosystem/grpc-gateway/pull/865) I forget to take into consideration this case causing in this particular case to fallback to condition protoc-gen-swagger/genswagger/template.go:1275. So basically I'm only changing condition protoc-gen-swagger/genswagger/template.go:1268 to the default fallback that existe before the changes but keeping the code needed to support the comments override behavior.

I don't have a clear idea how to make a test to test particular condition thats why I added an example case on examples/proto/examplepb/a_bit_of_everything.proto:166 to be sure that the problem was fixed but keeping past changes.

> Firstly, I don't understand what this fixes. Why does the applied change fix a comment that ends in a dot?
> 
> Could we add some tests for this case as well?
> 
> Thanks

johanbrandhorst:
Thanks for your contribution :)

Issue #756, opened 2018-09-14
Swagger get query param documentation shows repeated fields incorrectly


veqryn:
### Bug reports:
The swagger documentation generated for query parameters shows repeated fields incorrectly.

### Steps you follow to reproduce the error:

Proto file example:
```
service Foo {
	rpc Bar (Params) returns (Series) {
		option (google.api.http) = {
			get: "/v2/foo/bar"
		};
	}
}
message Params {
	repeated string tags = 1;
}
```

This generates swagger documentation that looks correct at first glance, but if you hit "Try It Now" and then fill in multiple "tags", then select execute, it gives you a curl command like this.
(This was done with me filling in the following tags: `foo=bar` and `blah=blarg`.)
`curl -X GET "http://editor.swagger.io/v2/foo/bar?tags=foo%3Dbar,blah%3Dblarg" -H "accept: application/json"`

That is, it asks the user to comma separate the individual tags:
`tags=foo%3Dbar,blah%3Dblarg`

When my function runs, it receives a single element in the array equal to: `["foo=bar,blah=blarg"]`

### What did you expect to happen instead:

The documentation should actually generate an example like this:
`curl -X GET "http://editor.swagger.io/v2/foo/bar?tags=foo%3Dbar&tags=blah%3Dblarg" -H "accept: application/json"`

This correctly gives my function two elements: `["foo=bar", "blah=blarg"]`

johanbrandhorst:
Thanks for raising this issue! I think this is actually just an incompatibility between Swagger and the Gateways expectations. The gateway generates a correct swagger file and there's no way that I know to tell swagger to put these parameters in separate query parameters.

If you know of a different solution to this, please let me know, but I think this might just be a matter of documenting the problem. I think the gateway is doing the sensible thing here and Swagger is generating a nonsensical query.

johanbrandhorst:
Having looked into it, I think we might be able to fix this by emitting a `collectionFormat` value set to `multi`: https://swagger.io/docs/specification/2-0/describing-parameters/. Would you like to contribute this?

veqryn:
I guess I could try at least.
Is this where my change would go?
https://github.com/grpc-ecosystem/grpc-gateway/blob/ab0345bb328757bfef2f3d7d4e642e182eb985b9/protoc-gen-swagger/genswagger/template.go#L92

johanbrandhorst:
[...]

PR #902, opened 2019-03-10
Use collectionFormat multi for query params of repeated fields

bmperrea:
Solves #756 . Also formats `protoc-gen-swagger/genswagger/template_test.go` according to `go fmt`.

### DIFF
--- a/examples/clients/abe/a_bit_of_everything_service_api.go
+++ b/examples/clients/abe/a_bit_of_everything_service_api.go
@@ -595,20 +595,20 @@ func (a ABitOfEverythingServiceApi) GetQuery(uuid string, floatValue float32, si
 	localVarQueryParams.Add("sfixed64_value", a.Configuration.APIClient.ParameterToString(sfixed64Value, ""))
 	localVarQueryParams.Add("sint32_value", a.Configuration.APIClient.ParameterToString(sint32Value, ""))
 	localVarQueryParams.Add("sint64_value", a.Configuration.APIClient.ParameterToString(sint64Value, ""))
-	var repeatedStringValueCollectionFormat = "csv"
+	var repeatedStringValueCollectionFormat = "multi"
 	localVarQueryParams.Add("repeated_string_value", a.Configuration.APIClient.ParameterToString(repeatedStringValue, repeatedStringValueCollectionFormat))
 
 	localVarQueryParams.Add("oneof_string", a.Configuration.APIClient.ParameterToString(oneofString, ""))
 	localVarQueryParams.Add("nonConventionalNameValue", a.Configuration.APIClient.ParameterToString(nonConventionalNameValue, ""))
 	localVarQueryParams.Add("timestamp_value", a.Configuration.APIClient.ParameterToString(timestampValue, ""))
-	var repeatedEnumValueCollectionFormat = "csv"
+	var repeatedEnumValueCollectionFormat = "multi"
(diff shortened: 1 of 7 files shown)


achew22:
Looks like really great work! FYI, I think the way to fix the generate CI failure is to run

```
docker run -v $(pwd):/go/src/github.com/grpc-ecosystem/grpc-gateway --rm jfbrandhorst/grpc-gateway-build-env:1.12 \
    /bin/bash -c 'cd /go/src/github.com/grpc-ecosystem/grpc-gateway && \
        make realclean && \
        make examples SWAGGER_CODEGEN="${SWAGGER_CODEGEN}"'
```

which will regenerate the examples. I'm really looking forward to this!

bmperrea:
Thanks @achew22 ! A docker run to generate these files will be great - I was having trouble figuring out how to get the right version of `swagger-codegen` (2.2.2).

When I run the command you posted I get
```
can't load package: package github.com/golang/protobuf/protoc-gen-go: cannot find package "github.com/golang/protobuf/protoc-gen-go" in any of:
	/usr/local/go/src/github.com/golang/protobuf/protoc-gen-go (from $GOROOT)
	/go/src/github.com/golang/protobuf/protoc-gen-go (from $GOPATH)
```
I have that installed on my local, but I guess the docker run is missing a dependency.

johanbrandhorst:
That's very strange, this is the command run by CI when generating the diff that's failing. Are you sure you ran the command correctly?

johanbrandhorst:
Ah, there is a problem with the mount path in Andrews link. Try this:
```
docker run -v $(pwd):/src/grpc-gateway --rm jfbrandhorst/grpc-gateway-build-env:1.12 \ /bin/bash -c 'cd /src/grpc-gateway && \         make realclean && \         make examples SWAGGER_CODEGEN="${SWAGGER_CODEGEN}"'
```

johanbrandhorst:
FYI the source of truth for this command is CONTRIBUTING.md

johanbrandhorst:
(Andrews command is how we used to run it before we switched to go modules).

johanbrandhorst:
Note to self: tests are a bit flaky, we should consider increasing the wait for the server to start up.

johanbrandhorst:
Thanks for your contribution!

johanbrandhorst:
We should make a release with this new functionality

mmarod:
I believe that this may have inadvertently broken spec...

```
  "definitions": {
    "Foo": {
      "type": "object",
      "properties": {
        "foo": {
          "type": "array",
          "items": {
            "$ref": "#/definitions/Bar"
          },
          "description": "foo",
          "collectionFormat": "multi"
        }
...
```

This spits out the following error when I run `swagger generate server`

```
- definitions.Foo.properties.foo.collectionFormat in body is a forbidden property
```

johanbrandhorst:
Thanks for your report @mmarod, could you raise an issue please? That is unfortunate.

johanbrandhorst:
My immediate thought is that is happens when we have repeated fields in the body of a request instead of in the query parameters of the URL. Might just need a little extra logic.

johanbrandhorst:
I raised https://github.com/grpc-ecosystem/grpc-gateway/issues/906

c2nc:
I have same trouble when do swagger spec validation...
Fix please guys. Thanks

Look here
https://github.com/OAI/OpenAPI-Specification/issues/419#issuecomment-128788426

johanbrandhorst:
Since that makes two of you that have spoken up I'm going to revert this until we can fix #906 at the same time.

johanbrandhorst:
I've reverted the change in master and will make another release.

Issue #906, opened 2019-03-13
Invalid swagger generated for bodies with repeated fields


johanbrandhorst:
As reported in https://github.com/grpc-ecosystem/grpc-gateway/pull/902#issuecomment-472257493, it appears #902 means we're generating invalid swagger definitions for some methods which take repeated fields as parameters. I think this is only affecting fields which have a repeated field in the _body_ of the request as opposed to in the URL query parameters, so we may need some extra logic in the swagger generator to only set the collection format when we know the parameters are going into the URL query parameters.

johanbrandhorst:
Closing in favour of #756

PR #907, opened 2019-03-13
Revert "Use collectionFormat multi for query params of repeated fields (#902)"

johanbrandhorst:
This reverts commit 2b6cab672e62c65a1bbe3a7c835c7273607e795a. It was found
to have introduced a swagger spec violation, see #906.


johanbrandhorst:
I think we'll close #906 and reopen #756 instead.

mmarod:
Thanks @johanbrandhorst for the quick turnaround!

Issue #414, opened 2017-06-16
can't specify an empty path?


jmuk:
Input proto:
```protobuf
syntax = "proto3";

package foo;

import "google/api/annotations.proto";

service EchoService {
  rpc echo(EchoRequest) returns (EchoResponse) {
    option (google.api.http) = {
      get: "/{msg}"
      additional_bindings: {
        get: "/"
      };
    };
  }
}

message EchoRequest {
  string msg = 1;
}

message EchoResponse {
  string msg = 2;
}
```

This is for sending a message optionally, but it seems grpc-gateway doesn't accept the path like this? protoc fails with the following error:
```
--grpc-gateway_out: segment neither wildcards, literal or variable: expected "{" but got "\x00": /
```

An empty string (`get: ""`) goes into the panic:
```
panic: reflect: indirection through nil pointer to embedded struct [recovered]
        panic: reflect: indirection through nil pointer to embedded struct

goroutine 1 [running]:
text/template.errRecover(0xc4200e57c8)
        /usr/lib/google-golang/src/text/template/exec.go:146 +0x62
panic(0x5bc500, 0xc4201dde30)
        /usr/lib/google-golang/src/runtime/panic.go:489 +0x2e1
reflect.Value.FieldByIndex(0x5e1fa0, 0x0, 0x19, 0xc4201dde20, 0x2, 0x2, 0x0, 0x702a80, 0x5effa0)
        /usr/lib/google-golang/src/reflect/value.go:798 +0x20c
text/template.(*state).evalField(0xc4200e5748, 0x5e1fa0, 0x0, 0x19, 0xc4200ea081, 0x6, 0x6ffa80, 0xc4200dc390, 0x0, 0x0, ...)
        /usr/lib/google-golang/src/text/template/exec.go:569 +0xaef
text/template.(*state).evalFieldChain(0xc4200e5748, 0x5e1fa0, 0x0, 0x19, 0x5e1fa0, 0x0, 0x19, 0x6ffa80, 0xc4200dc390, 0xc4200ea0a0, ...)
        /usr/lib/google-golang/src/text/template/exec.go:523 +0x128
text/template.(*state).evalFieldNode(0xc4200e5748, 0x5e1fa0, 0x0, 0x19, 0xc4200dc390, 0xc4200e4ff8, 0x1, 0x1, 0x0, 0x0, ...)
        /usr/lib/google-golang/src/text/template/exec.go:490 +0x118
[...]

PR #913, opened 2019-03-26
Fixed empty path bug

brycematheson1234:
Fixes this issue: https://github.com/grpc-ecosystem/grpc-gateway/issues/414
Where "/" is not allowed.

Also added a test to verify this behaviour.

There was a previous test that prevented this behaviour. Despite the comment it was never really testing double slash, it was testing single slash.

### DIFF
--- a/protoc-gen-grpc-gateway/httprule/parse_test.go
+++ b/protoc-gen-grpc-gateway/httprule/parse_test.go
@@ -113,6 +113,12 @@ func TestParseSegments(t *testing.T) {
 				literal("v1"),
 			},
 		},
+        {
+            tokens: []string{"/", eof},
+            want: []segment{
+                wildcard{},
+            },
+        },
 		{
 			tokens: []string{"-._~!$&'()*+,;=:@", eof},
 			want: []segment{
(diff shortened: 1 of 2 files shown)


protoc-gen-grpc-gateway/httprule/parse_test.go
```diff
@@ -247,7 +253,7 @@ func TestParseSegmentsWithErrors(t *testing.T) {
 	}{
 		{
 			// double slash
-			tokens: []string{"/", eof},
+			tokens: []string{"//", eof},
```
johanbrandhorst: Is it worth keeping both of these tests instead of replacing the old one?

johanbrandhorst:
The bazel build failure looks transient to me, I'll try to rerun it.

johanbrandhorst:
> The bazel build failure looks transient to me, I'll try to rerun it.

Rebuild worked a-OK

protoc-gen-grpc-gateway/httprule/parse_test.go
brycematheson1234: IIUC the previous test verified that "/" threw an error, and as this PR now allows "/", keeping the previous test would cause it to fail as "/" will no longer throw an error.

johanbrandhorst:
Thanks for your contribution!

PR #919, opened 2019-04-05
Added comments to base_path to explain behavior

### LINKED ISSUE #918: Documentation: 'base_path' Swagger attribute confuses users

1. I start with a working REST API and with properly generating Swagger definitions
2. I added a 'base_path' attribute to the 'openapiv2_swagger' annotation in my proto file
3. I removed the equivalent base_path prefix from each 'google.api.http' annotation (base_path = "/api/v1")
4. I regenerated my Go code and restarted my API
5. All my REST APIs now return 404 errors

I expected the API to remain working

The generated code was previously generating this:
```go
pattern_Provider_EncryptClientSecret_0 = runtime.MustPattern(runtime.NewPattern(1, []int{2, 0, 2, 1, 2, 2, 2, 3}, []string{"api", "v1", "provider", "client_secret_encrypter"}, ""))
```
It is now generating this:
```go
pattern_Provider_EncryptClientSecret_0 = runtime.MustPattern(runtime.NewPattern(1, []int{2, 0, 2, 1}, []string{"provider", "client_secret_encrypter"}, ""))
```
It seems like the generator is not aware of the 'base_path' attribute and expects the 'google.api.http' to hold the entire path. However, adding this path back on (while retaining 'base_path' attribute) has the impact of: 

1) Adding that path to all endpoints displayed in the Swagger UI which makes it very 'noisy' and hard to read 
2) Breaking the Swagger UI "try it out" feature which doubles up the 'base_path' be prefixing base_path to all operations

nu11ptr:
Added comments to base_path to explain how it doesn't modify generated paths and has to be handled manually

Closes #918


johanbrandhorst:
Ah, looks like you still need to regenerate the files with this new comment. There's a one liner in CONTRIBUTING.md that should work.

johanbrandhorst:
Are you still interested in this PR? The CI build failure can be fixed by running two simple docker commands and committing the changes:

```
docker run -v $(pwd):/src/grpc-gateway --rm jfbrandhorst/grpc-gateway-build-env:1.12 \
    /bin/bash -c 'cd /src/grpc-gateway && \
        make realclean && \
        make examples SWAGGER_CODEGEN="${SWAGGER_CODEGEN}"'
    l.gcr.io/google/bazel -c 'bazel run :gazelle; bazel run :buildifier'
```

go.mod
```diff
@@ -11,6 +11,6 @@ require (
 	google.golang.org/genproto v0.0.0-20180817151627-c66870c02cf8
 	google.golang.org/grpc v1.19.0
 	gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127 // indirect
-	gopkg.in/resty.v1 v1.12.0
+	gopkg.in/resty.v1 v1.12.0 // indirect
```
johanbrandhorst: If you can just revert this we can merge this.

nu11ptr: It was generated by the first docker cmd - wasn't sure whether to commit it or not. I don't know anything about Go modules so wasn't sure why it did that or if that was a desired change or not. I'll revert if that is the desired state.

johanbrandhorst:
Thanks for your contribution!

PR #922, opened 2019-04-14
Delete redundant load statement

pcj:


achew22:
@pcj, thanks so much for your contribution!

PR #924, opened 2019-04-18
Use proto gen swagger with protos from external repository

elenadeneva92:
Currently the protoc_gen_swagger doesn't work with protos from external repositories.
The the swagger.json file is not stored at the expected place.
Example:
protoc_gen_swagger(
        name = "action_swagger",
        proto =  "@remote//proto:action_proto"
)
fails because the action.swagger.json file is written to proto/action.swagger.json and
is expected to be at the same place as the proto external/remote/proto/action.swagger.json. Bazel can't find the generated rule and fails.


drigz:
LGTM. @achew22, any comments?

johanbrandhorst:
Looks like the bazel tests failed for unrelated reasons, but the bazel formatter also failed. Could you try running the formatter please?

elenadeneva92:
@johanbrandhorst Sure. Done :)

johanbrandhorst:
Sorry, this totally fell between the chairs. Merging now.

PR #930, opened 2019-05-09
ability to customize stream errors

### LINKED ISSUE #584: Errors in response streams do not go through the registered error handler
I am using `runtime.WithProtoErrorHandler` to provide custom error handling in my service. In particular, my service discards the error message (and provides a default based on the gRPC code) unless the error includes details that specify a "safe" message (e.g. no stack trace or other internal details leaked). I also have a slightly modified mapping of gRPC codes to HTTP codes.

But if I have a bidi- or server-streaming endpoint, the error will get converted into a `StreamError` message in a way that is not pluggable/configurable. That means the `err.Error()` or `stat.Message()` is rendered in the response chunk with no ability to customize it (such as stripping away any inappropriate details that are present in the message).

I understand that the existing `ProtoErrorHandlerFunc` type is not necessarily appropriate for use with a streaming response, since it assumes it can set a response code and response headers. So I think a new `ServerMuxOption` is necessary for this.

Proposal (two-pronged):

1. Use `ProtoErrorHandlerFunc` when possible -- e.g. when no response messages have actually been sent back.
2. Add new `ServerMuxOption` for rendering errors, maybe a signature like so:
`func(context.Context, error) *StreamError`

That second part could even be extended to customizing the envelope that is used for streaming response messages. Instead of a function, it could be an interface:

```
type StreamChunkHandler interface {
    HandleResponse(context.Context, proto.Message) interface{}
    HandleError(context.Context, error) interface{}
}
```
This allows for using a custom envelope as well as using a custom error representation, too. The returned item from these methods would then be passed to the configured `Marshaler` for rendering to the HTTP stream.

jhump:
Fixes #584

### DIFF
--- a/runtime/handler.go
+++ b/runtime/handler.go
@@ -169,14 +169,11 @@ func handleForwardResponseOptions(ctx context.Context, w http.ResponseWriter, re
 }
 
 func handleForwardResponseStreamError(ctx context.Context, wroteHeader bool, marshaler Marshaler, w http.ResponseWriter, req *http.Request, mux *ServeMux, err error) {
+	serr := streamError(ctx, mux.streamErrorHandler, err)
 	if !wroteHeader {
-		// if we haven't already written header, use normal error handler to render
-		HTTPError(ctx, mux, marshaler, w, req, err)
-		return
+		w.WriteHeader(int(serr.HttpCode))
 	}
-
-	// otherwise, render the error as a stream chunk in the response
-	buf, merr := marshaler.Marshal(streamErrorChunk(ctx, mux.streamErrorHandler, err))
+	buf, merr := marshaler.Marshal(errorChunk(serr))
 	if merr != nil {
 		grpclog.Infof("Failed to marshal an error: %v", merr)
 		return
@@ -191,18 +188,22 @@ func handleForwardResponseStreamError(ctx context.Context, wroteHeader bool, mar
 // given errHandler is used to render an error chunk if result is nil.
 func streamChunk(ctx context.Context, result proto.Message, errHandler StreamErrorHandlerFunc) map[string]proto.Message {
 	if result == nil {
-		return streamErrorChunk(ctx, errHandler, errEmptyResponse)
+		return errorChunk(streamError(ctx, errHandler, errEmptyResponse))
 	}
 	return map[string]proto.Message{"result": result}
 }
 
-// streamErrorChunk returns the final chunk in a response stream that represents
-// the given err.
-func streamErrorChunk(ctx context.Context, errHandler StreamErrorHandlerFunc, err error) map[string]proto.Message {
+// streamError returns the payload for the final message in a response stream
+// that represents the given err.
+func streamError(ctx context.Context, errHandler StreamErrorHandlerFunc, err error) *StreamError {
 	serr := errHandler(ctx, err)
-	if serr == nil {
-		// TODO: log about misbehaving stream error handler?
-		serr = DefaultHTTPStreamErrorHandler(ctx, err)
+	if serr != nil {
+		return serr
 	}
-	return map[string]proto.Message{"error": (*internal.StreamError)(serr)}
+	// TODO: log about misbehaving stream error handler?
+	return DefaultHTTPStreamErrorHandler(ctx, err)
+}
+
+func errorChunk(err *StreamError) map[string]proto.Message {
+	return map[string]proto.Message{"error": (*internal.StreamError)(err)}
 }
--- a/runtime/handler.go
+++ b/runtime/handler.go
@@ -1,6 +1,7 @@
 package runtime
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -8,12 +9,11 @@ import (
 
 	"context"
 	"github.com/golang/protobuf/proto"
-	"github.com/golang/protobuf/ptypes/any"
 	"github.com/grpc-ecosystem/grpc-gateway/internal"
-	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/grpclog"
-	"google.golang.org/grpc/status"
 )
+
+var errEmptyResponse = errors.New("empty response")
 
 // ForwardResponseStream forwards the stream from gRPC server to REST client.
 func ForwardResponseStream(ctx context.Context, mux *ServeMux, marshaler Marshaler, w http.ResponseWriter, req *http.Request, recv func() (proto.Message, error), opts ...func(context.Context, http.ResponseWriter, proto.Message) error) {
@@ -61,7 +61,7 @@ func ForwardResponseStream(ctx context.Context, mux *ServeMux, marshaler Marshal
 			return
 		}
 
-		buf, err := marshaler.Marshal(streamChunk(resp, nil))
+		buf, err := marshaler.Marshal(streamChunk(ctx, resp, mux.streamErrorHandler))
 		if err != nil {
 			grpclog.Infof("Failed to marshal response chunk: %v", err)
 			handleForwardResponseStreamError(ctx, wroteHeader, marshaler, w, req, mux, err)
@@ -124,7 +124,7 @@ func ForwardResponseMessage(ctx context.Context, mux *ServeMux, marshaler Marsha
 
 	contentType := marshaler.ContentType()
 	// Check marshaler on run time in order to keep backwards compatability
-	// An interface param needs to be added to the ContentType() function on 
+	// An interface param needs to be added to the ContentType() function on
 	// the Marshal interface to be able to remove this check
 	if httpBodyMarshaler, ok := marshaler.(*HTTPBodyMarshaler); ok {
 		contentType = httpBodyMarshaler.ContentTypeFromMessage(resp)
@@ -176,19 +176,7 @@ func handleForwardResponseStreamError(ctx context.Context, wroteHeader bool, mar
 	}
 
 	// otherwise, render the error as a stream chunk in the response
-	var chunk map[string]proto.Message
-	if mux.streamErrorHandler != nil {
-		serr := mux.streamErrorHandler(ctx, err)
-		if serr == nil {
-			// TODO: log error about misbehaving error handler?
-			chunk = streamChunk(nil, err)
-		} else {
-			chunk = map[string]proto.Message{"error": (*internal.StreamError)(serr)}
-		}
-	} else {
-		chunk = streamChunk(nil, err)
-	}
-	buf, merr := marshaler.Marshal(chunk)
+	buf, merr := marshaler.Marshal(streamErrorChunk(ctx, mux.streamErrorHandler, err))
 	if merr != nil {
 		grpclog.Infof("Failed to marshal an error: %v", merr)
 		return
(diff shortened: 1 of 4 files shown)


runtime/mux.go
```diff
@@ -110,6 +111,20 @@ func WithDisablePathLengthFallback() ServeMuxOption {
 	}
 }
 
+// WithStreamErrorHandler returns a ServeMuxOption that will use the given custom stream
+// error handler, which allows for customizing the error trailer for server-streaming
+// calls.
+//
+// For stream errors that occur before any response has been written, the mux's
+// ProtoErrorHandler will be invoked. However, once data has been written, the errors must
+// be handled differently: they must included in the response body. The response body's
```
johanbrandhorst:
```suggestion
// be handled differently: they must be included in the response body. The response body's
```

runtime/proto_errors.go
```diff
@@ -8,8 +8,17 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/grpclog"
 	"google.golang.org/grpc/status"
+	"github.com/grpc-ecosystem/grpc-gateway/runtime/internal"
```
johanbrandhorst: Could you run `goimports` on this file please? I think this import should be above the grpc ones.

runtime/handler.go
```diff
@@ -168,18 +168,30 @@ func handleForwardResponseOptions(ctx context.Context, w http.ResponseWriter, re
...
+
+	// otherwise, render the error as a stream chunk in the response
+	var chunk map[string]proto.Message
+	if mux.streamErrorHandler != nil {
+		serr := mux.streamErrorHandler(ctx, err)
+		if serr == nil {
+			// TODO: log error about misbehaving error handler?
+			chunk = streamChunk(nil, err)
+		} else {
+			chunk = map[string]proto.Message{"error": (*internal.StreamError)(serr)}
```
johanbrandhorst: Not sure if this was discussed in #584, but this still doesn't give the user the power to fully control the envelope, right? Is that intentional?

runtime/handler.go
```diff
@@ -168,18 +168,30 @@ func handleForwardResponseOptions(ctx context.Context, w http.ResponseWriter, re
...
 		}
-		w.WriteHeader(HTTPStatusFromCode(s.Code()))
+	} else {
+		chunk = streamChunk(nil, err)
+	}
```
johanbrandhorst: Instead of this if-else, we could set the `mux.streamErrorHandler` to `streamChunk` by default (before iterating the options).

runtime/handler.go
jhump: I added a comment earlier to that issue about why I felt that was ultimately a bad idea: a custom envelope in the runtime would make it impossible for a protoc plugin to generate a correct swagger/OpenAPI spec for the service.

johanbrandhorst: Fair enough

jhump:
@johanbrandhorst, the build errors aren't making a lot of sense to me. I'm not a bazel user, but I understand the concepts (I've used blaze and pants). I tried to add a dependency from the runtime package to runtime/internal (surprised it was not already present), but that is still failing.

Any advice on getting the build green? I'm also really confused about the other errors, which appear to be related to Go modules. The confusing thing is that a few lines up from the error, it states that it resolved "github.com/grpc-ecosystem/grpc-gateway/runtime", but somehow can't resolve the "internal" sub-package therein :(

I can keep hacking, but if you can spot any simple issues in here, that would certainly speed things up.

johanbrandhorst:
> @johanbrandhorst, the build errors aren't making a lot of sense to me. I'm not a bazel user, but I understand the concepts (I've used blaze and pants). I tried to add a dependency from the runtime package to runtime/internal (surprised it was not already present), but that is still failing.
> 
> Any advice on getting the build green? I'm also really confused about the other errors, which appear to be related to Go modules. The confusing thing is that a few lines up from the error, it states that it resolved "github.com/grpc-ecosystem/grpc-gateway/runtime", but somehow can't resolve the "internal" sub-package therein :(
> 
> I can keep hacking, but if you can spot any simple issues in here, that would certainly speed things up.

Something about this import certainly seems to have got Go modules confused - I also have no idea what the hell is wrong here. I'll do some more investigating.

runtime/BUILD.bazel
```diff
@@ -30,6 +30,7 @@ go_library(
         "@com_github_golang_protobuf//jsonpb:go_default_library_gen",
         "@com_github_golang_protobuf//proto:go_default_library",
         "@com_github_golang_protobuf//protoc-gen-go/generator:go_default_library_gen",
+        "@com_github_grpc_ecosystem_grpc_gateway//runtime/internal:go_default_library",
```
johanbrandhorst: I don't think this should be necessary, I can't imagine we don't already import the internal package somewhere in here and it's worked in the past.

johanbrandhorst:
Wait, `runtime/internal` isn't a package. You must mean `github.com/grpc-ecosystem/grpc-gateway/internal`? The `internal` package moved. @jhump.

jhump:
> Wait, runtime/internal isn't a package. You must mean github.com/grpc-ecosystem/grpc-gateway/internal? The internal package moved. @jhump.

Doh! Thanks for the tip. (You can probably tell that this patch was rebased from an older version of the repo. We've had it in a fork that is apparently pretty old: 463c5eda4ce58dd590b94bd2ac931db6790a1d2c, from March 2018.)

johanbrandhorst:
> > Wait, runtime/internal isn't a package. You must mean github.com/grpc-ecosystem/grpc-gateway/internal? The internal package moved. @jhump.
> 
> Doh! Thanks for the tip. (You can probably tell that this patch was rebased from an older version of the repo. We've had it in a fork that is apparently pretty old: [463c5ed](https://github.com/grpc-ecosystem/grpc-gateway/commit/463c5eda4ce58dd590b94bd2ac931db6790a1d2c), from March 2018.)

I celebrate the implication that you're trying to go back to the source :). I hope we can make you feel at home!

runtime/handler.go
johanbrandhorst: Still relevant

runtime/proto_errors.go
```diff
@@ -1,15 +1,24 @@
 package runtime
 
 import (
+	"context"
 	"io"
 	"net/http"
 
-	"context"
+        "github.com/grpc-ecosystem/grpc-gateway/internal"
```
johanbrandhorst: :eyes: spaces for indentation in a go file? :scream:

jhump: Not sure what happened there. Ran `gofmt` on it.

runtime/handler.go
jhump: Just pushed changes to address this. WDYT?

runtime/proto_errors.go
```diff
@@ -1,15 +1,27 @@
...
 
+// StreamErrorHandlerFunc accepts an error as a gRPC error generated via status package and translates it into a
+// a proto struct used to represent error at the end of a stream.
+type StreamErrorHandlerFunc func(context.Context, error) *StreamError
+
+// StreamError is the payload for the final message in a server stream in the event that the server returns an
+// error after a response message has already been sent.
+type StreamError internal.StreamError
+
+var _ StreamErrorHandlerFunc = DefaultHTTPStreamErrorHandler
```
johanbrandhorst: This shouldn't be necessary, line 134 in mux.go guarantees this.

runtime/handler.go
```diff
@@ -168,48 +168,41 @@ func handleForwardResponseOptions(ctx context.Context, w http.ResponseWriter, re
 	return nil
 }
 
-func handleForwardResponseStreamError(wroteHeader bool, marshaler Marshaler, w http.ResponseWriter, err error) {
-	buf, merr := marshaler.Marshal(streamChunk(nil, err))
+func handleForwardResponseStreamError(ctx context.Context, wroteHeader bool, marshaler Marshaler, w http.ResponseWriter, req *http.Request, mux *ServeMux, err error) {
+	if !wroteHeader {
+		// if we haven't already written header, use normal error handler to render
+		HTTPError(ctx, mux, marshaler, w, req, err)
+		return
```
johanbrandhorst: Is this backwards compatibility breaking? I can see we used to do just `w.WriteHeader` here.

runtime/handler.go
johanbrandhorst: Much better :). Left a few more comments.

runtime/handler.go
jhump: I guess technically it is (though the behavior in this case isn't really specified, at least not that I can find).

Previously, it would emit a JSON response body like so:
```json
{
  "error": {
    "grpc_code": 13,
    "http_code": 500,
    "message": "some error occurred",
    "http_status": "Internal Server Error",
    "details": [],
  }
}
```

Now, it will emit a response body like this instead:

```json
{
  "code": 13,
  "message": "some error occurred",
  "details": [],
}
```

The actual HTTP status code sent back would be the same (in the example: "500 Internal Server Error").

IMO, this makes it cohesive and consistent: if the client gets a non-200 status code, the response format is the same (regardless of whether the method was a streaming or unary method). So it feels more like a fix and will also simplify the specification/documentation of error handling.

But if that is too great of a change, how about the following alternative logic:

1. Call `streamErrorChunk` to let custom handler produce custom stream error message.
2. Use the resulting error's `HttpCode` to set the response code (if `!wroteHeader`).

johanbrandhorst: The swagger generator defines the error format for streams including the envelope, so I would prefer the consistency, even though I understand your point about the non-200 error code response.

jhump: > The swagger generator defines the error format for streams including the envelope

Oh, it does? I guess that must be another fix that isn't in our old fork. (When I run the swagger protoc plugin in our repo, it emits the wrong definition.)

Okay, in that case, I'll tweak this behavior.

jhump: > The swagger generator defines the error format for streams including the envelope

Out of curiosity, is this spec meant for interpreting responses for _successful_ calls only? Does swagger even allow specifying different response types based on error-vs-ok conditions? Because currently, unary RPCs may not be correctly encoded in swagger for non-200 response codes. (Maybe that's only in my old fork of the repo?)

johanbrandhorst: It's an optional field in the response, so the swagger definition shows the "result" key and the "error" for all responses.

jhump:
@johanbrandhorst: I think this is ready for another pass, whenever you get a chance.

runtime/handler.go
```diff
@@ -168,48 +168,42 @@ func handleForwardResponseOptions(ctx context.Context, w http.ResponseWriter, re
...
+	if serr != nil {
+		return serr
+	}
+	// TODO: log about misbehaving stream error handler?
+	return DefaultHTTPStreamErrorHandler(ctx, err)
+}
+
+func errorChunk(err *StreamError) map[string]proto.Message {
+	return map[string]proto.Message{"error": (*internal.StreamError)(err)}
+}
```
johanbrandhorst: This is an opinion so feel free to push back; I think it would be nicer to merge this function into streamError. I might be wrong but it seems to me the value from streamError is only used as a parameter to errorChunk and this separation is only making things unnecessarily complicated. Again, please feel free to push back.

jhump: I originally had it that way. But above, I actually need the `*StreamError` to set the HTTP status code when no header has been written. So I either needed to break it up into two functions or have the function above try to dig it out of the chunk map and type-convert it (which seemed like the much grosser option).

jhump:
> How do you feel about my suggestion to add some documentation about this new feature too?

I added a section to the "Customizing your Gateway" doc page. PTAL.

PR #931, opened 2019-05-09
newline between JSON messages

jhump:
This updates `JSONPb` so that its encoder function delimits output messages with a newline, just like `Encoder` in `"encoding/json"` does.

### DIFF
--- a/runtime/marshal_jsonpb.go
+++ b/runtime/marshal_jsonpb.go
@@ -151,7 +151,15 @@ func (d DecoderWrapper) Decode(v interface{}) error {
 
 // NewEncoder returns an Encoder which writes JSON stream into "w".
 func (j *JSONPb) NewEncoder(w io.Writer) Encoder {
-	return EncoderFunc(func(v interface{}) error { return j.marshalTo(w, v) })
+	return EncoderFunc(func(v interface{}) error {
+		if err := j.marshalTo(w, v); err != nil {
+			return err
+		}
+		// mimic json.Encoder by adding a newline (makes output
+		// easier to read when it contains multiple encoded items)
+		_, err := w.Write([]byte("\n"))
+		return err
+	})
 }
 
 func unmarshalJSONPb(data []byte, v interface{}) error {
--- a/runtime/marshal_jsonpb.go
+++ b/runtime/marshal_jsonpb.go
@@ -157,7 +157,7 @@ func (j *JSONPb) NewEncoder(w io.Writer) Encoder {
 		}
 		// mimic json.Encoder by adding a newline (makes output
 		// easier to read when it contains multiple encoded items)
-		_, err := w.Write([]byte("\n"))
+		_, err := w.Write(j.Delimiter())
 		return err
 	})
 }
--- a/runtime/marshal_jsonpb_test.go
+++ b/runtime/marshal_jsonpb_test.go
@@ -278,6 +278,9 @@ func TestJSONPbEncoder(t *testing.T) {
 	}{
 		{
 			verifier: func(json string) {
+				// remove trailing delimiter before verifying
+				json = strings.TrimSuffix(json, "\n")
+
 				if strings.ContainsAny(json, " \t\r\n") {
 					t.Errorf("strings.ContainsAny(%q, %q) = true; want false", json, " \t\r\n")
 				}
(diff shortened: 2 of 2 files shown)


runtime/marshal_jsonpb.go
```diff
@@ -151,7 +151,15 @@ func (d DecoderWrapper) Decode(v interface{}) error {
...
-	return EncoderFunc(func(v interface{}) error { return j.marshalTo(w, v) })
+	return EncoderFunc(func(v interface{}) error {
+		if err := j.marshalTo(w, v); err != nil {
+			return err
+		}
+		// mimic json.Encoder by adding a newline (makes output
+		// easier to read when it contains multiple encoded items)
+		_, err := w.Write([]byte("\n"))
+		return err
+	})
```
johanbrandhorst: Could we use `j.Delimiter` instead?

jhump: Let me look. I think when I originally prepared this patch (in a fork of the repo that we're using at FullStory), it may not have had `Delimiter`. (So maybe this PR is no longer needed; I'll dig around and either update it or close it.)

jhump: So, from looking a little more closely, an optional `Delimiter()` method can be added to control the output of server-streaming methods.

However, the `JSONPb` type itself has no field named `Delimiter`. And, IMO, it does not need to be configurable but rather should just mirror `JSONBuiltin` (whose `Encoder` already adds newlines in between messages; and it also implements `Delimiter()` to return a newline).

So I think this part of the patch is still applicable. But that I also should add a `Delimiter()` method to this type, to mirror `JSONBuiltin`.

jhump:
> How does this interact with JSONPb.Delimiter?

It needs to use the same text, so that the `NewEncoder` ends up delimiting messages the same way that server-stream output will delimit them. (Server stream output doesn't use `NewEncode`. It instead directly calls `Marshal` and then writes `Delimiter`, if the marshaler has such a method.)

I've updated to make sure these are consistent (per what I _think_ you were recommending). PTAL.

runtime/marshal_jsonpb_test.go
```diff
@@ -278,6 +278,9 @@ func TestJSONPbEncoder(t *testing.T) {
 	}{
 		{
 			verifier: func(json string) {
+				// remove trailing delimiter before verifying
+				json = strings.TrimSuffix(json, "\n")
```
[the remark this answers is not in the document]
jhump: Here's the code in `encoding/json` that adds a newline with every message: https://golang.org/src/encoding/json/stream.go?s=5244:5262#L211

The `TestJSONBuiltinEncoderFields` test does the same thing as my change, adding explicit newlines when asserting the marshaled value: https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/marshal_json_test.go#L146

The reason `TestJSONBuiltinEncoder` doesn't have the same issue as this line of code is that it doesn't actually try to verify the exact JSON output. Instead it just round-trips the value: encode, then decode, and then verify the decoded value is deeply equal to the original.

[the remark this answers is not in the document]
johanbrandhorst: Thanks for this explanation, I think that's good enough of a precedent.

johanbrandhorst:
Thanks for your contribution!

runtime/marshal_jsonpb.go
```diff
@@ -151,7 +151,15 @@ func (d DecoderWrapper) Decode(v interface{}) error {
...
 // NewEncoder returns an Encoder which writes JSON stream into "w".
 func (j *JSONPb) NewEncoder(w io.Writer) Encoder {
-	return EncoderFunc(func(v interface{}) error { return j.marshalTo(w, v) })
+	return EncoderFunc(func(v interface{}) error {
+		if err := j.marshalTo(w, v); err != nil {
+			return err
+		}
+		// mimic json.Encoder by adding a newline (makes output
+		// easier to read when it contains multiple encoded items)
+		_, err := w.Write(j.Delimiter())
```
socketpair: In `etcd` this may be written to different HTTP chunk, right? As far as I know, the protocol must guarantee that 1 chunk = 1 message (JSON).

I don't understand go, but I have the code that triggers that. I mean, sometimes newline is seen as separate HTTP chunk.

socketpair: Moreover. What also stops from gluing two messages in one HTTP chunk?

What I propose: is to document that every message is separated strictly by one newline. In that case, there is possible to write stable stream parsing that does not rely on HTTP chunk boundary.

johanbrandhorst: Could you please raise a separate issue detailing your concerns? I don't think this change covers the behaviour you're asking about.

socketpair: #1086

PR #932, opened 2019-05-10
use error value for bad URI so custom error handler could treat it special

jhump:
When a custom `ProtoErrorHandlerFunc` is used, it is given an error with `Unimplemented` gRPC code for cases where the server gets a request for an unrecognized URI path.

When no such handler is in place, it defaults to a "404 Not Found" error, which is what basically any REST client actually expect for such an error.

Unfortunately, there is not currently any way to distinguish between  "the server received an unrecognized URI path" and an actual "Unimplemented" error returned by a gRPC server handler. So it is not possible for a custom handler to turn those into "404 NotFound" errors, as a REST client would actually expect.

This change hoists the error out into an exported error value: `ErrUnknownURI`. That way, a custom error handler can check if `err == runtime.ErrUnknownURI` and choose to return a 404 for those errors (instead of "501 Not Implemented", which can be used for other errors with an "Unimplemented" gRPC code).

### DIFF
--- a/runtime/mux_test.go
+++ b/runtime/mux_test.go
@@ -2,6 +2,7 @@ package runtime_test
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"net/http"
 	"net/http/httptest"
@@ -9,6 +10,8 @@ import (
 
 	"github.com/grpc-ecosystem/grpc-gateway/runtime"
 	"github.com/grpc-ecosystem/grpc-gateway/utilities"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 )
 
 func TestMuxServeHTTP(t *testing.T) {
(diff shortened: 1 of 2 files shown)


achew22:
This seems eminently reasonable. Would you be willing to add a test or two to confirm the differentiation between a 404 not found and a 404 unimplemented?

jhump:
@achew22, I added tests. PTAL.

runtime/mux_test.go
```diff
@@ -279,6 +325,17 @@ func TestMuxServeHTTP(t *testing.T) {
 	}
 }
 
+func unknownPathIs404(ctx context.Context, mux *runtime.ServeMux, m runtime.Marshaler, w http.ResponseWriter, r *http.Request, err error) {
+	if err == runtime.ErrUnknownURI {
+		w.WriteHeader(http.StatusNotFound)
+	} else {
+		stat, _ := status.FromError(err)
```
johanbrandhorst: You can use `status.Convert` for these cases. If we're not using a recent enough gRPC package, we should upgrade.

jhump: Done. The version this repo uses (1.19.0) does indeed include that function.

PR #935, opened 2019-05-14
Fix make test on MacOS

emilaasa:
`make test` errors out on MacOS because it uses bsd find instead of gnu
find which is standard in most Linux distributions. The fix just
provides an explicit folder . which bsd find requires but gnu find has
as default.

Verifying should be as simple as running make test on a clean repo with any version of gnu `find` (i.e running on linux).


johanbrandhorst:
@emilaasa welcome to the team!

PR #939, opened 2019-05-18
fix query params not populate if method is post

mingqing:
for http method post and content-type is application/x-www-form-urlencoded, the query parameters will be in Request.Form or Request.PostForm.


johanbrandhorst:
Hi @mingqing, thanks for your pull request. The failing CI test is because you need to regenerate the example files. Please see CONTRIBUTING.md in the root of the repo for a docker one liner.

Also, could you add a test that fails before this change and works after it?

Thanks!

mingqing:
@johanbrandhorst I regenerated the file and added it, while passing the CI test.

protoc-gen-grpc-gateway/gengateway/template.go
```diff
@@ -345,7 +345,11 @@ var (
 	{{end}}
 {{end}}
 {{if .HasQueryParam}}
-	if err := runtime.PopulateQueryParameters(&protoReq, req.URL.Query(), filter_{{.Method.Service.GetName}}_{{.Method.GetName}}_{{.Index}}); err != nil {
+	values := req.URL.Query()
+	if req.Method == http.MethodPost {
```
johanbrandhorst: I'm worried that this simple comparison will break existing POST requests using query parameters. Maybe also check that the content type is `x-www-form-urlencoded` as well?

protoc-gen-grpc-gateway/gengateway/template.go
```diff
@@ -345,7 +345,10 @@ var (
 	{{end}}
 {{end}}
 {{if .HasQueryParam}}
-	if err := runtime.PopulateQueryParameters(&protoReq, req.URL.Query(), filter_{{.Method.Service.GetName}}_{{.Method.GetName}}_{{.Index}}); err != nil {
+	if err := req.ParseForm(); err != nil {
```
johanbrandhorst: Isn't this going to fail every time this is used outside a POST request?

johanbrandhorst:
Thanks for the explanation, however, we still need a new test case that was failing before this change and is now passing. Do you need some help to write the test?

mingqing:
> Thanks for the explanation, however, we still need a new test case that was failing before this change and is now passing. Do you need some help to write the test?

Is the test case added in examples/integration/integration_test.go? I will go add.

johanbrandhorst:
> > Thanks for the explanation, however, we still need a new test case that was failing before this change and is now passing. Do you need some help to write the test?
> 
> Is the test case added in examples/integration/integration_test.go? I will go add.

Yes, that would be great, thanks.

examples/integration/integration_test.go
```diff
@@ -1466,3 +1467,58 @@ func testResponseStrings(t *testing.T, port int) {
 	}
 
 }
+
+func TestRequestQueryParams(t *testing.T) {
+	testRequestQueryParamsKind(t, 8080, "GET")
+	testRequestQueryParamsKind(t, 8080, "POST")
+	testRequestQueryParamsKind(t, 8080, "PostForm")
+}
```
johanbrandhorst: Just a matter of style, instead of doing this, have you considered writing subtests?

```go
func TestRequestQueryParams(t *testing.T) {
	t.Run("GET", func(t *testing.T) {
		...
	})
	t.Run("POST", func(t *testing.T) {
		...
	})
	...
}
```

examples/integration/integration_test.go
```diff
@@ -1466,3 +1467,58 @@ func testResponseStrings(t *testing.T, port int) {
...
+	if err != nil {
+		t.Errorf("ioutil.ReadAll(resp.Body) failed with %v; want success", err)
+		return
+	}
+	got = string(buf)
+
+	if got, want := resp.StatusCode, http.StatusOK; got != want {
+		t.Errorf("resp.StatusCode = %d; want %d", got, want)
+		t.Logf("%s", buf)
+	}
```
johanbrandhorst: Could you move this check above the Body check? The subscope `got` variable here is adding a bit of confusion.

examples/integration/integration_test.go
mingqing: I see other examples are written in this style.

johanbrandhorst: I don't quite agree. The way that you're passing in a string to switch on inside the function is pretty nasty. I would suggest separating out the shared content into a function, if at all. Just having 3 different tests would be fine too.

johanbrandhorst:
Thanks for adding the tests, I took the liberty of trying to run the tests without the changes applied and was pleased to see that the `PostForm` test failing, which makes it easier to understand the scope of the change.

examples/integration/integration_test.go
```diff
@@ -9,6 +9,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	neturl "net/url"
```
johanbrandhorst: This is a bit nickpicky, but please don't do this. If you want to name a variable `url`, consider just naming it `u` or `endpoint` or something.

examples/integration/integration_test.go
```diff
@@ -1466,3 +1467,76 @@ func testResponseStrings(t *testing.T, port int) {
...
+			httpMethod:     "POST",
+			contentType:    "application/x-www-form-urlencoded",
+			apiURL:         fmt.Sprintf("http://localhost:%d/v1/example/a_bit_of_everything/params/get/foo?double_value=%v&bool_value=%v", port, 1234.56, true),
+			wantContent:    `{"single_nested":{"name":"foo"},"double_value":1234.56,"bool_value":true,"string_value":"hello-world","repeated_string_value":["demo1","demo2"]}`,
+			requestContent: strings.NewReader(formValues.Encode()),
+		},
+	}
+
+	for idx, tc := range testCases {
+		t.Run(strconv.Itoa(idx), func(t *testing.T) {
```
johanbrandhorst: Better than using the index, how about adding a  `tc.Name`?

examples/integration/integration_test.go
mingqing: The previous case used a lot of url variables. Wouldn't it be confusing if didn't change the net/url, or rename these variables?

mingqing: Thanks for your review, the last commit I renamed the local url variable, so is it feasible?

johanbrandhorst:
Thanks for your contribution!

PR #942, opened 2019-06-07
allow skipping methods and fields for swagger json

gainsley:
Allows the protoc-gen-swagger program to take a yaml config file as input. The config file specifies which methods and fields to skip for swagger doc generation. We use this to filter internal APIs and fields when generating swagger doc for external consumption.


gainsley:
Sorry, wrong upstream repo.

Issue #448, opened 2017-08-30
Configurable AllowUnknownFields in jsonpb?


lantame:
Now unknown object fields are silently dropped.
If I want to handle such cases, I have to provide a custom unmarshaller (likely by copying all unmarshalling stuff from runtime/marshal_jsonpb.go and changing AllowUnknownFields to false). Isn't it better to make this parameter configurable?

P.S. There will probably be some API change:
type JSONPb jsonpb.Marshaler will likely change to
type JSONPb struct{...} 
Which will require existing code modification.

buchanae:
Does [WithMarshalerOption](https://godoc.org/github.com/grpc-ecosystem/grpc-gateway/runtime#WithMarshalerOption) solve your issue? 

We use this to configure a jsonpb.Marshaler the way we like:
```
customMar := jsonpb.Marshaler{
	Indent: "  ",
        EmitDefaults: true,
}
mar := runtime.JSONPb(customMar)
rpcMux := runtime.NewServeMux(runtime.WithMarshalerOption("*/*", &mar))
```

buchanae:
Oh, no, it does not solve your issue. The offending line is here:
https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/marshal_jsonpb.go#L114

That unmarhsaler should be configurable. Maybe `run.time.JSONPb.SetUnmarshaler()` should be added?

racevedoo:
@buchanae @Lantame any update on this? I'm having the exact same problem

PR #948, opened 2019-06-14
Add new option for the decoder - to disallow unknown fields

vsaveliev:
Hi, gRPC Gateway's Team. 

We glad to use gRPC Gateway in our production a lot. We are affected by the problem of unknown fields' default allowing. I found existing issue about it - #448 . So as a new bie in the project I have made easy fix for this functionality which doesn't affect full project but it will help some people like we. Please make a review of the change and give me feedback or just merge it :-)

### DIFF
--- a/runtime/marshal_jsonpb.go
+++ b/runtime/marshal_jsonpb.go
@@ -172,7 +172,7 @@ func decodeJSONPb(d *json.Decoder, v interface{}) error {
 	if !ok {
 		return decodeNonProtoField(d, v)
 	}
-	unmarshaler := &jsonpb.Unmarshaler{AllowUnknownFields: true}
+	unmarshaler := &jsonpb.Unmarshaler{AllowUnknownFields: allowUnknownFields}
 	return unmarshaler.UnmarshalNext(d, p)
 }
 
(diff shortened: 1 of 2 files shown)


vsaveliev:
I signed it!

johanbrandhorst:
Ah, yes, using the "accept suggestion" button breaks the CLA bot thingy. Did you see my question?

> Instead of doing this, how about we pass in the Marshaller in the DecoderWrapper and use the marshaller instead? Then users that need this functionality can configure their own marshaller.

vsaveliev:
yes, I did. I had tried to do an option in `Marshaler` struct before my current fix, but I didn't understand how to pass the option in `unmarshalJSONPb` / `decodeJSONPb` / `decodeNonProtoField` functions right.

johanbrandhorst:
OK I've had another look and I can see there's not much else we can do here with the existing API - we don't allow the fields on the unmarshaler to be set anywhere. This is fine, but could you add some tests? I'll have @achew22 fix the CLA errors (I hereby allow my commits to be added to this and I agree to the license under which it is contributed).

vsaveliev:
sure, I can add tests for it.

vsaveliev:
I have opened https://github.com/grpc-ecosystem/grpc-gateway/pull/959 (with the test) to solve a problem with CLA and have closed this PR.

PR #949, opened 2019-06-15
Support colon in final path segment, last match wins behavior (behind flags)

jfhamlin:
Addresses #224. Reintroduces changes from #708, this time behind a command line flag (`-allow_colon_final_segments`) and adds a new option to `runtime.ServeMux` to enable "last match wins" behavior. Please see discussion in #224 and #708 for context.

### DIFF
--- a/examples/proto/examplepb/a_bit_of_everything.pb.gw.go
+++ b/examples/proto/examplepb/a_bit_of_everything.pb.gw.go
@@ -1481,45 +1481,45 @@ func RegisterABitOfEverythingServiceHandlerClient(ctx context.Context, mux *runt
 }
 
 var (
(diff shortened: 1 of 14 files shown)


johanbrandhorst:
Hi @jfhamlin, thanks for the contribution! Looks like you need to regenerate some of the example files, please see CONTRIBUTING.md for two docker one liners you can run.

jfhamlin:
> please see CONTRIBUTING.md for two docker one liners you can run.

Done!

aysylu:
This appears to be breaking existing projects with:

```
proto/v1beta1/project_go_proto/project.pb.gw.go:244:75: too many arguments in call to runtime.NewPattern
proto/v1beta1/project_go_proto/project.pb.gw.go:244:135: undefined: runtime.AssumeColonVerbOpt
```

Test link: https://circleci.com/gh/grafeas/grafeas/1066?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link. Could you provide some insight into what's broken?

johanbrandhorst:
@aysylu I've just released 1.9.2, which should fix your build. As a general note, the error you're getting is because you're using the `master` version of the generator (`protoc-gen-grpc-gateway`), and a locked version of the runtime. Ideally, these should always be the same version. In reality, we know this confuses a lot of users, so please update to 1.9.2 and it should fix your issue.

aysylu:
@johanbrandhorst we were able to get around the issue updating our config and pointing to `1.9.0`. We did test whether `1.9.2` and it looks like it fails with the same issue, just an FYI. Happy to provide more information on the dependencies if it's helpful.

johanbrandhorst:
Hm, how are you generating the files and versioning the runtime? I'd be very interested in hearing of 1.9.2 is broken.

PR #954, opened 2019-06-18
docs/customizingyourgateway: add ?pretty example

srenatus:
As discussed in slack. The code is rough, but should illustrate the
point.

### DIFF
--- a/docs/_docs/customizingyourgateway.md
+++ b/docs/_docs/customizingyourgateway.md
@@ -28,6 +28,55 @@ The protocol buffer compiler generates camelCase JSON tags that can be used with
    mux := runtime.NewServeMux(runtime.WithMarshalerOption(runtime.MIMEWildcard, &runtime.JSONPb{OrigName:false}))
    ```
 
+### Pretty-print JSON responses when queried with ?pretty
+
+You can have Elasticsearch-style `?pretty` support in your gateway's endpoints as follows:
+
+1. Wrap the ServeMux using a stdlib [`http.HandlerFunc`](https://golang.org/pkg/net/http/#HandlerFunc)
+   that translates the provided query parameter into a custom `Accept` header, and
+2. Register a pretty-printing marshaler for that MIME code.
+
+For example:
+
+```go
+mux := runtime.NewServeMux(
+	runtime.WithMarshalerOption("application/json+pretty", &runtime.JSONPb{Indent: "  "}),
+)
+prettier := func(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// checking Values as map[string][]string also catches ?pretty and ?pretty=
+		// r.URL.Query().Get("pretty") would not.
+		if _, ok := r.URL.Query()["pretty"]; ok {
+			r.Header.Set("Accept", "application/json+pretty")
+		}
+		h.ServeHTTP(w, r)
+	})
+}
+http.ListenAndServe(":8080", prettier(mux))
+```
+
+Note that  `runtime.JSONPb{Indent: "  "}` will do the trick for pretty-printing: it wraps
+`jsonpb.Marshaler`:
+```go
+type Marshaler struct {
+	// ...
+
+	// A string to indent each level by. The presence of this field will
+	// also cause a space to appear between the field separator and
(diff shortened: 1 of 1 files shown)


docs/_docs/customizingyourgateway.md
```diff
@@ -28,6 +28,43 @@ The protocol buffer compiler generates camelCase JSON tags that can be used with
    mux := runtime.NewServeMux(runtime.WithMarshalerOption(runtime.MIMEWildcard, &runtime.JSONPb{OrigName:false}))
    ```
 
+### Pretty-print JSON responses when queried with ?pretty
+
+You can have Elasticsearch-style `?pretty` support in your gateway's endpoints as follows:
+
+1. Wrap the ServeMux using a stdlib [`http.HandlerFunc`](#TODO) that translates the provided query parameter
```
johanbrandhorst: Did you mean to add a link here?

srenatus: oops will fix

docs/_docs/customizingyourgateway.md
```diff
@@ -28,6 +28,43 @@ The protocol buffer compiler generates camelCase JSON tags that can be used with
...
+   into a custom `Accept` header, and
+2. Register a pretty-printing marshaller for that MIME code.
+
+For example:
+
+```go
+mux := runtime.NewServeMux(
+	runtime.WithMarshalerOption("application/json+pretty", &runtime.JSONPb{Indent: "  "}),
+	)
```
johanbrandhorst: Remove this tab, I think.

docs/_docs/customizingyourgateway.md
```diff
@@ -28,6 +28,43 @@ The protocol buffer compiler generates camelCase JSON tags that can be used with
...
+
+1. Wrap the ServeMux using a stdlib [`http.HandlerFunc`](#TODO) that translates the provided query parameter
+   into a custom `Accept` header, and
+2. Register a pretty-printing marshaller for that MIME code.
+
+For example:
+
+```go
+mux := runtime.NewServeMux(
+	runtime.WithMarshalerOption("application/json+pretty", &runtime.JSONPb{Indent: "  "}),
```
johanbrandhorst: Maybe add a small comment about the meaning of the `Indent` parameter.

docs/_docs/customizingyourgateway.md
```diff
@@ -28,6 +28,43 @@ The protocol buffer compiler generates camelCase JSON tags that can be used with
...
+	)
+prettier := func(h http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		defer h.ServeHTTP(w, r)
+		// if the "Accept" is set and NOT */* (sent by curl), keep it as-is
```
johanbrandhorst: Why does this need special casing around `*/*`?

srenatus: It doesn't, in general. However, all folks I know (small sample) would end up using `curl` on that endpoint, and `curl` sends `Accept: */*` by default. So, if we only check that an `Accept` header _was not provided_, this will not work for `curl` users.

srenatus: It doesn't, in general. However, all folks I know (small sample) would end up using `curl` on that endpoint, and `curl` sends `Accept: */*` by default. So, if we only check that an `Accept` header _was not provided_, this will not work for `curl` users.

srenatus: If figured, if you do not care (i.e. "accept `*/*`"), we might a well override your choice 😄

johanbrandhorst: What about removing this check altogether? Any `Accept` header could be overridden. when the `pretty` parameter is present, maybe? What cases would that mess with?

johanbrandhorst:
I think the bazel failures are because there's an update available - I'll get @drigz or @achew22 on sorting that (#955), we can still merge this change.

docs/_docs/customizingyourgateway.md
srenatus: Oh you're right. In this example, that's surely fine. I was thinking _better safe than sorry_; but it's fine here.

johanbrandhorst:
Thanks for your contribution!

PR #957, opened 2019-06-18
Auto-generate go_repository rules.

drigz:
The rules are autogenerated with:

```
bazel run //:gazelle -- update-repos -from_file=go.mod
```

And I manually rearranged and added the comment. This is quite messy, maybe @achew22 has an idea how it could be improved.


johanbrandhorst:
Will the existing CI tests pick up if this gets out of sync with `go.mod`?

drigz:
> Will the existing CI tests pick up if this gets out of sync with `go.mod`?

No. `update-repos` doesn't have a `diff` mode, but we could try home-brewing something similar?

johanbrandhorst:
> > Will the existing CI tests pick up if this gets out of sync with `go.mod`?
> 
> No. `update-repos` doesn't have a `diff` mode, but we could try home-brewing something similar?

How about rerunning the generation script in CI and git diffing? Not sure we care about keeping the comments. If there's a better way just do that but I think it'd be nice to ensure it's kept in sync.

achew22:
[update-repos](https://github.com/bazelbuild/bazel-gazelle#update-repos) has a great flag that can be set by the WORKSPACE directive

```
# gazelle:repository_macro dependencies%grpc_gateway_dependencies
```
then we could centralize all that logic into a file that is basically the same as a `.sum` file. Unfortunately this feature is not yet in a release of Gazelle (there hasn't been one since Feb and the feature went in [here](https://github.com/bazelbuild/bazel-gazelle/pull/480)) so we can't depend on it without pointing at a specific commit. @jayconrod, do you have any ideas on how to move forward here? Is it safe to depend on a specific commit of Gazelle or would you prefer we wait till the next release?

jayconrod:
Sorry for not responding earlier. If you want to depend on a specific commit, `2cf99d7` is probably best (or `master` after bazelbuild/bazel-gazelle#555 is fixed).

I've been holding the release until Go module support is more fleshed out. bazelbuild/bazel-gazelle#477 is the last important thing, but most people won't need that. It should be done soon.

achew22:
@jayconrod, looks like https://github.com/bazelbuild/bazel-gazelle/commit/72ba271916ca02aaaff72949dc0c0a63ab37d395 is the commit to pin to, is that correct?

jayconrod:
@achew22 Yep, that fixes the issue. Let me know if you run into any problems.

johanbrandhorst:
What is the status of this? @drigz @achew22

achew22:
I did it in https://github.com/grpc-ecosystem/grpc-gateway/commit/d63917fcb0d53f39184485b9b6a0893af18a5668. @drigz, thanks for the contribution!

PR #967, opened 2019-07-08
Fix query.go to avoid invalid protobuf assumptions

dsnet:
The golang/protobuf compatibility agreement reserves the right add/remove
XXX fields and methods.

Fix this code to avoid assuming:
* that XXX_WellKnownType always exists (it won't in the future).
Instead, it should determine whether a message is a well-known type
based on its full name.
* that Go struct field indexes are stable (they aren't).
An internal field may be added to the front of the message,
causing the current assumptions to be invalidated.
Instead, access the struct field by name, which will not change.

### DIFF
--- a/runtime/query.go
+++ b/runtime/query.go
@@ -202,103 +202,102 @@ func populateField(f reflect.Value, value string, props *proto.Properties) error
 	i := f.Addr().Interface()
 
 	// Handle protobuf well known types
-	type wkt interface {
-		XXX_WellKnownType() string
-	}
-	if wkt, ok := i.(wkt); ok {
-		switch wkt.XXX_WellKnownType() {
-		case "Timestamp":
-			if value == "null" {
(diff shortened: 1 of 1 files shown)


johanbrandhorst:
I've triggered a rerun of the flaky test.

johanbrandhorst:
Thanks for your contribution! When will these changes be breaking? I.e. what version of golang/protobuf will require this commit to work?

dsnet:
> When will these changes be breaking?

When `google.golang.org/protobuf` is released, we will also tag a `github.com/golang/protobuf@v1.4.0` that depends on the new implementation. We're targeting a release at the end of July, but it's likely that the date will slip, unfortunately.

> what version of golang/protobuf will require this commit to work?

Thus, anything that depends on the new implementation or v1.4.0+ will be broken.

PR #985, opened 2019-08-08
Use json camel case when GetUseJSONNamesForFields is enabled

xin-au:
Use json camel case when GetUseJSONNamesForFields is enabled

PR #988, opened 2019-08-10
Camel case backward compatible in Swagger doc

xin-au:
Hi @johanbrandhorst,
In the PR, https://github.com/grpc-ecosystem/grpc-gateway/pull/985, the snake case arguments will be converted to lower camel case when the `json_names_for_fields ` is enabled.
Now, I noticed that there is a backward compatible use case may need to support, which is `/` could be inside `{}`, for example, `"/{user.name=prefix/*}:customMethod"`. (Sorry I did not notice that)
Also, when calculating `jsonSnakeCaseName`, if the name includes `/`, then it could cause `out of range` potentially. So, fixed it too.
Sorry for about.

### DIFF
--- a/protoc-gen-swagger/genswagger/template.go
+++ b/protoc-gen-swagger/genswagger/template.go
@@ -626,7 +626,7 @@ func templateToSwaggerPath(path string, reg *descriptor.Registry) string {
 			buffer += string(char)
 			if reg.GetUseJSONNamesForFields() &&
 				len(jsonBuffer) > 1 {
-				jsonSnakeCaseName := string(jsonBuffer[1 : len(buffer)-1])
+				jsonSnakeCaseName := string(jsonBuffer[1:])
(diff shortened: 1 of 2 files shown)


xin-au:
@johanbrandhorst , yes, it covered the test case  (`\` inside `{}`) 
In `TestFQMNtoSwaggerName` and `TestTemplateToSwaggerPath` with enabling `SetUseJSONNamesForFields` by setting it as `true`, which means to enable the JSON camle case, otherwise it is `false` by default. The following links are the tests with `SetUseJSONNamesForFields(true)`
https://github.com/grpc-ecosystem/grpc-gateway/pull/988/files#diff-20461b822a903046234d0fa5073ba2c4R924 
https://github.com/grpc-ecosystem/grpc-gateway/pull/988/files#diff-20461b822a903046234d0fa5073ba2c4R1009

johanbrandhorst:
Thanks for your contribution!

xin-au:
Thank you so much @johanbrandhorst !

PR #1003, opened 2019-08-23
Allow overriding the primitive type of fields

snowzach:
Allows you to specify a custom type in case you've altered the default json handling behavior.


snowzach:
Do I need to generate everything with bazel to make this pass CI?

johanbrandhorst:
> Do I need to generate everything with bazel to make this pass CI?

Yes, but there's a handy oneliner in CONTRIBUTING.md

protoc-gen-swagger/genswagger/template.go
```diff
@@ -1599,6 +1599,9 @@ func updateSwaggerObjectFromJSONSchema(s *swaggerSchemaObject, j *swagger_option
 	s.MaxProperties = j.GetMaxProperties()
 	s.MinProperties = j.GetMinProperties()
 	s.Required = j.GetRequired()
+    if overrideType := j.GetType(); len(overrideType) == 1 {
```
johanbrandhorst: Is this file gofmt'ed? It looks out of place in this diff.

examples/proto/examplepb/a_bit_of_everything.proto
```diff
@@ -172,7 +172,7 @@ message ABitOfEverything {
 	repeated Nested nested = 2;
 	float float_value = 3 [(grpc.gateway.protoc_gen_swagger.options.openapiv2_field) = {description: "Float value field", default: "0.2", required: ['float_value']}];
 	double double_value = 4;
-	int64 int64_value = 5;
+	int64 int64_value = 5 [(grpc.gateway.protoc_gen_swagger.options.openapiv2_field) = {type: INTEGER}];;
```
johanbrandhorst: Could we add another int64 field instead of overwriting the properties of the existing one? We still want to test that it's a string by default.

snowzach: Sure, I added another one.

protoc-gen-swagger/genswagger/template.go
snowzach: I assumed VSCode was formatting it right. Apparently not.. This has been fixed.

examples/proto/examplepb/a_bit_of_everything.swagger.json
```diff
@@ -138,6 +139,7 @@
           "404": {
             "description": "Returned when the resource does not exist.",
             "schema": {
+              "type": "string",
```
johanbrandhorst: These additions make me a bit nervous. Have you checked this is still a valid swagger file? Is it correct to have these here now?

examples/proto/examplepb/a_bit_of_everything.swagger.json
```diff
@@ -421,6 +423,13 @@
               "TRUE"
             ],
             "default": "FALSE"
+          },
+          {
+            "name": "int64_override_type",
+            "in": "query",
+            "required": false,
+            "type": "integer",
+            "format": "int64"
```
johanbrandhorst: Is this format value well defined?

examples/proto/examplepb/a_bit_of_everything.swagger.json
snowzach: I honestly wonder how it was correct to not have the `type` field there. I will admit, I have done very little with swagger but from what I have seen it seems more typical to always have a type field and only sometimes have the format field. 
I was able to paste the generated swagger `a_bit_of_everything.swagger.json` into https://editor.swagger.io/ and it seemed happy with the `type` related stuff. It complained about something unrelated to that though.

examples/proto/examplepb/a_bit_of_everything.swagger.json
snowzach: It appears to be according to this: https://swagger.io/docs/specification/data-models/data-types/#numbers

From what I have read, the `type` field must consist of one of the defined types (for which there are enums) and the format field is somewhat arbitrary. 
Ref: https://swagger.io/docs/specification/data-models/data-types/#string
At least for strings it says that `format` is an open value. Under the `integer` it says that format is a hint for whatever program is using it.

PR #1016, opened 2019-08-30
Fix hand-crafted protobuf message

dsnet:
errorBody is an improperly hand-crafted protobuf Message.
In particular, field number 1 is used twice, causing this to
message to crash in the v2 re-implementation of Go protobufs.

This PR modifies the field numbers to be unique.
Since Error is a special field not in status.proto,
we choose a large field number to avoid conflicting with
any new field that may be added to the real Status message.
We also fix the field numbers to truly match up with Status.

### DIFF
--- a/runtime/errors.go
+++ b/runtime/errors.go
@@ -66,12 +66,12 @@ var (
 )
 
 type errorBody struct {
-	Error string `protobuf:"bytes,1,name=error" json:"error"`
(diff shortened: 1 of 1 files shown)


dsnet:
\cc @johanbrandhorst

achew22:
@dsnet, hope that wasn't too hard to track down. Sorry about that!

PR #1021, opened 2019-08-31
invoke in process server need attach md to incoming context

hb-chen:
`rpc` mode attach metadata use NewOutgoingContext(), while invoke in process need attach metadata use NewIncomingContext()

### DIFF
--- a/runtime/context.go
+++ b/runtime/context.go
@@ -60,7 +60,8 @@ func AnnotateContext(ctx context.Context, mux *ServeMux, req *http.Request) (con
 	ctx, md, err := annotateContext(ctx, mux, req)
 	if err != nil {
 		return nil, err
-	} else if md == nil {
+	}
+	if md == nil {
 		return ctx, nil
 	}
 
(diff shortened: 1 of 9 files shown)


runtime/context.go
```diff
@@ -57,13 +57,37 @@ except that the forwarded destination is not another HTTP service but rather
 a gRPC service.
 */
 func AnnotateContext(ctx context.Context, mux *ServeMux, req *http.Request) (context.Context, error) {
+	ctx, md, err := annotateContext(ctx, mux, req)
+	if err != nil {
+		return nil, err
+	} else if md == nil {
```
johanbrandhorst: Using `else` unnecessarily is often a bit frowned upon. How about 
```suggestion
	}
	if md == nil {
```

runtime/context.go
```diff
@@ -57,13 +57,37 @@ except that the forwarded destination is not another HTTP service but rather
...
+	return metadata.NewOutgoingContext(ctx, md), nil
+}
+
+// AnnotateIncomingContext adds context information such as metadata from the request.
+// Attach metadata as incoming context.
+func AnnotateIncomingContext(ctx context.Context, mux *ServeMux, req *http.Request) (context.Context, error) {
+	ctx, md, err := annotateContext(ctx, mux, req)
+	if err != nil {
+		return nil, err
+	} else if md == nil {
```
johanbrandhorst: Same here
```suggestion
	}
	if md == nil {
```

hb-chen: tks

johanbrandhorst:
Thanks for the bug fix! I'll give it a few days if you discover anything else before making a bug fix release.

Issue #1019, opened 2019-08-30
Fuzzit CI job is failing unexpectedly


johanbrandhorst:
Two recent jobs (https://circleci.com/gh/grpc-ecosystem/grpc-gateway/3946 and https://circleci.com/gh/grpc-ecosystem/grpc-gateway/3948) suffered from fuzzit failures that were unrelated to the changes made. We should investigate and fix the issue.

cc @yevgenypats

yevgenypats:
I'm on it. I think this is connected to go-fuzz and go modules support. I'll push a fix now.
https://github.com/dvyukov/go-fuzz/pull/263

johanbrandhorst:
Fixed by #1022

PR #1022, opened 2019-08-31
Fuzzit: use go-fuzz with go-modules support

yevgenypats:
Testing a fix which happens because go-fuzz doesn't support go-modules. current fix use Fuzzit fork which supports go-modules until this support will be integrated into go-fuzz.

https://github.com/grpc-ecosystem/grpc-gateway/issues/1019

cc @johanbrandhorst


yevgenypats:
@johanbrandhorst I think this works now. I currently use fuzzit fork until this  or other commit merges into go-fuzz and I'll update again to point to original go-fuzz.

PR #1026, opened 2019-09-03
docs need type assertion on proto.Message

zachgersh:
makes a bit more sense with the type assertion and not using the `proto.Message` directly.

### DIFF
--- a/docs/_docs/customizingyourgateway.md
+++ b/docs/_docs/customizingyourgateway.md
@@ -104,7 +104,7 @@ Or you might want to mutate the response messages to be returned.
 1. Write a filter function.
    ```go
    func myFilter(ctx context.Context, w http.ResponseWriter, resp proto.Message) error {
-        t, ok := resp.(*external.Tokenizer)
+        t, ok := resp.(*externalpb.Tokenizer)
 
 	if ok {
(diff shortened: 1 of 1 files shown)


docs/_docs/customizingyourgateway.md
```diff
@@ -104,8 +104,13 @@ Or you might want to mutate the response messages to be returned.
 1. Write a filter function.
    ```go
    func myFilter(ctx context.Context, w http.ResponseWriter, resp proto.Message) error {
-   	w.Header().Set("X-My-Tracking-Token", resp.Token)
-   	resp.Token = ""
+        t, ok := resp.(*external.Tokenizer)
```
johanbrandhorst: Maybe even something like `externalpb` to make it more clear what general type this is.
```suggestion
        t, ok := resp.(*externalpb.Tokenizer)
```

zachgersh: @johanbrandhorst - yeah I like this. I wondered if we should place the type right above the filter?

johanbrandhorst:
Fuzzit failure is fixed on master, I'll merge this.

johanbrandhorst:
Thanks for the contribution!

PR #1030, opened 2019-09-04
nitpick a_bit_of_everything.proto

srenatus:


examples/proto/examplepb/a_bit_of_everything.proto
```diff
@@ -131,8 +131,7 @@ option (grpc.gateway.protoc_gen_swagger.options.openapiv2_swagger) = {
 };
 
 
-// Intentionaly complicated message type to cover many features of Protobuf.
-// NEXT ID: 30
```
srenatus: It wasn't up to date, so I've rather just removed it. 😃

Issue #992, opened 2019-08-12
Go integration tests are somewhat flaky


johanbrandhorst:
This doesn't happen with anything like the regularity of the bazel tests as described in #968, but it is still happening for unrelated changes. See https://circleci.com/gh/grpc-ecosystem/grpc-gateway/3630 for an example test failure:

```
--- FAIL: TestResponseBody (0.01s)
    --- FAIL: TestResponseBody/0 (0.00s)
        integration_test.go:1509: http.Get("http://localhost:8081/responsestrings/foo") failed with Get http://localhost:8081/responsestrings/foo: dial tcp 127.0.0.1:8081: connect: connection refused; want success
```

I think this is a just a race condition on startup, and we might be able to fix it by just introducing a  longer sleep on startup.

johanbrandhorst:
Fixed in #1039

PR #1039, opened 2019-09-11
examples/integration: wait for secondary gateway

srenatus:
I haven't been able to reproduce #992, but this looks like a likely fix. ;)


johanbrandhorst:
Thanks!

PR #1049, opened 2019-09-25
Fix auto generated fieldmask case in patch requests

william-plano-oxb:
Fix up tests to check expected behaviour
Fix up documentation to match new behaviour


examples/server/a_bit_of_everything.go
```diff
@@ -83,7 +82,7 @@ func (s *_ABitOfEverythingServer) BulkCreate(stream examples.StreamService_BulkC
 			return err
 		}
 		count++
-		glog.Error(msg)
+		glog.Info(msg)
```
william-plano-oxb: This was outputting error level logging during the test suite, I suspect it's left over from somebody's debugging. It doesn't seem to be an error case.

examples/server/a_bit_of_everything.go
```diff
@@ -187,18 +186,8 @@ func (s *_ABitOfEverythingServer) UpdateV2(ctx context.Context, msg *examples.Up
 }
 
 // PatchWithFieldMaskInBody differs from UpdateV2 only in that this method exposes the field mask in the request body,
-// so that clients can specify their mask explicitly
+// so that clients can specify their mask explicitly. The difference is entirely in the http interface.
 func (s *_ABitOfEverythingServer) PatchWithFieldMaskInBody(ctx context.Context, request *examples.UpdateV2Request) (*empty.Empty, error) {
-	// low-effort attempt to modify the field mask to only include paths for the ABE struct. Since this is only for the
```
william-plano-oxb: This method could be removed entirely and the rest endpoint annotation in the proto file added an additional binding to the UpdateV2 method. But it's perhaps useful for documentation to have it separate. 🤷‍♂

examples/server/fieldmask_helper.go
```diff
@@ -39,7 +40,7 @@ func getField(obj interface{}, path string) (val reflect.Value) {
 		if v.Kind() == reflect.Ptr {
 			v = reflect.Indirect(v)
 		}
-		v = v.FieldByName(s)
+		v = v.FieldByName(generator.CamelCase(s))
```
william-plano-oxb: This is where I think the case fixing should go, in the user's server code. I had a quick go at doing something like golang/protobuf#457 (comment) to pull the name from the descriptor, but couldn't get it to work with a generic interface/reflection, I think it requires the concrete type to be passed in making this helper a lot less useful.

examples/server/a_bit_of_everything.go
johanbrandhorst: Nice find, I think we can leave it in since this is just used for testing and as an example.

examples/server/fieldmask_helper_test.go
```diff
@@ -81,12 +93,12 @@ func TestSetValue(t *testing.T) {
 }
 
 type a struct {
-	B *b
-	E int
-	F string
+	BeefCake *b
```
william-plano-oxb: I've modified these names to fully test the casing support

protoc-gen-grpc-gateway/gengateway/template.go
```diff
@@ -314,16 +314,14 @@ var (
 	if err := marshaler.NewDecoder(newReader()).Decode(&{{.Body.AssignableExpr "protoReq"}}); err != nil && err != io.EOF  {
 		return nil, metadata, status.Errorf(codes.InvalidArgument, "%v", err)
 	}
-	{{- if and $AllowPatchFeature (and (eq (.HTTPMethod) "PATCH") (.FieldMaskField))}}
-	if protoReq.{{.FieldMaskField}} != nil && len(protoReq.{{.FieldMaskField}}.GetPaths()) > 0 {
-		runtime.CamelCaseFieldMask(protoReq.{{.FieldMaskField}})
-	} {{if not (eq "*" .GetBodyFieldPath)}} else {
+	{{- if and $AllowPatchFeature (eq (.HTTPMethod) "PATCH") (.FieldMaskField) (not (eq "*" .GetBodyFieldPath)) }}
```
william-plano-oxb: Simplified logic here

protoc-gen-grpc-gateway/gengateway/template_test.go
```diff
@@ -628,15 +628,18 @@ func TestAllowPatchFeature(t *testing.T) {
 						Bindings: []*descriptor.Binding{
 							{
 								HTTPMethod: "PATCH",
-								Body:       &descriptor.Body{FieldPath: nil},
+								Body: &descriptor.Body{FieldPath: descriptor.FieldPath{descriptor.FieldPathComponent{
```
william-plano-oxb: Need to pass a body now to generate the code being tested

examples/server/a_bit_of_everything.go
johanbrandhorst: I think it'd be nicer to remove this and show that it's possible to use an additional binding instead. Would you mind changing it?

johanbrandhorst:
The bazel failure is real btw: https://circleci.com/gh/grpc-ecosystem/grpc-gateway/4409. We'll need to run something to re-generate the bazel files I think (@achew22, @drigz?).

william-plano-oxb:
> This is a great PR - thanks for all the comments.

Thanks :smile:

> Something I'm curious about, what happens if someone uses non-standard casing in their protobuf field names. Will using camelCase for your fields produce camelCased field masks?

The intention is that the field masks are taken from the json keys in the request as-is. So as long as the user passes camelCase names matching the protobuf field names in the request, it should result in camelCase field masks too. That said, this seems like a reasonable thing to add a test for.

william-plano-oxb:
> The bazel failure is real btw: https://circleci.com/gh/grpc-ecosystem/grpc-gateway/4409. We'll need to run something to re-generate the bazel files I think (@achew22, @drigz?).

I did have a look at that, but I've not used Bazel before and I wasn't sure what it was telling me (though there is a command in the error output that I'll try running locally and see if it makes any changes).
Oddly the diff that it seems to be complaining about doesn't look like a diff at all, maybe it's a line ending thing?

examples/server/a_bit_of_everything.go
william-plano-oxb: Sure thing, I was very much on-the-fence about whether to keep it.

drigz:
> Oddly the diff that it seems to be complaining about doesn't look like a diff at all, maybe it's a line ending thing?

You should be able to run the command or manually apply the diff, eg:

```
patch -p0 << EOF
--- examples/server/BUILD.bazel	1970-01-01 00:00:00.000000000 +0000
+++ examples/server/BUILD.bazel	1970-01-01 00:00:00.000000000 +0000
@@ -20,6 +20,7 @@
         "//examples/proto/sub2:go_default_library",
         "@com_github_golang_glog//:go_default_library",
+        "@com_github_golang_protobuf//protoc-gen-go/generator:go_default_library_gen",
         "@com_github_rogpeppe_fastuuid//:go_default_library",
         "@go_googleapis//google/rpc:errdetails_go_proto",
         "@io_bazel_rules_go//proto/wkt:duration_go_proto",
--- runtime/BUILD.bazel	1970-01-01 00:00:00.000000000 +0000
+++ runtime/BUILD.bazel	1970-01-01 00:00:00.000000000 +0000
@@ -29,7 +29,6 @@
         "//utilities:go_default_library",
-        "@com_github_golang_protobuf//protoc-gen-go/generator:go_default_library_gen",
         "@go_googleapis//google/api:httpbody_go_proto",
         "@io_bazel_rules_go//proto/wkt:any_go_proto",
         "@io_bazel_rules_go//proto/wkt:duration_go_proto",
EOF
```

The diff reflects the fact the changes in the `import "github.com/golang/protobuf/protoc-gen-go/generator"` lines.

william-plano-oxb:
> The diff reflects the fact the changes in the `import "github.com/golang/protobuf/protoc-gen-go/generator"` lines

🤦‍♂ Of course, it's 2 diffs. I was just reading it wrong.

william-plano-oxb:
> Could we add a test for the latter being transformed correctly as well?

Will do, fingers crossed it "just works"™️

william-plano-oxb:
> > Could we add a test for the latter being transformed correctly as well?
> 
> Will do, fingers crossed it "just works"™️

Unfortunately, if somewhat predictably, it doesn't work very well. In the simple cases it works just fine, however there a few edge cases that my testing has picked up where JSON's camelCasing and protoc-gen-go/generator's camelCasing disagree, for a concrete example "foo_Bar" becomes "fooBar" with the former, but remains "foo_Bar" with the latter.

The more serious issue is however that if you specify [json_name] options in the proto file, then all bets are off since you can completely rename things rather than just changing the case.

I'll have to have a bit of a think about how to sort this out.

johanbrandhorst:
> Unfortunately, if somewhat predictably, it doesn't work very well. In the simple cases it works just fine, however there a few edge cases that my testing has picked up where JSON's camelCasing and protoc-gen-go/generator's camelCasing disagree, for a concrete example "foo_Bar" becomes "fooBar" with the former, but remains "foo_Bar" with the latter.
> 
> The more serious issue is however that if you specify [json_name] options in the proto file, then all bets are off since you can completely rename things rather than just changing the case.
> 
> I'll have to have a bit of a think about how to sort this out.

I don't think we're going to be able to support all the cases, and that's okay. It's possible for users to turn off the FieldMask implementation, and for the vast majority of cases it should work. I think this fix is better than what we have at the moment as it should be closer to the source of truth (the proto field name). I'm OK with these shortcomings as long as we document them, of course.

william-plano-oxb:
Okay, sorry, I went down a bit of a rabbit hole trying to support the json names. I have left that out of this PR for now.
I do have something that works, but it's not very clean. Shall I check it in as a separate PR on my fork?

johanbrandhorst:
> Okay, sorry, I went down a bit of a rabbit hole trying to support the json names. I have left that out of this PR for now.
> I do have something that works, but it's not very clean. Shall I check it in as a separate PR on my fork?

Hm, no, go ahead and check it into this one. I'm curious now :). This is already one of the best PRs I've reviewed so the anticipation is great!

william-plano-oxb:
> Hm, no, go ahead and check it into this one. I'm curious now :). This is already one of the best PRs I've reviewed so the anticipation is great!

I'm a bit reluctant to mess this PR up if we decide not to take the change though. I guess we could always revert it with a force push to my branch, not sure if that would break the world though.

johanbrandhorst:
> > Hm, no, go ahead and check it into this one. I'm curious now :). This is already one of the best PRs I've reviewed so the anticipation is great!
> 
> I'm a bit reluctant to mess this PR up if we decide not to take the change though. I guess we could always revert it with a force push to my branch, not sure if that would break the world though.

Alright fine, lets leave it out for now and open a separate PR.

william-plano-oxb:
> Alright fine, lets leave it out for now and open a separate PR.

Okay, all synced up with master now, and I've created https://github.com/william-plano-oxb/grpc-gateway/pull/1 on my fork with my attempt at getting the json names supported.

johanbrandhorst:
> Could we add another section to the _docs/patch.md documenting the supported json <-> proto field name mappings? Just something like "If you define custom JSON names, you're SOL with regards to the automatic PATCH feature".

Cool, as I said in the other PR, it LGTM, but we can still merge it separately. I'd be happy to merge this after ^ has been addressed.

william-plano-oxb:
Hmm, as I was fixing up the documentation for this I've noticed something that didn't seem right: The second case where we say

> * The FieldMask is exposed to the REST request (as in the PatchWithFieldMaskInBody example). For this case, a check is made as to whether the FieldMask is nil/empty prior to populating with the request body

Isn't true at all, we just don't do anything for this case. I've checked the code before I started poking at it, and all we did then was the case translation, it didn't take anything from the Body of the message. And I have confirmed this by modifying the TestABEPatchBody test to add an extra field on the originalValues that isn't in the input values, in the case of nil/empty field mask we're just falling through to the full Update/PUT semantics. 

The question is; what do we want to do here? Shall I just fix up the docs, or modify the code to support this case properly? (And if the latter, should that be in another separate PR?)

william-plano-oxb:
In fact, I'm not sure how we would know what the body of the resource is called in that case. Maybe we can just say the request body can only have 2 fields, and one of them must be the field_mask so the body is necessarily the other one?

johanbrandhorst:
> In fact, I'm not sure how we would know what the resource is called in that case. Maybe we can just say the request body can only have 2 fields, and one of them must be the field_mask so the resource is necessarily the other one?
Yeah I think a prerequisite of using this feature is to design your PATCH requests following the Google API design convention (https://cloud.google.com/apis/design/standard_methods#update). I don't particularly care about the fieldmask-in-body request since then you can also just not use this feature, so we can remove it I think.

william-plano-oxb:
> following the Google API design convention

> * The request message field containing the resource **must** map to the request body.

I don't think I realised it was this strict, seems fair to exclude this case then. I'll fix up the docs while I'm there.

examples/integration/integration_test.go
```diff
@@ -719,18 +724,25 @@ func TestABEPatchBody(t *testing.T) {
...
 			}, UpdateMask: &field_mask.FieldMask{Paths: []string{"single_nested"}}},
-			want: gw.ABitOfEverything{StringValue: "rabbit", SingleNested: &gw.ABitOfEverything_Nested{Amount: 456}},
+			want: gw.ABitOfEverything{
+				Int32Value:   42,
+				StringValue:  "rabbit",
+				SingleNested: &gw.ABitOfEverything_Nested{Amount: 456},
+			},
 		},
 		{
+			// N.B. This case passes the empty field mask to the UpdateV2 method so falls back to PUT semantics as per the implementeation
```
johanbrandhorst:
```suggestion
			// N.B. This case passes the empty field mask to the UpdateV2 method so falls back to PUT semantics as per the implementation.
```

examples/integration/integration_test.go
```diff
@@ -744,8 +756,10 @@ func TestABEPatchBody(t *testing.T) {
 			},
 		},
 		{
+			// N.B. This case passes the nil field mask to the UpdateV2 method so falls back to PUT semantics as per the implementeation
```
johanbrandhorst:
```suggestion
			// N.B. This case passes the nil field mask to the UpdateV2 method so falls back to PUT semantics as per the implementation.
```

examples/integration/integration_test.go
```diff
@@ -1617,3 +1631,85 @@ func TestRequestQueryParams(t *testing.T) {
...
+	}
+}
+
+// This benchmark requires the number of iterations be fairly low i.e. `-benchtime 100x`
+// otherwise you'll start to see errors like: `dial tcp: lookup localhost: no such host`
+func BenchmarkPatchABE(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		testABEPatch(b)
+	}
+}
```
johanbrandhorst: Hm, this benchmark actually fires requests at a real server? The cost of the fieldmask function is surely going to be lost in the variance of I/O cost? Could we make the benchmark run the transformation directly or is this due to the circular dependency problem?

william-plano-oxb: Yes, you're right, let me have a go at aiming this more directly at the function in question.

examples/integration/integration_test.go
johanbrandhorst: Looks like this is back again?

william-plano-oxb: D'oh, too much reverting.

runtime/fieldmask_test.go
```diff
@@ -67,21 +67,66 @@ func TestFieldMaskFromRequestBody(t *testing.T) {
...
+	var r *field_mask.FieldMask
+	var err error
+	for i := 0; i < b.N; i++ {
+		r, err = FieldMaskFromRequestBody(bytes.NewReader([]byte(input)))
+	}
+	if err != nil {
+		b.Error(err)
 	}
+	result = r
 }
```
johanbrandhorst: This is great, but a bit degenerate, could we just have another, more representative sample as well? Something like a flat resource with 5 or so fields, in a separate benchmark.

william-plano-oxb: Okay, I've added another one, for reference on my mac I get:
```
BenchmarkABEFieldMaskFromRequestBody-12            	    5000	    306614 ns/op
BenchmarkNonStandardFieldMaskFromRequestBody-12    	  100000	     19285 ns/op
```

william-plano-oxb:
Thank you for being so responsive to the contribution. It's a real pleasure to help out with a great project like this.

runtime/fieldmask.go
```diff
@@ -48,23 +47,3 @@ type fieldMaskPathItem struct {
 	// a generic decoded json object the current item to inspect for further path extraction
 	node interface{}
 }
-
-// CamelCaseFieldMask updates the given FieldMask by converting all of its paths to CamelCase, using the same heuristic
-// that's used for naming protobuf fields in Go.
-func CamelCaseFieldMask(mask *field_mask.FieldMask) {
```
RJPercival: This func being removed, and subsequently released as a patch release (v1.11.3), broke semantic versioning guarantees (our code no longer builds simply by updating from v1.11.2 to v1.11.3). Could you please avoid doing this in the future?

johanbrandhorst: Terribly sorry @RJPercival, you're right about this breaking users using a generator from version v1.11.2 and a runtime library with version v1.11.3. _However_, we do always encourage our users to ensure their generator versions and runtime library versions match up, so this shouldn't be a problem for users that can upgrade both at the same time. Sorry for the inconvenience caused.

RJPercival: I don't think this can be guaranteed with Go Modules. You may import a dependency that used an older version of the generator and Go Modules will happily use a newer runtime library, so long as it's at the same major version, since that's supposed to mean it's backwards compatible.

johanbrandhorst: This can (and should) be guaranteed by go modules. Define a [tool dependency](https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module) for the generator, and run `go install github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway` before invoking `protoc`.

johanbrandhorst: I agree that it's not ideal though, and I apologize for the inconvenience caused.

RJPercival: I don't think that guarantees what you think it guarantees. We already define a tool dependency, and have grpc-gateway listed in our go.mod file with a pre-v1.11.3 version number. However, according to [the version selection rules of Go Modules](https://github.com/golang/go/wiki/Modules#version-selection), anyone can import our module and require a newer v1 version of grpc-gateway in their go.mod file and it'll use that newer version (because it's supposed to be backwards-compatible), resulting in a build error. They then have to figure out that v1.11.3 of grpc-gateway broke them and pin to an earlier version.

Thanks for the apology. I don't mean to labour the point, but I just want to make it clear that violating semantic versioning rules **will** break your users and so we'd very much appreciate you avoiding it, thanks! :slightly_smiling_face:

PR #1053, opened 2019-10-01
Updates grpc proto deps

zachgersh:
Wanted to just bump the two most important dependencies but then had to make some modifications to the makefile so that everything would run correctly.

- protobuf is now at 1.3.2
- grpc at 1.24.0
- go 1.13.1 (built with this and it makes a difference for the mod below)

This has also bumped gazelle up one minor to take advantage of the fact it supports `go.mod` replace directives.

Tested with:
- codegen-swagger 2.2.2
- protoc: 3.9.2
- protoc-gen-go: 1.3.2


zachgersh:
@johanbrandhorst - can we get `jfbrandhorst/grpc-gateway-build-env:1.12` bumped to `1.13.1`? Should be all it takes to get this back into working order (now that I got the bazel bits fixed).

johanbrandhorst:
Please rebase on master.

PR #1056, opened 2019-10-02
use Go templates in protofile comments

Jeremytjuh:
Introduce GO Templates into proto comments to allow more advanced documentation such as from
external (markdown) files and automatically generated content based on the proto definitions.

**For example:**
The comments in this proto file:
```proto
syntax = "proto3";

package LoginTest;

service LoginService {

// Logout
// 
// {{.MethodDescriptorProto.Name}} is a call with the method(s) {{$first := true}}{{range .Bindings}}{{if $first}}{{$first = false}}{{else}}, {{end}}{{.HTTPMethod}}{{end}} within the "{{.Service.Name}}" service.
// It takes in "{{.RequestType.Name}}" and returns a "{{.ResponseType.Name}}".
// 
// ## {{.RequestType.Name}}
// | Field ID    | Name      | Type                                                       | Description                  |
// | ----------- | --------- | ---------------------------------------------------------  | ---------------------------- | {{range .RequestType.Fields}}
// | {{.Number}} | {{.Name}} | {{if eq .Label.String "LABEL_REPEATED"}}[]{{end}}{{.Type}} | {{fieldcomments .Message .}} | {{end}}  
//  
// ## {{.ResponseType.Name}}
// | Field ID    | Name      | Type                                                       | Description                  |
// | ----------- | --------- | ---------------------------------------------------------- | ---------------------------- | {{range .ResponseType.Fields}}
// | {{.Number}} | {{.Name}} | {{if eq .Label.String "LABEL_REPEATED"}}[]{{end}}{{.Type}} | {{fieldcomments .Message .}} | {{end}}  
rpc Logout (LogoutRequest) returns (LogoutReply) {
  option (google.api.http) = {
    post: "/v1/example/ekko"
    body: "*"
    additional_bindings {
      post: "/test/test/test"
    }
    };
  }
}

message LogoutRequest {
  // This field only contains this title
  string timeoflogout = 1;
  // This is the title
  //
  // This is the "Description" of field test
  // you can use as many newlines as you want
  //
  //
  // it will still format the same in the table
  int32 test = 2;
  // Array title
  repeated string stringarray = 3;
}

message LogoutReply {
  // Message that displays whether the logout was succesful or not
  string message = 1;
}
```

would generate the following documentation in Swagger UI:

image

or when imported in Postman:

image


johanbrandhorst:
Hi Jeremy, thanks for this PR. A few thoughts:

1. The goal here is to have proto comments that are automatically kept in-sync with the code definitions, I assume?
1. Would it be possible to consider adding support for the other swagger documentation fields too?
   See https://github.com/grpc-ecosystem/grpc-gateway/blob/master/examples/proto/examplepb/a_bit_of_everything.proto#L187 for an example.

Jeremytjuh:
Thanks for the fast reply Johan!

1. The main purpose is that you can automatically generate documentation from the proto definitions and that it allows importing larger portions of documentation from external files.
2. Good idea! Just added support for the other swagger documentation fields

johanbrandhorst:
> The main purpose is that you can automatically generate documentation from the proto definitions and that it allows importing larger portions of documentation from external files.
How does the second part of this work? I'm not sure what you mean.

Could we also add an example of this to one of our example files? We usually put an example of how to use a new feature into one of the proto files in the examples folder. Since this one has to be explicitly turned on (thanks for that by the way!), you'll need to add a new file and ensure it is generated with the appropriate flags. Let me know if you need any help with this.

Lastly, how do you feel about adding a small documentation page for this new feature? I think that would greatly help with adoption. Take a look in https://github.com/grpc-ecosystem/grpc-gateway/tree/master/docs/_docs for other pages and examples of what they can look like.

achew22:
Can you help me understand why it is preferable to have an unstructured description of the request response object in the swagger definition? Wouldn't it be better to improve the logic that pulls this information to fill in the request and response objects with more definition?

Jeremytjuh:
Hi Andrew,
I want to introduce Go templates to protofile comments not just to be able to read the Request and Response objects, but also to make it possible to list information about an object you otherwise would not be able to list. With this feature you would also be able to use just one (for example) table "template" for multiple objects.

Jeremytjuh:
> The main purpose is that you can automatically generate documentation from the proto definitions and that it allows importing larger portions of documentation from external files.
> How does the second part of this work? I'm not sure what you mean.

I have added a funcmap to the Go template which contains a function called "import". This function makes it possible to import documentation from an external file. For example a markdown file which contains a long detailed documentation. This way your protofile doesn't have to be flooded with documentation.


> Could we also add an example of this to one of our example files? We usually put an example of how to use a new feature into one of the proto files in the examples folder. Since this one has to be explicitly turned on (thanks for that by the way!), you'll need to add a new file and ensure it is generated with the appropriate flags. Let me know if you need any help with this.

I'm trying to make an example file right now, but I have no experience with Bazel. So I would really appreciate if you could help me with that.

> Lastly, how do you feel about adding a small documentation page for this new feature? I think that would greatly help with adoption.

Will do!

johanbrandhorst:
> I'm trying to make an example file right now, but I have no experience with Bazel. So I would really appreciate if you could help me with that.

Unfortunately, I, too, don't really undestand Bazel so the best I can say is "just do what the other example files do". @drigz or @achew22 might be able to help more.

drigz:
> > I'm trying to make an example file right now, but I have no experience with Bazel. So I would really appreciate if you could help me with that.
> 
> Unfortunately, I, too, don't really undestand Bazel so the best I can say is "just do what the other example files do". @drigz or @achew22 might be able to help more.

My first question is whether it makes sense to add a new flag for this. Will anyone want to disable templating? On the one hand, maybe it would break people who have `{{` `}}` in their comments already, but on the other hand that seems like it would not be common, and adding flags make maintenance harder. I'll leave that decision up to the two of you, though.

If you continue with the flag, you can follow the example of #944. You can test by [installing Bazel](https://docs.bazel.build/versions/master/install.html), then running: (assuming you call the new target `examplepb:examplepb_protoc_gen_swagger_template`)

```
bazel build //examples/proto/examplepb:examplepb_protoc_gen_swagger_template
```

and inspecting `bazel-bin/examples/proto/examplepb/a_bit_of_everything.swagger.json`.

protoc-gen-swagger/main.go
```diff
@@ -27,6 +27,7 @@ var (
 	allowRepeatedFieldsInBody  = flag.Bool("allow_repeated_fields_in_body", false, "allows to use repeated field in `body` and `response_body` field of `google.api.http` annotation option")
 	includePackageInTags       = flag.Bool("include_package_in_tags", false, "if unset, the gRPC service name is added to the `Tags` field of each operation. if set and the `package` directive is shown in the proto file, the package name will be prepended to the service name")
 	useFQNForSwaggerName       = flag.Bool("fqn_for_swagger_name", false, "if set, the object's swagger names will use the fully qualify name from the proto definition (ie my.package.MyMessage.MyInnerMessage")
+	useGoTemplate              = flag.Bool("use_go_templates", false, "if set, you can use GO templates in protofile comments")
```
johanbrandhorst:
```suggestion
	useGoTemplate              = flag.Bool("use_go_templates", false, "if set, you can use Go templates in protofile comments")
```

This is a somewhat common misconception; the name of the language is Go or go, not GO and not Golang.

johanbrandhorst:
Hi again @Jeremytjuh, please rebase on master to fix CI.

Jeremytjuh:
Hi @johanbrandhorst,
I just rebased on master and the tests still fail. I'm not sure what I should do now.

johanbrandhorst:
You'll also need to pull the latest generator image before regenerating the files again. See CONTRIBUTING.md. Sorry for the inconvenience 😅.

examples/proto/examplepb/use_go_template.proto
```diff
@@ -0,0 +1,86 @@
...
+// ## {{.ResponseType.Name}}
+// | Field ID    | Name      | Type                                                       | Description                  |
+// | ----------- | --------- | ---------------------------------------------------------- | ---------------------------- | {{range .ResponseType.Fields}}
+// | {{.Number}} | {{.Name}} | {{if eq .Label.String "LABEL_REPEATED"}}[]{{end}}{{.Type}} | {{fieldcomments .Message .}} | {{end}}  
+rpc Login (LoginRequest) returns (LoginReply) {
+  option (google.api.http) = {
+    post: "/v1/example/login"
+    body: "*"
+    };
+  }
```
johanbrandhorst: Nit: the indentation here looks a bit off.

docs/_docs/usegotemplates.md
```diff
@@ -0,0 +1,33 @@
...
+  -I$GOPATH/src \
+  -I$GOPATH/src/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis \
+  -I$GOPATH/src/github.com/grpc-ecosystem/grpc-gateway \
+  --go_out=plugins=grpc:. \
+  --grpc-gateway_out=logtostderr=true:. \
+  --swagger_out=logtostderr=true,use_go_templates=true:. \
+  *.proto 
+```
+
+For an example of a protofile which has Go templates enabled, [click here](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/examples/proto/examplepb/use_go_template.proto "Example protofile with Go template").
```
johanbrandhorst: This is a great start! Do you think you could just add a protofile snippet and some screen shots as well to show what this looks like when rendered to swagger?

johanbrandhorst:
Ah, the node_test error is because our runner doesn't have `bzip2` installed. Might be a change in the underlying docker image we use, but should be easily fixable. I will fix it on master.

johanbrandhorst:
`node_test` failure is gone.

drigz:
> ```
> ERROR: file 'examples/proto/examplepb/stream.swagger.json' is generated by these conflicting actions:
> Label: //examples/proto/examplepb:examplepb_protoc_gen_swagger, //examples/proto/examplepb:examplepb_protoc_gen_swagger_template
> ```

I think this is due to a limitation/bug of `protoc_gen_swagger()`: it doesn't prefix its output files with the rule name, so it can't be used twice in the same directory (the `single_file` flag is an exception as it changes the shape of the output). My first suggestion would be to put the example in an adjacent subdirectory (eg `examples/proto/templatepb`) to avoid blocking this PR on finding a better solution.

Jeremytjuh:
Hi @johanbrandhorst,

All tests seem to be working now. Let me know if you want me to change anything.
Also, could you please let @googlebot know you consent as an author.
Thanks in advance!

johanbrandhorst:
I consent to have my contributions to this PR merged into the repo under the relevant license. @achew22 could you make the bot happy please.

Jeremytjuh:
@johanbrandhorst @achew22 Are we good to go?

johanbrandhorst:
> @johanbrandhorst @achew22 Are we good to go?

I'm afraid we still need the CLA situation manually resolved by someone with authority. I've sent some reminders around. Sorry for the delay.

johanbrandhorst:
@googlebot I consent

johanbrandhorst:
@googlebot I consent.

ivucica:
I am confused by the internal debugging state of the tool,  but as I think I have the power to make this go through and as the consent has been granted, let's try.

johanbrandhorst:
Alright we got there in the end. Thanks for your contribution @Jeremytjuh, sorry for the arcane bot treatment. I should know better than to make actual github suggestions on PRs.

ivucica:
I didn't open a new internal bug for the tool, but I did fire off an email. [Someone should do something!](https://youtu.be/o63_eL8yOnA?t=37)

PR #1058, opened 2019-10-03
Swagger 2.4.8

zachgersh:
- had to update the makefile as the generated
file names have all moved around a little bit
- moved bazel forward to use go 1.13.1 as well
- test that was using a deep reflect to do equality
has changed to be per struct field equality (deep
reflect no longer works with the enum types).
- swagger 2.4.8 actually removed the need for resty
as they use the standard lib for http requests.

Biggest bit - you cannot generate both an `rpc` and a regular `service` in the same directory at this version. Why? In our example they both use an optional type which used to be simple a string but now `swagger-codegen` creates a struct. The struct has the same name for both versions and you get a type redeclaration (sad). 

Went and had a look at swagger-codegens repo for any open issues around this and saw 0. I don't think users actually do this so it seemed alright to remove it from our test generation.


johanbrandhorst:
> swagger 2.4.8 actually removed the need for resty as they use the standard lib for http requests.
This is the best news I've heard in weeks.

examples/integration/integration_test.go
```diff
@@ -48,6 +48,11 @@ func TestEcho(t *testing.T) {
 }
 
 func TestForwardResponseOption(t *testing.T) {
+	if testing.Short() {
+		t.Skip()
+		return
+	}
```
johanbrandhorst: Are these changes all necessary?

.circleci/Dockerfile
```diff
@@ -1,15 +1,18 @@
...
-ENV SWAGGER_CODEGEN_VERSION=2.2.2
+ENV SWAGGER_CODEGEN_VERSION=2.4.8
 RUN apt-get install -y openjdk-11-jre wget && \
     wget http://repo1.maven.org/maven2/io/swagger/swagger-codegen-cli/${SWAGGER_CODEGEN_VERSION}/swagger-codegen-cli-${SWAGGER_CODEGEN_VERSION}.jar \
     -O /usr/local/bin/swagger-codegen-cli.jar && \
     apt-get remove -y wget
-ENV SWAGGER_CODEGEN="java -jar /usr/local/bin/swagger-codegen-cli.jar"
+
+# Wrap the jar for swagger-codgen
+RUN echo '#!/bin/bash\n java -jar /usr/local/bin/swagger-codegen-cli.jar "$@"' > /usr/local/bin/swagger-codegen && \
```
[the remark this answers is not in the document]
zachgersh: went ahead and fixed this 💃 plus did some more cleanup of the dockerfile in general. sorry you have to rebuild again ;)

examples/integration/integration_test.go
zachgersh: Necessary, sort of? I think it's nice to have just in case people do want to skip the integration tests (this integration suite runs so fast it doesn't seem to actually matter). We had the pattern partially applied previously (and I missed some when I tried to bring it fully up to date).

zachgersh:
Would need one more dockerfile rebuild here.

.circleci/Dockerfile
```diff
@@ -1,32 +1,35 @@
...
-    wget http://repo1.maven.org/maven2/io/swagger/swagger-codegen-cli/${SWAGGER_CODEGEN_VERSION}/swagger-codegen-cli-${SWAGGER_CODEGEN_VERSION}.jar \
-    -O /usr/local/bin/swagger-codegen-cli.jar && \
-    apt-get remove -y wget
-ENV SWAGGER_CODEGEN="java -jar /usr/local/bin/swagger-codegen-cli.jar"
+ENV SWAGGER_CODEGEN_VERSION=2.4.8
+RUN wget http://repo1.maven.org/maven2/io/swagger/swagger-codegen-cli/${SWAGGER_CODEGEN_VERSION}/swagger-codegen-cli-${SWAGGER_CODEGEN_VERSION}.jar \
+    -O /usr/local/bin/swagger-codegen-cli.jar
+
+# Wrap the jar for swagger-codgen
+RUN echo -e '#!/bin/bash\n java -jar /usr/local/bin/swagger-codegen-cli.jar "$@"' > /usr/local/bin/swagger-codegen && \
```
johanbrandhorst:
```suggestion
RUN echo -e '#!/bin/bash\njava -jar /usr/local/bin/swagger-codegen-cli.jar "$@"' > /usr/local/bin/swagger-codegen && \
```
What monster puts a single space at the start of a new line.

johanbrandhorst:
Lets merge this, I'll do a rebuild of the image on master.

Issue #441, opened 2017-08-06
Feature Request: OpenAPI 3 support


c4milo:
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md

achew22:
@c4milo, Can you elaborate on what it is you're interested in that isn't available today?

c4milo:
@achew22, feature wise, nothing. What's concerning me right now is that https://github.com/swagger-api/swagger-js/tree/2.x is not longer active. Most development effort is going towards supporting OpenAPI v3.

zxy198717:
+1

johanbrandhorst:
The OpenAPI 3.0 spec has added support for [Authentication (JWT, Oath, HTTP Basic)](https://github.com/swagger-api/swagger-ui/pull/3780), which is something that is missing from 2.0. I think this would make it worthwhile to migrate to generating 3.0 compliant definitions.

jsw:
Also `oneof`?

tamalsaha:
I suppose https://github.com/go-openapi/spec3 will be needed first.

veqryn:
OpenAPI v3.0.0 adds support for specifying multiple servers (hosts).

jones77:
OpenAPI 3.0 adds support for nullable types, 2.0 doesn't support them: https://github.com/swagger-api/swagger-editor/issues/1364#issuecomment-309530250

johanbrandhorst:
https://github.com/grpc-ecosystem/grpc-gateway/issues/900 is a real issue that would be covered by this.

ivucica:
FTR I *will* be *very* happy to review forks of the swagger generator, or elegant refactors of the swagger generator that allow for both openapiv2 and openapiv3.

misrab:
+1

lukaskellerstein:
+1

zachgersh:
Hey everyone,

So we are actually doing this (at least I am going to try). I pushed the first bits of OPENAPI v3 support and attached the WIP PR to this issue for all to review.

I'd love everyone to go have a look at the proto and WIP if they can. I WANT YOUR FEEDBACK ON THIS. Would love to know what sort of fields are required missing from the next version of this generator.

ivucica:
@zachgersh I still have the notification email for your comment marked as 'unread' in my inbox, this is exciting and I hope to take a look.

JugrajSripalJain:
[...]

Issue #675, opened 2018-06-19
cleanup: rename 'swagger' references to 'openapi'


tmc:
The OpenAPI community has largely moved on from the previous name and we should follow suit.

tmc:
@ivucica do you have thoughts here. The "grpc.gateway.protoc_gen_swagger" strings in protos will be a concern here.

ivucica:
I am comfortable with a change, aside from renaming the `Swagger` object, which is still the root object in OpenAPI v2.

On the other hand, how about this happens for the v3 generator (which may or may not coexist with the v2 generator)?

The problem you cite *might* apply only to people who want to customize their OpenAPI v2 `.json` by manually specifying fields that are not otherwise describable in proto format. If the rename swagger->openapi happens when they elect to use a new generator, e.g., `protoc-gen-openapiv3`, then it is up to them to include the new proto.

Finally, don't hold me to this, but I think package name for the options file might be a non-sequitur anyway; I don't think it applies to the wire format.

So what do you think: we don't do a rename, and instead we ask that the rename happens whenever someone contributes a v3 generator?

philsturgeon:
Hey there, so folks closed down #872 because duplicate, so let me apply some glue and post some ccomments from that issue:

> I've not yet found something that we can express in the Google http annotations that don't fit nicely into OpenAPI v2.

OpenAPI v2 didn't have the concept of nullable, which is a big pain in the backside for a lot of folks, and seeing as everything in protobuf is optional. OpenAPI v3 has that, so upgrading support would be a pretty good idea for that alone. 

> we'd probably still want to make sure we produce a protoc-gen-swagger 

Yeah I mean sure make a placeholder package which sends people off to the right package, that's pretty standard. 

As for SEO concerns, these days people are starting to understand that "Swagger" is just the name of the toolsuit from Smartbear, nothing else. Most modern projects will write "OpenAPI (formerly known as Swagger)" once somewhere in their intro, then just talk about OpenAPI from there on out. 

If the concerns is SEO, I created https://openapi.tools to help push folks in the right direction too and it's getting 1000 unique visitors a day, so folks are finding it. 

<img width="712" alt="Screen Shot 2019-07-01 at 18 03 34" src="https://user-images.githubusercontent.com/67381/60450718-b9b95080-9c2a-11e9-9380-a705a195d1f0.png">

[...]

PR #1059, opened 2019-10-04
WIP: openapi v3 support

zachgersh:
This is 100% a work in progress and is meant to be available for the rest of the community to view the progress towards supporting OPENAPI v3 mentioned in #441 

## Summary of the major changes to the proto between v2 and v3:

- `Swagger` top level key is gone, it's replace by `OpenAPI`
- Removal of all of the reserved fields that were placeholders for fields we may want to add later but never actually added (was confusing to look at and unused). A YAGNI approach
- the proto and annotations packages now end in `openapi` (bye bye swagger)
- `Servers` is a new top level key
- `extensions` have been added to all messages that support them as per the spec
- `Annotations` going to basically stay the same (missing tags right now)
- A tiny sprinkle of new fields were added like `Discriminator`
- `SecurityScheme` has seen some slight tweaks around `types` and `in` that are supported as field options.
- Most comments have been removed as it currently makes the proto a bit easier to read (they will go back in soon).

## Questions
- `Components` as a a Field has been added but it is missing quite a few fields that we could support (which do we want)

As a note - all of the fields that we supported in v2 were moved to v3 (seems like the base level of support we would need). If you see a field that is missing please ping this.

### DIFF
--- a/protoc-gen-swagger/options/annotations_v3.proto
+++ b/protoc-gen-swagger/options/annotations_v3.proto
@@ -1,0 +1,40 @@
+syntax = "proto3";
+
+package grpc.gateway.protoc_gen_openapi.options;
+
+option go_package = "github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger/options";
+
+import "protoc-gen-swagger/options/openapiv3.proto";
+import "google/protobuf/descriptor.proto";
+
+extend google.protobuf.FileOptions {
+  // ID assigned by protobuf-global-extension-registry@google.com for grpc-gateway project.
+  //
+  // All IDs are the same, as assigned. It is okay that they are the same, as they extend
+  // different descriptor messages.
+  OpenAPI openapiv3 = 1042;
+}
+
+extend google.protobuf.MethodOptions {
+  // ID assigned by protobuf-global-extension-registry@google.com for grpc-gateway project.
+  //
+  // All IDs are the same, as assigned. It is okay that they are the same, as they extend
+  // different descriptor messages.
+  Operation openapiv3_operation = 1042;
+}
+
+extend google.protobuf.MessageOptions {
+  // ID assigned by protobuf-global-extension-registry@google.com for grpc-gateway project.
+  //
+  // All IDs are the same, as assigned. It is okay that they are the same, as they extend
+  // different descriptor messages.
+  Schema openapiv3_schema = 1042;
+}
+
+extend google.protobuf.FieldOptions {
+  // ID assigned by protobuf-global-extension-registry@google.com for grpc-gateway project.
+  //
+  // All IDs are the same, as assigned. It is okay that they are the same, as they extend
+  // different descriptor messages.
+  JSONSchema openapiv3_field = 1042;
+}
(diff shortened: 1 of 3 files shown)


ivucica:
Looks like I didn't mention this in #441 (I thought I did, but maybe that's somewhere else).

Given git has no concept of tracking a file copy ("content copy tracking ought to be enough for everyone"), for purposes of review, I am ok with modifying `protoc-gen-swagger`. This will allow comparison of what the changes are compared to old code.

However, once reviewed, I'd like to see this *copied* to `protoc-gen-openapiv3`.

I hope to take a look at this this week.

protoc-gen-swagger/options/annotations_v3.proto
```diff
@@ -0,0 +1,40 @@
...
+
+import "protoc-gen-swagger/options/openapiv3.proto";
+import "google/protobuf/descriptor.proto";
+
+extend google.protobuf.FileOptions {
+  // ID assigned by protobuf-global-extension-registry@google.com for grpc-gateway project.
+  //
+  // All IDs are the same, as assigned. It is okay that they are the same, as they extend
+  // different descriptor messages.
+  OpenAPI openapiv3 = 1042;
```
ivucica: Despite what is may seem the comment says, it's actually not encouraging keeping the same ID for two completely different messages. This is particularly true for v2 to v3 transition; it's not only incompatible, but there's also no reason why two generators would not coexist.

protoc-gen-swagger/options/openapiv3.proto
```diff
@@ -0,0 +1,207 @@
+syntax = "proto3";
+
+package grpc.gateway.protoc_gen_openapi.options;
+
+option go_package = "github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger/options";
+
+import "google/protobuf/struct.proto";
+
+message OpenAPI {
```
ivucica: I would really like to see many more comments. OpenAPI itself can be a good source of comments.

zachgersh:
@ivucica - I am totally happy to do that and I think it will be easier to review for all of us. I think two separate binaries gives users choice and means we don't have to clutter it up with any sort of flags etc to switch back and forth between each version.

I assume you are still reviewing the proto? I wanted to see if anyone caught anything absolutely of base with the `v3` version of said proto. I am sure it will evolve as I start to work through the actual protoc-gen-swagger binary bits.

ivucica:
> @ivucica - I am totally happy to do that and I think it will be easier to review for all of us. I think two separate binaries gives users choice and means we don't have to clutter it up with any sort of flags etc to switch back and forth between each version.
> 
> I assume you are still reviewing the proto? I wanted to see if anyone caught anything absolutely of base with the `v3` version of said proto. I am sure it will evolve as I start to work through the actual protoc-gen-swagger binary bits.

I did not review the proto in great detail, as I trust you are actively looking at OpenAPI v3 specification. I find `template.go` changes to be way more interesting than the proto changes.

Either way you should sort out upstream a new ID for the openapiv3 annotations. The process is now PR-based instead of fire-off-an-email-based.

(FTR the other issue where I commented about keeping the `protoc-gen-swagger` name is #675)

johanbrandhorst:
I'm gonna have to close this as stale - thanks for the effort @zachgersh, I know priorities often change.

zachgersh:
@johanbrandhorst yeah, sorry I didn't circle back to this / feel bad 😢

PR #1062, opened 2019-10-13
adding build-scoping for builds on earlier go versions (<1.12)

### LINKED ISSUE #1061: protoc-gen-swagger/genswagger does not build on go1.11 and earlier versions

```html
<!-- Example steps
1.  I grab my catapult
2.  I load it with lettuce
3.  Press the fire button
4.  It falls over
-->
```

1. Using `go1.11`
1. import `github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger/genswagger`
1. use `genswagger.New()` in a `main` function
1. `go build .`
1. package fails to build because `strings` package does not have function `ReplaceAll`.

```html
<!-- Example answer
1.  It would have rained lettuce from the sky bringing forth a cuddly army of bunnies we could
    play with
-->
```

`go build` should exit with exit code `0`.

```html
<!-- Example answer
Evil wizards are hoarding the bunnies and don't want to share. The wizards are casting 
lettuce protection spells so the catapult won't work.
-->
```

This line uses `strings.ReplaceAll`, which was only introduced in `go1.12`.
https://github.com/grpc-ecosystem/grpc-gateway/blob/e652ba06e9067ef41c199af59b9c6c67724850d4/protoc-gen-swagger/genswagger/generator.go#L71

marcusljx:
Fixes #1061 

This PR moves the helper function using `strings.ReplaceAll` to a new file which has build tags for `go.1.12`, limiting it to `go1.12+` builds. Also added a file for `go` versions lesser than `1.12`, using `strings.Replace` in place of `strings.ReplaceAll`.

### DIFF
--- a/protoc-gen-swagger/genswagger/BUILD.bazel
+++ b/protoc-gen-swagger/genswagger/BUILD.bazel
@@ -7,6 +7,8 @@ go_library(
     srcs = [
         "doc.go",
         "generator.go",
+        "helpers.go",
+        "helpers_go111_old.go",
         "template.go",
         "types.go",
     ],
--- a/protoc-gen-swagger/genswagger/generator.go
+++ b/protoc-gen-swagger/genswagger/generator.go
@@ -67,10 +67,6 @@ func mergeTargetFile(targets []*wrapper, mergeFileName string) *wrapper {
 	return mergedTarget
 }
 
-func fieldName(k string) string {
-	return strings.ReplaceAll(strings.Title(k), "-", "_")
-}
-
 // Q: What's up with the alias types here?
 // A: We don't want to completely override how these structs are marshaled into
 //    JSON, we only want to add fields (see below, extensionMarshalJSON).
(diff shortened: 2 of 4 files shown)


marcusljx:
@googlebot I signed it!

johanbrandhorst:
Hi @marcusljx, this looks great. I think you need to run the generator for the bazel files. See `CONTRIBUTING.md`.

achew22:
Or if you want to do it manually, [this file](https://github.com/marcusljx/grpc-gateway/blob/ce2c1ac56d1778b4a4ee791eed82b9b4067b08e4/protoc-gen-swagger/genswagger/BUILD.bazel#L10) is a list of the `.go` files used in the build

marcusljx:
Thanks @johanbrandhorst and @achew22! I ended up going with the manual route, as I kept getting a `make:generate command not found` error with the first command from the `CONTRIBUTING` file (`docker run ....`).

achew22:
@marcusljx, thanks so much for fixing up the BUILD files! This looks good to me. I'm going to let @johanbrandhorst take one more look at it and merge it but looks good on my front.

johanbrandhorst:
Thanks for your contribution!

Issue #1066, opened 2019-10-21
Nested message names can collide silently in swagger generation


johnchildren:
Generated swagger files can silently have name collisions between message names when messages are nested. For example:

```protobuf
syntax = "proto3";

message FooBar {
  string value = 1;
}

message Foo {
  message Bar {
    string value = 1;
  }
  Bar bar = 1;
}

service FooBarService {
  rpc BarYourFoos(FooBar) returns (Foo);
}
```

Here you will _sometimes_ end up with a different definition of `FooBar` depending on how two go maps are presumably being unioned? For example:

```json
{
  "swagger": "2.0",
  "info": {
    "title": "test.proto",
    "version": "version not set"
  },
  "schemes": [
    "http",
    "https"
  ],
  "consumes": [
    "application/json"
  ],
  "produces": [
    "application/json"
  ],
  "paths": {},
  "definitions": {
    "Foo": {
      "type": "object",
      "properties": {
        "bar": {
          "$ref": "#/definitions/FooBar"
        }
      }
    },
    "FooBar": {
      "type": "object",
      "properties": {
        "value": {
          "type": "string"
        }
      }
    }
  }
}
```

Which contains a ref to the wrong field type. This behavior can happen across imports so if you have one nested message in another file you can end up with unexpected random collisions which lead to different file output.

This was causing a CI pipeline to randomly fail internally as one message field in one file randomly collided with another.

I would expect that a collision like this should instead error or have sufficient namespacing so as to be avoided.

Have prepared an example repo with commited generated files, you can clone it and run:

```
prototool generate
```

To reproduce and if you use nix a shell is provided to pin against the latest release.

https://github.com/johnchildren/grpc-gateway-namespacing-issue-example

johanbrandhorst:
> This was causing a CI pipeline to randomly fail internally as one message field in one file randomly collided with another.

This sounds incredibly annoying, sorry that this happened to you! Thanks for reporting it. Would you be interested in trying to fix this? I would accept something like a detection that gives an error at generate time, but ideally having some way to namespace the two would be best, of course.

johnchildren:
Sure thing, I don't think I have successfully managed to make this completely reproducible, but I think a good first step would be trying to detect name clashes like this just for nested messages and emitting an error. Hopefully this should be reasonably easy to implement and easily testable enough and I can look at doing more if this does not adequately prevent this problem from happening again.

I can take a look at trying to do this in the next few days depending on how busy I am.

[...]

PR #1067, opened 2019-10-23
Error on nested namespacing collisions for message

johnchildren:
Nested messages can currently have name collisions with un-nested
messages, this commit adds an error if a collision occurs, though
hopefully this should be a reasonably rare occurrence.

In the case of a message Bar with a nested field Foo colliding with a
message BarFoo the error message will look like:

```
<input>:1:1:protoc-gen-swagger: message schema name collision: testBarFoo would be defined twice due to message nesting
```

It is possible there may be other causes for collision but I am
reasonably sure this is the only way it can happen without a failure in
protoc first.

### DIFF
--- a/protoc-gen-swagger/genswagger/template.go
+++ b/protoc-gen-swagger/genswagger/template.go
@@ -17,9 +17,9 @@ import (
 	"github.com/golang/glog"
 	"github.com/golang/protobuf/jsonpb"
 	"github.com/golang/protobuf/proto"
-	structpb "github.com/golang/protobuf/ptypes/struct"
 	pbdescriptor "github.com/golang/protobuf/protoc-gen-go/descriptor"
 	gogen "github.com/golang/protobuf/protoc-gen-go/generator"
+	structpb "github.com/golang/protobuf/ptypes/struct"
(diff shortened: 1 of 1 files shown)


johnchildren:
Should prevent https://github.com/grpc-ecosystem/grpc-gateway/issues/1066 from occuring

johanbrandhorst:
Thanks for the contribution! Have you tested this branch against your error reproduction case?

johnchildren:
Yes! The error message in the config is actually from my reproduction case (slightly modified).

However it looks like some of the projects test cases would fail if this change was added so I will have to revise it.

johnchildren:
Trying to figure out why this is failing and I'm not sure, but it may be that there are cases where a definition is passed into this function twice, perhaps when a file is imported twice? Anyway I think perhaps there isn't a way for me to resolve this easily so would recommend closing this PR and reconsidering.

It may be that the only real solution is to add a delimiter unavailable in protobuf message names to the FQMN, but this would be a big breaking change and may break a lot of people's APIs. Alternatively this error could be downgraded to just a warning?

PR #1075, opened 2019-11-04
Support creating a Swagger def with no schemes

### LINKED ISSUE #1069: protoc-gen-swagger: support generating a Swagger definition with no schemes
It would be nice if `grpc-gateway` supported generating a Swagger definition with no `schemes` key. According to the [openapiv2 spec](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md) this means the following:

> If the schemes is not included, the default scheme to be used is the one used to access the Swagger definition itself.

This is currently not possible because the [default schemes is `["http","https"]`](https://github.com/grpc-ecosystem/grpc-gateway/blob/c3787b4d95d15f4c89484b44e108a6640c1943ca/protoc-gen-swagger/genswagger/template.go#L1021) and this default is only overridden when there is at [least one scheme sepecified](https://github.com/grpc-ecosystem/grpc-gateway/blob/c3787b4d95d15f4c89484b44e108a6640c1943ca/protoc-gen-swagger/genswagger/template.go#L1119).

I'm willing go contribute a change for this but I'd like some guidance on how to proceed. I can think of a few ways to do this:

1. Filter out [`UNKNOWN`](https://github.com/grpc-ecosystem/grpc-gateway/blob/c3787b4d95d15f4c89484b44e108a6640c1943ca/protoc-gen-swagger/options/openapiv2.proto#L27) values and thus make the list `[UNKNOWN]` mean no schemes. This is kind of hacky. Arguably it would be better for `protoc-gen-swagger` to error out if it hits `UNKNOWN` since it is the fallback for unknown enum values. Right now `UNKNOWN` will be written to the definition as "unknown" which is not valid.
2. Add a new enum value (e.g. `EMPTY`) that explicitly means there are no schemes.
3. Use another field to specify that `schemes` is set to disable the default value. E.g. (`bool schemes_set = 105`).

Or if there is another way I didn't think of let me know!

ailurarctos:
Make it so if you do not specify the schemes then no schemes
will be present in the generated Swagger definition. The
OpenAPIv2 spec says this means to use the same scheme that
was used to access the Swagger definition itself.

Fixes #1069

### DIFF
--- a/examples/clients/echo/api/swagger.yaml
+++ b/examples/clients/echo/api/swagger.yaml
@@ -4,9 +4,6 @@ info:
   description: "Echo Service API consists of a single service which returns\na message."
   version: "version not set"
   title: "Echo Service"
-schemes:
-- "http"
-- "https"
 consumes:
 - "application/json"
 produces:
--- a/examples/clients/echo/configuration.go
+++ b/examples/clients/echo/configuration.go
@@ -60,7 +60,7 @@ type Configuration struct {
 
 func NewConfiguration() *Configuration {
 	cfg := &Configuration{
-		BasePath:      "http://localhost",
+		BasePath:      "https://localhost",
 		DefaultHeader: make(map[string]string),
 		UserAgent:     "Swagger-Codegen/1.0.0/go",
 	}
--- a/examples/clients/responsebody/api/swagger.yaml
+++ b/examples/clients/responsebody/api/swagger.yaml
@@ -3,9 +3,6 @@ swagger: "2.0"
 info:
   version: "version not set"
   title: "examples/proto/examplepb/response_body_service.proto"
-schemes:
-- "http"
-- "https"
 consumes:
 - "application/json"
 produces:
--- a/examples/clients/responsebody/configuration.go
+++ b/examples/clients/responsebody/configuration.go
@@ -60,7 +60,7 @@ type Configuration struct {
 
 func NewConfiguration() *Configuration {
 	cfg := &Configuration{
-		BasePath:      "http://localhost",
+		BasePath:      "https://localhost",
 		DefaultHeader: make(map[string]string),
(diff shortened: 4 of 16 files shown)


achew22:
I remember from earlier threads that this has the affect that some generators will default to `http`. Is that something we want to blindly opt people into? What would you think about just defaulting it to `[]string{"https"}` and leaving people to override it if they want something else?

cc @johanbrandhorst I know you have more context on this

ailurarctos:
@achew22 unfortunately if the default is not empty there is currently no way to override the default with an empty value. I mentioned some alternatives to changing the default in https://github.com/grpc-ecosystem/grpc-gateway/issues/1069#issue-512903087 but after further discussion with @johanbrandhorst it seemed like changing the default might be okay after all.

> I remember from earlier threads that this has the affect that some generators will default to `http`.

Yes, I looked into three code generators in https://github.com/grpc-ecosystem/grpc-gateway/issues/1069#issuecomment-549091129 and unfortunately `go-swagger` will change from using HTTPS to HTTP. This change will also cause `swagger-codegen` to change from HTTP to HTTPS while `openapi-generator` will continue to use HTTP as it does with the current default.

johanbrandhorst:
I think, on balance, it is not our responsibility to limit the functionality of our project in order to avoid potential security shortcomings of other projects. The rules we should be bound by for the swagger generator should be the swagger spec. We'll have to exercise judgment, of course, but I think we can get the best of both worlds by contributing safer defaults to the downstream projects, too. Do you agree @achew22? I'm happy to discuss further, the security of our users is of course paramount.

achew22:
If the analysis has been done then I'm comfortable with it.

@ailurarctos thanks for your contribution and for putting up with me :wink:!

ailurarctos:
Thanks very much @johanbrandhorst and @achew22!

Issue #1084, opened 2019-11-13
protoc-gen-swagger: json_names_for_fields=true does not respect json_name for path parameters


brocaar:
I'm using the `json_names_for_fields=true` option and am seeing inconsistent results, specifically between the URL parameter names and the names in the request / response bodies. Example:

```proto
service DeviceService {
    rpc Get(GetDeviceRequest) returns (GetDeviceResponse) {
        option (google.api.http) = {
            get: "/api/devices/{dev_eui}"
        };
    }

message GetDeviceRequest {
    // Device EUI (HEX encoded).
    string dev_eui = 1 [json_name = "devEUI"];
}
```

I believe this is because the fieldnames in the request / response body are get using the `f.GetJsonName()` (e.g. https://github.com/grpc-ecosystem/grpc-gateway/blob/master/protoc-gen-swagger/genswagger/template.go#L342) where the URL parameter names are get using the `lowerCamelCase` function (e.g. https://github.com/grpc-ecosystem/grpc-gateway/blob/master/protoc-gen-swagger/genswagger/template.go#L772).

The `json_name` is respected by the `GetJsonName()` function (e.g. `[json_name = "devEUI"]`, but not by the `lowerCamelCase` function.

The generated Swagger JSON looks like:

```json
    "/api/devices/{devEui}": {
      "get": {
        "summary": "Get returns the device matching the given DevEUI.",
        "operationId": "Get",
        "responses": {
          "200": {
            "description": "A successful response.",
            "schema": {
              "$ref": "#/definitions/apiGetDeviceResponse"
            }
          }
        },
        "parameters": [
          {
            "name": "devEui",
            "description": "Device EUI (HEX encoded).",
            "in": "path",
            "required": true,
            "type": "string"
          }
        ],
        "tags": [
          "DeviceService"
        ]
      },
```

I would have expected:

```json
    "/api/devices/{devEUI}": {
      "get": {
        "summary": "Get returns the device matching the given DevEUI.",
        "operationId": "Get",
        "responses": {
          "200": {
            "description": "A successful response.",
            "schema": {
              "$ref": "#/definitions/apiGetDeviceResponse"
            }
          }
        },
        "parameters": [
          {
            "name": "devEUI",
            "description": "Device EUI (HEX encoded).",
            "in": "path",
            "required": true,
            "type": "string"
          }
        ],
        "tags": [
          "DeviceService"
        ]
      },
```

[...]

PR #1085, opened 2019-11-14
Support reserved json name and add tests

xin-au:
Support reserved JSON name and add tests.
Like: when a reserved JSON name is specified in protobuf, then the swagger doc will generate this reserved JSON name instead of a lower case camel variable name. 
Ref: https://github.com/grpc-ecosystem/grpc-gateway/issues/1084


xin-au:
@johanbrandhorst , Thank you so much for reviewing! Anytime!

PR #1099, opened 2019-12-16
Add an empty package to `grpc-gateway/third_party/googleapis` 

jingweno:
Add an empty package to `grpc-gateway/third_party/googleapis` to allow vendoring the folder. For example, a `tools.go` file may look like the followings:

```go
package your_package

import (
     _ "github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis"
)
```

And you vendor this package by running `go mod vendor`. To generate grpc gateway code, you can run with the followings:

```
protoc -I. \
   -I./vendor/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis \
   --plugin=protoc-gen-grpc=grpc_ruby_plugin \
   path/to/your_service.proto
```

### DIFF
--- a/third_party/googleapis/package.go
+++ b/third_party/googleapis/package.go
@@ -1,0 +1,18 @@
+package googleapis
+// This is an empty package that allows vendoring the folder with `go mod`.
+// For example, a `tools.go` file may look like the followings:
+//
+// package your_package
+//
+// import (
+//     _ "github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis"
+// )
+//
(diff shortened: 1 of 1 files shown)


johanbrandhorst:
This is an intriguing suggestion, but I'm hesitant to do this before we understand if this sort of thing emerges as a common way of dealing with the problem of versioning non-go files with modules. The pattern you've used here is normally referred to as a [tool dependency](https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module), which I would argue this is _not_. Our currently recommended way of users doing this is to manually version the proto file dependencies.

Issue #1101, opened 2019-12-16
grpc-gateway is incompatible with --incompatible_load_proto_rules_from_bzl


Yannic:
Hello maintainers of grpc-gateway 👋,

The Bazel team is in progress of migrating the native Protobuf rules to Starlark. As a first step towards this goal, starting with Bazel 3.0, all Protobuf rules will require explicit load statements.

You can use the following [buildifier](https://github.com/bazelbuild/buildtools/tree/master/buildifier) command to apply most of the required migrations for you:
```shell
buildifier --lint=fix --warnings=native-proto $(find . -name "BUILD" -o -name "BUILD.bazel" -o -name "*.bzl")
```

For further information, see https://github.com/bazelbuild/bazel/issues/8922 or ping me.

Thanks,
Yannic

PR #1102, opened 2019-12-16
Explicitly load rules_proto rules

achew22:
Fixes: #1101

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -3,11 +3,51 @@ workspace(name = "grpc_ecosystem_grpc_gateway")
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 http_archive(
-    name = "io_bazel_rules_go",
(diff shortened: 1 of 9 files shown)


johanbrandhorst:
> ERROR: error loading package 'internal': Label '//third_party/bazel_rules/rules_proto/proto:defs.bzl' is invalid because 'third_party/bazel_rules/rules_proto/proto' is not a package; perhaps you meant to put the colon here: '//:third_party/bazel_rules/rules_proto/proto/defs.bzl'?

achew22:
Sorry @johanbrandhorst I started this at work but then had to do work things. Will handle "soon"

drigz:
IIUC it's not possible to fix this as a rules_go update is not available. I pinged the reporter: https://github.com/googlecloudrobotics/core/issues/24#issuecomment-566999976

achew22:
@drigz, I don't think we depend on the codepath in `rules_go` that's also needs to be upgraded.

@johanbrandhorst PTAL (sorry it's all greek to you)

PR #1103, opened 2019-12-19
Add --omit_package_doc flag

RJPercival:
Makes it possible to omit the package comment from generated code. This is useful if the generated code is added to an existing package that already has package documentation (the generated package comment might be misleading in that case).

For an example of a package where this would be useful, see https://pkg.go.dev/github.com/google/trillian?tab=doc.


RJPercival:
I haven't added any tests for this yet, since I want to confirm it's something that could be added first (and I'm not sure this is the best way to implement it - I'm not familiar with the codebase). Please let me know whether it's worth completing this PR.

johanbrandhorst:
Hi Rob! Thanks for taking an interest in this project. Generally these generated files go into the same folder as the generated go files, and nothing else. Help me understand, are you saying that your use case includes manually adding package comments to this otherwise generated package?

RJPercival:
In the case of https://github.com/google/trillian, the generated grpc-gateway code lives in the root directory, alongside the generated protobuf code. This has the unfortunate effect of making the subtitle of "github.com/google/trillian" on https://pkg.go.dev/search?q=trillian show "Package trillian is a reverse proxy." This is misleading, since it suggests that Trillian is a reverse proxy, when it isn't (that's just what the grpc-gateway code does, and it's mostly irrelevant to users). We'd prefer to add a doc.go file to our root directory and write our top-level package comment there, but right now there's no way to suppress the grpc-gateway package comment. We could remove the package comment after code generation, but that'd be onerous (and clash with our presubmit checks that ensure checked-in generated code is pristine).

RJPercival:
As an example, [gomock has a --write_package_comment flag](https://github.com/golang/mock/blob/v1.3.1/mockgen/mockgen.go#L52) that does what we want for its generated code. We'd just like an equivalent for grpc-gateway.

johanbrandhorst:
Thanks for the explanation, I think what's confusing me is why the generated code lives in the root of the repo (and thus producing this confusing and misleading documentation for the root package). Not saying I'm opposed to merging this change, but why can't you move the generated code to another directory?

RJPercival:
In the end we've just decided to remove our use of grpc-gateway entirely.

PR #1112, opened 2020-01-18
#1109 Inline streamingDefinitions in generated swagger to better support codegen tools

### LINKED ISSUE #1109: Streaming responses are put in `x-stream-definitions` rather than `/components/schema`, which tools do not support
The gateway wraps streaming responses in a "result" object
```ts
{
  result?: <T>;
  error?: RuntimeStreamError;
}
```

In the generated swagger json, these intermediary wrappers are put into `/x-stream-definitions`. It was decided to put them there because there could potentially be a name conflict by putting them into `/components/schema` (see https://github.com/grpc-ecosystem/grpc-gateway/pull/850)

However, it's uncommon for codegen tools to have support for extension fields: for instance
- https://github.com/OpenAPITools/openapi-generator
- https://github.com/contiamo/restful-react

Since these schemas are not an extension to the openapi spec itself, I think a better developer experience would be to place them in `/components/schema`, and give them a prefixed name that is unlikely to collide. I.e `StreamResponseMyResponseType`.

seanlaff:
Fixes #1109

### DIFF
--- a/examples/proto/examplepb/stream.swagger.json
+++ b/examples/proto/examplepb/stream.swagger.json
@@ -18,7 +18,16 @@
           "200": {
             "description": "A successful response.(streaming responses)",
             "schema": {
-              "$ref": "#/x-stream-definitions/examplepbABitOfEverything"
+              "type": "object",
+              "properties": {
+                "result": {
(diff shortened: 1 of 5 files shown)


seanlaff:
@googlebot I signed it!

PR #1114, opened 2020-01-18
#1113 Fix broken parsing of camelCase oneof

### LINKED ISSUE #1113: Gateway does not parse oneof types correctly when using camelCase
Enums like 

```proto
message GetRolesRequest {
  oneof get {
    // roles's project
    string project_id = 4;
    // cluster roles
    bool cluster = 5;
  }
}
```

will not be parsed correctly through the gateway with the `camelCase` flag. `project_id` succeeds but `projectId` does not.

The issue looks to be here :
https://github.com/grpc-ecosystem/grpc-gateway/blob/7f57073b1543f1b5a5077437aa57b92b6cfc04a6/runtime/query.go#L120-L128

There's a map of `oneof`s that we loop over, however they key of the map is the name with the underscore, so given a camelCase param, it will never find it.

seanlaff:
Fixes #1113

### DIFF
--- a/runtime/query.go
+++ b/runtime/query.go
@@ -117,14 +117,16 @@ func fieldByProtoName(m reflect.Value, name string) (reflect.Value, *proto.Prope
 	props := proto.GetProperties(m.Type())
 
 	// look up field name in oneof map
-	if op, ok := props.OneofTypes[name]; ok {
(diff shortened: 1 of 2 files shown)


johanbrandhorst:
Thanks for your contribution!

PR #1121, opened 2020-01-23
Fix context leak identified by nogo linter

akolchinsky:
the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak.


akolchinsky:
Fixed: the cancel function returned by context.WithTimeout should be called, not discarded, to avoid a context leak.

johanbrandhorst:
Closing as incorrect

PR #1124, opened 2020-02-01
Fix #1117: Support response examples.

### LINKED ISSUE #1117: protoc-gen-swagger: Support Response-specific examples
Swagger 2.0 supports specifying a per-mimetype response example for a given operation:

https://github.com/OAI/OpenAPI-Specification/blob/3.0.0/versions/2.0.md#responseObject
https://github.com/OAI/OpenAPI-Specification/blob/3.0.0/versions/2.0.md#exampleObject

This allows displaying a per-operation example for a given response schema, typically overriding any example specified on the response schema type definition in generated documentation. This is useful for clearly documenting workflows where the same schema type is returned with varied or evolving state/status.

This capability seems to have been anticipated but left unimplemented in the protobuf annotation types used by grpc-gateway: https://github.com/grpc-ecosystem/grpc-gateway/blob/29dafe9f591bb820bdb08bce0ec9018ff1e75c14/protoc-gen-swagger/options/openapiv2.proto#L87

```
// `Response` is a representation of OpenAPI v2 specification's Response object.
//
// See: https://github.com/OAI/OpenAPI-Specification/blob/3.0.0/versions/2.0.md#responseObject
//
message Response {
  // `Description` is a short description of the response.
  // GFM syntax can be used for rich text representation.
  string description = 1;
  // `Schema` optionally defines the structure of the response.
  // If `Schema` is not provided, it means there is no content to the response.
  Schema schema = 2;
  // field 3 is reserved for 'headers'.
  reserved 3;
  // field 3 is reserved for 'example'.
  reserved 4;
  map<string, google.protobuf.Value> extensions = 5;
}
```

This issue is a feature request for the addition of that capability.

I propose that the added field be structured like
```
map<string, string> examples = 4;
```

to support the specification's requirement of mapping mime types to examples, and to match the pattern string-based example value specification elsewhere in the grpc-gateway annotations (see https://github.com/OAI/OpenAPI-Specification/blob/3.0.0/versions/2.0.md#example-object-example )

jgiles:
Add support for examples in operation annotation responses. Merge
custom response data into the default 200 response.


protoc-gen-swagger/genswagger/template.go
```diff
@@ -944,9 +944,16 @@ func renderServices(services []*descriptor.Service, paths swaggerPathsObject, re
 					}
 					if opts.Responses != nil {
 						for name, resp := range opts.Responses {
-							respObj := swaggerResponseObject{
-								Description: resp.Description,
-								Schema:      swaggerSchemaFromProtoSchema(resp.Schema, reg, customRefs, meth),
+							// Merge response data into default response if available.
```
jgiles: This is a potential point of discussion. It's a very slight behavior change for existing protos which set a code 200 response object - doing so previously would result in a missing description (instead of method-comment-based description) and/or return schema (instead of return-proto-based schema), unless each of those fields were manually specified in the response object.

It's very convenient to merge the response specification into the default 200 spec generated from the proto return value + comments. For the case where you just want to customize one field, you don't have to re-specify all the others.

johanbrandhorst: Thanks for raising this - I don't think we're too worried about breaking users here, it seems to me a generally better solution :+1:

examples/clients/abe/api/swagger.yaml
```diff
@@ -2189,8 +2198,6 @@ definitions:
     properties:
       value:
         type: "string"
-    example:
-      value: "value"
```
johanbrandhorst: Do you know why this disappeared? It doesn't look particularly useful, but it would be interesting to understand.

jgiles: I do not know why this disappeared.  It appears that this file is generated via this command:

```
swagger-codegen generate -i examples/proto/examplepb/a_bit_of_everything.swagger.json \
    -l go -o examples/clients/abe --additional-properties packageName=abe
```

And I see no corresponding change to the `a_bit_of_everything.swagger.json` input file.

Perhaps the swagger-codegen tool is being "smart" and omitting the default example on the schema, because all usages of the schema in this file (just the one Echo grpc method) have a custom response example?

jgiles:
Where in the documentation do you think this should live? I did a quick look-through, and wasn't sure where to put it.

For what it's worth, I generally look at `a_bit_of_everything.proto` when I want to see what I can do and how :-)

johanbrandhorst:
> Where in the documentation do you think this should live? I did a quick look-through, and wasn't sure where to put it.

I think we could do with a section on customising the swagger output. It could just be including some links to the big protofile. I appreciate your time is valuable so don't worry about writing too much.

> For what it's worth, I generally look at `a_bit_of_everything.proto` when I want to see what I can do and how :-)

Yeah I often tell users this and it's not obvious enough to beginners. We should've had a link to this file for a long time already.

examples/clients/abe/api/swagger.yaml
johanbrandhorst: Yeah, don't worry about it.

protoc-gen-swagger/options/openapiv2.proto
```diff
@@ -84,8 +84,9 @@ message Response {
   Schema schema = 2;
   // field 3 is reserved for 'headers'.
   reserved 3;
-  // field 3 is reserved for 'example'.
-  reserved 4;
+  // `Examples` gives per-mimetype response examples.
+  // See: https://github.com/OAI/OpenAPI-Specification/blob/3.0.0/versions/2.0.md#example-object
+  map<string, string> examples = 4;
```
johanbrandhorst: I had a brief discussion with the other maintainers about this again and we're curious if it's possible to do this with a `map<string, any.Any>`. Have you tried it? What does a literal look like? Don't worry about spending time on it, just wanted to explore the possibility.

jgiles: I'm not sure `google.protobuf.Any` makes sense semantically here: https://github.com/protocolbuffers/protobuf/blob/4cf5bfee9546101d98754d23ff378ff718ba8438/src/google/protobuf/any.proto#L122

The `string type_url` value interpretation appears grpc-specific and unlikely to be useful in a Swagger context. In any case, its intent would appear to duplicate the `schema` property in the containing `Response` message here (and that `schema` field, in contrast, *is* meaningfully interpreted by Swagger tooling).

Examining the existing uses of `google.protobuf.Any` in these annotations (in particular, `Schma.example` https://github.com/grpc-ecosystem/grpc-gateway/blob/f0d0de466d4242a713e95d64685c12064495284a/protoc-gen-swagger/options/openapiv2.proto#L152 ), it would appear that the example literals omit the `type_url` field (questionable) and then use the fact that the `bytes value` field can be specified with a string literal to give a JSON string example: https://github.com/grpc-ecosystem/grpc-gateway/blob/f0d0de466d4242a713e95d64685c12064495284a/examples/proto/examplepb/a_bit_of_everything.proto#L196

So the `google.protobuf.Any` application looks like a semantically-questionable, more-verbose way to achieve the effect of the current `map<string, string>`.

On the other hand, a pretty good argument could be made for `map<string, google.protobuf.Value>` here, similar to how it is used in the various `extensions` fields (e.g. https://github.com/grpc-ecosystem/grpc-gateway/blob/f0d0de466d4242a713e95d64685c12064495284a/protoc-gen-swagger/options/openapiv2.proto#L342 ) . The main argument against it is the verbosity relative to a straight JSON string representation (see e.g. https://github.com/grpc-ecosystem/grpc-gateway/blob/f0d0de466d4242a713e95d64685c12064495284a/examples/proto/examplepb/a_bit_of_everything.proto#L66 for a simple two-field case). The verbosity is particularly unfortunate because in the 99% case here you are writing an example that conforms to an already-defined schema definition, meaning the explicit structuring is not particularly helpful.

README.md
```diff
@@ -210,6 +214,16 @@ annotation to your .proto file
    ```
 
 ## Parameters and flags
+
+During code generation with `protoc`, flags to grpc-gateway tools must be passed
+through protoc using the `--<tool_suffix>_out=<flags>:<path>` pattern, for
```
jgiles: This is something I struggled with a fair bit when getting started, and I expect other new users might too.

README.md
```diff
@@ -95,6 +95,10 @@ annotation to your .proto file
     }
    ```
 
+   See [a_bit_of_everything.proto](examples/proto/examplepb/a_bit_of_everything.proto)
```
jgiles: Front and center in Usage seems like a good high-visibility place.

docs/_docs/usage.md
```diff
@@ -68,6 +68,10 @@ Make sure that your `$GOPATH/bin` is in your `$PATH`.
     }
    ```
 
+   See [a_bit_of_everything.proto](examples/proto/examplepb/a_bit_of_everything.proto)
```
jgiles: I don't know if there is some automated tooling syncing the README and the docs directory? I just put the content in both places.

protoc-gen-swagger/options/openapiv2.proto
johanbrandhorst: Thanks, that makes sense

README.md
johanbrandhorst: Great addition!

README.md
johanbrandhorst: Great, thanks!

docs/_docs/usage.md
johanbrandhorst: Yeah, we don't at the moment, the duplication is correct.

johanbrandhorst:
Great work on this! Really appreciate the thoroughness and responsiveness.

jgiles:
Thank you for being so responsive! You're doing great work as a maintainer in the ecosystem - we've seen that both in this project and in others (e.g. https://github.com/grpc-ecosystem/go-grpc-middleware ).

PR #1127, opened 2020-02-10
protoc-gen-swagger: add default error response

### LINKED ISSUE #1122: Generated spec missing grpc errors
The gateway has an error object which any rpc can return- however it doesn't end up in the generated swagger/openapi.

https://github.com/grpc-ecosystem/grpc-gateway/blob/2da5bfafe05e07c9401137126e55eb0ca4f76472/runtime/errors.go#L68-L76

OpenAPI's pattern for something like this is to include it as the `default` case. 
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#responses-object
> The default MAY be used as a default response object for all HTTP codes that are not covered individually by the specification.

Can we do the following?
```diff
 {
   "/api/v1/rbac/users": {
     "get": {
       "summary": "GetUsers",
       "description": "Get the users' information",
       "operationId": "GetUsers",
       "responses": {
         "200": {
           "description": "A successful response.",
           "content": {
             "application/json": {
               "schema": {
                 "$ref": "#/components/schemas/rbacUsers"
               }
             }
           }
+        },
+        "default": {
+          "description": "Error",
+          "content": {
+            "application/json": {
+              "schema": {
+                "$ref": "#/components/schemas/grpcError"
+              }
+            }
+          }
         }
       },
       "tags": [
         "sevone.rbac.RBAC"
       ]
     }
   }
 }
```
With an entry in `schemas` for the new type:
```json
{
  "grpcError": {
    "description": "Error",
    "properties": {
      "error": {
        "type": "string"
      },
      "code": {
        "type": "integer",
        "format": "int32"
      },
      "message": {
        "type": "string"
      },
      "details": {
        "items": {
          "$ref": "#/components/schemas/protobufAny"
        },
        "type": "array"
      }
    }
  }
}
```
I wrote some jq to handle this in the meantime:
```jq
def addGRPCErrorType:
    .components.schemas["grpcError"] = {"description":"gRPC Error","properties":{"error":{"type":"string"},"code":{"type":"integer","format":"int32"},"message":{"type":"string"},"details":{"items":{"$ref":"#/components/schemas/protobufAny"},"type":"array"}}}
;

def addGRPCErrorDefaultResponse:
    .paths[][].responses.default = {"description":"Error","content":{"application/json":{"schema":{"$ref":"#/components/schemas/grpcError"}}}}
;

def addGRPCErrors:
    addGRPCErrorType | addGRPCErrorDefaultResponse
;
```
Not sure all that's required to make this change. I think at the least the errorBody struct would need to be made public. Thoughts?

johanbrandhorst:
Adds a "default" error entry to all responses
in the swagger definitions.

Fixes #1122 

I can't think of a reason why this wouldn't be backwards compatible.


seanlaff:
@johanbrandhorst Thanks for this!

PR #1130, opened 2020-02-14
Configure response content type at the method level

nboukeffa:
This PR will add the possibility to configure the response content
type at the method level.

To do such the developer will have to set the `produces` property
like in the following example.

```
rpc OverwriteResponseContentType(google.protobuf.Empty) returns (google.protobuf.StringValue) {
	option (google.api.http) = {
		get: "/v2/example/overwriteresponsecontenttype"
	};
	option (grpc.gateway.protoc_gen_swagger.options.openapiv2_operation) = {
		produces: "application/text"
	};
}
```


nboukeffa:
@googlebot I signed it!

nboukeffa:
I'm not sure why the build is failing on Circle CI. It's passing locally. If someone can help...

johanbrandhorst:
The bazel failure should be fixed by running:
```bash
$ docker run -itv $(pwd):/grpc-gateway -w /grpc-gateway --entrypoint /bin/bash --rm \
    l.gcr.io/google/bazel -c 'bazel run :gazelle'
```
From https://github.com/grpc-ecosystem/grpc-gateway/blob/master/CONTRIBUTING.md

johanbrandhorst:
Thanks for your contribution!

nboukeffa:
All green. You were right the import was missing in the build definition. Your command did the trick.

Maybe adding it to the `CONTRIBUTING.md` could help.

Issue #1128, opened 2020-02-10
Support for custom query parameters parsers


wp0pw:
Hello, currently generated code handles query parameters by calling 
`runtime.PopulateQueryParameters(&protoReq, req.Form, filters)` and that works great but I'm currently trying to proto some deeply nested messages and I don't want to expose that complexity to the external users - would you be interested in making this more configurable - akin to the model used for Marshalers?

Unless there is a way to already intercept that code somehow that I don't see. I will be happy to contribute if this is something you would be interesed in adding to the gateway.

Thanks.

johanbrandhorst:
If we can find a nice way to integrate this without changing existing users it might be possible yeah. I'm weary of giving any promises until I've seen the damage but I'm optimistic that if it's not too heavy weight we could merge it :). I know there have been some discussion around alternative query parsers in the past too.

wp0pw:
Great. By not changing existing users - do you mean the generated code being the same? I was thinking about going the `runtime.ServerMuxOption` and hving this information on the mux so we would get the specific `PopulateQueryParameters` instance from mux in the Handler, does that sound plausible?

I may not know all the usecases however, so may be long way off here. I can give it a quick stab in coming days and do an early PR too if needed be.

johanbrandhorst:
No, I mean that whatever is the default now must remain the default in the future so existing users upgrading don't have their query parameter parsing changed.

wp0pw:
@johanbrandhorst sorry for radio silence, was quite busy - I've create pr with the changes - #1138 
let me know what you think.

PR #1138, opened 2020-02-24
Configurable query parameters parser

wp0pw:
Implementing changes for  #1128 
Added new mux Option to configure query parameter parser to use with instance of mux.


runtime/mux.go
```diff
@@ -55,6 +58,19 @@ func WithForwardResponseOption(forwardResponseOption func(context.Context, http.
 	}
 }
 
+// A QueryParameterParser populates message form HTTP query parameters values
+type QueryParameterParser func(msg proto.Message, values url.Values, filter *utilities.DoubleArray) error
```
johanbrandhorst: What do you think about making this an interface?

runtime/mux.go
```diff
@@ -55,6 +58,19 @@ func WithForwardResponseOption(forwardResponseOption func(context.Context, http.
 	}
 }
 
+// A QueryParameterParser populates message form HTTP query parameters values
```
johanbrandhorst:
```suggestion
// A QueryParameterParser populates the message from HTTP query parameter values
```

runtime/mux.go
```diff
@@ -55,6 +58,19 @@ func WithForwardResponseOption(forwardResponseOption func(context.Context, http.
...
+type QueryParameterParser func(msg proto.Message, values url.Values, filter *utilities.DoubleArray) error
+
+// WithQueryParametersPopulator returns a ServeMuxOption representing new populator used to populate message from query parameters
```
johanbrandhorst:
```suggestion
// WithQueryParameterParser sets the query parameter parser, used to populate message from query parameters, on this ServeMux.
```

runtime/mux.go
```diff
@@ -55,6 +58,19 @@ func WithForwardResponseOption(forwardResponseOption func(context.Context, http.
...
+//
+// queryParametersPopulator is an option that will be called on the proto.Message
+// url.Values, and utilities.DoubleArray on incoming requests.
```
johanbrandhorst: This isn't necessary

runtime/mux.go
```diff
@@ -55,6 +58,19 @@ func WithForwardResponseOption(forwardResponseOption func(context.Context, http.
...
+func WithQueryParametersPopulator(queryParametersPopulator QueryParameterParser) ServeMuxOption {
```
johanbrandhorst: Lets name this more consistently:
```suggestion
func WithQueryParameterParser(queryParameterParser QueryParameterParser) ServeMuxOption {
```

runtime/mux.go
```diff
@@ -55,6 +58,19 @@ func WithForwardResponseOption(forwardResponseOption func(context.Context, http.
...
+	return func(serveMux *ServeMux) {
+		serveMux.queryParameterParser = queryParametersPopulator
```
johanbrandhorst:
```suggestion
		serveMux.queryParameterParser = queryParameterParser
```

wp0pw:
Global parser is fine I think, if we don't want to change the generated code there is not much else we can do here - I'll try to make changes some time this week, thanks for quick feedback.

runtime/mux.go
wp0pw: can do, will make the changes for global parser and see if interface fits/is needed.

runtime/mux.go
wp0pw: removed

wp0pw:
Changed to use global parser. Added interface for parser implementations.
No changes to generated code.

runtime/mux.go
wp0pw: added interface in query.go

runtime/query.go
```diff
@@ -17,9 +17,25 @@ import (
...
+}
+
+// PopulateQueryParameters parses query parameters
+// into "msg" using current query parser
 func PopulateQueryParameters(msg proto.Message, values url.Values, filter *utilities.DoubleArray) error {
+	return currentQueryParser.Parse(msg, values, filter)
+}
+
+// DefaultQueryParser is a default implementation of QueryParametersParser
+type DefaultQueryParser struct{}
```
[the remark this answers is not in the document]
wp0pw: unexported

runtime/mux.go
```diff
@@ -55,6 +55,13 @@ func WithForwardResponseOption(forwardResponseOption func(context.Context, http.
 	}
 }
 
+// WithQueryParameterParser sets the query parameter parser, used to populate message from query parameters.
+func WithQueryParameterParser(queryParameterParser QueryParameterParser) ServeMuxOption {
```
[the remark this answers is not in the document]
wp0pw: it's a `Set` now

wp0pw:
Pleasure!

anilCSE:
Is there any example implementation for this? I was looking at the docs, didn't find any info

johanbrandhorst:
The default implementation in both master and v2 implements the interface. You'll just have to design your own with the same signature. See https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/query.go#L37

anilCSE:
Looks like there's some issue with `gogoproto` support. 
```
 cannot use &customQueryParser literal (type *customQueryParser) as type runtime.QueryParameterParser in assignment:
        *customQueryParser does not implement runtime.QueryParameterParser (wrong type for Parse method)
                have Parse("github.com/gogo/protobuf/proto".Message, url.Values, *utilities.DoubleArray) error
                want Parse(protoiface.MessageV1, url.Values, *utilities.DoubleArray) error
```

Any known workarounds for this?

johanbrandhorst:
No, you can't use it with Gogo protobuf

anilCSE:
Oh okay, thank you for the info @johanbrandhorst . We are trying to `convert` query parameters to grpc server specific types in `grpc-gateway`. Is there an easy way to achieve this? For the sake of `UX`, we want to keep `human-readable` parameters for clients and convert them to `bytes` in `grpc-gateway`'s middleware/interceptor (?).

johanbrandhorst:
Any parameters not captured in the body or the path automatically become query parameters, are you sure you need to do anything?

anilCSE:
I need to cast them before forwarding them to grpc server. We have some prefix encoded strings for clients and grpc server uses bytes (proto definitions are bytes). We want to support strings from client and cast them to bytes in grpc gateway middleware

achew22:
@anilCSE you should be able to write anything you want in some kind of Go middleware. grpc-gateway doesn't have any provisions for doing this directly so you will have to write something custom up.

PR #1141, opened 2020-02-26
protoc-gen-swagger: add flag to disable default errors

johanbrandhorst:
The new flag disable_default_errors can be used to disable
the generation of the default error types. This is suitable
for users that define their own error formatting.


johanbrandhorst:
Tested this by manually generating all the swagger files with this enabled and it produces the expected output.

PR #1144, opened 2020-03-01
Fix #1143: Support multiple error handlers.

### LINKED ISSUE #1143: cannot use multiple different error handlers in one gateway binary
Fill in the following sections with explanations of what's gone wrong.

1. Create ServeMux without a custom error handler. Register it to handle prefix "/v1"
2. Create a ServeMux with a custom error handler using `runtime.WithProtoErrorHandler`. Register it to handle prefix "/v2".
3. Make calls resulting in errors to a mix of v1 and v2 endpoints.

Observe that the v2 error handler runs for:
 * errors returned from the backends on *both* the v1 and v2 endpoints
 * errors returned from within runtime.ServeMux code (e.g. Unimplemented on unknown URIs) on the v2 endpoints

Errors from within runtime.ServeMux code (like unknown URIs) under v1 result in `unexpected use of OtherErrorHandler` errors.

 * v1 endpoints use the default error handler
 * v2 endpoints use the configured custom error handler.

 * The `WithProtoErrorHandler` option overwrites a global handler HTTPError and configures the global OtherErrorHandler to return "unexpected use" errors: https://github.com/grpc-ecosystem/grpc-gateway/blob/9aff3552347d758821bca2c628f55f8e11d0e0c7/runtime/mux.go#L169
 * Generated gateway handler code calls the global HTTPError regardless of the protoErrorHandler configured on the current ServeMux instance: https://github.com/grpc-ecosystem/grpc-gateway/blob/9aff3552347d758821bca2c628f55f8e11d0e0c7/protoc-gen-grpc-gateway/gengateway/template.go#L571
 * Without a configured protoErrorHandler, ServeMux falls back to the global OtherErrorHandler (which has been overwritten if any mux in the process has a protoErrorHandler configured).

An initial patch to fix this issue can be seen here: https://github.com/grpc-ecosystem/grpc-gateway/compare/master...paxosglobal:per-mux-error-handler

However, since this does change the generated code to rely on the updated runtime code, I'd appreciate guidance on the process for handling/merging this kind of breaking change.

jgiles:
Support specifying different custom error handlers on different muxes,
or specifying custom error handlers on some muxes and using the default
handlers on others.

Make the runtime.HTTPError use the mux-configured error handler if
present, and allow overriding the global default behavior by setting
GlobalHTTPErrorHandler.

Add commentary explaining the use of the different options for
controlling error handling.

### DIFF
--- a/docs/_docs/customizingyourgateway.md
+++ b/docs/_docs/customizingyourgateway.md
@@ -234,7 +234,24 @@ if err := pb.RegisterMyServiceHandlerFromEndpoint(ctx, mux, serviceEndpoint, opt
 ```
 
 ## Error handler
-http://mycodesmells.com/post/grpc-gateway-error-handler
+The gateway uses two different error handlers for non-streaming requests:
+
+ * `runtime.HTTPError` is called for errors from backend calls
+ * `runtime.OtherErrorHandler` is called for errors from parsing and routing client requests
+
+To override all error handling for a `*runtime.ServeMux`, use the
+`runtime.WithProtoErrorHandler` serve option.
+
+Alternatively, you can override the global default `HTTPError` handling by
+setting `runtime.GlobalHTTPErrorHandler` to a custom function, and override
+the global default `OtherErrorHandler` by setting `runtime.OtherErrorHandler`
+to a custom function.
+
+You should not set `runtime.HTTPError` directly, because that might break
+any `ServeMux` set up with the `WithProtoErrorHandler` option.
+
+See https://mycodesmells.com/post/grpc-gateway-error-handler for an example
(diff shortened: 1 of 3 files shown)


jgiles:
@johanbrandhorst I've tested this change and it resolves #1143

jgiles:
@johanbrandhorst I updated the relevant documentation. It was mostly just a link, but the link doesn't talk about `WithProtoErrorHandler`, so I added some explanation and then just cited that link as a an example for the actual handler function implementation.

johanbrandhorst:
A pleasure as always Josh :smile:.

karel-3d:
just randomly showing here and saying thanks :) great change

PR #1145, opened 2020-03-01
Fix #201: Exclude path params from body schemas.

### LINKED ISSUE #201: Fields bound by path template also in body (Swagger)
https://github.com/googleapis/googleapis/blob/master/google/api/http.proto#L123

| HTTP | RPC |
| --- | --- |
| `PUT /v1/messages/123456 { "text": "Hi!" }` | `UpdateMessage(message_id: "123456" message { text: "Hi!" })` |

The special name `*` can be used in the body mapping to define that
every field not bound by the path template should be mapped to the
request body.  This enables the following alternative definition of
the update method:

``` proto
service Messaging {
  rpc UpdateMessage(Message) returns (Message) {
    option (google.api.http) = {
      put: "/v1/messages/{message_id}"
      body: "*"
    };
  }
}
message Message {
  string message_id = 1;
  string text = 2;
}
```

---

Currently, this doesn't seem to work when generating the Swagger definition, as both fields are showing up. This is a snippet from my `.proto` file.

``` proto
service Application {
    rpc Update(UpdateApplicationRequest) returns (UpdateApplicationResponse) {
        option (google.api.http) = {
            put: "/api/v1/application/{appEUI}"
            body: "*"
        };
    }
}

message UpdateApplicationRequest {
    string appEUI = 1;
    string name = 2;
}

```

image

I would have expected to ony have `name` in the request body. Leaving out the `devEUI` does work however when making requests!

When setting `body: "name"` in the `.proto` file (see also [http.proto#L109](https://github.com/googleapis/googleapis/blob/master/google/api/http.proto#L109)), the API returns the following error (even if the `appEUI` is included in the body too):

``` json
{
  "Error": "json: cannot unmarshal object into Go value of type string",
  "Code": 3
}
```

jgiles:
When body is "*" and some request fields are captured by path
parameters, exclude those fields from the request body Swagger schema.

Define an inline request body schema with all the request fields NOT
captured by path parameters. Also filter path parameters out of the
required fields list for the body schema.


examples/clients/abe/api/swagger.yaml
```diff
@@ -2388,6 +2388,139 @@ definitions:
     properties:
       value:
         type: "string"
+  A bit of everything:
```
jgiles: This appears to be an artifact of swagger-codegen: It is adding a definition for the inline schema in the source Swagger spec, choosing a name based on the "title" of the inlined schema object.

examples/clients/abe/api_a_bit_of_everything_service.go
```diff
@@ -2519,7 +2519,7 @@ ABitOfEverythingServiceApiService
 
 @return interface{}
 */
-func (a *ABitOfEverythingServiceApiService) PostWithEmptyBody(ctx context.Context, name string, body ExamplepbBody) (interface{}, *http.Response, error) {
+func (a *ABitOfEverythingServiceApiService) PostWithEmptyBody(ctx context.Context, name string, body interface{}) (interface{}, *http.Response, error) {
```
jgiles: apparently how swagger-codegen handles empty request bodies.

examples/clients/abe/model_a_bit_of_everything.go
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+package abe
+
+import (
+	"time"
+)
+
+// Intentionaly complicated message type to cover many features of Protobuf.
+type ABitOfEverything struct {
```
jgiles: model file for the definition artifact mentioned above.

examples/proto/examplepb/a_bit_of_everything.swagger.json
```diff
@@ -1556,7 +1556,186 @@
...
+                }
+              },
+              "description": "Intentionaly complicated message type to cover many features of Protobuf.",
+              "title": "A bit of everything",
+              "externalDocs": {
+                "description": "Find out more about ABitOfEverything",
+                "url": "https://github.com/grpc-ecosystem/grpc-gateway"
+              },
+              "required": [
+                "int64_value",
```
jgiles: note that "uuid" is not in the required list.

examples/proto/examplepb/a_bit_of_everything.swagger.json
```diff
@@ -1556,7 +1556,186 @@
             "in": "body",
             "required": true,
             "schema": {
-              "$ref": "#/definitions/examplepbABitOfEverything"
+              "type": "object",
+              "example": {
+                "uuid": "0cf361e1-4b44-483d-a159-54dabdf7e814"
+              },
+              "properties": {
```
jgiles: note that the inlined properties list excludes "uuid"

examples/proto/examplepb/a_bit_of_everything.swagger.json
```diff
@@ -1556,7 +1556,186 @@
             "in": "body",
             "required": true,
             "schema": {
-              "$ref": "#/definitions/examplepbABitOfEverything"
+              "type": "object",
+              "example": {
+                "uuid": "0cf361e1-4b44-483d-a159-54dabdf7e814"
```
jgiles: This is a bit unfortunate - the example here is from the proto message definition, and that example has uuid in it. Arguably, the example should be updated since it's not really applicable to everywhere the message is used.

examples/proto/examplepb/a_bit_of_everything.swagger.json
```diff
@@ -2201,7 +2380,7 @@
             "in": "body",
             "required": true,
             "schema": {
-              "$ref": "#/definitions/examplepbBody"
+              "type": "object"
```
jgiles: the only field in the request object is captured by a path parameter.

protoc-gen-grpc-gateway/descriptor/registry.go
```diff
@@ -425,6 +425,14 @@ func (r *Registry) GetUseJSONNamesForFields() bool {
 	return r.useJSONNamesForFields
 }
 
+// FieldName returns the proto or JSON field name for the field, based on UseJSONNamesForFields.
+func (r *Registry) FieldName(f *Field) string {
```
jgiles: drive-by refactor, since I was adding another couple places where I needed to resolve a field name.

protoc-gen-swagger/genswagger/template.go
```diff
@@ -288,74 +279,101 @@ func renderMessagesAsDefinition(messages messageMap, d swaggerDefinitionsObject,
...
 		}
-		schema := swaggerSchemaObject{
-			schemaCore: schemaCore{
-				Type: "object",
-			},
+		d[fullyQualifiedNameToSwaggerName(msg.FQMN(), reg)] = renderMessageAsSchema(msg, reg, customRefs, nil)
+	}
+}
+
+func renderMessageAsSchema(msg *descriptor.Message, reg *descriptor.Registry, customRefs refMap, excludeFields []*descriptor.Field) swaggerSchemaObject {
```
jgiles: this was lifted and modified from the loop body in `renderMessagesAsDefinition` - I recommend reviewing with "Hide whitespace changes" set to avoid the indentation change obscuring the logic diff.

protoc-gen-swagger/genswagger/template.go
```diff
@@ -762,13 +780,29 @@ func renderServices(services []*descriptor.Service, paths swaggerPathsObject, re
...
 							schemaCore: schemaCore{},
 						}
 
 						wknSchemaCore, isWkn := wktSchemas[meth.RequestType.FQMN()]
 						if !isWkn {
-							schema.Ref = fmt.Sprintf("#/definitions/%s", fullyQualifiedNameToSwaggerName(meth.RequestType.FQMN(), reg))
+							var bodyExcludedFields []*descriptor.Field
+							if len(b.PathParams) != 0 {
+								for _, p := range b.PathParams {
+									// We only support excluding top-level fields captured by path parameters.
```
jgiles: We could try to support nested field refs, but it's a fair bit more work for pretty limited return I think.

jgiles:
@johanbrandhorst this is a first pass at fixing #201 , with the fix confirmed in the existing a_bit_of_everything example and in our application's generated documentation.

examples/clients/abe/api/swagger.yaml
johanbrandhorst: I think this should be fine, it shouldn't cause any breakages.

examples/proto/examplepb/a_bit_of_everything.swagger.json
johanbrandhorst: That sounds right - could you update the example to include both the required values `int64_value` and `double_value`?

examples/proto/examplepb/a_bit_of_everything.swagger.json
johanbrandhorst: Is there some way we could remove this `"in": "body"` entry altogether?

protoc-gen-grpc-gateway/descriptor/registry.go
johanbrandhorst: :car: :gun:

protoc-gen-swagger/genswagger/template.go
johanbrandhorst: Yeah, lets not do that right now.

protoc-gen-swagger/genswagger/template.go
```diff
@@ -762,13 +780,29 @@ func renderServices(services []*descriptor.Service, paths swaggerPathsObject, re
...
+									if len(p.FieldPath) == 1 {
+										bodyExcludedFields = append(bodyExcludedFields, p.FieldPath[0].Target)
+									}
+								}
+							}
+							if len(bodyExcludedFields) != 0 {
+								// If we are excluding some fields, define a schema inline.
+								schema = renderMessageAsSchema(meth.RequestType, reg, customRefs, bodyExcludedFields)
```
johanbrandhorst: If we have no fields left here, we could skip this part?

johanbrandhorst:
Hey Josh, sorry there hasn't been any activity on here, are you waiting on me?

johanbrandhorst:
We were happy to merge this but it seems the author disappeared. Would you be willing to contribute a fix for this? If so, could I ask that it be against the v2 branch only?

stijndehaes:
@jgiles are you still planning on working on this or are you fine with me using your work as a starting point and finish it?

PR #1146, opened 2020-03-02
Minimize API surface in anticipation of gorelease

johanbrandhorst:
The grpc-gateway was created long before things like go modules semver and with little concern for internal package management. This unfortunately means that technically the API surface is enormous, however most of these packages are not _really_ considered part of the public API.

This change moves everything that isn't part of the public API under various `internal` folder structures. This is obviously technically a breaking change, but I'm hoping we can get away with saying that our users really shouldn't have been importing these things anyway if any of them have been. I'm happy to discuss any concerns.

The end goal of this migration is to be able to use https://godoc.org/golang.org/x/exp/cmd/gorelease to ensure API stability across minor patches. The current structure prohibits this (because we make breaking changes to the examples all the time).

CC @achew22 @tmc  @ivucica


achew22:
I'm extremely supportive of this change. Thank you for taking the initiative

ivucica:
I am basically ok with this. Some tests are failing. I'll try to review this some time this week. Poke me using some out-of-band channel if I forget about it.

johanbrandhorst:
There seems to be a consensus that we should try this and see if it angers any users. Here goes!

karel-3d:
why did not `grpc-gateway` bump version to v2 when doing this? It did break backward compatibility and one of our modules broke, because it imported `httprule`.

this is why `v2` exists. grpc-gateway should have bumped that :( 

I agree with the reason of doing this (reducing the public API), but it should change major version and thus import path when doing it

karel-3d:
see https://blog.golang.org/v2-go-modules and https://github.com/golang/go/wiki/Modules - modules should be backwards compatible or change import path

johanbrandhorst:
Hi @karel-3d, in principle I agree with you, but we wanted to see how much we could get away with doing without bumping to v2 _yet_. If you have a dependency that depends on `httprule`, I can move it back into the public API and make a new release (see #1154 and #1161). Could you raise an issue?

karel-3d:
@johanbrandhorst hey sorry for the little confrontational tone at first, it was a bit angry at first :) but makes sense, and it's not that popular import so I get it

anyway I made an issue 

https://github.com/grpc-ecosystem/grpc-gateway/issues/1168

PR #1150, opened 2020-03-04
Use an interface for ContentTypeFromMessage

wtlangford:
This allows custom `Marshaler` implementations to vary the HTTP response's Content-Type based on the returned proto.

Because of the comments around adding the interface param to the existing `ContentType()` function, I left the interface unexported, so that it doesn't become part of the public API of the package.

Alternatively, it could be exported and used to eventually deprecate the old `Marshaler` interface in favor of a more flexible `ContentTypeFromMessage`.
```go
// A ContentTypeMarshaler is a Marshaler that can vary its Content-Type based on the response value
type ContentTypeMarshaler interface {
   Marshaler
   ContentTypeFromMessage(v interface{}) string
}
```

Is there a preference between the two approaches?

### DIFF
--- a/runtime/errors.go
+++ b/runtime/errors.go
@@ -123,9 +123,9 @@ func DefaultHTTPError(ctx context.Context, mux *ServeMux, marshaler Marshaler, w
 	// Check marshaler on run time in order to keep backwards compatability
 	// An interface param needs to be added to the ContentType() function on
 	// the Marshal interface to be able to remove this check
-	if httpBodyMarshaler, ok := marshaler.(*HTTPBodyMarshaler); ok {
+	if typeMarshaler, ok := marshaler.(contentTypeMarshaler); ok {
 		pb := s.Proto()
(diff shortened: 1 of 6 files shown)


johanbrandhorst:
Hi William, thanks for the PR! Do you mind explaining why you want this feature added? I appreciate that this is a pretty minimally disruptive change, so I don't have any reservations, but if you have a particular use case in mind it would help me understand why the existing solution isn't good enough.

wtlangford:
We've added a custom marshaler in our application that wraps the existing `HTTPBodyMarshaler` and provides some additional functionality (mapping the request body into an `google.api.HttpBody`). Unfortunately, when we use that marshaler, we no longer have the `Content-Type` header properly set on responses that use `google.api.HttpBody`.

I'm happy to look at separately contributing back the request-mapping functionality, but this felt like a useful way to allow other custom marshalers to use `ContentTypeFromMessage`.

johanbrandhorst:
Hi, please rebase on master, I've moved some folders around, but otherwise SGTM.

PR #1160, opened 2020-03-06
Add Routes method to return registered routes

bentranter:
### Overview

Adds a new `Routes` method to the `ServeMux` that allows a user to get all registered routes, that is, a map of HTTP verbs to URL paths.

### Motivation

I want to push my routes up to a proxy, in order to make that proxy aware of which routes go to my grpc-gateway powered service. Previously, there was no way to do that other than to manually write out each route. With this change, users can now call `Routes` on a `ServeMux` to get a `map[string][]string` of HTTP verbs to URL paths.

---

I know this PR may be lacking tests, and might not even be the right way to approach this, but I wanted to create this PR to open this up for discussion.

### DIFF
--- a/runtime/mux.go
+++ b/runtime/mux.go
@@ -187,6 +187,35 @@ func (s *ServeMux) Handle(meth string, pat Pattern, h HandlerFunc) {
 	}
 }
 
+// Routes returns all defined routes defined on the ServeMux. The returned map
+// maps HTTP verbs to route paths, ie, the returned map will look something
+// like,
+//
+//  map[string][]string{
+//      "GET", []string{
+//          "/users",
+//          "/users{id=*}",
+//      },
+//      "POST": []string{
+//          "/users",
+//      }
+//  }
+func (s *ServeMux) Routes() map[string][]string {
+	routes := make(map[string][]string)
+
+	for method, handlers := range s.handlers {
+		for _, handler := range handlers {
+			if path, ok := routes[method]; !ok {
+				routes[method] = []string{handler.pat.String()}
+			} else {
+				path = append(path, handler.pat.String())
+			}
+		}
+	}
+
+	return routes
+}
+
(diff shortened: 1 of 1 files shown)


johanbrandhorst:
Hi Ben,

As mentioned in #grpc-gateway, another way you might do this is by exposing the grpc reflection service over the gateway. There are upsides and downsides with both solutions:

| Method | Pros | Cons|
|--- | --- | --- |
| grpc reflection | Standard way of getting gRPC type and RPC information<br>No gateway changes required | Would require annotating reflection protofile/using YAML config<br>Would expose _all_ gRPC methods on a server, not just gateway RPCs<br>Would not include the REST paths, path parameters, query parameters |
| Routes() method | Would only expose routes registered to the serve mux<br>Could include path information, query parameters, etc | Requires (potentially complex) gateway changes<br>Leaves a lot of work up to the client |

On balance, I think adding a new method would be best, since it allows the user to only expose the route registered to the mux. However, it is not clear to me that the method, as suggested, is useful enough. Have you tried using this? How are you going to translate the routes to anything useful, or know about the types of the parameters? It basically becomes a matter of exposing an OpenAPI-equivalent API via this method. Maybe it would be easier to just generate a swagger file and serve that?

bentranter:
Hey @johanbrandhorst, thanks for all the feedback, and sorry for the delay.

For our use case, we're only interested in the routes registered to the ServeMux, or more simply, we just care about the HTTP side of things – not the gRPC methods.

In terms of translating the types to something useful, we're using this only for configuration, ie, we push the routes information up as JSON so we don't care about the types of the parameters – just the HTTP verb and path(s).

I realize that our use case is pretty narrow, so I totally understand that I haven't thought of too many use cases outside of our own. That said, I hadn't considered the option of using the generated Swagger file, so I'm going to explore that option over the next day or two to see if that satisfies our use case, and get back to you. Thanks again!

PR #1163, opened 2020-03-09
Move genswagger to public API again

### LINKED ISSUE #1162: Swagger definition is broken if there's no default error response set
Introduced by #1127 , refs [utrack/clay#81](https://github.com/utrack/clay/issues/81)

Right now gen-swagger generates broken Swagger def if there's no `.grpc.gateway.runtime.Error` set ([lookup code](https://github.com/grpc-ecosystem/grpc-gateway/pull/1127/files#diff-d419038d63db0fdabc61f0d6cc6683aeR1051)).
Every handler has a link to error definition as "default" response, even if `Error` was not discovered. 

I guess it's better to drop error response definition if `Error` was not found, but still emit a warning (`no message found: Error`). What do you think?
I can send a PR with fix if you're OK with that.

Also, the message can be made better (in fact, generator tries to resolve `.grpc.gateway.runtime.Error`, not just `Error`).

johanbrandhorst:
Fixes #1162


johanbrandhorst:
Goreleaser is reporting an error here, but I don't think it is an error, so I raised https://github.com/golang/go/issues/37756.

PR #1166, opened 2020-03-10
genswagger: don't emit default response if there's no Error definition

### LINKED ISSUE #1162: Swagger definition is broken if there's no default error response set

utrack:
Swagger: do not create default response definitions if we can't find Error/StreamError in Registry, even if `DisableDefaultErrors()` is false.

Logs are still emitted if default errors are enabled, but error definition was not found.

grep: `.grpc.gateway.runtime.Error`, `.grpc.gateway.runtime.StreamError`

Closes #1162


protoc-gen-swagger/genswagger/template_test.go
```diff
@@ -2074,3 +2075,86 @@ func TestMessageOptionsWithGoTemplate(t *testing.T) {
...
+	}
+
+	defRsp, ok := result.Paths["/v1/echo"].Post.Responses["default"]
+	if !ok {
+		return
+	}
+
+	ref := defRsp.Schema.schemaCore.Ref
+	// TODO utrack: is there a better way to find object by ref?
+	refName := strings.TrimPrefix(ref, "#/definitions/")
```
utrack: I'm not sure if there's any existing helper function to get a definition by its ref? strings.TrimPrefix looks ugly

protoc-gen-swagger/genswagger/template.go
```diff
@@ -518,14 +518,24 @@ func renderEnumerationsAsDefinition(enums enumMap, d swaggerDefinitionsObject, r
 
 // Take in a FQMN or FQEN and return a swagger safe version of the FQMN
 func fullyQualifiedNameToSwaggerName(fqn string, reg *descriptor.Registry) string {
+	ret, _ := fullyQualifiedNameToSwaggerNameOk(fqn, reg)
+	return ret
+}
+
+// Take in a FQMN or FQEN and return a swagger safe version of the FQMN and
+// a boolean indicating if FQMN was properly resolved.
+// TODO utrack: bad name?
```
johanbrandhorst: Just remove the other one and use the one that reports if it was found everywhere. It's a good change :+1:.

protoc-gen-swagger/genswagger/template_test.go
johanbrandhorst: Haha not that I know :joy:. This is fine. Remove the comment.

utrack:
@johanbrandhorst the code looks a bit messier now because of more if branches; however, I don't want to do refactoring in the same PR so the changes would be as small as possible. This calls for a separate PR :)

johanbrandhorst:
Would appreciate a further refactor in a separate PR if you have the time :).

PR #1169, opened 2020-03-11
Move httprule back into the public API

### LINKED ISSUE #1168: Missing httprule breaks our module
Missing httprule in 1.14 breaks our module.

Now I see there are only a few modules on whole github that use it. One is our module here

https://github.com/gunk/gunk/blob/af47d12e8be68589435b002f60f291f380aa7ef4/docgen/parser/parser.go#L14

we generate documentation from GRPC definition and this is a bit useful.

Other use I can find is here 

https://github.com/elliots/protoc-gen-twirp_swagger/blob/fda01df7c6eb134cca0579e1ce45056af36f4b1a/genswagger/template_test.go#L16

It also seems to be some kind of protobuf generator (it was actually already submitted here, but this is just in the tests).

I cannot find any other use of this module on github or elsewhere. 🤷‍♂

johanbrandhorst:
Fixes #1168


karel-3d:
Thanks a lot!

PR #1174, opened 2020-03-14
should be using method index instead of service index for method's pr…

coolyrat:
tested on my [use case ](https://github.com/coolyrat/grpc-gateway-merge-test)


coolyrat:
@googlebot I signed it!

achew22:
Can you regenerate the examples so that we can see how this change affects things? Thanks!

johanbrandhorst:
See CONTRIBUTING.md for instructions :)

coolyrat:
I'm not sure what's wrong, could you guys help and have a peek

johanbrandhorst:
@googlebot I consent.

examples/internal/proto/examplepb/unannotated_echo_service.swagger.json
```diff
@@ -47,8 +47,6 @@
     },
     "/v1/example/echo/{id}/{num}": {
       "get": {
-        "summary": "Echo method receives a simple message and returns it.",
-        "description": "The message posted as the id parameter will also be\nreturned.",
```
johanbrandhorst: This change is incorrect, so we need to figure out how we can fix your case and still keep this.

coolyrat:
@johanbrandhorst i've tried every thing i could to make the ci happy. But there's nothing I could do about it. Need your help to regenerate.

examples/internal/proto/examplepb/a_bit_of_everything.swagger.json
```diff
@@ -2037,6 +2037,8 @@
     },
     "/v2/example/empty": {
       "get": {
+        "summary": "Create a new ABitOfEverything",
+        "description": "This API creates a new ABitOfEverything",
```
johanbrandhorst: This comment is not correct either. There should not be a comment here. See https://github.com/grpc-ecosystem/grpc-gateway/blob/master/examples/internal/proto/examplepb/a_bit_of_everything.proto#L541-L547

johanbrandhorst:
Stale

Issue #1175, opened 2020-03-14
protoc-gen-grpc-gateway should warn or fail if a selector does not exist


andrascz:
### Steps you follow to reproduce the error:

1.  specify a service config with a selector which does not exist (mistype rpc service name)
2.  generate gateway code
3.  generation does not fail, but ignores the endpoint with a not matched selector

### What did you expect to happen instead:

Generation fails or at least warns me about a non-matched selector

### What's your theory on why it isn't working:

Currently non-matched selectors are gracefully ignored.

johanbrandhorst:
Hi, thanks for the bug report! I take it this affects the yaml configuration only?

Would you be interested in adding this capability?

andrascz:
I assume this can not happen when you annotate the proto file, as you define the binding directly on the rpc of your choice.
But probably the swagger generator should behave the same way.

I can raise a PR although first I need to familiarize myself with the code a bit.

andrascz:
I have a working patchset at https://github.com/Strivacity/grpc-gateway/tree/error-on-unbound-rules-or-methods, as soon as I solved the CLA issue I will raise a PR.
Currently our company is dealing with transitioning to home office because of COVID-19 so processes a bit slower than usual.

johanbrandhorst:
Thanks, no rush from our side, stay safe :slightly_smiling_face:.

PR #1178, opened 2020-03-17
Emit error on HTTP rules without a matching selector and log warning on unbound methods

andrascz:
Fixing: https://github.com/grpc-ecosystem/grpc-gateway/issues/1175

I could not emit error in case of unbound methods as that would have break all the examples.

### DIFF
--- a/protoc-gen-grpc-gateway/descriptor/registry.go
+++ b/protoc-gen-grpc-gateway/descriptor/registry.go
@@ -268,6 +268,29 @@ func (r *Registry) AddExternalHTTPRule(qualifiedMethodName string, rule *annotat
 	r.externalHTTPRules[qualifiedMethodName] = append(r.externalHTTPRules[qualifiedMethodName], rule)
 }
 
+// UnboundExternalHTTPRules returns the list of External HTTPRules
+// which does not have a matching method in the registry
+func (r *Registry) UnboundExternalHTTPRules() []string {
+	allServiceMethods := make(map[string]struct{})
+	for _, f := range r.files {
+		for _, s := range f.GetService() {
+			svc := &Service{File: f, ServiceDescriptorProto: s}
+			for _, m := range s.GetMethod() {
+				method := &Method{Service: svc, MethodDescriptorProto: m}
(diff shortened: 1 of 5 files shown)


andrascz:
> Excellent work! What does this it look like when you run it and there's a route mismatch?

In case of grpc method without HTTP rule:
```
W0317 16:42:06.689825  444590 services.go:38] No HttpRule found for method: CastleblackService.GetHealth
```

In case of HTTP Rule without matching grpc method:
```
--grpc-gateway_out: HTTP rules without a matching selector: .pb.CastleblackService.GetHealthz
make: *** [makefiles/protobuf.mk:18: pkg/pb] Error 1
```

andrascz:
We submitted the CLA, but not yet processed I guess, I will check daily and update when there is progress.

johanbrandhorst:
> We submitted the CLA, but not yet processed I guess, I will check daily and update when there is progress.

You trigger a check manually:
> Once you've signed (or fixed any issues), please reply here with @googlebot I signed it! and we'll verify it.

andrascz:
> You trigger a check manually:

Thanks, I know that, but the CLA page does not show me as signed yet and according to the doc: 

>If the contributor is covered by a corporate CLA, make sure that the corporation is listed at http://linkremoved/. If it’s not, it simply may not have been processed yet, which typically takes a few days.

johanbrandhorst:
My bad!

andrascz:
@googlebot I signed it!

johanbrandhorst:
Thanks for your contribution! A pleasure working with you 😁

yousseftelda:
@johanbrandhorst What is rationale behind the warning on unbound methods? This produces a lot of warnings as we only want to generate a gateway for a subset of our gRPC services.
Can we put this behind a flag? Happy to open a PR.

johanbrandhorst:
I'd be happy to remove the unbound method warning, it's much less likely to be unintentional than the HTTP route mismatch. Does that sound okay @andrascz?

andrascz:
Sure, it seems reasonable to make this configurable.

PR #1184, opened 2020-03-23
Use 8088 instead of 8080 in integration test.

drigz:
This should avoid a failure on Bazel's Buildkite CI, where 8080 cannot be used.

### DIFF
--- a/examples/internal/integration/client_test.go
+++ b/examples/internal/integration/client_test.go
@@ -16,7 +16,7 @@ func TestEchoClient(t *testing.T) {
 	}
 
 	cfg := echo.NewConfiguration()
-	cfg.BasePath = "http://localhost:8080"
+	cfg.BasePath = "http://localhost:8088"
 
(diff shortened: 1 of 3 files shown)


drigz:
https://buildkite.com/bazel/grpc-ecosystem-grpc-gateway/builds/331

Not sure why Buildkite doesn't post status to pull requests.

achew22:
@drigz, where would I file a bug to have someone look into it posting a comment on this repo?

drigz:
I think the problem is a conflict with this metrics server: https://github.com/bazelbuild/continuous-integration/blob/85b2f190c32b95adc3759c2c901b2d2244aa8d2a/metrics/main.go

You could file an issue on the continuous-integration repo suggesting they switch to less commonly-used port, but I can't think of a perfect solution.

srenatus:
Oooh finally the streak of red builds in turned green. 👏 Thanks a lot for looking into this. 😃

PR #1186, opened 2020-03-23
feat: add enumAsInts flag

### LINKED ISSUE #1177: protoc-gen-swagger: support outputting enum parameters as integers
grpc gateway support EnumsAsInts, however grpc swagger has not this option

rowe0x:
fix #1177

### DIFF
--- a/protoc-gen-grpc-gateway/descriptor/registry.go
+++ b/protoc-gen-grpc-gateway/descriptor/registry.go
@@ -76,6 +76,9 @@ type Registry struct {
 	// in your protofile comments
 	useGoTemplate bool
 
+	// enumsAsInts render enum as integer, as opposed to string
+	enumsAsInts bool
+
 	// disableDefaultErrors disables the generation of the default error types.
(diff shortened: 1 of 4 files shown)


rowe0x:
@johanbrandhorst resolved :）

nashikb:
@johanbrandhorst @rowe0x The enums are generated with an array of strings rather than an array of ints, which down-stream breaks swagger generators. Was this really the expected behavior?

johanbrandhorst:
It's an optional flag, if it breaks your generators just don't set it. Or am I misunderstanding something?

nashikb:
Absolutely, it's not an issue in the installed-base sense. But for the optional feature itself, I'm not sure how to use it since it's not a format supported by swagger, or at least by the "standard". I'd love to have enums as ints, and I was pleased to see this land 6 days before I wanted it :-) but the downstream support doesn't seem to be there.

johanbrandhorst:
Ok, I see what you're saying. Regardless, clearly some users wanted it and here we are 🤷‍♂️.

PR #1188, opened 2020-03-26
fix: tailer headers should be written before body

hydra-zim:


hydra-zim:
@googlebot I signed it!

achew22:
@hydra-zim, thanks so much for sending in your patch! Would it be possible to add a test to validate this behavior?

PR #1189, opened 2020-03-27
Propagate proto error

flrossetto:


johanbrandhorst:
Hi @flrossetto, thanks for the PR. However, I'm not sure what you're trying to do here. Few if any of these will be status errors (I think), and in any case, if this changes the error output we may not be able to make the change, since it could break our users. Could you please open an issue to discuss the problem before opening a PR?

flrossetto:
OK

johanbrandhorst:
Closing as stale

Issue #1202, opened 2020-04-08
response_body of http rule is ignored in case of server streaming


adasari:
HI Team,

response_body field of http rule is ignored in case of server streaming. i checked the template.go and handler.go, seems they are not checking ResponseBody type.

[template.go](https://github.com/grpc-ecosystem/grpc-gateway/blob/0c9d3eb7ca8e28a7c8db327d6ef63902bfbfd778/protoc-gen-grpc-gateway/internal/gengateway/template.go#L669)
[handler.go](https://github.com/grpc-ecosystem/grpc-gateway/blob/0c9d3eb7ca8e28a7c8db327d6ef63902bfbfd778/runtime/handler.go#L64)

Also noticed stream response is wrapped in a map [here](https://github.com/grpc-ecosystem/grpc-gateway/blob/0c9d3eb7ca8e28a7c8db327d6ef63902bfbfd778/runtime/handler.go#L193). is there any specific reason ?

Thanks

johanbrandhorst:
Hi Anil, thanks for the issue report! I think when we first added support for the `response_body` (https://github.com/grpc-ecosystem/grpc-gateway/pull/712) we didn't consider it in the light of streaming RPCs. There may be a sensible way to implement this for streaming APIs, so lets discuss!

The reason we're using a wrapper by default for streaming responses is so that users can use a single type to marshal the chunks, while still allowing us to return an error at any point in the stream. This means the error has to always be there.

If a user were to use the ResponseBody though, we would allow the user to stream whatever they want.

What ideas do you have for how to implement this? I can't promise it will be merged, but feel free to open a Pull Request and we can take a look together.

adasari:
At high level, below changes should work. (not tested yet)

1. Following code in [handler.go](https://github.com/grpc-ecosystem/grpc-gateway/blob/0c9d3eb7ca8e28a7c8db327d6ef63902bfbfd778/runtime/handler.go#L64)
```
var buf []byte
switch {
case resp == nil:
	buf, err = marshaler.Marshal(errorChunk(streamError(ctx, mux.streamErrorHandler, errEmptyResponse)))
default:
	if rb, ok := resp.(responseBody); ok {
		buf, err = marshaler.Marshal(rb.XXX_ResponseBody())
	} else {
		buf, err = marshaler.Marshal(map[string]proto.Message{"result": resp})
	}
}
```
[...]

PR #1205, opened 2020-04-09
responseBody changes for sever stream rpc

adasari:
HI,

The changes regarding https://github.com/grpc-ecosystem/grpc-gateway/issues/1202. 

what is changed ?
1. Added response body changes for runtime/handler.go and test cases
2. template.go
4. stream rpc with response body http rule field example and integration tests
5. Existing test cases are failing in local due to port busy issue. changed the port of test cases so that it never fail.
6. auto generated files from bazel sync

Please review. thanks.


adasari:
@googlebot I signed it!

PR #1208, opened 2020-04-13
http rule's response_body support to stream rpc

adasari:
The changes regarding https://github.com/grpc-ecosystem/grpc-gateway/issues/1202
what is changed ?

1. Added response body changes for runtime/handler.go and test cases
2. template.go
3. stream rpc with response body http rule field example and integration tests
4. Existing test cases are failing in local due to port busy issue. changed the port of test cases so that it never fail.
5. auto generated files from bazel sync
Please review. thanks.


adasari:
@johanbrandhorst finally all passed. able to figure out managing branches of fork in hard way :). got delayed. Please take a look at now. 

can we delete closed pull requests 1205 and 1206 ?

examples/internal/integration/integration_test.go
```diff
@@ -1526,20 +1546,20 @@ func testResponseBodies(t *testing.T, port int) {
...
 	go func() {
-		if err := runGateway(ctx, ":8081", runtime.WithMarshalerOption(runtime.MIMEWildcard, &runtime.JSONPb{EnumsAsInts: false, EmitDefaults: true})); err != nil {
+		if err := runGateway(ctx, fmt.Sprintf(":%d", port), runtime.WithMarshalerOption(runtime.MIMEWildcard, &runtime.JSONPb{EnumsAsInts: false, EmitDefaults: true})); err != nil {
 			ch <- fmt.Errorf("cannot run gateway service: %v", err)
 		}
 	}()
 
-	if err := waitForGateway(ctx, 8081); err != nil {
+	if err := waitForGateway(ctx, uint16(port)); err != nil {
 		t.Fatalf("waitForGateway(ctx, 8081) failed with %v; want success", err)
```
johanbrandhorst, line 1560: This comment is wrong now

examples/internal/integration/integration_test.go
```diff
@@ -1476,26 +1476,46 @@ func TestResponseBody(t *testing.T) {
...
+	}{{
+		name:         "unary case",
+		url:          "http://localhost:%d/responsebody/foo",
+		wantStatus:   http.StatusOK,
+		wantResponse: `{"data":"foo"}`,
+	}, {
+		name:         "stream case",
+		url:          "http://localhost:%d/responsebody/stream/foo",
+		wantStatus:   http.StatusOK,
+		wantResponse: `{"data":"foo"}`,
```
johanbrandhorst, line 1493: This is not the response I would expect from a streaming RPC. It should be wrapped in the result envelope:
```json
{"result": {"data":"foo"}, "error":""}
```

johanbrandhorst: I'd also like to see this test separated from the unary case and test that we can stream two separate messages.

examples/internal/proto/examplepb/response_body_service.swagger.json
```diff
@@ -44,6 +44,45 @@
...
+      "get": {
+        "operationId": "ResponseBodyService_GetResponseBodyStream",
+        "responses": {
+          "200": {
+            "description": "(streaming responses)",
+            "schema": {
+              "type": "object",
+              "properties": {
+                "result": {
+                  "$ref": "#/definitions/examplepbResponseBodyOutResponse"
```
johanbrandhorst, line 57: This definition is not correct now, it should point to a `string` field with name `data`.

examples/internal/server/responsebody.go
```diff
@@ -42,3 +42,11 @@ func (s *responseBodyServer) ListResponseStrings(ctx context.Context, req *examp
 		Values: []string{"hello", req.Data},
 	}, nil
 }
+
+func (s *responseBodyServer) GetResponseBodyStream(req *examples.ResponseBodyIn, stream examples.ResponseBodyService_GetResponseBodyStreamServer) error {
+	return stream.Send(&examples.ResponseBodyOut{
+		Response: &examples.ResponseBodyOut_Response{
+			Data: req.Data,
+		},
+	})
```
johanbrandhorst, line 51: Lets send two messages on the stream, to test the streaming.

runtime/handler.go
```diff
@@ -61,7 +61,18 @@ func ForwardResponseStream(ctx context.Context, mux *ServeMux, marshaler Marshal
...
 		}
 
-		buf, err := marshaler.Marshal(streamChunk(ctx, resp, mux.streamErrorHandler))
+		var buf []byte
+		switch {
+		case resp == nil:
+			buf, err = marshaler.Marshal(errorChunk(streamError(ctx, mux.streamErrorHandler, errEmptyResponse)))
+		default:
+			if rb, ok := resp.(responseBody); ok {
+				buf, err = marshaler.Marshal(rb.XXX_ResponseBody())
```
johanbrandhorst, line 70: Please wrap this with the result envelope
```suggestion
				buf, err = marshaler.Marshal(map[string]proto.Message{"result": rb.XXX_ResponseBody()})
```

runtime/handler_test.go
```diff
@@ -113,7 +140,21 @@ func TestForwardResponseStream(t *testing.T) {
...
+				var b []byte
+
+				if tt.responseBody {
+					// responseBody interface is in runtime package and test is in runtime_test package. hence can't use responseBody direclty
+					// So type casting to fakeReponseBodyWrapper struct to verify the data.
+					rb, ok := msg.pb.(fakeReponseBodyWrapper)
+					if !ok {
+						t.Errorf("stream responseBody failed %v", err)
+					}
+					b, err = marshaler.Marshal(rb.XXX_ResponseBody())
```
johanbrandhorst, line 153: Will need to be updated with the result envelope.

runtime/handler.go
adasari: May i know reason to wrap this with result envelope ?

examples/internal/proto/examplepb/response_body_service.swagger.json
adasari: It is auto-generated file. Should it require any other changes to correct it ?

examples/internal/integration/integration_test.go
adasari: Updated

runtime/handler.go
johanbrandhorst: Of course :). The result envelope is there so that we can return a consistent structure for streaming messages, which can fail at any point in the stream. The envelope always has the structure `{"result":<message>, "error": <any error}`. This guarantees consistent JSON parsing. I want to preserve this property for this feature. If you need complete control over the streaming output, you may want to consider the `google.protobuf.HttpBody` type which is printed verbatim to the output.

examples/internal/proto/examplepb/response_body_service.swagger.json
johanbrandhorst: Yeah, you will need to have a look at the logic in the swagger generator to make sure it uses the ResponseBody type when it is specified in the proto options. This may already be done for the unary case, so might not be too bad to port. See https://github.com/grpc-ecosystem/grpc-gateway/blob/6b0fe46fad37b6eef2b5c32ded5aed9af08911d6/protoc-gen-swagger/genswagger/template.go#L912-L919

runtime/handler.go
adasari: `rb.XXX_ResponseBody()` is not of proto.Message type. it will be `marshaler.Marshal(map[string]interface{}{"result": rb.XXX_ResponseBody()})`. is it ok ?

johanbrandhorst: What type is it? Can we make it anything other than `interface{}`? If it'

johanbrandhorst: Hm, I see, it's `interface{}`. Yeah I guess that will have to do :)

examples/internal/proto/examplepb/response_body_service.swagger.json
adasari: i checked the swagger json, it looks good to me.
```
"responses": {
          "200": {
            "description": "(streaming responses)",
            "schema": {
              "type": "object",
              "properties": {
                "result": {
                  "$ref": "#/definitions/examplepbResponseBodyOutResponse"
                },
                "error": {
                  "$ref": "#/definitions/runtimeStreamError"
                }
              },
              "title": "Stream result of examplepbResponseBodyOut"
            }
          },
          "default": {
            "description": "An unexpected error response",
            "schema": {
              "$ref": "#/definitions/runtimeError"
            }
          }
        }

"examplepbResponseBodyOutResponse": {
      "type": "object",
      "properties": {
        "data": {
          "type": "string"
        }
      }
    }
```
without response body:
```
"responses": {
          "200": {
            "description": "A successful response.(streaming responses)",
            "schema": {
              "type": "object",
              "properties": {
                "result": {
                  "$ref": "#/definitions/examplepbResponseBodyOut"
                },
                "error": {
                  "$ref": "#/definitions/runtimeStreamError"
                }
              },
              "title": "Stream result of examplepbResponseBodyOut"
            }
          },
          "default": {
            "description": "An unexpected error response",
            "schema": {
              "$ref": "#/definitions/runtimeError"
            }
          }
        },


"examplepbResponseBodyOut": {
      "type": "object",
      "properties": {
        "response": {
          "$ref": "#/definitions/examplepbResponseBodyOutResponse"
        }
      }
    }
```

johanbrandhorst: Nice, you're right! Sorry about that.

PR #1213, opened 2020-04-17
remove deprecated usage of github.com/golang/protobuf/protoc-gen-go/generator

yinzara:
The github.com/golang/protobuf/protoc-gen-go/generator package is now deprecated and causes warnings when used.  This causes this plugin to fail when used with some protoc related tools (like prototool) as they expect all plugins to not write to stdout/stderr except for returning the generation results.

We only use the "CamelCase" function from that package so I just copied it into the template.go


yinzara:
Hmmm so now it's failing for a reason I have no idea about.

johanbrandhorst:
Gorelease is failing because you're not up to date with master. Please rebase your branch. The bazel failure looks like a flaky test, unfortunately, I reran it.

yinzara:
There. Fixed :)

johanbrandhorst:
Thanks for your contribution! Could you please cherry pick this against the v2 branch as well?

PR #1216, opened 2020-04-18
Add option to disable new operationID generation

johanbrandhorst:
The new generation model, while correct, was causing
friction with some users. Add a flag to re-enable the old
behaviour.

CC @aselus-hub, @mouradmourafiq @dadgar


mouradmourafiq:
Amazing, thank you 👍 @johanbrandhorst

Issue #1226, opened 2020-04-21
Add WithUnmarshaler NewServeMux option.


johanbrandhorst:
Users often ask to be able to disallow fields, or control unmarshaling of the incoming JSON in some way. We've long had `WithMarshaler`, but not the equivalent for unmarshaling.

johanbrandhorst:
This might not be necessary after all.

PR #1228, opened 2020-04-21
add runtime.WithUnmarshalerOption

AyushG3112:
This adds `runtime.WithUnmarshalOption` to allow use of a custom Unmarshaller. 

- Adds `runtime.WithUnmarshalOption`
- Split `runtime.Marshaler` into `runtime.Marshaler` and `runtime.Unmarshaler`
- Make methods of `runtime.Marshaler` and `runtime.Unmarshaler` accept `context.Context`

Related: https://github.com/grpc-ecosystem/grpc-gateway/issues/1226

### DIFF
--- a/runtime/marshaler.go
+++ b/runtime/marshaler.go
@@ -1,20 +1,27 @@
 package runtime
 
 import (
+	"context"
 	"io"
 )
 
-// Marshaler defines a conversion between byte sequence and gRPC payloads / fields.
+// Marshaler defines a outbound conversion between byte sequence and gRPC payloads / fields.
 type Marshaler interface {
 	// Marshal marshals "v" into byte sequence.
-	Marshal(v interface{}) ([]byte, error)
+	Marshal(ctx context.Context, v interface{}) ([]byte, error)
+	// NewEncoder returns an Encoder which writes bytes sequence into "w".
+	NewEncoder(ctx context.Context, w io.Writer) Encoder
+	// ContentType returns the Content-Type which this marshaler is responsible for.
+	ContentType() string
+}
+
+// Unmarshaler defines a imbound conversion between byte sequence and gRPC payloads / fields.
+type Unmarshaler interface {
 	// Unmarshal unmarshals "data" into "v".
 	// "v" must be a pointer value.
-	Unmarshal(data []byte, v interface{}) error
+	Unmarshal(ctx context.Context, data []byte, v interface{}) error
 	// NewDecoder returns a Decoder which reads byte sequence from "r".
-	NewDecoder(r io.Reader) Decoder
-	// NewEncoder returns an Encoder which writes bytes sequence into "w".
-	NewEncoder(w io.Writer) Encoder
+	NewDecoder(ctx context.Context, r io.Reader) Decoder
 	// ContentType returns the Content-Type which this marshaler is responsible for.
 	ContentType() string
 }
--- a/runtime/marshaler.go
+++ b/runtime/marshaler.go
@@ -9,7 +9,7 @@ import (
 type Marshaler interface {
 	// Marshal marshals "v" into a byte sequence.
 	Marshal(ctx context.Context, v interface{}) ([]byte, error)
-	// NewEncoder returns an Encoder which writes bytes sequence into "w".
+	// NewEncoder returns an Encoder which writes byte sequences into "w".
 	NewEncoder(ctx context.Context, w io.Writer) Encoder
 	// ContentType returns the Content-Type which this marshaler is responsible for.
 	ContentType() string
@@ -20,7 +20,7 @@ type Unmarshaler interface {
 	// Unmarshal unmarshals "data" into "v".
 	// "v" must be a pointer value.
 	Unmarshal(ctx context.Context, data []byte, v interface{}) error
-	// NewDecoder returns a Decoder which reads byte sequence from "r".
+	// NewDecoder returns a Decoder which reads byte sequences from "r".
 	NewDecoder(ctx context.Context, r io.Reader) Decoder
 	// ContentType returns the Content-Type which this marshaler is responsible for.
 	ContentType() string
--- a/runtime/marshaler.go
+++ b/runtime/marshaler.go
@@ -5,9 +5,9 @@ import (
 	"io"
 )
 
-// Marshaler defines a outbound conversion between byte sequence and gRPC payloads / fields.
+// Marshaler defines an outbound conversion between a gRPC payload and a byte sequence.
 type Marshaler interface {
-	// Marshal marshals "v" into byte sequence.
+	// Marshal marshals "v" into a byte sequence.
 	Marshal(ctx context.Context, v interface{}) ([]byte, error)
 	// NewEncoder returns an Encoder which writes bytes sequence into "w".
 	NewEncoder(ctx context.Context, w io.Writer) Encoder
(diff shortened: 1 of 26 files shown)


runtime/marshaler.go
```diff
@@ -1,20 +1,27 @@
 package runtime
 
 import (
+	"context"
 	"io"
 )
 
-// Marshaler defines a conversion between byte sequence and gRPC payloads / fields.
+// Marshaler defines a outbound conversion between byte sequence and gRPC payloads / fields.
```
johanbrandhorst, line 8:
```suggestion
// Marshaler defines an outbound conversion between a gRPC payload and a byte sequence.
```

runtime/marshaler.go
```diff
@@ -1,20 +1,27 @@
...
 type Marshaler interface {
 	// Marshal marshals "v" into byte sequence.
-	Marshal(v interface{}) ([]byte, error)
+	Marshal(ctx context.Context, v interface{}) ([]byte, error)
+	// NewEncoder returns an Encoder which writes bytes sequence into "w".
```
johanbrandhorst, line 12:
```suggestion
	// NewEncoder returns an Encoder which writes byte sequences into "w".
```

runtime/marshaler.go
```diff
@@ -1,20 +1,27 @@
...
 
 import (
+	"context"
 	"io"
 )
 
-// Marshaler defines a conversion between byte sequence and gRPC payloads / fields.
+// Marshaler defines a outbound conversion between byte sequence and gRPC payloads / fields.
 type Marshaler interface {
 	// Marshal marshals "v" into byte sequence.
```
johanbrandhorst, line 10:
```suggestion
	// Marshal marshals "v" into a byte sequence.
```

runtime/marshaler.go
```diff
@@ -1,20 +1,27 @@
...
-	Marshal(v interface{}) ([]byte, error)
+	Marshal(ctx context.Context, v interface{}) ([]byte, error)
+	// NewEncoder returns an Encoder which writes bytes sequence into "w".
+	NewEncoder(ctx context.Context, w io.Writer) Encoder
+	// ContentType returns the Content-Type which this marshaler is responsible for.
+	ContentType() string
+}
+
+// Unmarshaler defines a imbound conversion between byte sequence and gRPC payloads / fields.
```
johanbrandhorst, line 18:
```suggestion
// Unmarshaler defines an inbound conversion between a byte sequence and a gRPC payload.
```

runtime/marshaler.go
```diff
@@ -1,20 +1,27 @@
...
+type Unmarshaler interface {
 	// Unmarshal unmarshals "data" into "v".
 	// "v" must be a pointer value.
-	Unmarshal(data []byte, v interface{}) error
+	Unmarshal(ctx context.Context, data []byte, v interface{}) error
 	// NewDecoder returns a Decoder which reads byte sequence from "r".
```
johanbrandhorst, line 23:
```suggestion
	// NewDecoder returns a Decoder which reads byte sequences from "r".
```

runtime/marshaler_registry.go
```diff
@@ -16,42 +16,67 @@ var (
 	defaultMarshaler = &JSONPb{OrigName: true}
 )
 
-// MarshalerForRequest returns the inbound/outbound marshalers for this request.
-// It checks the registry on the ServeMux for the MIME type set by the Content-Type header.
+// MarshalerForRequest returns the outbound marshalers for this request.
+// It checks the registry on the ServeMux for the MIME type set by the Accept header.
+// If it isn't set (or the request Accept is empty), it checks for the Content-Type header.
 // If it isn't set (or the request Content-Type is empty), checks for "*".
```
johanbrandhorst, line 22:
```suggestion
// If it isn't set (or the request Content-Type is empty), it checks for "*".
```

runtime/marshaler_registry.go
```diff
@@ -16,42 +16,67 @@ var (
...
 // If there are multiple Content-Type headers set, choose the first one that it can
 // exactly match in the registry.
-// Otherwise, it follows the above logic for "*"/InboundMarshaler/OutboundMarshaler.
-func MarshalerForRequest(mux *ServeMux, r *http.Request) (inbound Marshaler, outbound Marshaler) {
+// Otherwise, it follows the above logic for "*"/OutboundMarshaler.
```
johanbrandhorst, line 25: I know you only slightly rewrote this, but I'm not sure I understand what this means. If you understand what it means, could you explain? If you are also confused, please just remove this line.

runtime/marshaler_registry.go
```diff
@@ -16,42 +16,67 @@ var (
...
+		}
+	}
+	if outbound == nil {
+		outbound = mux.marshalers.mimeMap[MIMEWildcard]
+	}
+
+	return outbound
+}
+
+// UnmarshalerForRequest returns the inbound marshalers for this request.
```
johanbrandhorst, line 49:
```suggestion
// UnmarshalerForRequest returns the inbound marshaler for this request.
```

runtime/marshaler_registry.go
```diff
@@ -16,42 +16,67 @@ var (
...
+// It checks the registry on the ServeMux for the MIME type set by the Content-Type header.
+// If it isn't set (or the request Content-Type is empty), checks for "*".
+// If there are multiple Content-Type headers set, choose the first one that it can
+// exactly match in the registry.
+// Otherwise, it follows the above logic for "*"/InboundMarshaler/OutboundMarshaler.
```
johanbrandhorst, line 54: Same here

runtime/marshaler.go
AyushG3112: I think `byte sequence` makes more sense here, because it is a single sequence of bytes per call, WDYT?

runtime/marshaler.go
AyushG3112: Same here.

johanbrandhorst:
Please rebase on v2

AyushG3112:
@johanbrandhorst Done.

AyushG3112:
@johanbrandhorst this will need your consent because you're the co-author of https://github.com/grpc-ecosystem/grpc-gateway/pull/1228/commits/fef93928cec20aba92958513a7a88f20a1b5f3c0

runtime/marshaler.go
johanbrandhorst: I think in the general case that this `interface` describes, we want an encoder that _can_ write an arbitrary number of byte sequences (for example, when streaming).

AyushG3112: @johanbrandhorst I see. Commited these changes in https://github.com/grpc-ecosystem/grpc-gateway/pull/1228/commits/35d0c348983bd14912d292a75dbbac47803569a1

AyushG3112:
If someone only wants to implement a custom unmarshaler, a single interface will force them to implement the `Encode` and `Marshal` methods too. Multiple interfaces prevent that, and the user can pass the same implementation for both interfaces if they want.

johanbrandhorst:
> If someone only wants to implement a custom unmarshaler, a single interface will force them to implement stubs for the `Encode` and `Marshal` methods too. Multiple interfaces prevent that, and the user can pass the same implementation for both interfaces if they want.

That is true, but in the grand scheme of things, I think this are simpler with a single interface, and in the use case where a user only wants to implement the unmarshaling part, they can simply embed the default marshaler. I think most cases will be about changing the marshaler, and the user will modify embed the default marshaller to avoid modifying the unmarshalling.

It also makes me wonder if we already had this capability, why did we add `DisallowUnknownFields`? I'm looking at https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/marshaler_registry.go and it looks to me like all one has to do to get custom unmarshaling behaviour is to define your own marshaler and embed the default marshaler. Am I missing something?

johanbrandhorst:
I believe we concluded that this change was not necessary in the end, so I'm closing it. Looking forward to future contributions from you Ayush :).

PR #1234, opened 2020-04-22
v2: set rpcMethodName in request.Context

AyushG3112:
This sets  a value `rpcMethodName` in the request context. It follows the same pattern as the value set by gRPC as `/package.Service/Method`

PR #1236, opened 2020-04-22
v1: set rpcMethodName in request.Context

AyushG3112:
Same as https://github.com/grpc-ecosystem/grpc-gateway/pull/1234 but from the master branch as base.
Merging this will need the user to regenerate the definitions though.


AyushG3112:
Closing, backward incompatible as this will need user to generate the generated files.

PR #1242, opened 2020-04-25
#1098: remove error field from error messages

### LINKED ISSUE #1098: Remove "error" field from errors.
```go
body := &errorBody{
		Error:   s.Message(),
		Message: s.Message(),
		Code:    int32(s.Code()),
		Details: s.Proto().GetDetails(),
}
```

Here ([link](https://github.com/grpc-ecosystem/grpc-gateway/blob/3c06998610d4d2edea99d12228b1f6ea2d943ce4/runtime/errors.go#L109))

The Error and Message both field got s.Message(), I think it's not clean to duplicate the same thing. I don't know if there's any special intention behind it.

Please let us know. Thanks.

srenatus:
Fixes #1098.

⚠️ Doesn't remove `internal/errors.proto`'s `StreamError`. That can be done next ;)


internal/errors.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.21.0
-// 	protoc        v3.10.1
+// 	protoc        v3.7.0
```
srenatus, line 4: I can remove this, or commit it from all the other `.pb.go` files where the bazel run has changed just this line. What do you think?

johanbrandhorst: I think we should be able to remove this

srenatus: (How) do we control the protoc version used?

johanbrandhorst: It's in the build environment [Dockerfile](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/.circleci/Dockerfile#L20) at the moment. Clearly this file isn't being regenerated.

internal/errors.proto
```diff
@@ -4,17 +4,6 @@ option go_package = "internal";
...
-	// https://github.com/grpc/grpc/blob/master/src/proto/grpc/status/status.proto
-	// It should be the exact same message as the Error field.
-	int32 code = 2;
-	string message = 3;
-	repeated google.protobuf.Any details = 4;
-}
-
 // StreamError is a response type which is returned when
 // streaming rpc returns an error.
 message StreamError {
```
johanbrandhorst, line 9: Can we delete this too? Not sure we need it to be anything other than a `status` either.

srenatus: Sure! I'll get rid of it once I've dealt with the fallout of removing `Error`. 👍

srenatus:
Looks like removing the Error message type has implications on the openapi spec:
- it's processed in [protoc-gen-swagger/genswagger/template.go](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/protoc-gen-swagger/genswagger/template.go#L966-L977)
- and put into the specs, like this example https://github.com/grpc-ecosystem/grpc-gateway/blob/v2/examples/internal/clients/echo/api/swagger.yaml#L525-L538

Not a blocker, just something I hadn't had on my radar. I'll see into replacing that with the status message (similar to `protobufAny`).

srenatus:
So far, the [generate failures](https://circleci.com/gh/grpc-ecosystem/grpc-gateway/7412?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) look reasonable. Let's drop `StreamError` next.

srenatus:
I haven't fully understood how the StreamError handling works -- I think I'd rather we got this in, and I (or someone else) pick up removing StreamError next. What do you think?

examples/internal/integration/integration_test.go
```diff
@@ -32,7 +32,7 @@ import (
 )
 
 type errorBody struct {
-	Error   string        `json:"error"`
+	Message string        `json:"message"`
 	Code    int           `json:"code"`
 	Details []interface{} `json:"details"`
 }
```
johanbrandhorst, line 38: Could this type be a `status.Status`?

runtime/errors.go
```diff
@@ -116,6 +115,7 @@ func DefaultHTTPError(ctx context.Context, mux *ServeMux, marshaler Marshaler, w
 	if !ok {
 		s = status.New(codes.Unknown, err.Error())
 	}
+	pb := s.Proto()
```
johanbrandhorst, line 118: We can replace the code above this like with `status.Convert(err)` (it's equivalent AFAIK).

runtime/errors.go
```diff
@@ -116,6 +115,7 @@ func DefaultHTTPError(ctx context.Context, mux *ServeMux, marshaler Marshaler, w
 	if !ok {
```
johanbrandhorst, line 115: The docstring of this function is no longer accurate.

examples/internal/integration/integration_test.go
srenatus: I believe we've got a test case doing that already. I'll double-check and either remove this one or change it.

srenatus: 👉 https://github.com/grpc-ecosystem/grpc-gateway/blob/e6d05361fee32e0f75b57de7b98b9455742384a3/examples/internal/integration/proto_error_test.go#L71 

Looks like we can remove this one, then. Unless I'm mistaking their intention.

srenatus:
@johanbrandhorst I've done a quick grep, but I'm afraid I wasn't very thorough.

johanbrandhorst:
Great work Stephan! Lets get this in :). You've reminded me that we need to have a general look at the error functions, I believe we have too many ways to configure error handling at the moment.

ahrtr:
This PR is a breaking change, could you please add the error field back? Please refer to https://github.com/etcd-io/etcd/issues/21100#issuecomment-3760942241

Thanks

PR #1252, opened 2020-04-29
ADOPTERS: add Chef

srenatus:


johanbrandhorst:
The CI job is failing because we don't give environment secrets to forked PRs (for obvious reasons). I think we may have to find a different way of authenticating to github. I'm gonna try creating a dummy user, since read access should be open to anyone with a token.

johanbrandhorst:
https://github.com/grpc-ecosystem/grpc-gateway/pull/1253 _sigh_. Might have to revert the build package change.

johanbrandhorst:
For now I've created https://github.com/grpc-ecosystem/grpc-gateway/pull/1255 so we can get this in

PR #1253, opened 2020-04-29
Use public key to authenticate for docker registry

johanbrandhorst:
The token belongs to a user with no membership anywhere,
and it is limited to only reading github packages.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,34 +1,45 @@
 version: 2
 jobs:
   build:
+    environment:
+    ## Split key to avoid github revoking it
+      password0: '99544cdcb19ad4e3fd64'
+      password1: '3ec86b2e5a431be2d72c'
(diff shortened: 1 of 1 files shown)


johanbrandhorst:
Interestingly, it seems Github detected that I committed a valid token and automatically revoked it. Nice work Github, but actually I _did_ know what I was doing this time :cry:. I've reached out to github support to see what I can do.

achew22:
We could switch to github actions which are able to do this, I believe

srenatus:
https://github.com/nektos/act looks useful when doing github actions...

johanbrandhorst:
I'll definitely give github actions a try - the overall win of not depending on my personal docker hub is well worth it.

johanbrandhorst:
Ha, splitting the token, as suggested by github support, works! Good enough.

PR #1262, opened 2020-04-30
runtime: replace StreamError with status.Status

### LINKED ISSUE #1098: Remove "error" field from errors.

johanbrandhorst:
Following on from #1242, this replaces the StreamError
with a status.Status type. Also, remove the ability to configure
the stream error handler. The existing handler was specific to the
old type, and we can add something better back in later
if necessary.

Fixes #1098


johanbrandhorst:
I would like some feedback on whether we think this is going to impact users, since we now don't included the translated HTTP status code anymore. Personally, I think it makes sense to align with `status.Status` here.

PR #1265, opened 2020-05-02
Consolidate error handling configuration

johanbrandhorst:
All unary responses are now handled with a single function,
and are configured in a single place.

Would appreciate some feedback on this as it removes a lot of API.


runtime/mux.go
```diff
@@ -239,34 +226,22 @@ func (s *ServeMux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 				return
 			}
-			if s.protoErrorHandler != nil {
-				_, outboundMarshaler := MarshalerForRequest(s, r)
-				s.protoErrorHandler(ctx, s, outboundMarshaler, w, r, ErrUnknownURI)
-			} else {
-				OtherErrorHandler(w, r, http.StatusText(http.StatusMethodNotAllowed), http.StatusMethodNotAllowed)
-			}
+			_, outboundMarshaler := MarshalerForRequest(s, r)
+			s.errorHandler(ctx, s, outboundMarshaler, w, r, ErrUnknownURI)
```
johanbrandhorst, line 238: The handler now returns `http.StatusNotImplemented` rather than `http.StatusMethodNotAllowed` when a request with no matching Method is received.

runtime/mux.go
```diff
@@ -239,34 +226,22 @@ func (s *ServeMux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 	}
 
-	if s.protoErrorHandler != nil {
-		_, outboundMarshaler := MarshalerForRequest(s, r)
-		s.protoErrorHandler(ctx, s, outboundMarshaler, w, r, ErrUnknownURI)
-	} else {
-		OtherErrorHandler(w, r, http.StatusText(http.StatusNotFound), http.StatusNotFound)
-	}
+	_, outboundMarshaler := MarshalerForRequest(s, r)
+	s.errorHandler(ctx, s, outboundMarshaler, w, r, ErrUnknownURI)
```
johanbrandhorst, line 244: The handler now returns `http.StatusNotImplemented` rather than `http.StatusNotFound` when no path was found. This is more consistent with gRPC in general.

johanbrandhorst:
@jgiles I would appreciate your thoughts on this redesign as a user.

johanbrandhorst:
This will need documentation changes

runtime/mux.go
jhump: > This is more consistent with gRPC in general.

But it is totally inconsistent with typical REST APIs. The HTTP client for grpc-gateway is using a REST API so those idioms feel much more appropriate than trying to mirror gRPC's behavior.

docs/_docs/customizingyourgateway.md
```diff
@@ -234,91 +234,19 @@ if err := pb.RegisterMyServiceHandlerFromEndpoint(ctx, mux, serviceEndpoint, opt
...
-
-If no custom handler is provided, the default stream error handler
-will include any gRPC error attributes (code, message, detail messages),
-if the error being reported includes them. If the error does not have
-these attributes, a gRPC code of `Unknown` (2) is reported. The default
-handler will also include an HTTP code and status, which is derived
-from the gRPC code (or set to `"500 Internal Server Error"` when
-the source error has no gRPC attributes).
+to RPC methods that have a unary response. It is currently
+not possible to configure the stream error handler.
```
jhump, line 249: This would be a disappointing outcome. We may have to maintain a fork at FullStory, to continue to tweak the error output for a server-stream call.

runtime/mux.go
johanbrandhorst: I'm happy to reverse this change, it wasn't clear to me which way to go. The same argument can be extended to error formats in general, do you think it makes sense to rely on the status.Status type as an error still?

runtime/errors.go
```diff
@@ -10,6 +10,9 @@ import (
 	"google.golang.org/grpc/status"
 )
 
+// ErrorHandlerFunc handles the error as a gRPC error generated via status package and replies to the request.
+type ErrorHandlerFunc func(context.Context, *ServeMux, Marshaler, http.ResponseWriter, *http.Request, error)
```
jhump, line 14: A single signature for error handling is great, but to support streaming errors, this would need more info -- mainly it needs to know whether or not the HTTP status and headers have already been sent. That way the handler can choose to write to the body in this case.

runtime/mux.go
jhump: I think using `status.Status` for every error is nice for uniformity. But since it's was designed to convey the result of an RPC operation, it can be awkward for certain kinds of errors.

Perhaps when this status is constructed, you could add a gateway-specific error detail to it, making it clear that the error was caused by not being able to map the given URI to a known RPC method. Then, as long as the hook for customizing the error response is used for these errors too, you could do the "not implemented" thing by default. Those that prefer to render the errors as "not found" can still do so with a custom error handler while also being able to preserve "not implemented" errors from other situations (like when that is actually the code in the RPC server's response).

johanbrandhorst: I think you're right that this error code is going to be confusing for most users, and I'm simply going to revert the change in behaviour. Thanks for your feedback!

docs/_docs/customizingyourgateway.md
johanbrandhorst: Reinstated the stream error handler.

runtime/mux.go
johanbrandhorst: This is still true, because there wasn't a better translation within the scope of the protobuf `codes`

PR #1272, opened 2020-05-07
runtime: rewrite query.go in new protoreflect

johanbrandhorst:
Replaces lots of go reflection with protoreflect from APIv2. This is an essential part of rooting out the old API from our packages. Unfortunately, we also lose the capability to support gogoproto `stdtime` and `stdduration` types, since they cannot be represented in the new reflection interface.

If you do want to review this, I suggest opening the new file separately, the diff is mostly unhelpful since it's a complete rewrite.


runtime/query.go
```diff
@@ -45,362 +46,294 @@ func (*defaultQueryParser) Parse(msg proto.Message, values url.Values, filter *u
...
-	var name string
-	switch m := i.(type) {
-	case interface{ XXX_WellKnownType() string }:
-		name = m.XXX_WellKnownType()
-	case proto.Message:
-		const wktPrefix = "google.protobuf."
-		if fullName := proto.MessageName(m); strings.HasPrefix(fullName, wktPrefix) {
-			name = fullName[len(wktPrefix):]
-		}
+	if msgDesc.FullName().Parent() != "google.protobuf" {
```
bufdev, line 248: Is `Parent` documented to not use "actual" fully-qualified names, ie `.google.protobuf`? You might have to handle this case as well.

Also technically, this doesn't check for WKT - you need to check the actual files. There's lists at various places, such as https://github.com/bufbuild/buf/blob/master/internal/gen/embed/wkt/wkt.gen.go, you could make a `map[string]struct{}` with the WKT file names.

runtime/query.go
```diff
@@ -45,362 +46,294 @@ func (*defaultQueryParser) Parse(msg proto.Message, values url.Values, filter *u
...
 
 		t, err := time.Parse(time.RFC3339Nano, value)
 		if err != nil {
-			return fmt.Errorf("bad Timestamp: %v", err)
+			return protoreflect.Value{}, fmt.Errorf("parsing Timestamp: %w", err)
 		}
-		f.FieldByName("Seconds").SetInt(int64(t.Unix()))
-		f.FieldByName("Nanos").SetInt(int64(t.Nanosecond()))
-		return nil
+		msg.Set(msgDesc.Fields().ByName("seconds"), protoreflect.ValueOfInt64(t.Unix()))
```
bufdev, line 263: Is there no other way to do this, ie convert message to a proper `Timestamp` and then set the fields?

runtime/marshal_jsonpb.go
```diff
@@ -260,3 +260,17 @@ var allowUnknownFields = true
 func DisallowUnknownFields() {
 	allowUnknownFields = false
 }
+
+var (
+	convFromType = map[reflect.Kind]reflect.Value{
```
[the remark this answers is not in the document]
johanbrandhorst: This was simply moved from the query.go file, and is referenced elsewhere in this file. Renaming it is outside the scope of this change.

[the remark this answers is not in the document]
bufdev: Fair enough - it is a pet peeve of mine when stuff that is moved gets commented on, so carry on :-)

runtime/query.go
```diff
@@ -45,362 +46,294 @@ func (*defaultQueryParser) Parse(msg proto.Message, values url.Values, filter *u
...
-		if !isLast && m.Kind() != reflect.Struct {
-			return fmt.Errorf("non-aggregate type in the mid of path: %s", strings.Join(fieldPath, "."))
+		fields := msgValue.Descriptor().Fields()
+
+		// Get field by name
+		fieldDesc = fields.ByName(protoreflect.Name(fieldName))
+		if fieldDesc == nil {
+			fieldDesc = fields.ByJSONName(fieldName)
+			if fieldDesc == nil {
+				grpclog.Infof("field not found in %q: %q", msgValue.Descriptor().FullName(), strings.Join(fieldPath, "."))
```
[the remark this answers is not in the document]
johanbrandhorst: This was put here to preserve the old behaviour, and I think the intention is to simply ignore extraneous query parameters (which could include tracking parameters that aren't necessarily part of the gateway request).

runtime/query.go
```diff
@@ -45,362 +46,294 @@ func (*defaultQueryParser) Parse(msg proto.Message, values url.Values, filter *u
...
+
+	switch {
+	case fieldDesc.IsList():
+		return populateRepeatedField(fieldDesc, msgValue, values)
+	case fieldDesc.IsMap():
+		return populateMapField(fieldDesc, msgValue, values)
+	}
+
+	if len(values) > 1 {
+		grpclog.Infof("too many field values for %q: %q", fieldDesc.FullName().Name(), strings.Join(fieldPath, "."))
```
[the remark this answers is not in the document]
johanbrandhorst: Yeah this probably should be an error, though it will be a change of behaviour.

runtime/query.go
```diff
@@ -45,362 +46,294 @@ func (*defaultQueryParser) Parse(msg proto.Message, values url.Values, filter *u
...
-	for i, v := range values {
-		result := conv.Call([]reflect.Value{reflect.ValueOf(v)})
-		if err := result[1].Interface(); err != nil {
-			return err.(error)
+func parseField(fieldDesc protoreflect.FieldDescriptor, msgValue protoreflect.Message, value string) (protoreflect.Value, error) {
+	switch fieldDesc.Kind() {
+	case protoreflect.BoolKind:
+		v, err := strconv.ParseBool(value)
+		if err != nil {
+			return protoreflect.Value{}, fmt.Errorf("parsing boolean: %w", err)
```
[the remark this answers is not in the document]
johanbrandhorst: Yep, this is superfluous

runtime/query.go
```diff
@@ -45,362 +46,294 @@ func (*defaultQueryParser) Parse(msg proto.Message, values url.Values, filter *u
...
-		case reflect.Struct:
-			m = f
-			continue
-		case reflect.Map:
-			if !isLast {
-				return fmt.Errorf("unexpected nested field %s in %s", fieldPath[i+1], strings.Join(fieldPath[:i+1], "."))
-			}
-			return populateMapField(f, values, props)
+		switch fieldDesc.Kind() {
+		case protoreflect.MessageKind, protoreflect.GroupKind:
```
dsnet, line 90:
```go
if fieldDesc.Message() != nil && field.Cardinality() != protoreflect.Repeated
```

It seems that the intent is to ensure that this is a *singular* message field. The current check will let lists of messages and maps through the cracks.

runtime/query.go
dsnet, line 114: This error seems a bit redundant. It gives the short name of the current field and also the full fieldPath, where the last segment of the fieldPath should be more or less identical to the name of the current field.

runtime/query.go
dsnet: The protoreflect.FullName type is documented to never include the preceding "." by virtue of the type itself being called "FullName".

dsnet: > Also technically, this doesn't check for WKT - you need to check the actual files.

At least in the protobuf module itself, we only identify WKT by name. https://github.com/protocolbuffers/protobuf-go/blob/master/internal/detectknown/detect.go

Is there a case you're concerned about if we don't also check the names?

runtime/query.go
```diff
@@ -45,362 +46,294 @@ func (*defaultQueryParser) Parse(msg proto.Message, values url.Values, filter *u
...
+	if len(fieldPath) < 1 {
+		return fmt.Errorf("no field path")
 	}
-	var props *proto.Properties
-	m = m.Elem()
+	if len(values) < 1 {
+		return fmt.Errorf("no value provided")
+	}
+
+	var fieldDesc protoreflect.FieldDescriptor
```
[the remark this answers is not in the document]
johanbrandhorst: Sounds good

runtime/query.go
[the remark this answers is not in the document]
bufdev: *behavior

runtime/query.go
dsnet: Can either use [`ptypes.TimestampProto`](https://pkg.go.dev/github.com/golang/protobuf/ptypes?tab=doc#TimestampProto) or wait until https://golang.org/cl/225298 is submitted (I've been distracted by many things lately).

runtime/query.go
johanbrandhorst: I could reword the error message, because it doesn't cover all WKT (obviously, descriptors aren't here), and it doesn't try to. It handles the few we support right now. I'll remove this and make the type lookup more specific and fall through to the error in the switch instead.

runtime/query.go
johanbrandhorst: Good spot! I added a test for this and it led me to find another bug in the code and the new code manages to get rid of passing the messageValue a lot, which is much nicer. Thanks!

runtime/query.go
johanbrandhorst: Yeah, this was wrong, I've changed the wording and this and made it an error. Thanks!

runtime/query.go
johanbrandhorst: Changed to instantiate all messages explicitly, this avoids the need to pass around `msgValue` which was a pain point. The new code looks much nicer!

johanbrandhorst:
Seeing a really strange bazel build error:
```
[396 / 639] 1 / 10 tests; GoCompilePkg external/org_golang_google_grpc/internal/channelz/linux_amd64_race_stripped/go_default_library%/google.golang.org/grpc/internal/channelz.a; 1s processwrapper-sandbox ... (16 actions running)

ERROR: /src/grpc-gateway/runtime/BUILD.bazel:5:1: GoCompilePkg runtime/linux_amd64_race_stripped/go_default_library%/github.com/grpc-ecosystem/grpc-gateway/v2/runtime.a failed (Exit 1) builder failed: error executing command bazel-out/host/bin/external/go_sdk/builder compilepkg -sdk external/go_sdk -installsuffix linux_amd64_race -tags race -src runtime/context.go -src runtime/convert.go -src runtime/doc.go -src ... (remaining 77 argument(s) skipped)



Use --sandbox_debug to see verbose messages from the sandbox

compilepkg: error running subcommand: exit status 2

/root/.cache/_grpc_gateway_bazel/sandbox/processwrapper-sandbox/534/execroot/grpc_ecosystem_grpc_gateway/runtime/query.go:260:12: cannot assign *timestamp.Timestamp to msg (type protoreflect.ProtoMessage) in multiple assignment:

	*timestamp.Timestamp does not implement protoreflect.ProtoMessage (missing ProtoReflect method)

/root/.cache/_grpc_gateway_bazel/sandbox/processwrapper-sandbox/534/execroot/grpc_ecosystem_grpc_gateway/runtime/query.go:272:7: cannot use ptypes.DurationProto(d) (type *duration.Duration) as type protoreflect.ProtoMessage in assignment:

	*duration.Duration does not implement protoreflect.ProtoMessage (missing ProtoReflect method)

/root/.cache/_grpc_gateway_bazel/sandbox/processwrapper-sandbox/534/execroot/grpc_ecosystem_grpc_gateway/runtime/query.go:329:9: undefined: "github.com/golang/protobuf/proto".MessageV2

INFO: Elapsed time: 92.311s, Critical Path: 73.69s
```

Meanwhile, `go build ./...` works fine. It's like `bazel` is using the wrong version of `github.com/golang/protobuf` (pre-v1.4.0), but [`repositories.bzl`](https://github.com/grpc-ecosystem/grpc-gateway/blob/v2/repositories.bzl#L44) seems to be up to date, so what's going on?

@drigz please help me like you have with all my other bazel issues :joy:.

runtime/BUILD.bazel
```diff
@@ -28,6 +28,7 @@ go_library(
         "@com_github_golang_protobuf//descriptor:go_default_library_gen",
         "@com_github_golang_protobuf//jsonpb:go_default_library_gen",
         "@com_github_golang_protobuf//proto:go_default_library",
+        "@com_github_golang_protobuf//ptypes:go_default_library_gen",
```
johanbrandhorst, line 31: This is suspect. What does it mean that it's `go_default_library_gen`? How do I know which version of `protoc-gen-go` is used to generate these files?

johanbrandhorst: The plot thickens: https://github.com/bazelbuild/rules_go/blob/master/third_party/com_github_golang_protobuf-extras.patch. This seems to be hardcoded. Maybe we need to override it?

johanbrandhorst: This might've been fixed upstream, but I can't tell: https://github.com/bazelbuild/rules_go/commit/2634df77091483c4e63b8523d5d3815bfbf3f2eb. Probably worth just waiting for the next `rules_go` release and see.

johanbrandhorst: Worked around this issue by avoiding `ptypes`.

johanbrandhorst:
Great, it's actually not fixed, because now we're getting this error instead:
```
/root/.cache/_grpc_gateway_bazel/sandbox/processwrapper-sandbox/9/execroot/grpc_ecosystem_grpc_gateway/runtime/query.go:343:9: undefined: "github.com/golang/protobuf/proto".MessageV2
```

I'll revert the change and go back to `ptypes` and just wait for the next release of `rules_go`.

drigz:
Did you find https://nullset.xyz/2020/04/20/bazel-go-and-protobufs/? Does it help?

https://github.com/bazelbuild/rules_go/issues/2395 was closed a week and a half ago, so maybe using rules_go from master is easier than excising uses of the new APIv2.

johanbrandhorst:
> Did you find https://nullset.xyz/2020/04/20/bazel-go-and-protobufs/? Does it help?
> 
> [bazelbuild/rules_go#2395](https://github.com/bazelbuild/rules_go/issues/2395) was closed a week and a half ago, so maybe using rules_go from master is easier than excising uses of the new APIv2.

Thanks Rodrigo, I took some time to investigate the error and landed on the same conclusion as you. I actually upgraded to rules_go from master to test it, and it does work (the bazel build is now failing for other reasons), but I'm going to back out the change again and simply wait for the next release to drop before merging this.

Thanks again for all your help :heart_eyes:.

PR #1268, opened 2020-05-04
HttpBody mesage feature for stream RPC

adasari:
Hi,

HttpBody message Feature is not working with stream. the changes include fix the behavior

This possibly fix https://github.com/grpc-ecosystem/grpc-gateway/issues/1254

PR #1273, opened 2020-05-07
HttpBody mesage feature for stream RPC

adasari:
It is same as https://github.com/grpc-ecosystem/grpc-gateway/pull/1268, but target branch is v2.


adasari:
@johanbrandhorst fuzzit and node_test failures are not related to changes. can you take a look at them ? thanks.

adasari:
@johanbrandhorst each stream header includes from http body now. please review.
Also, can you elaborate https://github.com/grpc-ecosystem/grpc-gateway/pull/1268#discussion_r419740248 ? i couldn't understand what you are referring to.

adasari:
fuzzit and node_tests successful now. not sure what caused the issues.

examples/internal/integration/integration_test.go
```diff
@@ -59,10 +59,11 @@ func TestForwardResponseOption(t *testing.T) {
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
+	port := 7079
```
johanbrandhorst, line 62: Why did we change this port number?

examples/internal/integration/integration_test.go
```diff
@@ -980,6 +982,27 @@ func testABEList(t *testing.T, port int) {
...
+	if err != nil {
+		t.Fatalf("readAll(resp.Body) failed with %v; want success", err)
+	}
+
+	want := []string{"Hello 1", "Hello 2"}
+	if !reflect.DeepEqual(body, want) {
+		t.Errorf("testABEDownload failed: got %v, want %v", body, want)
+	}
+
+}
```
johanbrandhorst, line 1004: Could we check that the content type header is correct as well please?

runtime/handler.go
```diff
@@ -64,12 +65,18 @@ func ForwardResponseStream(ctx context.Context, mux *ServeMux, marshaler Marshal
...
+			if httpBody, ok := resp.(*httpbody.HttpBody); ok {
+				buf = httpBody.GetData()
+				w.Header().Set("Content-Type", httpBody.GetContentType())
+			} else {
+				result := map[string]interface{}{"result": resp}
+				if rb, ok := resp.(responseBody); ok {
+					result["result"] = rb.XXX_ResponseBody()
+				}
+
+				buf, err = marshaler.Marshal(result)
```
johanbrandhorst, line 77: Could you merge this if-else into the encapsulating switch?

examples/internal/integration/integration_test.go
adasari: test is failing with port busy error.

examples/internal/integration/integration_test.go
adasari: added check if content type is correct or not. but not sure how to check content type of each chunk using http client. do you have any idea ?

adasari:
can we take updating HTTPBody documentation as separate PR/MR ? can also share me where i can update the doc ? thanks.

runtime/handler.go
adasari: As discussed, resolving the review comment.

examples/internal/integration/integration_test.go
```diff
@@ -74,10 +75,10 @@ func TestForwardResponseOption(t *testing.T) {
 			return
 		}
 	}()
-	if err := waitForGateway(ctx, 8081); err != nil {
+	if err := waitForGateway(ctx, uint16(port)); err != nil {
 		t.Errorf("waitForGateway(ctx, 8081) failed with %v; want success", err)
```
johanbrandhorst, line 79:
```suggestion
		t.Errorf("waitForGateway(ctx, %d) failed with %v; want success", port, err)
```

runtime/handler.go
johanbrandhorst: I wonder if this would work:
```go
var buf []byte
httpBody, isHttpBody := resp.(*httpbody.HttpBody)
switch {
case resp == nil:
	buf, err = marshaler.Marshal(errorChunk(status.New(codes.Internal, "empty response")))
case isHttpBody:
	buf = httpBody.GetData()
	w.Header().Set("Content-Type", httpBody.GetContentType())
default:
	result := map[string]interface{}{"result": resp}
	if rb, ok := resp.(responseBody); ok {
		result["result"] = rb.XXX_ResponseBody()
	}
	buf, err = marshaler.Marshal(result)
}
```
?

I'd like to avoid the `if-else`, it's a bit messy to me. What do you think?

adasari:
the httpbody.md is not updated as per v2. do you want me add stream example to the same one ?

johanbrandhorst:
> the httpbody.md is not updated as per v2. do you want me add stream example to the same one ?

Yes please

adasari:
added example to httpbody readme. please review

PR #1309, opened 2020-05-09
runtime: rewrite fieldmask logic with protoreflect

johanbrandhorst:
Replace the use of the descriptorpb when we can use reflection
instead. Also add more thorough test and support repeated fields.


johanbrandhorst:
Bazel is failing for the same reason as https://github.com/grpc-ecosystem/grpc-gateway/pull/1272, will have to wait for the next release of `rules_go`.

vtolstov:
does it possible to with rewriting not import glog logger package to minimize deps (i'm use parts of runtime package in other project)

johanbrandhorst:
> does it possible to with rewriting not import glog logger package to minimize deps (i'm use parts of runtime package in other project)

Sure, would you like to submit a PR to remove this? I don't think it's relevant to this PR.

PR #1375, opened 2020-05-22
Add missing documentation for openapiv2 proto definition

bvwells:
Add missing documentation for openapiv2 proto definition.

### DIFF
--- a/protoc-gen-swagger/options/openapiv2.proto
+++ b/protoc-gen-swagger/options/openapiv2.proto
@@ -11,16 +11,27 @@ import "google/protobuf/struct.proto";
 //
 // See: https://github.com/OAI/OpenAPI-Specification/blob/3.0.0/versions/2.0.md#swaggerObject
 //
-// TODO(ivucica): document fields
 message Swagger {
+  // Specifies the Swagger Specification version being used. It can be
+  // used by the Swagger UI and other clients to interpret the API listing. The 
+  // value MUST be "2.0".
   string swagger = 1;
+  // Provides metadata about the API. The metadata can be used by the 
+  // clients if needed.
   Info info = 2;
+  // The host (name or ip) serving the API. This MUST be the host only and does 
+  // not include the scheme nor sub-paths. It MAY include a port. If the host is
+  // not included, the host serving the documentation is to be used (including
+  // the port). The host does not support path templating.
   string host = 3;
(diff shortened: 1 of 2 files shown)


bvwells:
@johanbrandhorst  - this PR adds some missing documentation for open API fields. I've taken the OpenAPI spec definitions for them.

bvwells:
Guess that I will need to regenerate some code before completing. Will push updates.

johanbrandhorst:
Please see https://github.com/grpc-ecosystem/grpc-gateway/blob/master/CONTRIBUTING.md for the regeneration steps

bvwells:
Yep, on it. Just had to setup access to docker.pkg.github.com.

bvwells:
All re-generated now.

protoc-gen-swagger/options/openapiv2.proto
```diff
@@ -94,13 +148,20 @@ message Response {
...
+  // A short description of the application. GFM syntax can be used for rich
+  // text representation.
   string description = 2;
+  // The Terms of Service for the API.
   string terms_of_service = 3;
+  // The contact information for the exposed API.
   Contact contact = 4;
+  // The license information for the exposed API.
   License license = 5;
+  // Required Provides the version of the application API (not to be confused
```
johanbrandhorst, line 163:
```suggestion
  // Required. Provides the version of the application API (not to be confused
```

bvwells: Agreed, I was a little unsure about adding this as obviously there is a difference between what the swagger spec defines as required and what is required for protoc-gen-swagger. I'll remove the required statement. I guess we could document the defaults at some point.

bvwells: It looks like there are a number of other descriptions in the proto definition which look like they have been copied from the open api spec verbatim with 'required'. Do you think we should also remove this?

johanbrandhorst: Haha, yeah, just left a comment to this effect.

bvwells: I've not removed all the other instances of required.

johanbrandhorst:
Thanks a ton, this is great! Could you please cherry pick this against v2?

bvwells:
Added PR #1378. Thanks for the review and merge!

PR #1381, opened 2020-05-23
Fix various misspellings

bvwells:
Fix various misspellings


johanbrandhorst:
Thanks! Could you please cherry pick this against v2?

PR #1383, opened 2020-05-23
Fix various misspellings (#1381)

bvwells:
Fix various misspellings (#1381)


johanbrandhorst:
Thanks for your contribution!

PR #1384, opened 2020-05-23
Implement last-match-wins behaviour in mux

johanbrandhorst:
In accordance with the http.proto spec, we should
use last-match-wins to determine which handler
is chosen if two are matched.

See https://github.com/grpc-ecosystem/grpc-gateway/issues/224
for more discussion.


examples/internal/integration/integration_test.go
```diff
@@ -1053,7 +1053,7 @@ func testABEBulkEcho(t *testing.T, port int) {
 	go func() {
 		defer wg.Done()
 		defer reqw.Close()
-		for i := 0; i < 1000; i++ {
+		for i := 0; i < 10; i++ {
```
johanbrandhorst, line 1056: Every time this fails it would fill the log, so lowered the number

examples/internal/proto/examplepb/a_bit_of_everything.proto
```diff
@@ -486,7 +486,7 @@ service ABitOfEverythingService {
 	}
 	rpc DeepPathEcho(ABitOfEverything) returns (ABitOfEverything) {
 		option (google.api.http) = {
-			post: "/v1/example/a_bit_of_everything/{single_nested.name}"
+			post: "/v1/example/deep_path/{single_nested.name}"
```
johanbrandhorst, line 489: This was overlapping with the bulk POST in stream.proto in a way which failed the tests, so I changed this since I can't imagine it's intentional.

PR #1390, opened 2020-05-24
Rename protoc-gen-swagger to protoc-gen-openapi

johanbrandhorst:


achew22:
Do you want to specify v2?

achew22:
To be clear openapi v2?

johanbrandhorst:
That would force us to rewrite it with a new name if we wanted to support v3 right?

johanbrandhorst:
> To be clear openapi v2?

Thinking more about this, I think you're right. I'll rename the binary

PR #1391, opened 2020-05-25
Adding docs about controlling http response status codes

zhughes3:
#### References to other Issues or PRs

Adding notes to documentation after repeated discussion in https://github.com/grpc-ecosystem/grpc-gateway/issues/517#issuecomment-633359394

#### Brief description of what is fixed or changed
Added documentation about controlling http response status codes.

PR #1397, opened 2020-05-26
Adding docs about controlling http response status codes (#1391)

zhughes3:
* Adding docs about controlling http response status codes

* PR changes based on review


johanbrandhorst:
Thanks for your contribution!

PR #1399, opened 2020-05-26
Fixes 1398: deprecate Regiter_XXX_HandlerServer

joesonw:
Gives deprecated notice when try to use Regiter_XXX_HandlerServer. Encourage using Register_XXX_FromEndpoint for full feature set and simplicity.

### DIFF
--- a/protoc-gen-grpc-gateway/internal/gengateway/template.go
+++ b/protoc-gen-grpc-gateway/internal/gengateway/template.go
@@ -564,7 +564,7 @@ func local_request_{{.Method.Service.GetName}}_{{.Method.GetName}}_{{.Index}}(ct
 // Register{{$svc.GetName}}{{$.RegisterFuncSuffix}}Server registers the http handlers for service {{$svc.GetName}} to "mux".
 // UnaryRPC     :call {{$svc.GetName}}Server directly.
 // StreamingRPC :currently unsupported pending https://github.com/grpc/grpc-go/issues/906.
-// Deprecated   :use with caution. This will not provide grpc.ServerTransportStream https://github.com/grpc-ecosystem/grpc-gateway/issues/1398. This implementation will be missing certain features to behave like a standard stanalone grpc.Server.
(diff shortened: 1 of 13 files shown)


joesonw:
@googlebot I signed it!

protoc-gen-grpc-gateway/internal/gengateway/template.go
```diff
@@ -564,6 +564,7 @@ func local_request_{{.Method.Service.GetName}}_{{.Method.GetName}}_{{.Index}}(ct
 // Register{{$svc.GetName}}{{$.RegisterFuncSuffix}}Server registers the http handlers for service {{$svc.GetName}} to "mux".
 // UnaryRPC     :call {{$svc.GetName}}Server directly.
 // StreamingRPC :currently unsupported pending https://github.com/grpc/grpc-go/issues/906.
+// Deprecated   :consider using Register{{$svc.GetName}}{{$.RegisterFuncSuffix}}FromEndpoint instead. This will not provide grpc.ServerTransportStream https://github.com/grpc-ecosystem/grpc-gateway/issues/1398. This implementation will be missing certain features to behave like a standard stanalone grpc.Server.
```
johanbrandhorst, line 567: This is a little heavy handed, I don't think we need to use the word "deprecated", this feature was only added earlier this year, and it does serve some users needs.

```suggestion
// Note that using this registration option will cause many gRPC library features (such as grpc.SendHeader, etc) to stop working . Consider using Register{{$svc.GetName}}{{$.RegisterFuncSuffix}}FromEndpoint instead.
```

joesonw: I wonder if there is a better way of warning people. In Goland, for example, It is really hard to notice the warnings, until you hover your mouse over the method for over a second or so.

johanbrandhorst: I don't think we need anything more than this. It's part of the documentation of the function, it should be one of the first places the user looks when using it.

johanbrandhorst:
Thanks for your contribution! Could you cherry pick this against the v2 branch please?

PR #1400, opened 2020-05-26
Add module generator option

Humpheh:
The go protobuf compiler recently added a `module` option define a module prefix which is trimmed from the name of go package to determine the output directory. Currently when using `paths=import` a protobuf file with the `go_package` being `example.com/path/to/example` would be output under `./example.com/path/to/example/myfile.pb.go`. This would usually be fine with `GOROOT` but with go modules it's less useful as the output directory structure does not have to be `example.com/path/to` etc.

I have tried to mirror the behaviour from the protobuf generator. The commit for which is here: https://go-review.googlesource.com/c/protobuf/+/219298/

### DIFF
--- a/protoc-gen-grpc-gateway/internal/gengateway/generator.go
+++ b/protoc-gen-grpc-gateway/internal/gengateway/generator.go
@@ -32,11 +32,12 @@ type generator struct {
 	useRequestContext  bool
 	registerFuncSuffix string
 	pathType           pathType
+	modulePath         string
 	allowPatchFeature  bool
 }
 
 // New returns a new generator which generates grpc gateway files.
-func New(reg *descriptor.Registry, useRequestContext bool, registerFuncSuffix, pathTypeString string, allowPatchFeature bool) gen.Generator {
+func New(reg *descriptor.Registry, useRequestContext bool, registerFuncSuffix, pathTypeString, modulePathString string, allowPatchFeature bool) gen.Generator {
 	var imports []descriptor.GoPackage
 	for _, pkgpath := range []string{
 		"context",
@@ -84,6 +85,7 @@ func New(reg *descriptor.Registry, useRequestContext bool, registerFuncSuffix, p
 		useRequestContext:  useRequestContext,
 		registerFuncSuffix: registerFuncSuffix,
 		pathType:           pathType,
+		modulePath:         modulePathString,
 		allowPatchFeature:  allowPatchFeature,
 	}
 }
@@ -105,9 +107,10 @@ func (g *generator) Generate(targets []*descriptor.File) ([]*plugin.CodeGenerato
 			glog.Errorf("%v: %s", err, code)
 			return nil, err
 		}
-		name := file.GetName()
-		if g.pathType == pathTypeImport && file.GoPkg.Path != "" {
-			name = fmt.Sprintf("%s/%s", file.GoPkg.Path, filepath.Base(name))
+		name, err := g.getFilePath(file)
+		if err != nil {
+			glog.Errorf("%v: %s", err, code)
+			return nil, err
 		}
 		ext := filepath.Ext(name)
 		base := strings.TrimSuffix(name, ext)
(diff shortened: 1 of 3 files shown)


Humpheh:
@googlebot I signed it!

johanbrandhorst:
Hi Humphrey, thanks for your interest in the project! Let me understand how this overlaps with `paths=source_relative`, which has been my go-to for the issue with the default output files. Do you have a use case which is not covered by `paths=source_relative`? I had a quick look at the CL linked and I'm confused about why they added it, unless they want to deprecate `paths=source_relative`. Do you know?

Humpheh:
Hey Johan, thanks for looking at this so quickly!

Recently I've been trying to upgrade our protobuf dependencies to later versions and I hit some snags with the output directories due to our current directory structure (which needs changing anyway). The following is an example which uses the `go_out` option but it was the same behaviour for the `grpc-gateway`.

Currently we have a folder which contains quite a few of our apis called `apis`, which contain a number of protobuf files (e.g. `apis/a.proto` and `apis/b.proto`) with imports relative to the root (e.g. `apis/a.proto` has `import "apis/b.proto"`. We were previously generating the files within this directory via something like:
```
protoc -I . -I ../ --go_out=paths=source_relative:../mymessages a.proto
``` 
...which included `../` as a include so that the `apis/b.proto` would resolve. We have had some issues though so I'm trying to move it so we generate files from the root instead using something like
```
protoc -I . --go_out=paths=source_relative:mymessages apis/a.proto
```

The issue I then found was that the `source_relative` option here would output the files in `mymessages/apis/myfile.go` which didn't match the `go_package` defined in the protobuf spec. In this instance as I was using go modules I wasn't able to use the default paths behaviour either (although if my directory structure was from the `GOROOT` then it probably would've been fine.

This change allows me to run:
```
protoc -I . --go_out=module=github.com/Humpheh/example:. apis/a.proto` 
```
which respects the `go_package` and outputs the file in the directory we expect.

We've been moving towards a better directory structure for the apis so that they're not all under `apis` and instead under the correct directory structure but at the moment this has been slow work. The change in the go protobuf generator fixed that issue so I've had a go at applying it here also.

protoc-gen-grpc-gateway/internal/gengateway/generator.go
```diff
@@ -121,6 +124,24 @@ func (g *generator) Generate(targets []*descriptor.File) ([]*plugin.CodeGenerato
...
+		}
+		trimPath, pkgPath := g.modulePath+"/", file.GoPkg.Path+"/"
+		if !strings.HasPrefix(pkgPath, trimPath) {
+			return "", fmt.Errorf("%v: file go path does not match module prefix: %v", file.GoPkg.Path, trimPath)
+		}
+		name = filepath.Join(strings.TrimPrefix(pkgPath, trimPath), filepath.Base(name))
+	} else if g.pathType == pathTypeImport && file.GoPkg.Path != "" {
+		// Only respect the path value if module is not supplied
+		name = fmt.Sprintf("%s/%s", file.GoPkg.Path, filepath.Base(name))
+	}
```
johanbrandhorst, line 141: Whenever I see something like `else if` I think it can usually be done neater with a switch statement. What do you think about:
```suggestion
	switch {
	case g.modulePath != "" && g.pathType != pathTypeImport:
		return "", errors.New("cannot use module= with paths=source_relative")
	case g.modulePath != "":
		trimPath, pkgPath := g.modulePath+"/", file.GoPkg.Path+"/"
		if !strings.HasPrefix(pkgPath, trimPath) {
			return "", fmt.Errorf("%v: file go path does not match module prefix: %v", file.GoPkg.Path, trimPath)
		}
		name = filepath.Join(strings.TrimPrefix(pkgPath, trimPath), filepath.Base(name))
	case g.pathType == pathTypeImport && file.GoPkg.Path != "":
		// Only respect the path value if module is not supplied
		name = fmt.Sprintf("%s/%s", file.GoPkg.Path, filepath.Base(name))
	}
```

johanbrandhorst:
Thanks for your contribution! Could you please cherry pick this against the v2 branch?

Humpheh:
Awesome, thanks @johanbrandhorst! I've made a cherry pick on the v2 branch here: #1409

PR #1403, opened 2020-05-26
docs: use default marshaler in examples

johanbrandhorst:


johanbrandhorst:
> Thank you! clap
> 
> A last step could be to mention that we're _not_ wrapping JSONPb in HTTPBodyMarshaler in the `google.api.HttpBody` section. But it might also stand out on its own. thought_balloon

Good idea, I've added a little blurb about the difference. Thanks for your feedback!

PR #1437, opened 2020-06-04
protoc-gen-swagger: Fix message recursion detection

doylezdead:
- Fixes a bug introduced during 9e25b2b which added recursion
  detection within messages. Briefly, the nature of the bug was that
  references to messages not in the current recursion traversal were
  remaining in the map used to track those occurrences.
- Includes a test case outlining the edge case that prompted this fix.

#### Brief description of what is fixed or changed
Briefly, the nature of the bug was that references to messages not in the current recursion traversal path were remaining in the map used to track those occurrences.  The fix just ensures those references don't hang around while recursing and are passed by value only.

#### Other comments
Testing the new case before fix...
```
(protoc-gen-swagger:)[16:32] ryan@psyduck ✘+ */grpc-gateway/protoc-gen-swagger $> go test ./genswagger/                                                                     
--- FAIL: TestMessageToQueryParametersNoRecursive (0.00s)
    template_test.go:518: No recursion error should be thrown: Recursive types are not allowed for query parameters, cycle found on ".example.NonRecursiveMessage"
```
And after the fix...
```
[1](protoc-gen-swagger:)[16:32] ryan@psyduck ✘+ */grpc-gateway/protoc-gen-swagger $> go test ./genswagger/
ok      github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger/genswagger    0.005s
```

PR #1439, opened 2020-06-04
protoc-gen-openapiv2: Fix message recursion detection (v2)

doylezdead:
Cherry-picked + merged from https://github.com/grpc-ecosystem/grpc-gateway/pull/1437

- Fixes a bug introduced during 9e25b2b which introduced recursion
  detection within messages. Briefly, the nature of the bug was that
  references to messages not in the current recursion traversal were
  remaining in the map used to track those occurrences.
- Includes a test case outlining the edge case that prompted this fix.


johanbrandhorst:
Thanks for your contribution!

PR #1440, opened 2020-06-05
add mux.HandlePath method  (v2)

ti:
in v2, we can not use `internal/httprule` from outside, so we need add this new method for add custom pattern from out side.

with this feature, People can add custom pattern From outsiede

exmaple

```go
mux.HandlePath("GET", "/v1/user/{user_id}/data", ...)
```

#### Other comments

in v1 we can use bellow code, but in v2 the `httprule` in internal, we can not use it from outside

```go
// HandlePath associates "h" to the pair of HTTP method and path pattern.
func HandlePath(method string, path string, h runtime.HandlerFunc) {
	compiler, err := httprule.Parse(path)
	if err != nil {
		log.Fatalf("Parse path to compiler failed: %v", err)
	}
	tp := compiler.Compile()
	pattern := runtime.MustPattern(runtime.NewPattern(tp.Version, tp.OpCodes, tp.Pool, tp.Verb))
	mux.Handle(method, pattern, h)
}
// exmaple
HandlePath("GET", "/v1/user/{user_id}/data", ...)
```

### DIFF
--- a/docs/_docs/inject_router.md
+++ b/docs/_docs/inject_router.md
@@ -25,7 +25,7 @@ import (
 func main() {
 	ctx := context.TODO()
 	mux := runtime.NewServeMux()
-	// Register proto's GET /say/{name} url
+	// Register generated routes to mux
 	_ = pb.RegisterGreeterHandlerServer(ctx, mux, &GreeterServer{})
 	// Register custom route for  GET /hello/{name}
 	_ = mux.HandlePath("GET", "/hello/{name}", func(w http.ResponseWriter, r *http.Request, pathParams map[string]string) {
--- a/docs/_docs/inject_router.md
+++ b/docs/_docs/inject_router.md
@@ -7,8 +7,6 @@ title: Adding custom routes to the mux
 
 The gRPC-gateway allows you to add custom routes to the serve mux, for example if you want to support a use case that isn't supported by the grpc-gateway, like file uploads.
 
-It may become a good alternative to `gorilla/mux` or other golang router matcher.
-
 ## Example
 
 ```go
@@ -47,5 +45,3 @@ func (h *GreeterServer) SayHello(ctx context.Context, req *pb.HelloRequest) (*pb
 	}, nil
 }
 ```
-
-> Note: You may need change the `pb "github.com/grpc-ecosystem/grpc-gateway/v2/examples/internal/helloworld"` to your code
--- a/docs/_docs/inject_router.md
+++ b/docs/_docs/inject_router.md
@@ -27,7 +27,7 @@ func main() {
 	mux := runtime.NewServeMux()
 	// Register proto's GET /say/{name} url
 	_ = pb.RegisterGreeterHandlerServer(ctx, mux, &GreeterServer{})
-	// Register custom router urls GET /hello/{name} url
+	// Register custom route for  GET /hello/{name}
 	_ = mux.HandlePath("GET", "/hello/{name}", func(w http.ResponseWriter, r *http.Request, pathParams map[string]string) {
 		w.Write([]byte("hello " + pathParams["name"]))
 	})
--- a/docs/_docs/inject_router.md
+++ b/docs/_docs/inject_router.md
@@ -1,0 +1,50 @@
+---
+category: documentation
+---
+
+# How do I use this?
+
+The grpc gateway v2 allows you inject custom router in grpc gateway.
+
+It may become a good alternative to `gorilla/mux` or other golang router matcher.
+
+## Example
+
+```go
+package main
+
+import (
+	"context"
+	"net/http"
+	
+	pb "github.com/grpc-ecosystem/grpc-gateway/v2/examples/internal/helloworld"
+	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
+)
+
+func main() {
+	ctx := context.TODO()
+	mux := runtime.NewServeMux()
+	// Register proto's GET /say/{name} url
+	_ = pb.RegisterGreeterHandlerServer(ctx, mux, &GreeterServer{})
+	// Register custom router urls GET /hello/{name} url
+	_ = mux.HandlePath("GET", "/hello/{name}", func(w http.ResponseWriter, r *http.Request, pathParams map[string]string) {
+		w.Write([]byte("hello " + pathParams["name"]))
+	})
+	panic(http.ListenAndServe(":8080", mux))
+}
+
+
+// GreeterServer is the server API for Greeter service.
+type GreeterServer struct {
+	
+}
+
+// SayHello implement to say hello
+func (h *GreeterServer) SayHello(ctx context.Context, req *pb.HelloRequest) (*pb.HelloReply, error) {
+	return &pb.HelloReply{
+		Message: "hello " + req.Name,
+	}, nil
+}
+```
+
+> Note: You may need change the `pb "github.com/grpc-ecosystem/grpc-gateway/v2/examples/internal/helloworld"` to your code
--- a/docs/_docs/inject_router.md
+++ b/docs/_docs/inject_router.md
@@ -24,11 +24,17 @@ func main() {
 	ctx := context.TODO()
 	mux := runtime.NewServeMux()
 	// Register generated routes to mux
-	_ = pb.RegisterGreeterHandlerServer(ctx, mux, &GreeterServer{})
+	err := pb.RegisterGreeterHandlerServer(ctx, mux, &GreeterServer{})
+	if err != nil {
+		panic(err)
+	}
 	// Register custom route for  GET /hello/{name}
-	_ = mux.HandlePath("GET", "/hello/{name}", func(w http.ResponseWriter, r *http.Request, pathParams map[string]string) {
+	err = mux.HandlePath("GET", "/hello/{name}", func(w http.ResponseWriter, r *http.Request, pathParams map[string]string) {
 		w.Write([]byte("hello " + pathParams["name"]))
 	})
+	if err != nil {
+		panic(err)
+	}
 	http.ListenAndServe(":8080", mux)
 }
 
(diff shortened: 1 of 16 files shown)


ti:
@googlebot I signed it!

ti:
@johanbrandhorst  Is this merger ok?

johanbrandhorst:
Hi @ti, thanks for your interest in the project! I'm initially hesitant to add this, as it's not one of the primary use cases of the gateway to allow custom paths to be added to the mux. Instead of this, have you considered wrapping the gateway mux (which implements `http.Handler`) in another mux? Something like

```go
import (
	"github.com/gorilla/mux"
	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
)

func main() {
	gwMux := runtime.NewServeMux()

	r := mux.NewRouter()
	r.HandleFunc("/v1/user/{user_id}/data", ...).Methods("GET")
	r.HandleFunc("/", gwMux)
}
```

I haven't tested this, but I think it could probably be made to work, and it would mean we don't have to expose this internal function. What do you think?

ti:
The `gorilla/mux` is google router for go, but in this case it may not be `Suitable`

1.  it does not have  the grpc's  `/v1/go:action`, `:action` feature
2. it makes two router matcher in the project

and for your concern,  this `merge request` does not `expose any internal function`,  it just hide all internal function behind the `HandlePath`

in the v2's code it has 
```go
func (s *ServeMux) Handle(meth string, pat Pattern, h HandlerFunc) 
```

 function, but we can not use this `Handle` outside, because the `compile Pattern` function is internal.

refer "https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/mux.go#L182"

I think the `Handle` and  `HandlePath` is the sample level, It does not increase the complexity of the project function.

this is my code use v1's `exmaple`

https://github.com/ti/noframe/blob/master/grpcmux/_exmaple/main.go

my go pakage is

`https://github.com/ti/noframe/blob/master/grpcmux`

ti:
@johanbrandhorst I think the `Handle` and `HandlePath` function is the sample level, It does not increase the complexity of the project function. it does not really expose any internal function

ti:
@johanbrandhorst do you think we need change this code to `PatternFromPath`
```
// PatternFromPath compile path to Pattern
func PatternFromPath(pathPattern string) Pattern {
	compiler, err := httprule.Parse(pathPattern)
	if err != nil {
		grpclog.Fatalf("Parse path to compiler failed: %v", err)
	}
	tp := compiler.Compile()
	return MustPattern(NewPattern(tp.Version, tp.OpCodes, tp.Pool, tp.Verb))
}
```

runtime/mux.go
```diff
@@ -164,6 +166,17 @@ func (s *ServeMux) Handle(meth string, pat Pattern, h HandlerFunc) {
 	s.handlers[meth] = append([]handler{{pat: pat, h: h}}, s.handlers[meth]...)
 }
 
+// HandlePath Handle path pattern form outside
```
johanbrandhorst, line 169:
```suggestion
// HandlePath allows users to configure custom path handlers.
```

runtime/mux.go
```diff
@@ -164,6 +166,17 @@ func (s *ServeMux) Handle(meth string, pat Pattern, h HandlerFunc) {
...
+func (s *ServeMux) HandlePath(meth string, pathPattern string, h HandlerFunc) {
+	compiler, err := httprule.Parse(pathPattern)
+	if err != nil {
+		grpclog.Fatalf("Parse path to compiler failed: %v", err)
+	}
+	tp := compiler.Compile()
+	pattern := MustPattern(NewPattern(tp.Version, tp.OpCodes, tp.Pool, tp.Verb))
+	s.Handle(meth, pattern, h)
+}
```
johanbrandhorst, line 178: Lets make this fail instead of panic if the pattern is invalid
```suggestion
func (s *ServeMux) HandlePath(meth string, pathPattern string, h HandlerFunc) error {
	compiler, err := httprule.Parse(pathPattern)
	if err != nil {
		return fmt.Errorf("parsing path pattern: %w", err)
	}
	tp := compiler.Compile()
	pattern, err := NewPattern(tp.Version, tp.OpCodes, tp.Pool, tp.Verb)
	if err != nil {
		return fmt.Errorf("creating new pattern: %w", err)
	}
	s.Handle(meth, pattern, h)
}
```

runtime/mux.go
johanbrandhorst: After adding the documentation example, lets add a link here to the docs.

ti:
@johanbrandhorst 

how about this docs

https://github.com/ti/grpc-gateway/blob/patch-1/docs/_docs/inject_router.md

The grpc gateway v2 may become a good alternative to gorilla/mux or other golang router matcher.

johanbrandhorst:
> The grpc gateway v2 may become a good alternative to `gorilla/mux` or other golang router matcher.

To be clear, I never want this to be the case. The grpc-gateway has a specific usecase that it tries to cover. We're never going to add more general purpose muxing/routing features.

docs/_docs/inject_router.md
```diff
@@ -0,0 +1,50 @@
+---
+category: documentation
```
johanbrandhorst, line 2:
```suggestion
category: documentation
title: Adding custom routes to the mux
```

docs/_docs/inject_router.md
```diff
@@ -0,0 +1,50 @@
...
+---
+
+# How do I use this?
```
johanbrandhorst, line 5:
```suggestion
# Adding custom routes to the mux
```

docs/_docs/inject_router.md
```diff
@@ -0,0 +1,50 @@
...
+
+The grpc gateway v2 allows you inject custom router in grpc gateway.
+
+It may become a good alternative to `gorilla/mux` or other golang router matcher.
```
johanbrandhorst, line 9: Remove this please, we don't want this to be the case.

docs/_docs/inject_router.md
```diff
@@ -0,0 +1,50 @@
+---
+category: documentation
+---
+
+# How do I use this?
+
+The grpc gateway v2 allows you inject custom router in grpc gateway.
```
johanbrandhorst, line 7:
```suggestion
The gRPC-gateway allows you to add custom routes to the serve mux, for example if you want to support a use case that isn't supported by the grpc-gateway, like file uploads.
```

docs/_docs/inject_router.md
```diff
@@ -0,0 +1,50 @@
...
+	"net/http"
+	
+	pb "github.com/grpc-ecosystem/grpc-gateway/v2/examples/internal/helloworld"
+	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
+)
+
+func main() {
+	ctx := context.TODO()
+	mux := runtime.NewServeMux()
+	// Register proto's GET /say/{name} url
```
johanbrandhorst, line 27:
```suggestion
	// Register generated routes to mux
```

docs/_docs/inject_router.md
```diff
@@ -0,0 +1,50 @@
...
+	_ = pb.RegisterGreeterHandlerServer(ctx, mux, &GreeterServer{})
+	// Register custom router urls GET /hello/{name} url
```
johanbrandhorst, line 29:
```suggestion
	// Register custom route for  GET /hello/{name}
```

docs/_docs/inject_router.md
```diff
@@ -0,0 +1,50 @@
...
+	_ = mux.HandlePath("GET", "/hello/{name}", func(w http.ResponseWriter, r *http.Request, pathParams map[string]string) {
+		w.Write([]byte("hello " + pathParams["name"]))
+	})
+	panic(http.ListenAndServe(":8080", mux))
```
johanbrandhorst, line 33:
```suggestion
	http.ListenAndServe(":8080", mux)
```

docs/_docs/inject_router.md
```diff
@@ -0,0 +1,50 @@
...
+
+// SayHello implement to say hello
+func (h *GreeterServer) SayHello(ctx context.Context, req *pb.HelloRequest) (*pb.HelloReply, error) {
+	return &pb.HelloReply{
+		Message: "hello " + req.Name,
+	}, nil
+}
+```
+
+> Note: You may need change the `pb "github.com/grpc-ecosystem/grpc-gateway/v2/examples/internal/helloworld"` to your code
```
johanbrandhorst, line 50: Remove this, this should be clear already

runtime/mux.go
```diff
@@ -164,6 +165,23 @@ func (s *ServeMux) Handle(meth string, pat Pattern, h HandlerFunc) {
 	s.handlers[meth] = append([]handler{{pat: pat, h: h}}, s.handlers[meth]...)
 }
 
+
+// HandlePath allows users to configure custom path handlers.
+// refer : ../docs/_docs/inject_router.md
```
johanbrandhorst, line 170: Lets link to the docs website: https://grpc-ecosystem.github.io/grpc-gateway/docs/inject_router.html

johanbrandhorst: (This link won't work until we have released v2)

ti:
@johanbrandhorst Are there any other problems?

johanbrandhorst:
Could we also add a test for this method to `mux_test.go` please?

ti:
> Could we also add a test for this method to `mux_test.go` please?
the main logic is in httprule.parse, so the test file is in the following file.

https://github.com/grpc-ecosystem/grpc-gateway/blob/v2/internal/httprule/parse_test.go#L29

johanbrandhorst:
> > Could we also add a test for this method to `mux_test.go` please?
> > the main logic is in httprule.parse, so the test file is in the following file.
> 
> https://github.com/grpc-ecosystem/grpc-gateway/blob/v2/internal/httprule/parse_test.go#L29

I appreciate that, but this is now part of the public API, and we should have a test for this function, however simple it is. Add one successful and one failing test please.

ti:
> > The grpc gateway v2 may become a good alternative to `gorilla/mux` or other golang router matcher.
> 
> To be clear, I never want this to be the case. The grpc-gateway has a specific usecase that it tries to cover. We're never going to add more general purpose muxing/routing features.

👌。The test file has been added.

johanbrandhorst:
Superceded by #1449

ti:
the merge request move to: https://github.com/grpc-ecosystem/grpc-gateway/pull/1449

for I lost some data when rebase.

sorry。

PR #1449, opened 2020-06-09
add mux.HandlePath method (v2)

ti:
in v2, we can not use `internal/httprule` from outside, so we need add this new method for add custom pattern from out side.

with this feature, People can add custom pattern From outsiede

exmaple

```go
mux.HandlePath("GET", "/v1/user/{user_id}/data", ...)
```

#### Other comments

in v1 we can use bellow code, but in v2 the `httprule` in internal, we can not use it from outside

```go
// HandlePath associates "h" to the pair of HTTP method and path pattern.
func HandlePath(method string, path string, h runtime.HandlerFunc) {
	compiler, err := httprule.Parse(path)
	if err != nil {
		log.Fatalf("Parse path to compiler failed: %v", err)
	}
	tp := compiler.Compile()
	pattern := runtime.MustPattern(runtime.NewPattern(tp.Version, tp.OpCodes, tp.Pool, tp.Verb))
	mux.Handle(method, pattern, h)
}
// exmaple
HandlePath("GET", "/v1/user/{user_id}/data", ...)
```

### DIFF
--- a/docs/_docs/inject_router.md
+++ b/docs/_docs/inject_router.md
@@ -1,0 +1,53 @@
+---
+category: documentation
+title: Adding custom routes to the mux
+---
+
+# Adding custom routes to the mux
+
+The gRPC-gateway allows you to add custom routes to the serve mux, for example if you want to support a use case that isn't supported by the grpc-gateway, like file uploads.
+
+## Example
+
+```go
+package main
+
+import (
+	"context"
+	"net/http"
+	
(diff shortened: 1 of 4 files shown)


ti:
@johanbrandhorst 
sorry for  I lost some data when rebase from v2, I create a new patch with all the codes and docs from the previous merge request

refer : https://github.com/grpc-ecosystem/grpc-gateway/pull/1440

johanbrandhorst:

ti:
Thank you for your patience. @johanbrandhorst

PR #1450, opened 2020-06-10
primitiveSchema: switch case to map

khanhhhh:


johanbrandhorst:
Why do you want to make this change?

khanhhhh:
Hi, I think this change will make the code more readability since the map can be searched faster comparing to switch-case

johanbrandhorst:
> Hi, I think this change will make the code more readability since the map can be searched faster comparing to switch-case

Firstly let me say that I appreciate that you wanted to improve the project. I think this actually hurts readability, because now we have to define a separate type, and you're using unkeyed literals here, which doesn't make it any easier either (e.g.:
```go
{"number", "double"}
```

instead of

```
{ftype: "number", format: "double"}
```
).

I think the current structure is actually easier to read, since there is a direct mapping in the logic. Next time you want to make a change, maybe consider asking in the support channel first, or raising an issue?

PR #1455, opened 2020-06-12
Create tracing.md

erikperttu:
This is a first basic draft for adding tracing. This is for using OpenCensus and aws x-ray so maybe it should be added to another section/file. There are no screenshots in this one, I could use the ones from https://github.com/johanbrandhorst/grpc-gateway-boilerplate/tree/add-tracing however that is from another repo. It would not be to hard to add an actual example of this in /examples/internal. Please let me know what to change/remove.

### DIFF
--- a/docs/_docs/tracing.md
+++ b/docs/_docs/tracing.md
@@ -1,0 +1,119 @@
+---
+category: documentation
+---
+
+# Tracing
+
+## With [OpenCensus.io](https://opencenus.io) and [AWS X-ray](https://aws.amazon.com/xray/)
+
+#### Adding tracing using AWS-Xray as the exporter
+
+This example uses the AWS-Xray exporter with a global trace setting. Note that AWS X-ray exporter does not handle any metrics only tracing. 
+
+1. Add the following imports
+
+```
+xray "contrib.go.opencensus.io/exporter/aws"
+"go.opencensus.io/plugin/ocgrpc"
+"go.opencensus.io/plugin/ochttp"
+"go.opencensus.io/trace"
+```
+
+2. Register the AWS X-ray exporter for the GRPC server
+
+```
+xrayExporter, err := xray.NewExporter(
+    xray.WithVersion("latest"),
+    // Add your AWS region.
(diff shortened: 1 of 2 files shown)


erikperttu:
@googlebot I signed it!

johanbrandhorst:
There's a section on using `OpenTracing` in [`customizingyourgateway`](https://github.com/grpc-ecosystem/grpc-gateway/blob/v2/docs/_docs/customizingyourgateway.md#opentracing-support), could you move that across to the new file as well?

erikperttu:
> There's a section on using `OpenTracing` in [`customizingyourgateway`](https://github.com/grpc-ecosystem/grpc-gateway/blob/v2/docs/_docs/customizingyourgateway.md#opentracing-support), could you move that across to the new file as well?

Move and remove from this one or leave it here and also add to the customization file?

johanbrandhorst:
> > There's a section on using `OpenTracing` in [`customizingyourgateway`](https://github.com/grpc-ecosystem/grpc-gateway/blob/v2/docs/_docs/customizingyourgateway.md#opentracing-support), could you move that across to the new file as well?
> 
> Move and remove from this one or leave it here and also add to the customization file?

Move the existing section in customizing your gateway to this new file.

docs/_docs/tracing.md
```diff
@@ -0,0 +1,177 @@
...
+        }
+        return startOptions
+    },
+}
+```
+
+4. No global configuration means we have to use the [per span sampler](https://opencensus.io/tracing/sampling/#per-span-sampler)
+
+
+##### A method we __want__ to trace
```
johanbrandhorst, line 101: Will this be correctly rendered on the docs site? I'm not sure. Maybe best to remove the italics?

erikperttu: removed it, it does not look like github does that type of bold text

johanbrandhorst:
Thanks for your contribution! Could you cherry pick this against v2 please?

achew22:
I believe this was merged again v2 already, could you do it against master?

PR #1474, opened 2020-06-22
support jsonp

t2krew:
support jsonp request from browser


t2krew:
@googlebot I signed it!

PR #1497, opened 2020-07-02
docs: fix broken season of docs links

dufcrule:
Fixed the link to the issue templates as it was broken.

Corrected the link to CONTRIBUTING.md as it was pointing at CHANGELOG.md.

PR #1498, opened 2020-07-03
docs: fix broken season of docs links (#1497)

dufcrule:
References https://github.com/grpc-ecosystem/grpc-gateway/pull/1497

### DIFF
--- a/docs/_docs/season_of_docs.md
+++ b/docs/_docs/season_of_docs.md
@@ -89,13 +89,13 @@ Material:
 ### Improve the contributor's guide
 
 This is currently split between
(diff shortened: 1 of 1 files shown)


johanbrandhorst:
Just a note about links in this document; obviously this is on the v2 branch but we're linking to the `master` version of the issue template, etc. This is intentional, I think what we want to do when we do finally cut v2 is to make a large merge into master, so that v2 releases can be made from the master branch, and these links will then be correct.

johanbrandhorst:
Thanks for your contribution!

PR #1502, opened 2020-07-06
Corrects an invalid test

### LINKED ISSUE #1501: Invalid test cases for runtime.Marshaler
## 🐛 Bug Report

Taking address of any element in this slice produces the same value (because the element is an empty struct):

https://github.com/grpc-ecosystem/grpc-gateway/blob/855204b63e5a0977e67eeedfecb64f6fec7228cf/runtime/marshaler_registry_test.go#L30-L52

So, these test cases almost won't work as expected.

As per [the last line of golang spec](https://golang.org/ref/spec#Size_and_alignment_guarantees):

> A struct or array type has size zero if it contains no fields (or elements, respectively) that have a size greater than zero.
> 
> Two distinct zero-size variables may have the same address in memory. 

## To Reproduce

See the code at [playground](https://play.golang.org/p/HPwRXhccsU9):

```go
package main

import (
	"fmt"
)

type S struct {
}

type T struct {
	b bool
}

func main() {
	var s [3]S
	var t [3]T

	fmt.Printf("%p,%p,%p\n", &s[0], &s[1], &s[2])
	fmt.Printf("%p,%p,%p\n", &t[0], &t[1], &t[2])
}
```

Outputs:

```-
0x58fd18,0x58fd18,0x58fd18
0xc00009400b,0xc00009400c,0xc00009400d
```

## Your Environment

Any.

yugui:
Fixes #1501 

Yes. I wrote it.

The expectations in the test was implemented wrong. But the wrong test
input has been hiding the wrong expectation, as explained in #1501.

Gave a non-zero size to `dummyMarshaler` to avoid the issue.

### DIFF
--- a/runtime/marshaler_registry_test.go
+++ b/runtime/marshaler_registry_test.go
@@ -2,6 +2,7 @@ package runtime_test
 
 import (
 	"errors"
+	"fmt"
 	"io"
 	"net/http"
 	"testing"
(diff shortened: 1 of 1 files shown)


runtime/marshaler_registry_test.go
```diff
@@ -27,23 +28,27 @@ func TestMarshalerForRequest(t *testing.T) {
...
 		},
+		// You can specify a marshaler for a specific MIME type.
+		// The output marshaler follows the input one unless specified.
 		{
 			opt:     runtime.WithMarshalerOption("application/x-in", &marshalers[1]),
 			wantIn:  &marshalers[1],
-			wantOut: &marshalers[0],
+			wantOut: &marshalers[1],
 		},
+		// You can also separately specify a output marshaler
```
johanbrandhorst, line 51: Nit
```suggestion
		// You can also separately specify an output marshaler
```

yugui: Thank you for your review. Fixed.

johanbrandhorst:
Thanks Yugui for the fix, could you please cherry pick this against the v2 branch as well?

Issue #559, opened 2018-02-26
`protoc-gen-swagger` does not parse query params of a DELETE HTTP rpc definition into swagger.json file 


zwtj:
Hi,

I defined a proto file called `abc.proto` as follows:

```proto
syntax="proto3";

import "google/api/annotations.proto";
import "google/protobuf/empty.proto";

service ABC {
    // delete all ABC elements
    //
    // delete all ABC elements 
    rpc DelConfigs(DelConfigsRequest) returns (google.protobuf.Empty) {
        option (google.api.http) = {
            delete: "/api/v1/abc"
        };
    }
}

message DelConfigsRequest {
    int32 whatever = 1;
}
```

When I used `protoc -I . -I/usr/include -I $GOPATH/src/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis --go_out=plugins=grpc:. --grpc-gateway_out=logtostderr=true:. --swagger_out=logtostderr=true:. ./abc.proto` to get `abc.swagger.json`, I expected that the `DelConfigsRequest` was parsed by `protoc-gen-swagger` to generate query parameters for the `delete` method. In fact it's not parsed at all. The `abc.swagger.json` I got is the following:

```json
{
  "swagger": "2.0",
  "info": {
    "title": "abc.proto",
    "version": "version not set"
  },
  "schemes": [
    "http",
    "https"
  ],
  "consumes": [
    "application/json"
  ],
  "produces": [
    "application/json"
  ],
  "paths": {
    "/api/v1/abc": {
      "delete": {
        "summary": "delete all ABC elements",
        "description": "delete all ABC elements",
        "operationId": "DelConfigs",
        "responses": {
          "200": {
            "description": "",
            "schema": {
              "$ref": "#/definitions/protobufEmpty"
            }
          }
        },
        "tags": [
          "ABC"
        ]
      }
    }
  },
  "definitions": {
    "protobufEmpty": {
      "type": "object",
      "description": "service Foo {\n      rpc Bar(google.protobuf.Empty) returns (google.protobuf.Empty);\n    }\n\nThe JSON representation for `Empty` is empty JSON object `{}`.",
      "title": "A generic empty message that you can re-use to avoid defining duplicated\nempty messages in your APIs. A typical example is to use it as the request\nor the response type of an API method. For instance:"
    }
  }
}
```

Is this an expected behavior to ignore the arguments of the `delete` method?

[...]

PR #1514, opened 2020-07-10
protoc-gen-swagger: add example similar to AIP-133

jonathaningram:
This example illustrates that the `book_id` field is not converted into a query string parameter in the generated swagger.json file.

This PR contains an example required for a fix for #559.

Yes.

I added an example to `examples/internal/proto/examplepb/a_bit_of_everything.proto` that is very similar to the example in https://google.aip.dev/133#user-specified-ids that allows users to provide their own ID. The user-specified ID needs to become a query string parameter per https://github.com/googleapis/aip/issues/542#issuecomment-656138118.

The gateway generator is fine, but the Swagger generator does not generate query string parameters when there is a body on the binding.

#### Other comments

I started to fix this locally and I have a rough idea of what needs doing, but I wanted to start the PR with a broken example first and was hoping that I could have some support if I have any questions along the way!

BTW if you are happy with including a broken example feel free to merge this PR. I'm happy to provide a subsequent PR that fixes the issue. Not sure if that's a better way to tackle the issue rather than a massive PR?

Thanks for reading :)


examples/internal/proto/examplepb/a_bit_of_everything.proto
```diff
@@ -322,6 +322,38 @@ message UpdateV2Request {
...
+	string parent = 1;
+
+	// The book to create.
+	Book book = 2;
+
+	// The ID to use for the book.
+	//
+	// This must start with an alphanumeric character.
+	string book_id = 3;
+}
```
johanbrandhorst, line 355: Could you reference the AIP in a comment here to explain these two messages please?

examples/internal/proto/examplepb/a_bit_of_everything.swagger.json
```diff
@@ -2776,6 +2834,26 @@
...
+          "readOnly": true
+        },
+        "create_time": {
+          "type": "string",
+          "format": "date-time",
+          "description": "Output only. Creation time of the book.",
+          "readOnly": true
+        }
+      }
+    },
```
johanbrandhorst, line 2856: What part of this is incorrect?

jonathaningram: Parameters is currently:

```json
{
    "parameters": [
          {
            "name": "parent",
            "description": "The publisher in which to create the book.\n\nFormat: `publishers/{publisher}`\n\nExample: `publishers/1257894000000000000`",
            "in": "path",
            "required": true,
            "type": "string"
          },
          {
            "name": "body",
            "description": "The book to create.",
            "in": "body",
            "required": true,
            "schema": {
              "$ref": "#/definitions/examplepbBook"
            }
          }
    ],
}
```

Notice that there is no mention of the `book_id`. For clients generated off the swagger.json there is literally no way for them to know they need to provide a `book_id`.

Thus `parameters` should include this:

```json
{
  "name": "book_id",
  "description": "The ID to use for the book.\n\nThis must start with an alphanumeric character",
  "in": "query",
  "required": true,
  "type": "string"
}
```

Note: this is correlated by the reply on my AIP issue here: https://github.com/googleapis/aip/issues/542#issuecomment-656138118

jonathaningram:
> Hi, thanks for your interest in the project! If I understand correctly this is meant to exhibit a bug, but I'm not sure I understand where it is. Regardless, it'd be great to have the fix included too in the same PR. Thanks!

@johanbrandhorst Thanks. Hopefully my inline comment reply has helped clear up what the actual issue is. Apologies for not being clearer originally. Generally speaking, the main issue is that for requests with a body non-body fields need to be added as query parameters. I think this is the same conclusion that @Wotzhs came to in their comment at https://github.com/grpc-ecosystem/grpc-gateway/issues/559#issuecomment-437964684. If that does not sound like the correct behaviour please let me know.

Happy to also push forward a fix in this PR.

johanbrandhorst:
Perfect, thanks for the clarification. I'd love to see this fixed in the swagger generator. Note that we're very close to tagging a stable v2, and this issue is reported on v1, can I ask that you test v2 as well? I don't think the logic has changed much in this area so I don't expect it to be working, but I'd rather see this fixed on v2 first and backported than the other way around if there are differences that make it hard to do it for both.

jonathaningram:
@johanbrandhorst sure, can do. If the issue exists in v2 I'll open a PR against that and go from there. Note: heading into the weekend here so you'll likely not hear from me until next week.

johanbrandhorst:
> @johanbrandhorst sure, can do. If the issue exists in v2 I'll open a PR against that and go from there. Note: heading into the weekend here so you'll likely not hear from me until next week.

Have a good weekend :smile:. If you want more synchronous communication to discuss the changes, we have a support channel on the Gophers Slack (#grpc-gateway), join via https://invite.slack.golangbridge.org/.

jonathaningram:
@johanbrandhorst this is ready for review now.

johanbrandhorst:
Thanks for your contribution!

PR #1528, opened 2020-07-16
Adds fqn_for_swagger_name to bazel defs

chriserickson:

https://github.com/grpc-ecosystem/grpc-gateway/pull/881

Support for setting the `fqn_for_swagger_name` flag was not exposed through the bazel rules. This follows the same pattern used for `json_names_for_fields`, but it begs the question to me as to whether there should be a more generic way to set these arguments, as there are a lot more flags that can be set that are not exposed.

### DIFF
--- a/protoc-gen-swagger/defs.bzl
+++ b/protoc-gen-swagger/defs.bzl
@@ -53,7 +53,8 @@ def _run_proto_gen_swagger(
         protoc_gen_swagger,
         grpc_api_configuration,
         single_output,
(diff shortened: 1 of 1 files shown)


johanbrandhorst:
Thanks for your contribution! Could you please cherry pick this against the v2 branch?

chriserickson:
Sure: https://github.com/grpc-ecosystem/grpc-gateway/pull/1529

PR #1546, opened 2020-07-22
Removed the "invalid remote addr" log  (#1543)

### LINKED ISSUE #1538: Log spam during attempt to append remote address to the XFF header when the transport is a unix domain socket
## 🐛 Bug Report

The `annotateContext` function cannot append the request's remote address if the request arrives via a Unix Domain Socket. The [code](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/context.go#L125) attempts to parse the `req.remoteAddr` field and fails. This is not really a big problem except that an [info level log message](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/context.go#L132) is printed every time the `req.remoteAddr` cannot be parsed.

## To Reproduce

1. Send a request to the gateway via a Unix Domain Socket (abstract in my case)

## Expected behavior

No info level log, maybe update the `X-Forwarded-For` header (though I'm not sure if appending the Unix Domain Socket path to the header contents is valid)

## Actual Behavior

While running a high traffic service, a lot of log messages are generated that look like this: `invalid remote addr: @`

## Your Environment

Linux (container running in kubernetes)

janzantinge:
Fixes #1538

Cherry-picked from master!

### DIFF
--- a/runtime/context.go
+++ b/runtime/context.go
@@ -12,7 +12,6 @@ import (
 	"time"
 
 	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/grpclog"
 	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
 )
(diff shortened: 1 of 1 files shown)


janzantinge:
@johanbrandhorst Not sure what the cherry-picking process entails so I created a PR for the v2 branch 😸

johanbrandhorst:
Thanks for your contribution!

PR #1566, opened 2020-08-03
Private annotation

klevin-vistar:


johanbrandhorst:
Hi, did you mean to open this against our project or was it meant for some internal fork?

klevin-vistar:
Hey maintainers! This PR was originally intended to add a `private` field annotation to protoc-gen-swagger and allow partial documentation of existing protos.

However, I hit 2 major snags when integrating with an existing repo:

1) Bazel's incompatibility with python targets that have `-` characters (renamed protoc-gen-swagger directory to protoc_gen_swagger so that python protobuf targets can import protoc-gen-swagger's options.proto)
2) To document services defined across multiple .proto files, it was necessary to bundle multiple top-level proto_library targets into an umbrella proto_library, then use ProtoInfo's transitive_sources AND hackily avoid duplicate well-known-protos at runtime.

Both of these are unfortunate workarounds, but I'm hoping that your feedback will lead to an ideal solution that could be merged.

Thank you!

johanbrandhorst:
Hi, thanks for your interest in the project. I'm terribly sorry, but I don't think we're likely to want to add this. The current schema options are all based on fields in the openapi v2 spec (https://swagger.io/specification/v2/), and there'd have to be an extraordinarily good reason for us to add fields that aren't in that spec. I'm not convinced this qualifies, and we certainly can't merge the change as-is as it would be breaking users on many levels. I understand that there is an underlying problem that you're trying to solve, maybe we can go back to the drawing board in an issue?

Sorry to disappoint you, I recommend raising an issue next time so we can discuss new features such as this to save everyone involved time.

PR #1602, opened 2020-08-17
protoc-gen-swagger: fix map field options (#1601)

### LINKED ISSUE #1601: protoc-gen-swagger map type filed doesn't support options.openapiv2_field
## 🐛 Bug Report

```proto
message GetSubjectPermissions2DReply{
  map<string, ListOfString> permissions = 1 [(grpc.gateway.protoc_gen_swagger.options.openapiv2_field) = {title: "展开的二级权限列表", description:"展开的二级权限列表"}];
}
```

generated swagger json:

```
    "HodorGetSubjectPermissions2DReply": {
      "type": "object",
      "properties": {
        "permissions": {
          "type": "object",
          "additionalProperties": {
            "$ref": "#/definitions/HodorListOfString"
          }
        }
      },
      "title": "GetSubjectPermissions2DReply"
    },
```

permissions.title and permissions.description is not set.

## Your Environment

grpc-gateway v1.14.6

## Other Info

I dig into the source code of protoc-gen-swagger, found the problem is in `template.go:schemaOfField`：

```go
// schemaOfField returns a swagger Schema Object for a protobuf field.
func schemaOfField(f *descriptor.Field, reg *descriptor.Registry, refs refMap) swaggerSchemaObject {
	const (
		singular = 0
		array    = 1
		object   = 2
	)
	var (
		core      schemaCore
		aggregate int
	)

	fd := f.FieldDescriptorProto
	if m, err := reg.LookupMsg("", f.GetTypeName()); err == nil {
		if opt := m.GetOptions(); opt != nil && opt.MapEntry != nil && *opt.MapEntry {
			fd = m.GetField()[1]
			aggregate = object
		}
	}
	if fd.GetLabel() == pbdescriptor.FieldDescriptorProto_LABEL_REPEATED {
		aggregate = array
	}
        ……
}
```

When the field is map type, `fd = m.GetField()[1]` use the value's FieldDescriptorProto. Because the mapentry is auto generate, so the fd.Options is always nil.

yangxikun:

Fixes #1601

Yes

Fix map type filed doesn't support options.

### DIFF
--- a/protoc-gen-swagger/genswagger/template.go
+++ b/protoc-gen-swagger/genswagger/template.go
@@ -504,7 +504,7 @@ func schemaOfField(f *descriptor.Field, reg *descriptor.Registry, refs refMap) s
 		}
 	}
 
-	if j, err := extractJSONSchemaFromFieldDescriptor(fd); err == nil {
+	if j, err := extractJSONSchemaFromFieldDescriptor(f.FieldDescriptorProto); err == nil {
 		updateSwaggerObjectFromJSONSchema(&ret, j, reg, f)
 	}
 
--- a/protoc-gen-swagger/genswagger/template_test.go
+++ b/protoc-gen-swagger/genswagger/template_test.go
@@ -2066,7 +2066,16 @@ func TestSchemaOfField(t *testing.T) {
 	type test struct {
 		field    *descriptor.Field
 		refs     refMap
-		expected schemaCore
+		expected swaggerSchemaObject
+	}
+
+	var fieldOptions = new(protodescriptor.FieldOptions)
+	err := proto.SetExtension(fieldOptions, swagger_options.E_Openapiv2Field, &swagger_options.JSONSchema{
+		Title:       "field title",
+		Description: "field description",
+	})
+	if err != nil {
+		t.Errorf("proto.SetExtension() failed with %v; want success", err)
 	}
 
 	tests := []test{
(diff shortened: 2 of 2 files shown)


protoc-gen-swagger/genswagger/template.go
```diff
@@ -504,7 +504,7 @@ func schemaOfField(f *descriptor.Field, reg *descriptor.Registry, refs refMap) s
 		}
 	}
 
-	if j, err := extractJSONSchemaFromFieldDescriptor(fd); err == nil {
+	if j, err := extractJSONSchemaFromFieldDescriptor(f.FieldDescriptorProto); err == nil {
```
johanbrandhorst, line 507: Is this really the correct change? It looks like we're doing some sort of map parsing on line 443-445. Is that still right?

johanbrandhorst (commented):
Tests look great, just had a question about the logic change.

protoc-gen-swagger/genswagger/template.go
yangxikun: 443 - 445 is just for get `FieldDescriptorProto` of map's value, then populate `swaggerSchemaObject.AdditionalProperties`.

From map's value `FieldDescriptorProto`, we cann't extract JSONSchema(options.openapiv2_field), because it's always nil.

The JSONSchema(options.openapiv2_field) should extract from `f.FieldDescriptorProto`.

johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution! Could you please cherry pick this against v2?

PR #1606, opened 2020-08-21
Control warning behavior on unbound methods using a flag

yousseftelda:
The current behavior since #1178 is to emit a warning log for every unbound RPC method, causing a lot of noise in case the intention is to generate the gateway from a subset of the RPC methods. This PR allows us to control the logging behavior in case of encountering an unbound method, with the default behavior being log with severity `Info`.


yousseftelda:
@googlebot I signed it!

johanbrandhorst:
Thanks for your contribution! Could you please cherry pick this against v2?

PR #1628, opened 2020-08-31
fix: the default value of required property

### LINKED ISSUE #1627: protoc-gen-swagger: the default value of the `required` property in parameter object should be false
## 🐛 Bug Report

The value of the `required` property in parameter object should be false by default if it is not in `path` 
https://github.com/grpc-ecosystem/grpc-gateway/blob/5a432166606fee3c3686df2be3c5cf914d3bc30b/protoc-gen-swagger/genswagger/template.go#L910

> Determines whether this parameter is mandatory. If the parameter is in "path", this property is required and its value MUST be true. Otherwise, the property MAY be included and its default value is false.

It is described in the swagger specification document.
https://swagger.io/specification/v2/

## To Reproduce

`protoc -I/path/to/proto/ -I/gp/src/github.com/grpc-ecosystem/grpc-gateway/third_party/googleapis -I/go/src/github.com/grpc-ecosystem/grpc-gateway --swagger_out=allow_repeated_fields_in_body=true,json_names_for_fields=true:out /path/to/proto/spec.proto`

```spec.proto
syntax = "proto3";

service FooService {
  rpc PostMessage(Message) returns (Success) {
    option (google.api.http) = {
      post: "/v1/messages"
      body: "*"
    };
  }
}

message Message {
  string message_id = 1;
  string text = 2;
}

message Success {
  bool success = 1;
}
```

## Expected behavior

Expected output (parameters only)

```spec.swagger.json
{
  "paths": {
    "/v1/messages": {
      "post": {
        "parameters": [
          {
            "name": "body",
            "in": "body",
            "required": false,
            "schema": {
              "$ref": "#/definitions/Message"
            }
          }
        ]
      }
    }
  }
}
```

## Actual Behavior

Actual output (parameters only)

```spec.swagger.json
{
  "paths": {
    "/v1/messages": {
      "post": {
        "parameters": [
          {
            "name": "body",
            "in": "body",
            "required": true,
            "schema": {
              "$ref": "#/definitions/Message"
            }
          }
        ]
      }
    }
  }
}
```

## Your Environment

I execute this in a Docker container. It uses `golang:1.14-alpine` as a base image.
The version of grpc-gateway is `v1.14.7`.

algas:

Fixes #1627

Yes

The value of the required property in parameter object should be false by default if it is not in path.

> Determines whether this parameter is mandatory.
If the parameter is in "path", this property is required and its value MUST be true.
Otherwise, the property MAY be included and its default value is false.

https://swagger.io/specification/v2/

#### Other comments

I may need to fix some tests.

### DIFF
--- a/examples/internal/clients/abe/api/swagger.yaml
+++ b/examples/internal/clients/abe/api/swagger.yaml
@@ -37,7 +37,7 @@ paths:
       parameters:
       - in: "body"
         name: "body"
-        required: true
+        required: false
         schema:
           $ref: "#/definitions/examplepbABitOfEverything"
         x-exportParamName: "Body"
@@ -702,7 +702,7 @@ paths:
         x-exportParamName: "StringValue"
       - in: "body"
         name: "body"
-        required: true
+        required: false
         schema:
           $ref: "#/definitions/ABitOfEverythingNested"
         x-exportParamName: "Body"
@@ -1441,7 +1441,7 @@ paths:
         x-exportParamName: "SingleNestedName"
       - in: "body"
         name: "body"
-        required: true
+        required: false
         schema:
           $ref: "#/definitions/examplepbABitOfEverything"
         x-exportParamName: "Body"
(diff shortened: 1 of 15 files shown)


johanbrandhorst:
Hey, as I commented on the issue, I'm not sure we can change this in v1 without breaking users, but could you cherry pick this against the v2 branch instead? We're still allowed breaking changes there.

PR #1637, opened 2020-09-02
Deprecate example field

johanbrandhorst:
The example field using google.protobuf.Any
was pretty silly and just happened to work because
of how we were using it. It's confusing to users and breaks some tooling,
so it is best deprecated.

This is an unfortunate break from the "each field maps to the openapiv2 schema"
precedent we've maintained so far, but there is
no other good way that I can think of to safely deprecate
this.

Any thoughts @ivucica?


ivucica (commented):
LGTM, I just wish this PR had included generated .swagger.json examples.

johanbrandhorst:
There is no diff in the swagger examples because the output is exactly the same as before.

ivucica:
After writing my comment, I took an even closer look at the source code --
I see that it's still putting the output into 'example' field.

Thank you for doing this!

Issue #517, opened 2018-01-08
Override HTTP response code for particular service method?


anweiss:
Is there any way to choose specific HTTP status codes in a response rather than relying on the pre-mapped values defined [here](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/errors.go#L15)? Similar to #240

anweiss:
Found https://github.com/grpc-ecosystem/grpc-gateway/wiki/How-to-customize-your-gateway#replace-a-response-forwarder-per-method, but would love to be able to do this using an proto `option`

achew22:
Maybe the correct thing to do is to fix the central list of these mappings. I think somewhere in the bowels of the beast we have a mapping like that. What is the mapping you would prefer?

OscarHanzely:
Well I ended on same problem. My Create method must return HTTP status 201 Created. gRPC has no code for that, in fact gRPC is all 200 OK or 4xx+ if error occurs. 
So the grpc-gateway just has complete missing implementation of any 2xx codes. 

I am ending up with implementing own forwarder for each of the Create end point, however it is poorly designed, because I have to implement entire forwarder, since the status code comes as last method of Writer and after that headers cannot be modified. So I cannot even call the generic runtime.ForwardResponseMessage because the headers would be already sent. This is a big gap in implementation of grpc-gateway

robinbryce:
What is the problem with providing a sensible means to set the response codes directly ? 206 partial content is a pretty reasonable ask isn't it ?

johanbrandhorst:
There is only one successful response code in gRPC, and there is an argument to be made that if you need something that is much more than just mapping gRPC onto HTTP/REST, maybe the grpc-gateway isn't what you should be using.

Having said that, I think we'd all be open to exploring any ideas you or anyone else might have for how to implement this sort of ask, be it custom proto options and what not. What we will not do, however, is put our own valuable time into it, unless we have a personal need for it.

deefstes:
> There is only one successful response code in gRPC, and there is an argument to be made that if you need something that is much more than just mapping gRPC onto HTTP/REST, maybe the grpc-gateway isn't what you should be using.

I don't think a requirement to provide HTTP/REST codes when responding on HTTP/REST and gRPC codes when responding on gRPC could really be described as "much more than just mapping gRPC onto HTTP/REST".

[...]

PR #1638, opened 2020-09-02
fix SendHeader/SetTrailer when use grpc-gateway standalone server (v2)

ghostiam:
Fix https://github.com/grpc-ecosystem/grpc-gateway/issues/517#issuecomment-684625645 when use gateway without gRPC.

### DIFF
--- a/protoc-gen-grpc-gateway/internal/gengateway/template.go
+++ b/protoc-gen-grpc-gateway/internal/gengateway/template.go
@@ -577,7 +577,7 @@ func local_request_{{.Method.Service.GetName}}_{{.Method.GetName}}_{{.Index}}(ct
 // Register{{$svc.GetName}}{{$.RegisterFuncSuffix}}Server registers the http handlers for service {{$svc.GetName}} to "mux".
 // UnaryRPC     :call {{$svc.GetName}}Server directly.
 // StreamingRPC :currently unsupported pending https://github.com/grpc/grpc-go/issues/906.
-// Note that using this registration option will cause many gRPC library features (such as grpc.SendHeader, etc) to stop working. Consider using Register{{$svc.GetName}}{{$.RegisterFuncSuffix}}FromEndpoint instead.
+// Note that using this registration option will cause many gRPC library features to stop working. Consider using Register{{$svc.GetName}}{{$.RegisterFuncSuffix}}FromEndpoint instead.
 func Register{{$svc.GetName}}{{$.RegisterFuncSuffix}}Server(ctx context.Context, mux *runtime.ServeMux, server {{$svc.InstanceName}}Server) error {
 	{{range $m := $svc.Methods}}
 	{{range $b := $m.Bindings}}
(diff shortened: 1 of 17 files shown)


ghostiam:
@johanbrandhorst bazel failed by timeout. Very slow compiling...

protoc-gen-grpc-gateway/internal/gengateway/template.go
```diff
@@ -595,13 +596,16 @@ func Register{{$svc.GetName}}{{$.RegisterFuncSuffix}}Server(ctx context.Context,
 		ctx, cancel := context.WithCancel(ctx)
 	{{- end }}
 		defer cancel()
+		var stream runtime.ServerTransportStream
```
johanbrandhorst, line 599: We can update the comment on line 580 now.

runtime/context.go
```diff
@@ -165,6 +166,59 @@ func ServerMetadataFromContext(ctx context.Context) (md ServerMetadata, ok bool)
 	return
 }
 
+// ServerTransportStream implement grpc.ServerTransportStream
```
johanbrandhorst, line 169:
```suggestion
// ServerTransportStream implements grpc.ServerTransportStream.
// It should only be used by the generated files to support grpc.SendHeader
// outside of gRPC server use.
```

runtime/context.go
```diff
@@ -165,6 +166,59 @@ func ServerMetadataFromContext(ctx context.Context) (md ServerMetadata, ok bool)
...
+}
+
+// Method returns the method for the stream.
+func (s *ServerTransportStream) Method() string {
+	return ""
+}
+
+// Header returns the header metadata of the stream.
+func (s *ServerTransportStream) Header() metadata.MD {
+	c := s.header.Copy()
```
johanbrandhorst, line 183: Doesn't this need to be mutex protected?
```suggestion
	s.mu.Lock()
	defer s.mu.Unlock()
	c := s.header.Copy()
```

runtime/context.go
```diff
@@ -165,6 +166,59 @@ func ServerMetadataFromContext(ctx context.Context) (md ServerMetadata, ok bool)
...
+}
+
+// SendHeader sets the header metadata.
+func (s *ServerTransportStream) SendHeader(md metadata.MD) error {
+	return s.SetHeader(md)
+}
+
+// Trailer returns the cached trailer metadata.
+func (s *ServerTransportStream) Trailer() metadata.MD {
+	c := s.trailer.Copy()
```
johanbrandhorst, line 206: Doesn't this need to be mutex protected?
```suggestion
	s.mu.Lock()
	defer s.mu.Unlock()
	c := s.trailer.Copy()
```

johanbrandhorst (changes requested):

runtime/context.go
ghostiam: Maybe, but the original library does not use a mutex in its implementation
https://github.com/grpc/grpc-go/blob/master/internal/transport/transport.go#L386

johanbrandhorst: Interesting, seems pretty specific to their design though. This design definitely needs it.

protoc-gen-grpc-gateway/internal/gengateway/template.go
ghostiam: I couldn't find any other examples, so I just removed the mention of grpc.SendHeader.

johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution!

PR #1652, opened 2020-09-08
Generate unbound methods

### LINKED ISSUE #1649: Support service definitions without any annotations.
## 🚀 Feature

When a service definition does not use any REST / transcoding annotations, it should still be possible to generate a gateway and OpenAPI spec by using the default gRPC path.

As per the [gRPC spec](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md), the path of a particular operation is well-defined: `/<service name>/<method name>`. Moreover, the request and response are serialized in the body and there's no use of any request parameter or header to pass information around. Lastly, the HTTP verb is also well-defined: `POST`.

Thus, there's a clear default to use when a service has no api annotation. In fact, this is what "Cloud Endpoint" does https://cloud.google.com/endpoints/docs/grpc/transcoding#where_to_configure_transcoding

This would be useful because:

* running the generators on any source would "just work" with no additional effort on the end user;
* provides an alternative, which requires less effort, to the external configuration file when the proto definition is not controlled by the end user;
* allows mapping directly to the underlying (well-defined) gRPC protocol (simplest possible thing that works, no indirection);

This is probably a breaking change since users can no longer "selectively" pick the rpc methods to expose in the gateway. That said, having the default behaviour be exposing all methods instead of none of them seems less surprising. Furthermore, users can have separate service definitions or perhaps additional annotations could be used to "turn off" gateway generation per method. 

Could such a feature be considered for v2?

plaflamme:

Fixes #1649 

Yes.

This is a first pass at implementing the new option to generate methods for methods that have no http annotation.
When the option is enabled and a method has no options (either through annotations or external configuration), default options are generated and applied.

#### Other comments

This is `WIP` since it probably needs more tests and would need a corresponding change to `gen-swagger`


protoc-gen-grpc-gateway/descriptor/services.go
```diff
@@ -205,6 +214,18 @@ func extractAPIOptions(meth *descriptor.MethodDescriptorProto) (*options.HttpRul
...
 
+func defaultAPIOptions(svc *Service, md *descriptor.MethodDescriptorProto) (*options.HttpRule, error) {
+	// FQSN prefixes the service's full name with a '.', e.g.: '.example.ExampleService'
+	fqsn := strings.TrimPrefix(svc.FQSN(), ".")
+	rule := &options.HttpRule{
+		Pattern: &options.HttpRule_Post{
+			Post: fmt.Sprintf("/%s/%s", fqsn, md.GetName()),
+		},
+		Body: "*",
+	}
```
johanbrandhorst, line 225: Could we add a reference to the gRPC HTTP/2 spec here (https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md#requests), so that we know what we're basing this default pathing on?

johanbrandhorst (changes requested):
This is a great start, it seems even simpler than I had anticipated., great work! You're right that we'll need to enable a similar thing in the swagger generator so that they can still be used together, and ideally I'd like to see a new example proto generation test that makes use of this generator flag. See [the Makefile](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/Makefile#L122-L127) for examples of this today.

plaflamme:
@johanbrandhorst For the example proto, should this new example not affect any of the existing ones? I'm asking because if I add `generate_unbound_methods=true` to the generator, it'll affect the `a_bit_of_everything` example (it has an unbound method). I can make a completely separate example if you think that's more appropriate.

johanbrandhorst:
> @johanbrandhorst For the example proto, should this new example not affect any of the existing ones? I'm asking because if I add `generate_unbound_methods=true` to the generator, it'll affect the `a_bit_of_everything` example (it has an unbound method). I can make a completely separate example if you think that's more appropriate.

Yes, sorry for not being clear, I think it should be a separate example. Don't enable it across the board.

plaflamme:
@johanbrandhorst I've added the requested example in https://github.com/grpc-ecosystem/grpc-gateway/pull/1652/commits/5339757076378ae10e663b24d3eae22b7c3fcf81 and the resulting generated files in https://github.com/grpc-ecosystem/grpc-gateway/pull/1652/commits/498c5a4fe0e0b8f81f534ddc7df1427bdeb1df36

I have to admit that the changes to the `Makefile` are pretty verbose. Any suggestions to simplify things are welcome, I'm not particularly knowledgeable of `Makefile` composability.

examples/internal/clients/generateunboundmethods/.gitignore
```diff
@@ -0,0 +1,24 @@
...
+*.cgo2.c
+_cgo_defun.c
+_cgo_gotypes.go
+_cgo_export.*
+
+_testmain.go
+
+*.exe
+*.test
+*.prof
```
johanbrandhorst, line 24: I think this file is generated by `swagger-codegen` if there isn't one already, and it looks like we're using a simpler version in our other generated clients: https://github.com/grpc-ecosystem/grpc-gateway/blob/master/examples/internal/clients/abe/.gitignore. Maybe we could rewrite this one too?

examples/internal/clients/generateunboundmethods/.swagger-codegen-ignore
```diff
@@ -0,0 +1,23 @@
...
+
+# You can recursively match patterns against a directory, file or extension with a double asterisk (**):
+#foo/**/qux
+# This matches foo/bar/qux, foo/baz/qux, and foo/bar/baz/qux
+
+# You can also negate patterns with an exclamation (!).
+# For example, you can ignore all files in a docs folder with the file extension .md:
+#docs/*.md
+# Then explicitly reverse the ignore rule for a single file:
+#!docs/README.md
```
johanbrandhorst, line 23: Ditto simpler ignore: https://github.com/grpc-ecosystem/grpc-gateway/blob/master/examples/internal/clients/abe/.swagger-codegen-ignore

examples/internal/proto/examplepb/generate_unbound_methods.proto
```diff
@@ -0,0 +1,35 @@
+syntax = "proto3";
+option go_package = "examplepb";
+
+// Unannotated Echo Service
+// Similar to echo_service.proto but without annotations and without external configuration.
+//
+// Echo Service API consists of a single service which returns
+// a message.
+package grpc.gateway.examples.internal.examplepb;
```
johanbrandhorst, line 9: This comment is out of date.

examples/internal/proto/examplepb/generate_unbound_methods.proto
```diff
@@ -0,0 +1,35 @@
...
+
+// Do not need annotations.proto, can still use well known types as usual
+import "google/protobuf/duration.proto";
+
+// UnannotatedSimpleMessage represents a simple message sent to the unannotated Echo service.
+message GenerateUnboundMethodsSimpleMessage {
```
johanbrandhorst, line 15: This comment is out of date

examples/internal/proto/examplepb/generate_unbound_methods.proto
```diff
@@ -0,0 +1,35 @@
...
+	// Id represents the message identifier.
+	string id = 1;
+	int64 num = 2;
+	google.protobuf.Duration duration = 3;
+}
+
+// Echo service responds to incoming echo requests.
```
johanbrandhorst, line 22: This comment is out of date

Makefile
```diff
@@ -136,8 +142,14 @@ RESPONSE_BODY_EXAMPLE_SRCS=$(EXAMPLE_CLIENT_DIR)/responsebody/client.go \
...
-
-EXAMPLE_CLIENT_SRCS=$(ECHO_EXAMPLE_SRCS) $(ABE_EXAMPLE_SRCS) $(UNANNOTATED_ECHO_EXAMPLE_SRCS) $(RESPONSE_BODY_EXAMPLE_SRCS)
+GENERATE_UNBOUND_METHODS_EXAMPLE_SPEC=examples/internal/proto/examplepb/generate_unbound_methods.swagger.json
+GENERATE_UNBOUND_METHODS_EXAMPLE_SRCS=$(EXAMPLE_CLIENT_DIR)/generateunboundmethods/client.go \
+		 $(EXAMPLE_CLIENT_DIR)/generateunboundmethods/response.go \
+		 $(EXAMPLE_CLIENT_DIR)/generateunboundmethods/configuration.go \
+		 $(EXAMPLE_CLIENT_DIR)/generateunboundmethods/model_examplepb_generate_unbound_methods_simple_message.go \
+		 $(EXAMPLE_CLIENT_DIR)/generateunboundmethods/api_generate_unbound_methods.go
+
+EXAMPLE_CLIENT_SRCS=$(ECHO_EXAMPLE_SRCS) $(ABE_EXAMPLE_SRCS) $(UNANNOTATED_ECHO_EXAMPLE_SRCS) $(RESPONSE_BODY_EXAMPLE_SRCS) $(GENERATE_UNBOUND_METHODS_EXAMPLE_SRCS)
```
johanbrandhorst, line 152: I don't mind if we want to skip the swagger client generation altogether for this new service, it doesn't add much.

Makefile
```diff
@@ -210,8 +229,13 @@ $(RESPONSE_BODY_EXAMPLE_SRCS): $(RESPONSE_BODY_EXAMPLE_SPEC)
 	    -l go -o examples/internal/clients/responsebody --additional-properties packageName=responsebody
 	@rm -f $(EXAMPLE_CLIENT_DIR)/responsebody/README.md \
 		$(EXAMPLE_CLIENT_DIR)/responsebody/git_push.sh
+$(GENERATE_UNBOUND_METHODS_EXAMPLE_SRCS): $(GENERATE_UNBOUND_METHODS_EXAMPLE_SPEC)
+	$(SWAGGER_CODEGEN) generate -i $(GENERATE_UNBOUND_METHODS_EXAMPLE_SPEC) \
+	    -l go -o examples/internal/clients/generateunboundmethods --additional-properties packageName=generateunboundmethods
+	@rm -f $(EXAMPLE_CLIENT_DIR)/generateunboundmethods/README.md \
+		$(EXAMPLE_CLIENT_DIR)/generateunboundmethods/git_push.sh
```
johanbrandhorst, line 236: We can skip this

johanbrandhorst (changes requested):
Great work! Lets remove the swagger client generation, it doesn't add much value.

Makefile
plaflamme: Yeah, it was cumbersome to add, but it does test that the generated swagger file is valid. Unless I'm missing something?

johanbrandhorst: That is a fair point, and I agree that does add value. Lets keep it!

Makefile
plaflamme: Do you mean skip the `swagger-codegen` step altogether? I think it makes sense to leave it in to confirm that we're still generating valid swagger, but I also don't mind removing it if you prefer.

plaflamme:
@johanbrandhorst I've added a few commits to address your comments. The only outstanding thing is whether we should run `swagger-codegen` on the swagger file. I think either way is fine, so please let me know what you prefer.

Makefile
plaflamme: I just noticed this comment after adding [this one](https://github.com/grpc-ecosystem/grpc-gateway/pull/1652#issuecomment-692041348). So then I think this is gtg!

johanbrandhorst:
> @johanbrandhorst I've added a few commits to address your comments. The only outstanding thing is whether we should run `swagger-codegen` on the swagger file. I think either way is fine, so please let me know what you prefer.

You've convinced me that it is valuable to have, since it checks that our automatic annotations must produce valid swagger files, as defined by the third party generator we're using, so lets leave them in.

johanbrandhorst (changes requested):
The last thing we need to do is take a pass over the existing documentation and see if there's anything that could do with amending with this change. I can't think of anything off the top of my head, but I think it would be cool to add a section to both the intro and the more detailed pages that you can now use the gateway generator without _any_ annotations, which hopefully opens it up to some new users. The docs are in Markdown files in the `docs` directory.

plaflamme:
@johanbrandhorst I've added https://github.com/grpc-ecosystem/grpc-gateway/pull/1652/commits/aaf2c61db1ad7d1d5e2721668eabd0d9865508e5 for the documentation, let me know what you think

johanbrandhorst (commented):
The docs additions look great - just one more thought; I think we should add this option straight to the README too. Step 2 in the getting started guide talks about adding the annotations - we can add a step (2.a) that mentions how to use the gateway without adding any annotations at all, and move the existing step 2 to (2.b). What do you think?

Really great work on this Philippe, very happy to have you as a contributor :smile:.

plaflamme:
@johanbrandhorst Glad I can help! Please take a look, I took a stab at re-jiggering the README to put this in more prominently. Here's the rendered version: https://github.com/plaflamme/grpc-gateway/blob/generate-unbound-methods/README.md

README.md
```diff
@@ -96,8 +96,58 @@ Make sure that your `$GOBIN` is in your `$PATH`.
...
-annotation to your .proto file
+2. Generate gRPC stubs
+
+    This step generates the gRPC stubs that you can use to implement the service and consume from clients:
+
+    Here's an example of what a `protoc` command might look like to generate Go stubs:
+
+    ```sh
+    protoc -I. --go_out=plugins=grpc,paths=source_relative:./gen/go/ your/service/v1/your_service.proto
+    ```
```
johanbrandhorst, line 107: Thanks for starting this, lets clean this up while we're here:
```suggestion
    ```sh
    protoc -I . --go_out ./gen/go/ --go_opt plugins=grpc --go_opt paths=source_relative your/service/v1/your_service.proto
    ```
```

README.md
```diff
@@ -96,8 +96,58 @@ Make sure that your `$GOBIN` is in your `$PATH`.
...
+
+3. Implement your service in gRPC as usual
+
+   1. (Optional) Generate gRPC stub in the [other programming languages](https://grpc.io/docs/).
+
+     For example, the following generates gRPC code for Ruby based on `your/service/v1/your_service.proto`:
+     ```sh
+     protoc -I. --ruby_out=./gen/ruby your/service/v1/your_service.proto
```
johanbrandhorst, line 115:
```suggestion
     protoc -I . --ruby_out ./gen/ruby your/service/v1/your_service.proto
```

README.md
```diff
@@ -96,8 +96,58 @@ Make sure that your `$GOBIN` is in your `$PATH`.
...
+
+     protoc -I. --grpc-ruby_out=./gen/ruby your/service/v1/your_service.proto
```
johanbrandhorst, line 117:
```suggestion
     protoc -I . --grpc-ruby_out ./gen/ruby your/service/v1/your_service.proto
```

README.md
```diff
@@ -96,8 +96,58 @@ Make sure that your `$GOBIN` is in your `$PATH`.
...
+
+    This requires no additional modification to the `.proto` file, but does require enabling a specific option when executing the plugin.
+    The `generate_unbound_methods` should be enabled.
+
+    Here's what a `protoc` execution might look like with this option enabled:
+
+    ```sh
+       protoc -I. --grpc-gateway_out=logtostderr=true,paths=source_relative:./gen/go \
+         --grpc-gateway_opt=generate_unbound_methods=true \
+         your/service/v1/your_service.proto
```
johanbrandhorst, line 144:
```suggestion
    ```sh
       protoc -I . --grpc-gateway_out ./gen/go \
         --grpc-gateway_opt logtostderr=true \
         --grpc-gateway_opt paths=source_relative \
         --grpc-gateway_opt generate_unbound_methods=true \
```

README.md
```diff
@@ -130,46 +180,30 @@ annotation to your .proto file
    See [a_bit_of_everything.proto](examples/internal/proto/examplepb/a_bit_of_everything.proto)
    for examples of more annotations you can add to customize gateway behavior
    and generated Swagger output.
+   
+   Here's what a `protoc` execution might look like:
+
+    ```sh
+       protoc -I. --grpc-gateway_out=logtostderr=true,paths=source_relative:./gen/go \
+         your/service/v1/your_service.proto
```
johanbrandhorst, line 188:
```suggestion
    ```sh
       protoc -I . --grpc-gateway_out ./gen/go \
         --grpc-gateway_opt logtostderr=true \
         --grpc-gateway_opt paths=source_relative \
```

README.md
```diff
@@ -130,46 +180,30 @@ annotation to your .proto file
...
-     ```
-   2. Add the googleapis-common-protos gem (or your language equivalent) as a dependency to your project.
-   3. Implement your gRPC service stubs
+   Here's what a `protoc` execution might look like with this option enabled:
 
-5. Generate reverse-proxy using `protoc-gen-grpc-gateway`
+    ```sh
+       protoc -I. --grpc-gateway_out=logtostderr=true,paths=source_relative:./gen/go \
+         --grpc-gateway_opt=grpc_api_configuration=path/to/config.yaml \
+        your/service/v1/your_service.proto
```
johanbrandhorst, line 203:
```suggestion
       protoc -I . --grpc-gateway_out ./gen/go \
         --grpc-gateway_opt logtostderr=true \
         --grpc-gateway_opt paths=source_relative \
         --grpc-gateway_opt grpc_api_configuration=path/to/config.yaml \
         your/service/v1/your_service.proto
```

README.md
```diff
@@ -220,12 +254,14 @@ annotation to your .proto file
    }
    ```
 
-7. (Optional) Generate swagger definitions using `protoc-gen-swagger`
+6. (Optional) Generate swagger definitions using `protoc-gen-swagger`
 
    ```sh
    protoc -I. --swagger_out=logtostderr=true:./gen/swagger your/service/v1/your_service.proto
    ```
```
johanbrandhorst, line 261: Lets update the rest of the invocations in the README while we're here, I think you can see the pattern.
```suggestion
   ```sh
   protoc -I . --swagger_out ./gen/swagger --swagger_opt logtostderr=true your/service/v1/your_service.proto
   ```
```

johanbrandhorst (changes requested):
Great job, just a few comments on how we want to format the `protoc` invocations. I think these new invocations are a bit clearer (removes a lot of `=`). What do you think?

plaflamme:
@johanbrandhorst I've added your suggestions and made another README fixup in https://github.com/grpc-ecosystem/grpc-gateway/pull/1652/commits/3249fb0f4ef048174d5a9f212bd15984309fc834 please take a look

johanbrandhorst (approved):
Awesome, thanks again for this Philippe, it's been a privilege to work with you!

johanbrandhorst:
Could you please cherry pick this against v2? I expect there'll be a few merge conflicts, so let me know if you need some help.

johanbrandhorst:
Hi Philippe, just checking, do you intend to cherry pick this against v2? I'm waiting for this before making another beta release of v2, and if you're busy I can complete the cherry picking myself, with your permission.

plaflamme:
@johanbrandhorst I did do the cherry pick locally, but it turned out with a lot of conflicts in places that I'm not familiar with. I don't think I can invest the necessary time to learn the codebase enough to confidently resolve the conflicts... Please go ahead and do the cherry-picking, I would appreciate it.

Also, you had mentioned that it could potentially become the default behaviour in v2. Do you plan on making turning this into a different option, perhaps something like `no_unbound_methods`?

johanbrandhorst:
I see, thanks, I'll take care of the cherry pick. I'll also see if we need to change the name if we change the default behavior, but I don't think so, we'll just add a note to the migration guide.

PR #1665, opened 2020-09-14
Add support for OpenAPI YAML annotations

### LINKED ISSUE #701: YAML support for Swagger annotations
Hi there, first of all great project thanks for all your hard work.

I'm experimenting with the swagger doc generation and have got things to work fine in the proto file. I am interested in keeping all of the swagger / rest stuff out of the proto file and use the associated YAML file that can be used to configure endpoints, as per
https://cloud.google.com/endpoints/docs/grpc/grpc-service-config

Is it possible to add the protogen-swagger annotation messages in the YAML file instead? I'd like to write something like:
```
# API Reference: https://cloud.google.com/endpoints/docs/grpc-service-config/reference/rpc/google.api#google.api.Service

type: google.api.Service
config_version: 3

name: api.vega.trading# API Reference: https://cloud.google.com/endpoints/docs/grpc-service-config/reference/rpc/google.api#google.api.Service

type: google.api.Service
config_version: 3
info:
  title: "A Bit of Everything";
  version: "1.0";
  contact: 
    name: "gRPC-Gateway project";
    url: "https://github.com/grpc-ecosystem/grpc-gateway";
    email: "none@example.com";
```

jasonewang:

Fixes #701

Yep

Added support for passing OpenAPI/Swagger annotations in YAML file to protoc-gen-openapiv2

* Extended annotations.proto to include OpenAPI option messages
* Extended descriptor.Registry to hold OpenAPI options
* Extended template to look up options in Registry if proto descriptors don't have options
#### Other comments
Extended template tests that tested proto options to set options via registry.


jasonewang:
I think this is ready for review, though I can't quite figure how to fix the generate CI test.

protoc-gen-openapiv2/main.go
```diff
@@ -113,6 +114,13 @@ func main() {
 		return
 	}
 
+	if *openAPIConfiguration != "" {
+		if err := reg.LoadOpenAPIConfigFromYAML(*openAPIConfiguration); err != nil {
+			emitError(err)
+			return
+		}
+	}
```
johanbrandhorst, line 122: One thing I was wondering was how this will interact with existing annotations in the file, I take it loading this before we go through the files means any file annotations take precedence? I think that's what we want, anyway.

johanbrandhorst (changes requested):
This PR is an amazing achievement! Well done! I'd like to see an example of these new annotations in action, do you think you could add some annotations to the `unannotated_echo_service.yaml` to show users how to use these new annotations, and as a test to see that it generates correctly?

The generation error is because of a dependency issue in the generation job - `make realclean` deletes all the generated files, and then `make examples` tries to install the generators before regenerating the files - but now the grpc-gateway generator depends on the generated openapi options, so we can't build it. We'll probably have to split up cleaning and building, so that we can still have the confidence that the generated files are the same as are checked into the repo (by deleting and regenerating). There is a bit of an interesting chicken-and-egg problem here though in that we can't be sure that the files we regenerated are correct until we've regenerated them and reinstalled the generators and generated them again, since what if the generator was built with an old generated file?

Another option which is tempting is to remove the dependency on protoc-gen-openapiv2 options from `protoc-gen-grpc-gateway`. This seems a bit of a weird dependency anyway, and the part of apiconfig that needs the new options is actually unused by `protoc-gen-grpc-gateway`. We may need to split the packages.

What do you think?

jasonewang:
> ...do you think you could add some annotations to the `unannotated_echo_service.yaml` to show users how to use these new annotations, and as a test to see that it generates correctly?

Sure. Should I also add options inline to echo_service.proto to maintain parity between the annotated and unannotated versions for the messages and services that are in both the annotated and unannotated protos?

> Another option which is tempting is to remove the dependency on protoc-gen-openapiv2 options from `protoc-gen-grpc-gateway`. This seems a bit of a weird dependency anyway, and the part of apiconfig that needs the new options is actually unused by `protoc-gen-grpc-gateway`. We may need to split the packages.

Splitting the package up makes sense to me. Would moving the OpenAPI config options to github.com/grpc-ecosystem/grpc-gateway/v2/internal/descriptor/openapiconfig" suffice?

protoc-gen-openapiv2/main.go
jasonewang: The extraction of OpenAPI options looks for the inline options first and looks up annotated registered from the YAML as a fallback (example:  [protoc-gen-openapiv2/internal/genopenapi/template.go:getMethodOpenAPIOption](https://github.com/grpc-ecosystem/grpc-gateway/pull/1665/files#diff-caf612205311291b1a186109ab244a23R1794)). It's easily changed to have the YAML options take precedence, and I think we should have the same precedence as the service annotations, though I confess to not knowing what that is.

internal/descriptor/registry.go
```diff
@@ -582,3 +605,111 @@ func (r *Registry) packageIdentityName(f *descriptorpb.FileDescriptorProto) stri
...
+
+// RegisterOpenAPIOptions registers OpenAPI options
+func (r *Registry) RegisterOpenAPIOptions(opts *apiconfig.OpenAPIOptions) error {
+	if opts == nil {
+		return nil
+	}
+
+	for _, opt := range opts.File {
+		if _, ok := r.files[opt.File]; !ok {
+			return fmt.Errorf("no file %s found", opt.File)
```
jasonewang, line 617: I think these errors should instead be warnings, like when you add an HTTP for a non-existent method. Is that correct?

jasonewang: I misremebered and unbound HTTP rules are errors (https://github.com/grpc-ecosystem/grpc-gateway/blob/v2/protoc-gen-grpc-gateway/main.go#L112).

jasonewang:
Looks like `protoc-gen-grpc-gateway` pulls in descriptor.Registry which depends on "internal/descriptor/openapiconfig/openapiconfig.pb.go" and thus "github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2/options", so splitting out the package doesn't have the desired effect of breaking the dependency.

I'm a bit confused at how `protoc-gen-openapiv2` is being built because that plugin definitely requires the package github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2/options.

jasonewang:
> I'm a bit confused at how `protoc-gen-openapiv2` is being built because that plugin definitely requires the package github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2/options.

Turns out that `make realclean` deletes the Go stubs at github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2/options. Adding those stubs as a dependency for `protoc-gen-grpc-gateway` (https://github.com/grpc-ecosystem/grpc-gateway/pull/1665/files#diff-b67911656ef5d18c4ae36cb6741b7965R160) solves it.

johanbrandhorst (changes requested):
The final step would be to add a little blurb about these new capabilities to the docs - we already have a page for the grpc api configuration and it would make sense to add this to that page. The docs can be found in the /docs subfolder, and are simple markdown documents.

Thanks again for this work, it's really appreciated!

protoc-gen-openapiv2/main.go
johanbrandhorst: Yeah, I think this is fine.

johanbrandhorst:
Hi, I merged a generator update so you'll need to rebase on v2, sorry for the inconvenience :grimacing:.

jasonewang:
> Hi, I merged a generator update so you'll need to rebase on v2, sorry for the inconvenience 😬.

No problem! I rebased and updated the docs.

docs/_docs/grpcapiconfiguration.md
```diff
@@ -61,7 +63,23 @@ The following is equivalent to the basic [usage example](usage.html) but without
...
     ```
-   
+
    This will generate a reverse proxy `gen/go/your/service/v1/your_service.pb.gw.go` that is identical to the one produced for the annotated proto.
 
+6. Generate the optional your_service.swagger.json
+
+    ```sh
+    protoc -I. --swagger_out=grpc_api_configuration=path/to/your_service.yaml:./gen/go \
+      your/service/v1/your_service.proto
```
johanbrandhorst, line 73: Lets update the protoc invocations in this file to make this a bit easier to read:
```suggestion
    protoc -I . --swagger_out ./gen/go \
      --swagger_opt grpc_api_configuration=path/to/your_service.yaml \
      your/service/v1/your_service.proto
```

docs/_docs/grpcapiconfiguration.md
```diff
@@ -61,7 +63,23 @@ The following is equivalent to the basic [usage example](usage.html) but without
...
+    ```
+
+    or using an OpenAPI configuration file
+
+    ```sh
+    protoc -I. --swagger_out=grpc_api_configuration=path/to/your_service.yaml,openapi_configuration=path/to/your_service_swagger.yaml:./gen/go \
+      your/service/v1/your_service.proto
+    ```
```
johanbrandhorst, line 81:
```suggestion
    ```sh
    protoc -I . --swagger_out ./gen/go \
      --swagger_opt grpc_api_configuration=path/to/your_service.yaml \
      --swagger_opt openapi_configuration=path/to/your_service_swagger.yaml \
    ```
```

johanbrandhorst (changes requested):
Just some small thoughts on the new docs and how we want to display the protoc invocations. Great work on this Jason!

jasonewang:
What do you think about moving openapiconfig.proto to an exported package instead of under internal? My idea is that it would make generating OpenAPI annotations easier if you could fill out the struct in Go, then marshal to JSON and then to YAML instead of having to work with templates.

johanbrandhorst:
> What do you think about moving openapiconfig.proto to an exported package instead of under internal? My idea is that it would make generating OpenAPI annotations easier if you could fill out the struct in Go, then marshal to JSON and then to YAML instead of having to work with templates.

I'd like to limit our API surface as much as possible. I see the benefit of what you're saying, but I'd be more inclined to let users copy the generated struct for their internal use.

johanbrandhorst (approved):
This is great, thank you so much for your work!

johanbrandhorst:
I'm going to make another beta release of v2 once we've cherry picked the new unbound method generator, we've had some incredible contributions lately, yours definitely among them!

jasonewang:
Thanks for being so responsive to review requests, Johan! Happy to contribute.

Issue #1612, opened 2020-08-25
protoc-gen-grpc-gateway: can't generate gateway stubs for multiple packages in one go


adambabik:
## 🐛 Bug Report

I would like to execute `protoc` with `protoc-gen-grpc-gateway` to generate stubs for all my packages in one go. Unfortunately, I am getting an error: `--grpc-gateway_out: inconsistent package names: av1 bv1`.

## To Reproduce

1. Organize the packages like this:
```
proto
└── my-org
    ├── a
    │   └── v1
    │       └── a.proto
    └── b
        └── v1
            └── b.proto

```

where `a.proto`'s package is `av1` and `b.proto`'s package name is `bv1`.

2. Execute `protoc` in the following way:
```
protoc \
    -I=/proto \
    -I=/third_party \
    --go_opt=paths=source_relative \
    --go_out=/gen \
    --go-grpc_opt=paths=source_relative \
    --go-grpc_out=/gen \
    --grpc-gateway_out=paths=source_relative,logtostderr=true:/gen \
    --validate_out=paths=source_relative,lang=go:/gen \
    $(find /proto -name '*.proto')
```

## Expected behavior

I expect to have grpc-gateway stubs generated in corresponding directories.

## Actual Behavior

I am getting an error `--grpc-gateway_out: inconsistent package names: av1 bv1`. However, when executing the `protoc` command package-by-package, it seems that the stubs are generated properly in the corresponding directories in accordance with `paths=source_relative`.

## Your Environment

I execute this in a Docker container. It uses `golang:1.14-stretch` as a base image. 

The version of `protoc-gen-grpc-gateway` is `v1.14.7`.

johanbrandhorst:
This would be cool to support, but it's definitely a bigger piece of work. I'd love to hepl you bring this in if you have the spare cycles but we're not going to be able to add this unless it's as a community contribution. Note that `protoc-gen-go` doesn't support this (https://github.com/golang/protobuf/issues/39), so you usually end up writing separate protoc invocations for each package anyway, but I can see this being a request for non-go users.

adambabik:
@johanbrandhorst I do have spare cycles and actually started digging into the code. If you could give me some introduction how this feature can be added that would be awesome.

[...]

PR #1668, opened 2020-09-14
refactor: upgrade protoc-gen-grpc-gateway to use the new Go code generator

adambabik:

This is initial work to refactor how the gRPC Gateway files are generated in order to fix a problem described in https://github.com/grpc-ecosystem/grpc-gateway/issues/1612.

This change upgrades protoc-gen-grpc-gateway to use google.golang.org/protobuf/compiler/protogen which is the new package for writing protoc plugins generating Go code.

It also fixes all places to use packages from google.golang.org/protobuf module instead of github.com/golang/protobuf.

#### Other comments

I tested it by running `go test ./...` and regenerating all `examples/internal/proto/examplepb/*.gw.go` files using `protoc` with a new `bin/protoc-gen-grpc-gateway` binary. There were no changes.

### DIFF
--- a/internal/codegenerator/BUILD.bazel
+++ b/internal/codegenerator/BUILD.bazel
@@ -10,8 +10,8 @@ go_library(
     ],
     importpath = "github.com/grpc-ecosystem/grpc-gateway/v2/internal/codegenerator",
     deps = [
-        "@io_bazel_rules_go//proto/wkt:compiler_plugin_go_proto",
         "@org_golang_google_protobuf//proto:go_default_library",
+        "@org_golang_google_protobuf//types/pluginpb:go_default_library",
     ],
 )
 
@@ -21,8 +21,8 @@ go_test(
     embed = [":go_default_library"],
     deps = [
         "@com_github_google_go_cmp//cmp:go_default_library",
-        "@io_bazel_rules_go//proto/wkt:compiler_plugin_go_proto",
         "@org_golang_google_protobuf//proto:go_default_library",
         "@org_golang_google_protobuf//testing/protocmp:go_default_library",
+        "@org_golang_google_protobuf//types/pluginpb:go_default_library",
     ],
 )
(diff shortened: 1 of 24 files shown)


adambabik:
@googlebot I signed it!

adambabik:
It seems like I need to update Bazel BUILD files. Is there a way to regenerate them or I should fix them manually?

achew22:
@adambabik thanks for your contribution! I'm extremely excited for this one. Assuming the major structure hasn't changed (looking through the PR it doesn't look like it but I won't rule it out entirely), you should be able to invoke http://godoc.org/github.com/bazelbuild/bazel-gazelle/cmd/gazelle in the root of the repo and it will generate all the `BUILD` file changes for you. If that doesn't work, please post back here and I'll fix them manually on your PR

johanbrandhorst:
This is a really impressive piece of work. I haven't looked at it yet but I think you'll find that we can't replace the old github.com/golang/protobuf imports yet because of how bazel imports them. If you revert the import changes (for now) it should be possible to regenerate the files using the commands listed in CONTRIBUTING.md.

adambabik:
@achew22 thanks, it seems that it worked and the CI pipeline is happy :)

@johanbrandhorst Thanks! It's just a first step actually, there will be more changes to fully embrace `protogen.Plugin`. Here, I merely wrapped the existing code into it.

I actually managed to run the commands from CONTRIBUTING.md regenerating files with the existing changes and they did not complain. Do you think I still should revert to `github.com/golang/protobuf` imports due to some reason?

johanbrandhorst (approved):
This looks great, thank you so much for taking the time to get this done! I understand it's just the start, but great work!

johanbrandhorst:
Thanks for your contribution!

achew22:
To echo Johan even more, thank you so much! I think this will really help cut down on the bugs we get where there is some weird flag parsing thing that I implement incorrectly 5 years ago 😁

adambabik:
Thank you folks! :) I will continue working on the next PR to fully embrace [protogen](https://pkg.go.dev/google.golang.org/protobuf/compiler/protogen?tab=doc) package.

PR #1684, opened 2020-09-22
Support generating unbound methods

johanbrandhorst:
Cherry picks https://github.com/grpc-ecosystem/grpc-gateway/commit/141dba338026b98a1d03e09df8fbcac248b9cd8e
from master.

Co-authored-by: Philippe Laflamme <484152+plaflamme@users.noreply.github.com>

@plaflamme please confirm that you consent for this to be merged.


johanbrandhorst:
I'll look into the generation errors tomorrow, could be a mistake in the cherry picking.

plaflamme:
I consent

plaflamme:
Nice, thanks @johanbrandhorst for the cherry-pick and helping move this feature forward; very much appreciated!

johanbrandhorst:
https://github.com/grpc-ecosystem/grpc-gateway/releases/tag/v2.0.0-beta.5 is now available with this feature included :). I didn't change the default behaviour (yet) as I wasn't convinced it would be the right thing.

plaflamme:
My 2 cents on why changing the default behaviour would be good:

* default behaviour "just works" on any input;
* default behaviour similar to existing tools, i.e. Cloud Endpoints;
* safe if an option to disable is also provided:
    * users on v1 not using `generate_unbound_methods` can use the new option to get the old behaviour back
    * users on v1 using `generate_unbound_methods` can simply drop the argument (which could be still available and simply print a deprecation warning)

That said, I don't have a lot of experience with this tool, so perhaps others with more experience may think differently... Thanks again!

PR #1687, opened 2020-09-23
Add devcontainer.json

johanbrandhorst:
Adds support for VS Code dev containers, which allows using the
build environment in VS Code and Github Codespaces.

See https://code.visualstudio.com/docs/remote/containers and
https://github.com/features/codespaces for more information.


.bazelversion
```diff
@@ -0,0 +1 @@
+3.5.0
```
jadekler, line 1: Screen Shot 2020-09-23 at 07 46 34

PR #118, opened 2016-03-01
Replace glog with grpclog

cuongdo:
This allows projects with forked versions of `glog` (e.g. CockroachDB) to use `grpc-gateway` without a flag collision (runtime panic) on the `glog` flags such as `-logtostderr`. This also allows integration with any logging systems that have been integrated with `grpclog`.

Note that I only replaced `glog` in the `runtime` package and in the files generated by `protoc-gen-grpc-gateway`. The use of `glog` elsewhere in this repo doesn't seem to affect ability to use `grpc-gateway` in projects with forked `glog` versions.

Fixes #92.

[Review on Reviewable](https://reviewable.io/reviews/gengo/grpc-gateway/118)

PR #1694, opened 2020-09-24
Remove indirect github.com/golang/glog dependency of runtime package

jasonewang:
Similar intended fix as #118

Yep

Remove glog in dependency of runtime package so projects that have forked `glog` can use grpc-gateway without a flag collision. 

#1449 introduced the httprule import to mux.go which causes `glog` to be imported when runtime is imported.

#### Other comments
There are other glog imports in internal/, protoc-gen-grpc-gateway/, and protoc-gen-openapiv2/ that don't seem to cause flag collisions when creating a mux and using the generated .pb.gw.go files. I can remove or update those to use grpclog instead if desired.

### DIFF
--- a/internal/httprule/BUILD.bazel
+++ b/internal/httprule/BUILD.bazel
@@ -10,10 +10,7 @@ go_library(
         "types.go",
     ],
     importpath = "github.com/grpc-ecosystem/grpc-gateway/v2/internal/httprule",
-    deps = [
-        "//utilities:go_default_library",
-        "@com_github_golang_glog//:go_default_library",
-    ],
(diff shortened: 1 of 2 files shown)


johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution!

johanbrandhorst:
Could you please cherry pick this against master?

jasonewang:
> Could you please cherry pick this against master?

This dependency as far as I can tell doesn't exist in master because httprule/parse.go is only used in protoc-gen-grpc-gateway which isn't imported by any user code.

Issue #1697, opened 2020-09-25
Support WithOutgoingTrailerMatcher


maja42:
## 🚀 Feature

grpc-gateway offers the possibility do define custom `outgoingHeaderMatcher` to decide which grpc-headers should be forwarded as http-headers via a ServeMuxOption. However, grpc trailer-headers are always forwarded in the http-trailers.

In my case, I don't want to forward any grpc metadata to the http API. So I defined a headerMatcher that always returns false. But I can't do the same for trailers. It would be nice to have that possibility.

(As a side note: sending http trailers also removes the Content-Length Header from the response)

johanbrandhorst:
This sounds reasonable! A side note; we disabled headers altogether for clients that don't include the `"TE"` header, see https://github.com/grpc-ecosystem/grpc-gateway/blob/v2/runtime/errors.go#L113, but it seems we only disabled it in our error handler, not in our general purpose message handler, see https://github.com/grpc-ecosystem/grpc-gateway/blob/v2/runtime/handler.go#L140. I'd love to have the previous logic added to this section of the code as well, to be compliant with RFC 7230. Would this be enough of a fix for your use case? Would you be willing to add this functionality?

maja42:
This sounds good, but it doesn't solve my use-case completely. I'd still like to have a outgoingTrailerMatcher.

Reason: We are using the grpc-gateway to provide an external API for our grpc-based microservice architecture. So it acts as a single-point-of-entry and we must be careful about which internal information we want to expose.

We use grpc-headers and trailers to transmit internal information between services, and their content should generally not be exposed via the http API. So even if the browser/user would send a "TE" header (request trailer data), I don't want to return any internal trailer information.

johanbrandhorst:
Fair enough, sorry for hijacking this issue, I'll change the title and open a separate one for the bug I mentioned.

PR #1721, opened 2020-10-05
added trailer match capability #1709

### LINKED ISSUE #1709: Allow client to disable sending of trailers by omitting "TE" header.
## 🐛 Bug Report

We shouldn't send trailers unless the client sends the "TE" trailer.

See https://github.com/grpc-ecosystem/grpc-gateway/issues/1697 for more context

bytecook:
Related to #1697

Fixes #1709 
(hopefully)

### DIFF
--- a/runtime/mux.go
+++ b/runtime/mux.go
@@ -33,6 +33,7 @@ type ServeMux struct {
 	marshalers                marshalerRegistry
 	incomingHeaderMatcher     HeaderMatcherFunc
 	outgoingHeaderMatcher     HeaderMatcherFunc
+	outgoingTrailerMatcher	  TrailerMatcherFunc
 	metadataAnnotators        []func(context.Context, *http.Request) metadata.MD
 	streamErrorHandler        StreamErrorHandlerFunc
 	protoErrorHandler         ProtoErrorHandlerFunc
@@ -67,6 +68,11 @@ func SetQueryParameterParser(queryParameterParser QueryParameterParser) ServeMux
 // HeaderMatcherFunc checks whether a header key should be forwarded to/from gRPC context.
 type HeaderMatcherFunc func(string) (string, bool)
 
+// TrailerMatcherFunc checks wherther a trailer should be forwarded from gRPC context.
+//
+// Best practice: omit "TE" header if no header or trailers need be forwarded from gRPC context.
+type TrailerMatcherFunc func(string) (string, bool)
+
 // DefaultHeaderMatcher is used to pass http request headers to/from gRPC context. This adds permanent HTTP header
 // keys (as specified by the IANA) to gRPC context with grpcgateway- prefix. HTTP headers that start with
 // 'Grpc-Metadata-' are mapped to gRPC metadata after removing prefix 'Grpc-Metadata-'.
@@ -98,6 +104,16 @@ func WithIncomingHeaderMatcher(fn HeaderMatcherFunc) ServeMuxOption {
 func WithOutgoingHeaderMatcher(fn HeaderMatcherFunc) ServeMuxOption {
 	return func(mux *ServeMux) {
 		mux.outgoingHeaderMatcher = fn
+	}
+}
+
+//WithOutgoingTrailerMatcher returns a ServeMuxOption representing a trailerMatcher for outgoing response from gateway.
+//
+//Similar to func WithOutgoingHeaderMatcher
+//Best practice: omit "TE" header if no header or trailers need be forwarded from gRPC context.
+func WithOutgoingTrailerMatcher(fn TrailerMatcherFunc) ServeMuxOption {
+	return func(mux *ServeMux) {
+		mux.outgoingTrailerMatcher = fn
 	}
 }
 


bytecook:
> @googlebot

I signed it!

runtime/mux.go
```diff
@@ -101,6 +107,16 @@ func WithOutgoingHeaderMatcher(fn HeaderMatcherFunc) ServeMuxOption {
 	}
 }
 
+//WithOutgoingTrailerMatcher returns a ServeMuxOption representing a trailerMatcher for outgoing response from gateway.
+//
+//Similar to func WithOutgoingHeaderMatcher
+//Best practice: omit "TE" header if no header or trailers need be forwarded from gRPC context.
```
johanbrandhorst, line 113: Please add a space after the slashes
```suggestion
// WithOutgoingTrailerMatcher returns a ServeMuxOption representing a trailerMatcher for the outgoing response from the gateway.
//
// This matcher will be called with each trailer in the response trailer metadata. If matcher returns true, that header will be
// passed to the http response returned from the gateway. To transform the trailer before passing to response,
// the matcher should return a modified header.
// Best practice: omit "TE" header if no trailers need to be forwarded from the gRPC server.
```

runtime/mux.go
```diff
@@ -67,6 +68,11 @@ func SetQueryParameterParser(queryParameterParser QueryParameterParser) ServeMux
 // HeaderMatcherFunc checks whether a header key should be forwarded to/from gRPC context.
 type HeaderMatcherFunc func(string) (string, bool)
 
+// TrailerMatcherFunc checks wherther a trailer should be forwarded from gRPC context.
+//
+// Best practice: omit "TE" header if no header or trailers need be forwarded from gRPC context.
```
johanbrandhorst, line 73:
```suggestion
// TrailerMatcherFunc checks whether a trailer should be forwarded from the gRPC server.
//
// Best practice: omit "TE" header if no trailers need to be forwarded from the gRPC server.
```

johanbrandhorst (changes requested):
Thanks for this PR! It seems that you're missing the logic for actually using the new field to filter the trailers. We'll need to add this somewhere in the ServeMux logic, I suggest looking at how `outgoingHeaderMatcher` is used.

Also I'd like a test for this before we can merge it. Again I suggest looking at how `outgoingHeaderMatcher` is tested.

Let me know if you need anymore pointers!

bytecook:
I am unsure of how to test the handler. Should I modify the TestForwardResponseMessage func to include tests with trailers in handler_test.go? Or append a spec to the TestMuxServeHTTP func in mux_test.go? I'm assuming the latter, even though the changes were made in handler.go.

johanbrandhorst:
> I am unsure of how to test the handler. Should I modify the TestForwardResponseMessage func to include tests with trailers in handler_test.go? Or append a spec to the TestMuxServeHTTP func in mux_test.go? I'm assuming the latter, even though the changes were made in handler.go.

Sorry if I wasn't clear. I thought we had some tests for `outgoingHeaderMatcher` but it doesn't look like it :joy:. I think these tests would probably live in `handler_test`, maybe in a new test that can test the functionality of the outgoing trailer matcher. You can copy one of the existing tests and play around with it until it does what you want, but at the very least we should have a test without the trailer matcher set, and another test with a custom trailer matcher and check that the behaviour changes as expected.

Does that sound okay?

bytecook:
Got it. No worries. I didn't add the logic at first because I mistakenly thought the implementation of the interface would be done by the client, but even back then I still felt like this was incorrect. I will continue to work on this until it is up to standard.

bytecook:
Still unsure of how to test in handler_test.go, the only test headers I see are in mux_test.go.I feel like both test files need to be modified in order to test this functionality. The modified handler was the non-exported func handleForwardResponseServerMetadata(w http.ResponseWriter, mux *ServeMux, md ServerMetadata)

bytecook:
The reason why the CI tests pass is because if there is no specified outgoingTrailerMatcher, "TE" will be passed. The original issue wanted to be agnostic to whether or not "TE" was passed. 
However, gRPC trailers will be forwarded by default, so I believe this is in conflict with #1709

johanbrandhorst:
I'm not sure what you're trying to say. There are two different concerns here:

- Trailers should not be sent unless the "TE" header is sent in the request. This is #1709.
- There should be a way for users to customize how to return headers. This is #1697.

These are separate concerns, there is no conflict between the two. This PR should tackle #1697. Don't worry about the other issue for the purposes of this PR.

Re: testing, I think you should be able to copy one of the tests in `handler_test.go` and adjust it so that it can test the behaviour of the outgoingTrailerMatcher. Have you tried doing this?

bytecook:
I am not necessarily saying there is a conflict between the concerns. I'm saying that the way I resolved issue #1697 the "TE" header is set by default, with or without values. Not sure if what I did is the correct approach, but it is why it passes the CI tests.

Ok. I'll spend more time attempting to modify the handler_test.go.

Also, If you have any tips for me to improve my communication, please let me know.

johanbrandhorst:
Hi, could you please rebase this on `master`? We've merged v2 into master.

johanbrandhorst:
Are you still intending to merge this? Otherwise I will close it soon.

PR #1754, opened 2020-10-14
Fix -module option for protoc-gen-grpc-gateway

### LINKED ISSUE #1753: Option -module does not work anymore for protoc-gen-grpc-gateway v2
## 🐛 Bug Report

The option module fails for protoc-gen-grpc-gateway v2.

Actually, the issue is that the module prefix is removed a [first time](https://github.com/grpc-ecosystem/grpc-gateway/blob/1aab03caf0bfadbf99a107ffdd86467ac9db7caa/protoc-gen-grpc-gateway/internal/gengateway/generator.go#L142-L145) by protoc-gen-grpc-gateway, then protogen attempts to remove the prefix a [second time](https://github.com/protocolbuffers/protobuf-go/blob/f62736dede0a74248d862247dbcb53ab69582a05/compiler/protogen/protogen.go#L431-L436). It then fails with an error "generated file does not match prefix".

## To Reproduce

1. Create a service file `service.proto`:

   ```proto
   syntax = "proto3";
   package your.service.v1;
   option go_package = "github.com/yourorg/yourproject/your/service/v1";
   message StringMessage {
     string value = 1;
   }
   service YourService {
     rpc Echo(StringMessage) returns (StringMessage) {}
   }
   ```

1. Generate protobuf definitions and go-grpc with option `module=`:

   ```
   protoc -I . \
      --go_out . \
      --go_opt module=github.com/yourorg/yourproject \
      --go-grpc_out . \
      --go-grpc_opt module=github.com/yourorg/yourproject \
      service.proto
   ```

   It generates definitions in the right directory:
   image

1. Then, use protoc-gen-grpc-gateway with the same `module=` option:

   ```
   protoc -I . \
      --grpc-gateway_out . \
       --grpc-gateway_opt module=github.com/yourorg/yourproject \
      --grpc-gateway_opt logtostderr=true \
      --grpc-gateway_opt generate_unbound_methods=true \
      service.proto
   ```
   It fails with:
   ```
   --grpc-gateway_out: your/service/v1/service.pb.gw.go: generated file does not match prefix "github.com/yourorg/yourproject"
   ```

## Expected behavior

I'd expect the last command to generate `your/service/v1/service_grpc.pb.go`.

## Actual Behavior

See above

## Your Environment

Fedora 32

```
➜  protoc --version
libprotoc 3.11.2
➜  protoc-gen-go-grpc -version
protoc-gen-go-grpc 1.0
```

olivierlemasle:

Fixes #1753

Yes

When option `module=...` is used with protoc-gen-grpc-gateway v2, the module prefix is removed a [first time](https://github.com/grpc-ecosystem/grpc-gateway/blob/1aab03caf0bfadbf99a107ffdd86467ac9db7caa/protoc-gen-grpc-gateway/internal/gengateway/generator.go#L142-L145) by protoc-gen-grpc-gateway, then protogen attempts to remove the prefix a [second time](https://github.com/protocolbuffers/protobuf-go/blob/f62736dede0a74248d862247dbcb53ab69582a05/compiler/protogen/protogen.go#L431-L436). It then fails with an error "generated file does not match prefix".

This commit fixes this issue by removing the first prefix handling.

### DIFF
--- a/protoc-gen-grpc-gateway/main.go
+++ b/protoc-gen-grpc-gateway/main.go
@@ -28,7 +28,7 @@ var (
 	allowDeleteBody            = flag.Bool("allow_delete_body", false, "unless set, HTTP DELETE methods may not have a body")
 	grpcAPIConfiguration       = flag.String("grpc_api_configuration", "", "path to gRPC API Configuration in YAML format")
 	pathType                   = flag.String("paths", "", "specifies how the paths of generated files are structured")
-	modulePath                 = flag.String("module", "", "specifies a module prefix that will be stripped from the go package to determine the output directory")
+	_                          = flag.String("module", "", "specifies a module prefix that will be stripped from the go package to determine the output directory")
 	allowRepeatedFieldsInBody  = flag.Bool("allow_repeated_fields_in_body", false, "allows to use repeated field in `body` and `response_body` field of `google.api.http` annotation option")
 	repeatedPathParamSeparator = flag.String("repeated_path_param_separator", "csv", "configures how repeated fields should be split. Allowed values are `csv`, `pipes`, `ssv` and `tsv`.")
 	allowPatchFeature          = flag.Bool("allow_patch_feature", true, "determines whether to use PATCH feature involving update masks (using google.protobuf.FieldMask).")
@@ -90,7 +90,7 @@ func main() {
 			targets = append(targets, f)
 		}
 
-		g := gengateway.New(reg, *useRequestContext, *registerFuncSuffix, *pathType, *modulePath, *allowPatchFeature, *standalone)
+		g := gengateway.New(reg, *useRequestContext, *registerFuncSuffix, *pathType, *allowPatchFeature, *standalone)
 		files, err := g.Generate(targets)
 		for _, f := range files {
 			glog.V(1).Infof("NewGeneratedFile %q in %s", f.GetName(), f.GoPkg)
--- a/protoc-gen-grpc-gateway/internal/gengateway/BUILD.bazel
+++ b/protoc-gen-grpc-gateway/internal/gengateway/BUILD.bazel
@@ -32,7 +32,9 @@ go_test(
     deps = [
         "//internal/descriptor:go_default_library",
         "//internal/httprule:go_default_library",
+        "@org_golang_google_protobuf//compiler/protogen:go_default_library",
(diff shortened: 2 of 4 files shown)


protoc-gen-grpc-gateway/main.go
```diff
@@ -28,7 +28,7 @@ var (
 	allowDeleteBody            = flag.Bool("allow_delete_body", false, "unless set, HTTP DELETE methods may not have a body")
 	grpcAPIConfiguration       = flag.String("grpc_api_configuration", "", "path to gRPC API Configuration in YAML format")
 	pathType                   = flag.String("paths", "", "specifies how the paths of generated files are structured")
-	modulePath                 = flag.String("module", "", "specifies a module prefix that will be stripped from the go package to determine the output directory")
+	_                          = flag.String("module", "", "specifies a module prefix that will be stripped from the go package to determine the output directory")
```
johanbrandhorst, line 31: I understand this flag is handled separately by `protogen`, could we just remove it instead of leaving it in like this? Is the `protogen` flag different? What other flags do you think we may need to worry about having duplicated, if any?

johanbrandhorst (commented):
Thanks a lot for finding this issue, and for the quick fix! I've left a simple question about what to do about the old flag, but I'm ready to merge this without further changes if you think it makes sense.

protoc-gen-grpc-gateway/main.go
olivierlemasle: Actually, `protogen` [uses](https://github.com/protocolbuffers/protobuf-go/blob/f62736dede0a74248d862247dbcb53ab69582a05/compiler/protogen/protogen.go#L180) the flag `module` but do not declare it.
Removing this line result in a failure:
> Cannot set flag module=github.com/yourorg/yourproject
> --grpc-gateway_out: protoc-gen-grpc-gateway: Plugin failed with status code 255

johanbrandhorst: Ha, weird. Alright, I'm happy to merge this now, I think we're going to be changing this part of the code soonish anyway.

olivierlemasle:
Oh, you were quick to merge it! I was about to write that #1756 might supersede this PR...
#1756 uses completely `protogen` for code generation, so it removes the need for this fix (however, it is a bigger PR so it might require a more thorough review)

PR #1755, opened 2020-10-14
Adapt usage instructions to use protoc-gen-go-grpc

olivierlemasle:

N/A

Yes

This reflect recent changes in https://grpc.io/docs/languages/go/quickstart/ to use new protoc-gen-go-grpc.

Also remove a duplicated paragraph.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -4,13 +4,13 @@
 - [ ] Update your protoc to the [latest version](https://github.com/google/protobuf/releases).
 - [ ] Update your copy of the `grpc-gateway` library to the latest version from github:
   ```sh
-  go get -u github.com/grpc-ecosystem/grpc-gateway
+  go get -u github.com/grpc-ecosystem/grpc-gateway/v2
   ```
 - [ ] Delete the `protoc-gen-grpc-gateway` and `protoc-gen-openapiv2` binary from your `PATH`,
   and reinstall the latest versions:
   ```sh
-  go get -u github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway
-  go get -u github.com/grpc-ecosystem/grpc-gateway/protoc-gen-openapiv2
+  go get -u github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway
+  go get -u github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2
   ```
   
 ## I still have a problem!
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -4,13 +4,13 @@
 - [ ] Update your protoc to the [latest version](https://github.com/google/protobuf/releases).
 - [ ] Update your copy of the `grpc-gateway` library to the latest version from github:
   ```sh
-  go get -u github.com/grpc-ecosystem/grpc-gateway/v2
+  go get github.com/grpc-ecosystem/grpc-gateway/v2@latest
   ```
 - [ ] Delete the `protoc-gen-grpc-gateway` and `protoc-gen-openapiv2` binary from your `PATH`,
   and reinstall the latest versions:
   ```sh
-  go get -u github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway
-  go get -u github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2
+  go get github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway
+  go get github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2
   ```
   
 ## I still have a problem!
--- a/README.md
+++ b/README.md
@@ -60,10 +60,10 @@ to track the versions of the following executable packages:
 package tools
 
 import (
-    _ "github.com/golang/protobuf/protoc-gen-go"
     _ "github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway"
     _ "github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2"
     _ "google.golang.org/grpc/cmd/protoc-gen-go-grpc"
+    _ "google.golang.org/protobuf/cmd/protoc-gen-go"
 )
 ```
 
@@ -73,7 +73,7 @@ Run `go mod tidy` to resolve the versions. Install by running
 $ go install \
     github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway \
     github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2 \
-    github.com/golang/protobuf/protoc-gen-go \
+    google.golang.org/protobuf/cmd/protoc-gen-go \
     google.golang.org/grpc/cmd/protoc-gen-go-grpc
 ```
 
--- a/README.md
+++ b/README.md
@@ -60,9 +60,10 @@ to track the versions of the following executable packages:
 package tools
 
 import (
+    _ "github.com/golang/protobuf/protoc-gen-go"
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -60,9 +60,10 @@ to track the versions of the following executable packages:
 package tools
 
 import (
+    _ "github.com/golang/protobuf/protoc-gen-go"
```
johanbrandhorst, line 63: If we make this documentation change, we should also change this to `google.golang.org/protobuf/cmd/protoc-gen-go`.

johanbrandhorst (changes requested):

johanbrandhorst:
Thanks a lot for this update, and for removing the duplicate installation instructions :joy:. I'm worried what other merge mistakes I made :grimacing:.

README.md
olivierlemasle: Right! I just copied the instructions from [gRPC quickstart](https://grpc.io/docs/languages/go/quickstart/) but it should be updated.
I'll update the PR.

olivierlemasle:
@johanbrandhorst I did not find any other duplicated paragraph in the README ;-)

johanbrandhorst (changes requested):
Having made this switch, we should replace any uses of `plugins=grpc` in the repo. I can see it's still there in a few docs files. Sorry I missed it on the first review.

olivierlemasle:
I've updated the PR; I hope instructions are now correct.
I did not test Cygwin instructions, as I do not have a Windows environment at hand, but they might need to be updated.

docs/_docs/cygwin.md
```diff
@@ -35,17 +35,18 @@ From an elevated cmd.exe prompt set the GOPATH variable in Windows and add the `
 
 Then `go get -u -v` the following packages:
 
-    go get -u -v github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway
-    go get -u -v github.com/grpc-ecosystem/grpc-gateway/protoc-gen-openapiv2
-    go get -u -v github.com/golang/protobuf/protoc-gen-go
+    go get -u -v github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway
+    go get -u -v github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2
+    go get -u -v google.golang.org/protobuf/cmd/protoc-gen-go
+    go get -u -v google.golang.org/grpc/cmd/protoc-gen-go-grpc
```
johanbrandhorst, line 41: We should drop the `-u -v` as well, this is not how things are installed anymore.

docs/_docs/grpcapiconfiguration.md
```diff
@@ -66,7 +66,10 @@ The following is equivalent to the basic [usage example](usage.html) but without
 3. Generate gRPC stub as before
 
     ```sh
-    protoc -I. --go_out=plugins=grpc,paths=source_relative:./gen/go/ your/service/v1/your_service.proto
+    protoc -I . \
+      --go_out=paths=source_relative:./gen/go/ \
+      --go-grpc_out=paths=source_relative:./gen/go/ \
+      your/service/v1/your_service.proto
```
johanbrandhorst, line 72: I know you didn't write this, but do you think we could rewrite this in the `--x-opt` format?
```suggestion
    protoc -I . \
      --go_out ./gen/go/ \
      --go_opt paths=source_relative \
      --go-grpc_out ./gen/go/ \
      --go-grpc_opt paths=source_relative
      your/service/v1/your_service.proto
```
I'd like to use this consistently across our documentation because I think it's much clearer. What do you think?

olivierlemasle: I also prefer the `--plugin-opt` format. I did not want to change the instructions, but if you also prefer it, I'll do it!

olivierlemasle:
@johanbrandhorst I've updated the PR with the _opt format, and fixed some other small issues (missing `v2` or some `swagger` left).

However, I did not update Cygwin instructions. I've installed Cygwin on a Windows VM to test the Cygwin instructions with an up-to-date Golang version (1.15 instead of 1.8). The majority of the instructions were now unnecessary (the errors mentioned with `go get` seem to be solved now with Go modules, and I did not have to adapt `protoc` commands). I even doubt if this page is still needed, but as I have no experience with Cygwin and even less with Go/protoc on cygwin/Windows, I do not want to handle this.

johanbrandhorst:
> @johanbrandhorst I've updated the PR with the _opt format, and fixed some other small issues (missing `v2` or some `swagger` left).
> 
> However, I did not update Cygwin instructions. I've installed Cygwin on a Windows VM to test the Cygwin instructions with an up-to-date Golang version (1.15 instead of 1.8). The majority of the instructions were now unnecessary (the errors mentioned with `go get` seem to be solved now with Go modules, and I did not have to adapt `protoc` commands). I even doubt if this page is still needed, but as I have no experience with Cygwin and even less with Go/protoc on cygwin/Windows, I do not want to handle this.

:joy: I can't believe you booted a windows VM to test this, that's some serious dedication. Yeah I'd be happy to remove the Cygwin instructions as well, they have surely bitrotted beyond usefulness. Thanks so much for your help!

docs/_docs/usegotemplates.md
```diff
@@ -28,9 +33,11 @@ Example of a bash script with the `use_go_templates` flag set to true:
 
 ```shell
 $ protoc -I. \
-    --go_out=plugins=grpc:. \
-    --grpc-gateway_out=logtostderr=true:. \
-    --openapiv2_out=logtostderr=true,use_go_templates=true:. \
+    --go_out=. --go-grpc_out=. \
```
johanbrandhorst, line 36: Nit; we can remove these equal signs
```suggestion
    --go_out . --go-grpc_out . \
```

.github/ISSUE_TEMPLATE.md
```diff
@@ -4,13 +4,13 @@
...
-  go get -u github.com/grpc-ecosystem/grpc-gateway
+  go get -u github.com/grpc-ecosystem/grpc-gateway/v2
   ```
 - [ ] Delete the `protoc-gen-grpc-gateway` and `protoc-gen-openapiv2` binary from your `PATH`,
   and reinstall the latest versions:
   ```sh
-  go get -u github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway
-  go get -u github.com/grpc-ecosystem/grpc-gateway/protoc-gen-openapiv2
+  go get -u github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway
+  go get -u github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2
```
johanbrandhorst, line 13: Could we remove the `-u` from these too? Sorry for being so picky, this is really great.

johanbrandhorst (commented):
Some more picky comments while we're at it, thank you so much for this cleanup!

.github/ISSUE_TEMPLATE.md
olivierlemasle: As this is the instructions to update binaries to the latest version, shouldn't we keep the `-u` flag?

> The -u flag instructs get to use the network to update the named packages
and their dependencies. By default, get uses the network to check out
missing packages but does not use it to look for updates to existing packages.

johanbrandhorst: Fair point, but I'd rather we did `go get github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway@latest` than use the older flag. I don't think it means what people think it means for modules.

olivierlemasle: Ok, done. I've used `@latest` only for `github.com/grpc-ecosystem/grpc-gateway/v2`, afterwards, it should not be necessary when re-installing the binaries.

And I've just seen that... this file is not used anymore! It it is a legacy GitHub issue template, and the new templates, in `.github/ISSUE_TEMPLATE` directory, are used instead. I guess both should be merged.

johanbrandhorst (approved):
Great, thanks so much for this Olivier :smile:

.github/ISSUE_TEMPLATE.md
johanbrandhorst: Doh! Feel free to follow up with another PR.

PR #1752, opened 2020-10-14
fix empty parameter fatal in protoc-gen-grpc-gateway 

danielhochman:
Fixes #1751

#### Have you read the [Contributing Guidelines](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/CONTRIBUTING.md)?

Yes

#### Brief description of what is fixed or changed

`Split()` on an empty parameter set for `protoc-gen-grpc-gateway` would cause a fatal since the for loop woulds still execute with an empty string.

Fixed with an early return and added some small tests.

```
F1014 11:23:06.564537  195394 main.go:115] Cannot set flag 
goroutine 1 [running]:
github.com/golang/glog.stacks(0xc000628000, 0xc0001ae680, 0x3c, 0x40)
        /home/danielhochman/go/pkg/mod/github.com/golang/glog@v0.0.0-20160126235308-23def4e6c14b/glog.go:769 +0xb8
github.com/golang/glog.(*loggingT).output(0xb8c4c0, 0xc000000003, 0xc0001fbf80, 0xb5bcf1, 0x7, 0x73, 0x0)
        /home/danielhochman/go/pkg/mod/github.com/golang/glog@v0.0.0-20160126235308-23def4e6c14b/glog.go:720 +0x372
github.com/golang/glog.(*loggingT).printf(0xb8c4c0, 0x3, 0x843bf2, 0x12, 0xc000295bc8, 0x1, 0x1)
        /home/danielhochman/go/pkg/mod/github.com/golang/glog@v0.0.0-20160126235308-23def4e6c14b/glog.go:655 +0x14b
github.com/golang/glog.Fatalf(...)
        /home/danielhochman/go/pkg/mod/github.com/golang/glog@v0.0.0-20160126235308-23def4e6c14b/glog.go:1148
main.parseFlags(0xc0001b4750, 0x0, 0x0)
        /home/danielhochman/go/src/github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway/main.go:115 +0x444
main.main.func1(0xc00039d5e0, 0x0, 0xc0001fa000)
        /home/danielhochman/go/src/github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway/main.go:68 +0x79
google.golang.org/protobuf/compiler/protogen.run(0xc0001977a0, 0x0, 0xc000295f30, 0xb8ce20, 0x7f326cb7ee98)
[...]

PR #1756, opened 2020-10-14
Refactor registry and generator to use protogen

adambabik:

This is continuation of work submitted and merged in https://github.com/grpc-ecosystem/grpc-gateway/pull/1668.

This PR switches fully to [google.golang.org/protobuf/compiler/protogen](https://pkg.go.dev/google.golang.org/protobuf/compiler/protogen) which is a new plugin to protoc to generate Go code.

The most important changes are:
* rely on [google.golang.org/protobuf/compiler/protogen](https://pkg.go.dev/google.golang.org/protobuf/compiler/protogen) to figure out Go package names and file names,
* removal of flags that are no longer support by [protoc-gen-go](https://pkg.go.dev/google.golang.org/protobuf/cmd/protoc-gen-go)
* moving handling of some flags like `-Mpath/to/another.proto=example.com/example` or `-paths` to [protoc-gen-go](https://pkg.go.dev/google.golang.org/protobuf/cmd/protoc-gen-go)

#### Other comments

The follow up work:
- [ ] test whether it solves the problem described in https://github.com/grpc-ecosystem/grpc-gateway/issues/1612
- [ ] disable warnings from protoc-gen-go in tests


protoc-gen-openapiv2/internal/genopenapi/template_test.go
```diff
@@ -1902,11 +1902,10 @@ func TestApplyTemplateRequestWithBodyQueryParameters(t *testing.T) {
...
 			},
 			{
-				Name:     proto.String("book"),
-				Label:    descriptorpb.FieldDescriptorProto_LABEL_REQUIRED.Enum(),
-				Type:     descriptorpb.FieldDescriptorProto_TYPE_STRING.Enum(),
-				TypeName: proto.String("Book"),
-				Number:   proto.Int32(2),
+				Name:   proto.String("book"),
+				Label:  descriptorpb.FieldDescriptorProto_LABEL_REQUIRED.Enum(),
+				Type:   descriptorpb.FieldDescriptorProto_TYPE_STRING.Enum(),
```
adambabik, line 1907: @johanbrandhorst I have a question about this line. The original implementation uses `FieldDescriptorProto_TYPE_STRING` and I kept this unchanged, however, it also provides `TypeName` equal to `Book`. The new protogen-based implementation fails with this configuration which makes sense because if we want `TypeName` equal to `Book` it clearly is a message type.

I am not sure what was the original intention because either `TypeName` is not needed or it is needed but then the `Type` should be `FieldDescriptorProto_TYPE_MESSAGE`. If the latter is the original intention, the code fails in the same way in this PR and on the master branch, hence, I assumed the former is the right one.

johanbrandhorst: That sounds right to me, frankly I haven't looked at this part of the code for a while, so I trust you have it right.

johanbrandhorst:
Note that #1752 and #1757 may require you to rebase this PR a few times, I hope that's okay. Thanks for all your work Adam, it's greatly appreciated.

adambabik:
That's cool, this PR is taking long time so it's expected. I think I will have it ready today/tomorrow. I just fixed the last test, yay! 

No problem at all, it's a great open source tool and I am glad I can give something to the community as well :)

adambabik:
@johanbrandhorst I can't reproduce this failed bazel build. I run both `docker run` commands from `CONTRIBUTING.md` and they worked fine. Is it possible it's a flake?

johanbrandhorst:
Is it possible that `FieldMaskFromRequestBody` returns the old type field mask? It looks like a build error, does it not reproduce with `go test ./...`?

adambabik:
Oh yeah, that might be it, thanks! I refactored code to `protobuf/types/known/fieldmaskpb` but it looks like only `*_test.go` files were changed.

adambabik:
I needed to revert usage of `google.golang.org/protobuf/types/known/fieldmaskpb`. It looks like there is some problem with package versions when building with bazel. I can't reproduce it locally unfortunately.

In generated `.pb.go`, it looks like this:

```
wrappers "github.com/golang/protobuf/ptypes/wrappers"
field_mask "google.golang.org/genproto/protobuf/field_mask"
```

`github.com/golang/protobuf/ptypes/wrappers` has aliases to `google.golang.org/protobuf/types/known/wrapperspb`. Similarly `google.golang.org/genproto/protobuf/field_mask` has aliases to `google.golang.org/protobuf/types/known/fieldmaskpb` but in bazel build in the CI, there was an error that `google.golang.org/genproto/protobuf/field_mask` needs to be update to a version released after 26 May, 2020 which actually is a case currently.

johanbrandhorst:
Yeah, I had a similar experience, I think it's going to take until rules_go has switched to a reasonably recent version of golang/protobuf. Just leave it for now, I will be making that change when we can.

adambabik:
Got it. In my opinion, it's good for a thorough review. There are still two todo items in the description but I think whatever will result from them, it will be a separate issue.

johanbrandhorst:
Could you rebase on master again please? I just merged another PR that affects these files.

johanbrandhorst (commented):
Really nice to see all the logic we can move out and consume from protogen, nice work! I'm curious about some of the flag removals and the test removals, should we worry that this will break any users? The removal of the `module` flag for example, does that now get consumed by protogen?

adambabik:
I removed four flags:
* `paths`, `import_path`, and `module` are all [handled by protogen](https://github.com/protocolbuffers/protobuf-go/blob/42cc4c592f87f9c4fa8c52df6a1c0abd6c7fb571/compiler/protogen/protogen.go#L178-L190),
* `import_prefix` [is deprecated in protobuf-go](https://github.com/protocolbuffers/protobuf-go/blob/95ba8ac24031e0290a9916e6e6c94030a1e4b0f1/cmd/protoc-gen-go/main.go#L42) and I followed the same logic.

The problem might be with different interpretation of these flags. I tried to figure this out and it looked the same but that might not be 100% true.

Regarding the removed tests, I removed everything related to methods that were also removed. The main important thing that got removed was generating the output path (for example `TestGenerateOutputPath`) which now is a responsibility of protogen. Some other tests were testing removed flags like `TestLoadFileWithPrefix` which tested `import_prefix` or `TestLoadSetInputPath` and `TestLoadGoPackageInputPath ` which tested `import_path`.

johanbrandhorst:
Ugh, we really should've got that `import_prefix` removal in before v2.0.0 but I guess the protoc-gen-go precedent might be good enough, and the workaround is to stay on v1.

PR #1757, opened 2020-10-14
fix import alias issue with protoc-gen-grpc-gateway

danielhochman:

Yes

Starting in v2, if any [base imports](https://github.com/grpc-ecosystem/grpc-gateway/blob/1aab03caf0bfadbf99a107ffdd86467ac9db7caa/protoc-gen-grpc-gateway/internal/gengateway/generator.go#L45-L55) collide with imported packages in the proto, the base import will be aliased and `pb.gw.go` will not compile. Compilation fails because the rest of the template does not account for the import alias.

Proto file import (whose go_package is [google.golang.org/genproto/googleapis/rpc/status](https://github.com/googleapis/googleapis/blob/215c12ade72d9d9616457d9b8b2f8a37f38e79f3/google/rpc/status.proto#L22)):
```
...
import "google/rpc/status.proto";

message BatchResponse {
  repeated google.rpc.Status statuses = 1;
}
```

Generated `batch.pb.gw.go`:
```
import (
	...
	status_0 "google.golang.org/grpc/status"
	...
)
```

Build fails.
```
$ go build
api/v1/batch.pb.gw.go:22:2: imported and not used: "google.golang.org/grpc/status" as status_0
api/v1/batch.pb.gw.go:29:7: undefined: status
```

The regression occurred as part of the move to v2 because the files are loaded into the registry in v2 before the generator set up base imports in the registry.

#### Other comments

I'm sure there are many possible fixes here. However, I went ahead and just had the registry returned by the generator New instead of being an input parameter to enforce the call order previously used in v1. Happy to try another approach if this is building on a brittle foundation.

(note: working on getting corp CLA signed)


johanbrandhorst (commented):
This fix looks fine, and I think it will change in the long term with the work @adambabik is doing in https://github.com/grpc-ecosystem/grpc-gateway/pull/1756 (:heart:), but I was wondering if we could add a test for this, or a test file that when generated without this fix exhibits the failure?

Thanks again for your contributions!

johanbrandhorst:
Looks like we'll need to rebase these changes on master since #1752 was merged.

protoc-gen-grpc-gateway/main.go
```diff
@@ -65,8 +63,11 @@ func main() {
 		ParamFunc: flag.CommandLine.Set,
 	}.Run(func(plugin *protogen.Plugin) error {
 		// FIXME: still needed to parse request parameter and apply flags manually, see the comment above.
-		parseFlags(reg, plugin.Request.GetParameter())
-		if err := applyFlags(reg); err != nil {
+		pkgMap := parseFlags(plugin.Request.GetParameter())
+
+		gen, reg := gengateway.New(*useRequestContext, *registerFuncSuffix, *pathType, *allowPatchFeature, *standalone)
```
johanbrandhorst, line 68: Could we keep creating the `reg` outside of `gengateway` and just keep this new order of operations? E.g. still pass in the registry and mutate it, but do it before running `applyFlags`? It feels a bit dirty for `gengateway.New` to return the registry. Looks to me like this should be solved by just moving it above `parseFlags` and `applyFlags`. Am I missing something?

danielhochman: yep, no problem. i was originally trying to enforce structurally in the code that it couldn't be placed out of order again, but now that we have the generation path exercised with ABE the same issue would be caught by compilation during testing.

johanbrandhorst (approved):

johanbrandhorst:
Hi Daniel, I just merged https://github.com/grpc-ecosystem/grpc-gateway/pull/1756 so this will need a rebase, sorry for the inconvenience.

johanbrandhorst:
Thanks for your contribution!

Issue #669, opened 2018-06-12
protoc-gen-swagger: should well known types be nullable


zheng1:
- OpenAPI 3.0 support `nullable` [[link](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#schemaNullable)] field to define schema can be `null` value.
- `go-swagger` support vendor extensions for `x-nullable` [[link](https://github.com/go-swagger/go-swagger/blob/master/docs/generate/spec.md)]

Should `protoc-gen-swagger` convert well know types to nullable type?

tmc:
@ivucica do you want to chime in here?

ivucica:
@zheng1, I must admit I am confused at what's being proposed. Which well-known types would you make nullable and under what _specific_ criteria?

Closest could be protobuf's built-in `optional`, but that's not quite the same semantically.

Finally, I am not aware of PRs that make `protoc-gen-swagger` generate OpenAPI v3. I believe it currently generates OpenAPI v2, so until that happens, the point is moot. I would not mind the proposal being better documented for when we get a v3 generator, of course.

johanbrandhorst:
I think actually this `x-nullable` extension should probably be enabled on all non-scalar types. I assume message types are already nullable, so maybe the safest thing is to have a whitelist in the generator that inserts this property. Is there anything I can do to help you get a PR with this in @marcusljx?

marcusljx:
> I think actually this `x-nullable` extension should probably be enabled on all non-scalar types. I assume message types are already nullable, so maybe the safest thing is to have a whitelist in the generator that inserts this property. Is there anything I can do to help you get a PR with this in @marcusljx?

I don't mind trying my hand at this @johanbrandhorst, but I'm afraid I'm not sure where to begin.

johanbrandhorst:
I think @ivucica should be able to give some pointers.

ivucica:
Not sure when I'll be getting around to studying the problem space. That is, here's some necessary things I don't know off the top of my head:

- what the `x-nullable` extension actually is, and how to make use of it
- where it needs to be written in the output file
- how to adapt `template.go` to do this
- how to specifically check if a field -- whatever its type -- is considered `required`

Each of these items is relatively short, I'd just need to spend the time to figure them out. Maybe just enumerating them helps you?

However, note: OP says `x-nullable` is an OpenAPI v3 feature. As I stated in my June 26 comment, I think we are only generating OpenAPI v2 at this time.

[...]

PR #1806, opened 2020-11-10
add support for field_behavior options on swagger plugin

ewhauser:

There is a discussion in #669 about adding support for `x-nullable` using `FieldBehavior.OPTIONAL`. While I haven't implemented that behavior here yet, it should be relatively trivial to add.

This adds support for using `google.api.field_behavior` options to specify whether fields are required or output only. The syntax is a little nicer than `grpc.gateway.protoc_gen_swagger.options.openapiv2_schema` because you can define the values on the individual fields as opposed to the message level.


protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -1836,6 +1856,7 @@ func getFileOpenAPIOption(reg *descriptor.Registry, file *descriptor.File) (*ope
 	return opts, nil
 }
 
+
```
johanbrandhorst, line 1859: Nit: remove this line please

johanbrandhorst (changes requested):
This is really interesting, thanks for the PR! I'd like to understand the consequences of this change, as it may change the expectations of the users. For example:

* Should we add any handling to `protoc-gen-grpc-gateway` of these annotations? What would that mean?
* Adding this file as a dependency of the library runs the risk of annoying all current users who do not have `field_behavior.proto` vendored (as it wasn't a reqiurement), is this going to break users that don't vendor this file? From a cursory reading I don't think it should since we're just checking if the extensions are set.

Also, could we update one of the fields in `examples/internal/proto/a_bit_of_everything.proto` to add some of these descriptions to show users how they're used?

Thanks!

ewhauser:
> This is really interesting, thanks for the PR! I'd like to understand the consequences of this change, as it may change the expectations of the users. For example:
> 
> * Should we add any handling to `protoc-gen-grpc-gateway` of these annotations? What would that mean?

You could certainly add runtime checking for some of the options - `REQUIRED`, `INPUT_ONLY`, `OUTPUT_ONLY` - but I don't believe gateway does anything with the existing `required` or `readonly` fields at runtime today? It would have to be an opt in feature because it could potentially be a breaking change for people who documented that behavior but didn't enforce it.

I think it's a reasonable discussion but probably would be a separate issue. 

> * Adding this file as a dependency of the library runs the risk of annoying all current users who do not have `field_behavior.proto` vendored (as it wasn't a reqiurement), is this going to break users that don't vendor this file? From a cursory reading I don't think it should since we're just checking if the extensions are set.

Confirmed that there is no compile time dependency on `field_behavior.proto`. 

> 
> Also, could we update one of the fields in `examples/internal/proto/a_bit_of_everything.proto` to add some of these descriptions to show users how they're used?
> 

Will do.

> Thanks!

examples/internal/clients/abe/api/swagger.yaml
```diff
@@ -373,6 +373,19 @@ paths:
         format: "int64"
         x-exportParamName: "Int64OverrideType"
         x-optionalDataType: "Int64"
+      - name: "requiredStringViaFieldBehaviorAnnotation"
+        in: "query"
+        description: "mark a field as required in Open API definition."
+        required: true
```
ewhauser, line 379: Noting that this field is marked as `required` as expected.

examples/internal/clients/abe/api/swagger.yaml
```diff
@@ -2464,6 +2516,13 @@ definitions:
       int64OverrideType:
         type: "integer"
         format: "int64"
+      requiredStringViaFieldBehaviorAnnotation:
+        type: "string"
+        title: "mark a field as required in Open API definition"
+      outputOnlyStringViaFieldBehaviorAnnotation:
+        type: "string"
+        title: "mark a field as readonly in Open API definition"
+        readOnly: true
```
ewhauser, line 2525: Noting that this field is marked as `readonly` as expected.

examples/internal/proto/examplepb/a_bit_of_everything.swagger.json
```diff
@@ -2998,6 +3054,18 @@
         "int64OverrideType": {
           "type": "integer",
           "format": "int64"
+        },
+        "requiredStringViaFieldBehaviorAnnotation": {
+          "type": "string",
+          "title": "mark a field as required in Open API definition",
+          "required": [
+            "required_string_via_field_behavior_annotation"
```
ewhauser, line 3062: I believe this is a bug in the existing implementation as the other fields that are marked as required as also using snakecase for the field names here.

ewhauser: This behavior is the same as today (see the `float_value` field) - but I don't believe it's actually correct. The OpenAPI spec is a little hard to read, but I think that what should happen here is that the `required` parameter should not be set on the field and instead the name of the field should go in the `required` array at the object level.

johanbrandhorst: Hm, we have generators that run against our generated files as an oracle for our output - with your suggested change do they respond correctly? Would you have time to fix this at the same time?

ewhauser: If you take the OpenAPI spec that we are [generating](https://raw.githubusercontent.com/grpc-ecosystem/grpc-gateway/1f17ac3600b1bc59ecdc9e34a176ab9dd050deb6/examples/internal/proto/examplepb/a_bit_of_everything.swagger.json) and compare it against the [OpenAPI 2.0 JSON Schema](https://github.com/OAI/OpenAPI-Specification/blob/master/schemas/v2.0/schema.json), it [validates](https://www.jsonschemavalidator.net/) just fine. However, JSON Schema validation isn't strict - so the schema validation won't fail when the optional `required` field that is being added to `definitions` property.

What's not entirely clear to me in reading the specification is which of the fields for a [Schema Object](https://swagger.io/specification/v2/#schemaObject) are allowed at the top level vs. the field level. The specification says that these parameters derive from the [JSON Schema Validation](https://tools.ietf.org/html/draft-fge-json-schema-validation-00) spec which [clearly specifies](https://tools.ietf.org/html/draft-fge-json-schema-validation-00#section-5.4.3) that `required` should be an array at an object level. Given that it does not specifically state that `required` can be used with numerics or strings, I think that answer is that `required` should be omitted there (but again doesn't hurt anything since it still passes validation).

I think there are 2 things that need to be fixed here:

1) When using camel case, the `required` field names need to be modified to account for it
2) My change is only setting `required` at the field level but is not adding it to the `required` array for the object which needs addressed

And this one is optional, but I think we can just ignore it:

3) The required property should not be set on the field definition for objects; as the required array (http://swagger.io/specification/#schemaObject) at the object level is actually responsible for setting this

ewhauser: This is now addressed.

examples/internal/proto/examplepb/a_bit_of_everything.swagger.json
```diff
@@ -3008,8 +3076,10 @@
       },
       "required": [
         "uuid",
-        "int64_value",
-        "double_value"
+        "int64Value",
+        "doubleValue",
+        "floatValue",
```
ewhauser, line 3081: Note that the casing has been fixed for `int64_value` and `double_value`. `float_value` is also added to the array now where previously it was not.

examples/internal/proto/examplepb/unannotated_echo_service.swagger.json
```diff
@@ -449,7 +449,8 @@
         "url": "https://github.com/grpc-ecosystem/grpc-gateway"
       },
       "required": [
-        "id"
+        "id",
+        "num"
```
ewhauser, line 453: This was a bug where num was marked as `required` but didn't show up in the object level array

ewhauser:
@johanbrandhorst - I think we are all set here.

In relation to the original runtime behavior question, I think there are a couple options for anyone who wants automatic runtime validation against the Swagger spec:

1) Integrate with `go-openapi` which has an existing validator for OpenAPI schemas: https://godoc.org/github.com/go-openapi/validate#hdr-Validating_a_schema. This would have to be done at the HTTP level since it's validates against JSON which wouldn't be ideal for performance reasons since you'd be parsing the request twice.
2) Code generate validators against the Swagger spec (a la https://github.com/mwitkow/go-proto-validators or https://github.com/envoyproxy/protoc-gen-validate) and use [validator middleware](https://github.com/grpc-ecosystem/go-grpc-middleware/blob/master/validator/validator.go) to validate at runtime.

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -2074,3 +2135,13 @@ func getReservedJSONName(fieldName string, messageNameToFieldsToJSONName map[str
...
+func find(a []string, x string) int {
+	// This is a linear search but we are dealing with a small number of fields
+	for i, n := range a {
+		if x == n {
+			return i
+		}
+	}
+	return -1
+}
```
johanbrandhorst, line 2147: Nit: newline at the end of the file please.

johanbrandhorst (changes requested):
Hey, one small nit, but I have a final request - now that we've added this behaviour, lets write a little docs page about how to use it! Our docs pages are in `docs/docs/`. We could either add it to "customizing your gateway" (though that page is getting very big) or consider adding a new one for customizing the swagger output (we should probably have a page for this anyway!).

Thanks for your patience!

protoc-gen-openapiv2/internal/genopenapi/template.go
ewhauser: No problem.

Do you format with `go fmt`? If so, might be good to have CI run a check to see if all files are formatted property so it's easier for you the reviewer and for contributors.

ewhauser:
I'll update the docs sometime in the next few days.

protoc-gen-openapiv2/internal/genopenapi/template.go
johanbrandhorst: You're absolutely right, we should do that. I haven't messed around with the CI jobs for a while but I might just set up a new golangci-lint job.

johanbrandhorst: Done, you may want to rebase on master in case there are other formatting errors: https://github.com/grpc-ecosystem/grpc-gateway/pull/1817

ewhauser:
@johanbrandhorst I've opened another PR to cover the docs piece since the changes were a bit beyond the scope of just field behavior support. Figured we could get this merged and iterate on any docs changes there.

Thanks.

johanbrandhorst (approved):
Perfect, SGTM!

johanbrandhorst:
Thanks a ton for your contribution!

Issue #1816, opened 2020-11-13
`generate_unbound_methods` option and FieldMask fields don't work together.


zoido:
## 🐛 Bug Report

There seems to be inconsistency between generated OpenAPI specs and jsonproto Unmarshaler specs.

`wktSchemas` do says that:
> The schemas must match with the behavior of the JSON unmarshaler in https://github.com/protocolbuffers/protobuf-go/blob/v1.25.0/encoding/protojson/well_known_types.go

Docs in the Umarshaler specification say that:
> The JSON representation for a FieldMask is a JSON string where paths are separated by a comma.

When the `generate_unbound_methods=true` option is used, generated OpenAPI spec has fieldmask fields as:
```json
{
	"type": "array",
	"items": {
		"type": "string"
	}
}
```

This causes when failures when requests that adhere to the generated specification arrive to the server (see the reproduction step below).
When the generated client code is type-checked, there's usually no way to override the content of the request body.

I am willing to help with this issue. I am not sure at this point what else the change that would change the type in the OpenAPI spec affect.

### References

https://github.com/grpc-ecosystem/grpc-gateway/blob/26104fd424bf45f44d498149b071ef657d604f2b/protoc-gen-openapiv2/internal/genopenapi/template.go#L27-L36

https://github.com/protocolbuffers/protobuf-go/blob/3f7a61f89bb6813f89d981d1870ed68da0b3c3f1/encoding/protojson/well_known_types.go#L833-L836

## To Reproduce

1. Have a minimal proto:
	```proto
	// api/v1/example.proto
	syntax = "proto3";
	
	package zoido.example;
	option go_package = "api/v1;example";
	
	import "google/protobuf/field_mask.proto";
	
	service Example {
	  rpc Update(UpdateRequest) returns(UpdateResponse) {}
	}
	
	message UpdateRequest {
	  string name = 1;
	  google.protobuf.FieldMask update_mask = 2;
	}
	message UpdateResponse {}
	
	```

1. Generate code:
	```sh
	protoc --go_out=:. --go-grpc_out=:. \
    	--grpc-gateway_out=:. --grpc-gateway_opt=generate_unbound_methods=true \
	    --openapiv2_out=:. --openapiv2_opt=generate_unbound_methods=true \
	    api/v1/*.proto
	
	```
1. Inspect the swagger definition:
	relevant part:
	```json
	{
      "definitions": {
        "exampleUpdateRequest": {
          "type": "object",
          "properties": {
[...]

PR #1820, opened 2020-11-13
Change FieldMask OpenAPI type to string.

zoido:

RESOLVES https://github.com/grpc-ecosystem/grpc-gateway/issues/1816

Yes

 - sets type for `FieldMask` to string
 - adjust test expectations accordingly
 - update examples generated code accordingly
 - updates `patch.md` doc to contains the corect format for the FieldMask in the example
	- also changes the case to smallCamel that is the what the Marshaler accepts

### DIFF
--- a/docs/docs/mapping/patch.md
+++ b/docs/docs/mapping/patch.md
@@ -94,7 +94,7 @@ that is what we specify in the field_mask.
 
 ```shell
 $ curl \
-	--data '{"abe":{"single_nested":{"amount":457},"string_value":"some value that won't get updated because not in the field mask"},"update_mask":{"paths":["single_nested"]}}' \
+	--data '{"abe":{"single_nested":{"amount":457},"string_value":"some value that will not get updated because not in the field mask"},"update_mask":"singleNested"}}' \
 	-X PATCH \
 	http://address:port/v2a/example/a_bit_of_everything/1
 ```
--- a/docs/docs/mapping/patch.md
+++ b/docs/docs/mapping/patch.md
@@ -82,7 +82,7 @@ else in our resource the same.
 
 ```shell
 $ curl \
-	--data '{"string_value": "strprefix/foo"}' \
+	--data '{"stringValue": "strprefix/foo"}' \
 	-X PATCH \
 	http://address:port/v2/example/a_bit_of_everything/1
 ```
@@ -94,7 +94,7 @@ that is what we specify in the field_mask.
 
 ```shell
 $ curl \
-	--data '{"abe":{"single_nested":{"amount":457},"string_value":"some value that will not get updated because not in the field mask"},"update_mask":"singleNested"}}' \
+	--data '{"abe":{"singleNested":{"amount":457},"stringValue":"some value that will not get updated because not in the field mask"},"updateMask":"singleNested"}}' \
 	-X PATCH \
 	http://address:port/v2a/example/a_bit_of_everything/1
 ```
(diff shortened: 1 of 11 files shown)


johanbrandhorst (commented):
Hi! Thanks for your contribution, could you confirm that this works when using it with the generated swagger file? Thanks!

johanbrandhorst:
Related to https://github.com/grpc-ecosystem/grpc-gateway/issues/1816

vtolstov:
Very needed feature

сб, 14 нояб. 2020 г., 14:41 Johan Brandhorst-Satzkorn <
notifications@github.com>:

johanbrandhorst (approved):
Hi! This looks fine, could you just confirm that this fixes your issue before I merge it?

zoido:
@johanbrandhorst Hi. It fixes it for us. Would you recommend the place where to include some kind of test to make sure it's covered?

I have also spotted that `patch.md` doc contains the example that seems to use even older format. I'd like also adjust test that before I mark this as ready for review.

johanbrandhorst:
I don't think we ever actually encourage the user to specify a field mask, we _automatically_ create the field mask from input JSON when possible, so the input format has never been tested. If you want to add a test for this explicit specification, it'd make sense to add it to the integration tests in https://github.com/grpc-ecosystem/grpc-gateway/blob/master/examples/internal/integration/integration_test.go#L801. Looks like we rely on the marshaler, maybe a new test which does it explicitly would make sense.

zoido:
@johanbrandhorst Checked and updated the `patch.md`. PTAL.

docs/docs/mapping/patch.md
```diff
@@ -94,7 +94,7 @@ that is what we specify in the field_mask.
 
 ```shell
 $ curl \
-	--data '{"abe":{"single_nested":{"amount":457},"string_value":"some value that won't get updated because not in the field mask"},"update_mask":{"paths":["single_nested"]}}' \
+	--data '{"abe":{"single_nested":{"amount":457},"string_value":"some value that will not get updated because not in the field mask"},"update_mask":"singleNested"}}' \
```
johanbrandhorst, line 97: Shouldn't this be `single_nested`? I generally use the proto field name casing to identify fields in the mask.
```suggestion
	--data '{"abe":{"single_nested":{"amount":457},"string_value":"some value that will not get updated because not in the field mask"},"update_mask":"single_nested"}}' \
```

johanbrandhorst (commented):
Small question on the new docs.

docs/docs/mapping/patch.md
zoido: This is unfortunate but ... expects paths to be in the `smallCamelCase`.

> …
> Fields name in each path are converted to/from lower-camel naming conventions.
> …

https://github.com/protocolbuffers/protobuf-go/blob/0fd4f3a506894c05a9cad2d52c521a8ce953b49d/encoding/protojson/well_known_types.go#L837-L840


Using `snake_case` leads to error:
```txt
{"code":3, "message":"proto: (line 1:136): google.protobuf.FieldMask.paths contains invalid path: \"single_nested\"", "details":[]}
```

johanbrandhorst: TIL! That makes sense on some level - the official protobuf json convention is to use lower camelCase. We should probably update the rest of the fields to use lower camel case instead. What do you think?

zoido: Agree. Changed.

johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution!

PR #1866, opened 2020-12-08
protoc-gen-openapiv2: Add openapiv2_tag support

### LINKED ISSUE #690: Support openapiv2_tag option in swagger generator
Hey guys, 

I'm trying to use grpc-gateway to create a swagger.json directly from a .proto file
I was wondering if the tag option
option (grpc.gateway.protoc_gen_swagger.options.openapiv2_tag)
was working?

I pasted this tag option inside a service and it should change the description for the service "tag" if I'm not mistaken. When I do this and generate a swagger.json file, the services do not have descriptions appended to them. (Snippet Below)

service APIService {

  option (grpc.gateway.protoc_gen_swagger.options.openapiv2_tag) = {
	description: "Some Description";
  };
 rpc someService(Request) returns (Response)
}

In addition, for passing parameters, is there any kind of header support? Reading through the documentation, I've figured out that you have an optional body field and an automatic query parameter generation from the requests you pass into the rpc. However, is there any header parameter support that isn't in the documentation?

Thank you for the help.

rcgoodfellow:

Fixes #690

Added a `renderServiceTag` method to pick up `openapiv2_tag` options. The change is pretty surgical aside from the fact that the return type for `renderServices` was changed to propagate discovered tags back to the caller. So a small pile of return statements had to be updated.

#### Other comments

Thanks to the grpc-gateway team, it's a really great tool.


examples/internal/proto/examplepb/echo_service.swagger.json
```diff
@@ -5,6 +5,7 @@
     "description": "Echo Service API consists of a single service which returns\na message.",
     "version": "version not set"
   },
+  "tags": null,
```
johanbrandhorst, line 8: Could we omit this, rather than set it to null, in these instances?

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -782,15 +782,46 @@ func isResourceName(prefix string) bool {
 	return field == "parent" || field == "name"
 }
 
-func renderServices(services []*descriptor.Service, paths openapiPathsObject, reg *descriptor.Registry, requestResponseRefs, customRefs refMap, msgs []*descriptor.Message) error {
+func renderServiceTag(svc *descriptor.Service) []openapiTagObject {
+
```
johanbrandhorst, line 786: Nit: remove this newline
```suggestion
```

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -782,15 +782,46 @@ func isResourceName(prefix string) bool {
 	return field == "parent" || field == "name"
 }
 
-func renderServices(services []*descriptor.Service, paths openapiPathsObject, reg *descriptor.Registry, requestResponseRefs, customRefs refMap, msgs []*descriptor.Message) error {
+func renderServiceTag(svc *descriptor.Service) []openapiTagObject {
```
johanbrandhorst, line 785: Is there a good reason this returns a slice rather than just a `openapiTagObject`?
```suggestion
func renderServiceTag(svc *descriptor.Service) openapiTagObject {
```

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -1196,9 +1227,11 @@ func applyTemplate(p param) (*openapiSwaggerObject, error) {
 	// and create entries for all of them.
 	// Also adds custom user specified references to second map.
 	requestResponseRefs, customRefs := refMap{}, refMap{}
-	if err := renderServices(p.Services, s.Paths, p.reg, requestResponseRefs, customRefs, p.Messages); err != nil {
+	tags, err := renderServices(p.Services, s.Paths, p.reg, requestResponseRefs, customRefs, p.Messages)
+	if err != nil {
 		panic(err)
 	}
+	s.Tags = append(s.Tags, tags...)
```
johanbrandhorst, line 1234: I think it'd be neater to just do the tag calculation in a separate loop here than add it to `renderServices`. Yes, it would make this O(2n) rather than O(n) but it's hardly significant, and it means we can avoid changing `renderServices`. What do you think?

protoc-gen-openapiv2/internal/genopenapi/types.go
```diff
@@ -54,6 +61,7 @@ type extension struct {
 type openapiSwaggerObject struct {
 	Swagger             string                              `json:"swagger"`
 	Info                openapiInfoObject                   `json:"info"`
+	Tags                []openapiTagObject                  `json:"tags"`
```
johanbrandhorst, line 64: Making this `omitempty` should omit empty objects.

johanbrandhorst (changes requested):
Thanks for taking this on! It doesn't look too bad. Some minor cosmetic suggestions and a question around the use of `renderServices` is all.

examples/internal/proto/examplepb/echo_service.swagger.json
rcgoodfellow: Yup. Fixed via `omitempty` in struct tags as suggested in other comments.

protoc-gen-openapiv2/internal/genopenapi/template.go
rcgoodfellow: Done

protoc-gen-openapiv2/internal/genopenapi/template.go
rcgoodfellow: Obviated by factoring service tag rendering into it's own loop as suggested in other comments.

protoc-gen-openapiv2/internal/genopenapi/template.go
rcgoodfellow: Yes, this is much cleaner and implicitly solves some of the other issues. Done.

protoc-gen-openapiv2/internal/genopenapi/types.go
rcgoodfellow: Done. A few other `omitempty` tags also added where applicable.

rcgoodfellow:
Failed CI task appears to be an issue with the fuzzit service

> 2020/12/08 17:08:33 Get https://app.fuzzit.dev/getStorageLinkV3?path=orgs%2Fgrpc-gateway%2Ftargets%2Fparse-http-rule%2Fcorpus.tar.gz&api_key=&action=read: x509: certificate is valid for c.sni.fastly.net, not app.fuzzit.dev

rcgoodfellow:
Thanks for the feedback and quick response @johanbrandhorst. I believe I've addressed the comments.

johanbrandhorst (approved):
LGTM

johanbrandhorst:
Thanks for your contribution!

Issue #1909, opened 2021-01-19
Adding custom routes to the mux. Cannot add root route. ("/")


AP0rPi7uQ9:
Hello, I have feedback on slack. I tested in my app and I think this is a bug

> https://grpc-ecosystem.github.io/grpc-gateway/docs/operations/inject_router/ 

Follow this example, I want to add some routes of my own.

> err := mux.HandlePath("GET", "/abc", xxxx)
> fmt.Println(err)

This is normal.  err == nil, I can successfully access through the browser www.my-url.com/abc

But!!!!!!
> err := mux.HandlePath("GET", "/", xxxx)
> fmt.Println(err)

This is abnormal.   I cannot access through www.my-url.com/, and the `err` return
```
parsing path pattern: segment neither wildcards, literal or variable: expected "{" but got "\x00": /
```

AP0rPi7uQ9:
This is an example. Edit by https://grpc-ecosystem.github.io/grpc-gateway/docs/operations/inject_router/.

This example will panic. and show error message.


main.go:

```go
package main

import (
	"context"
	"net/http"
	"testgw/helloworld"

	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
)

func main() {
	ctx := context.TODO()
	mux := runtime.NewServeMux()
	// Register generated routes to mux
	err := helloworld.RegisterGreeterHandlerServer(ctx, mux, &GreeterServer{})
	if err != nil {
		panic(err)
	}
	// Register custom route for  GET /hello/{name}
	// This route is OK!
	err = mux.HandlePath("GET", "/hello/{name}", func(w http.ResponseWriter, r *http.Request, pathParams map[string]string) {
		w.Write([]byte("hello " + pathParams["name"]))
	})
	if err != nil {
		panic(err)
	}
	// This route is OK!
	err = mux.HandlePath("GET", "/test/{name}", func(w http.ResponseWriter, r *http.Request, pathParams map[string]string) {
		w.Write([]byte("test with name " + pathParams["name"]))
	})
	if err != nil {
		panic(err)
	}
	// This route is OK!
	err = mux.HandlePath("GET", "/test", func(w http.ResponseWriter, r *http.Request, pathParams map[string]string) {
		w.Write([]byte("this is test " + pathParams["name"]))
	})
	if err != nil {
		panic(err)
	}
	// This route will be PANIC
[...]

PR #1916, opened 2021-01-21
fix "/" route

AP0rPi7uQ9:
This is a dumb method, but try not to modify the underlying code.

### DIFF
--- a/runtime/mux.go
+++ b/runtime/mux.go
@@ -180,6 +180,10 @@ func (s *ServeMux) Handle(meth string, pat Pattern, h HandlerFunc) {
 // HandlePath allows users to configure custom path handlers.
 // refer: https://grpc-ecosystem.github.io/grpc-gateway/docs/inject_router.html
 func (s *ServeMux) HandlePath(meth string, pathPattern string, h HandlerFunc) error {
+	if pathPattern == "/" {
+		index, _ := NewPattern(1, []int{2,0}, []string{""}, "")
+		s.Handle(meth, index, h)
+	}
 	compiler, err := httprule.Parse(pathPattern)
 	if err != nil {
 		return fmt.Errorf("parsing path pattern: %w", err)
--- a/runtime/mux.go
+++ b/runtime/mux.go
@@ -183,6 +183,7 @@ func (s *ServeMux) HandlePath(meth string, pathPattern string, h HandlerFunc) er
 	if pathPattern == "/" {
 		index, _ := NewPattern(1, []int{2,0}, []string{""}, "")
 		s.Handle(meth, index, h)
+		return nil
 	}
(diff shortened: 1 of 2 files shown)


achew22:
Thank you so much for sending this in, really excited to get this this fixed. Do you think you could add a test that demonstrates that you have solved the problem?

AP0rPi7uQ9:
The test is in this link
https://github.com/grpc-ecosystem/grpc-gateway/issues/1909

runtime/mux.go
```diff
@@ -180,6 +180,10 @@ func (s *ServeMux) Handle(meth string, pat Pattern, h HandlerFunc) {
 // HandlePath allows users to configure custom path handlers.
 // refer: https://grpc-ecosystem.github.io/grpc-gateway/docs/inject_router.html
 func (s *ServeMux) HandlePath(meth string, pathPattern string, h HandlerFunc) error {
+	if pathPattern == "/" {
+		index, _ := NewPattern(1, []int{2,0}, []string{""}, "")
+		s.Handle(meth, index, h)
```
johanbrandhorst, line 185: I think this is missing a return below this line

johanbrandhorst (changes requested):
Nice! Agree with Andrew, could you please add a test to the repository? Something like https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/mux_test.go#L425.

AP0rPi7uQ9:
Okay, let me learn about adding tests and reply later.

runtime/mux.go
AP0rPi7uQ9: yes, forgot it.... return nil. i will fix it

AP0rPi7uQ9:
Is this test correct?

And I don't know why this item keeps failing.

> golangci-lint / lint (pull_request) Failing after 54s — lint

johanbrandhorst:
> Is this test correct?
> 
> And I don't know why this item always keeps failing.
> 
> > golangci-lint / lint (pull_request) Failing after 54s — lint

Yes, it's saying that the file changes you made aren't formatted correctly. You need to install and run `goimports` on the source code: https://pkg.go.dev/golang.org/x/tools/cmd/goimports

johanbrandhorst:
Also, the CLA bot error is because you used a different account to make the test commit. You may want to make sure you're using the same github email in your commit as you have on your account.

AP0rPi7uQ9:
> Also, the CLA bot error is because you used a different account to make the test commit. You may want to make sure you're using the same github email in your commit as you have on your account.

Thank you very much! you are very kind and glad to share with others.

johanbrandhorst:
Hi, please rebase on `master`, we've changed our generation strategy quite a bit with https://github.com/grpc-ecosystem/grpc-gateway/pull/1971.

Issue #1766, opened 2020-10-20
grpc-gateway v2 misreads grpc protobuf field_mask behavior, breaks existing valid behavior


irridia:
## 🐛 Bug Report

In v2, repeated fields are now only allowed once per global scope:

[https://github.com/grpc-ecosystem/grpc-gateway/blob/4a35071321dc8a3fc19e8042ed54110fd1a42835/runtime/fieldmask.go#L77-L82](https://github.com/grpc-ecosystem/grpc-gateway/blob/4a35071321dc8a3fc19e8042ed54110fd1a42835/runtime/fieldmask.go#L77-L82)

There is a single comment in the protobuf spec that mentions this behavior _on a per-path basis_, not globally. Additionally, [further reading of that same document](https://github.com/protocolbuffers/protobuf/blob/6b0ff74ecf63e26c7315f6745de36aff66deb59d/src/google/protobuf/field_mask.proto#L152-L153) explains the specifics of this behavior, and that this behavior can be modified by an implementation.

While it may make sense that repeated fields can't be duplicated per path, enforcing a single repeated field globally IMHO contradicts both the standard and the explicitly described flexibility.

Multiple repeated fields have been allowed and properly supported by grpc and grpc-gateway for years, and changing this behavior, as it does in my case, makes the protobuf-supported use of field_masks impossible with grpc-gateway, effectively removing it as a reliable feature and breaking existing code bases.

AFAICT, aligning to this behavior would require multiple requests for a single update under logic described by grpc-gateway. This would require API consumers to know that they're hitting a specific new grpc-gateway implementation, and add implementation-specific logic and additional round-trips to work around the issue.

Just as a single person/company data point, this is a permanent hard-blocker for grpc-gateway v2. I absolutely understand the value of grpc-gw v2 as well as the middleware v2 work that I was slightly involved with, so I'd much rather not use v1 in the long term.

Thanks in advance for feedback!

## To Reproduce

Specify an update request with multiple repeated fields, e.g.,

```
{
  "entity0": {
    "field0": {
      "name": "Field Name 0",
      "entries" [
        "entry0",
        "entry1"
      ],
      "category": "entity"
    },
    "field1": {
      "name": "Field Name 1",
      "entries" [
        "entry2",
        "entry3"
      ],
      "category": "otherEntity"
    },
    "requestType": "update"
}
```

## Expected behavior

grpc-gateway v2 should populate field masks in the same compliant and functional way in which v1 performed this operation.

[...]

PR #1921, opened 2021-01-24
Multiple repeated fields are allowed

topherCantrell:
The protobuffer comments have been corrected here:
https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/field_mask.proto#L85-L86

The restriction was intended for the path string itself and not the
entire mask. There can be only one repeated field in a path string, and
when we get to it, the path string is done.

We found ourselves using multiple repeated fields in a message.

This fixes:

https://github.com/grpc-ecosystem/grpc-gateway/issues/1766

### DIFF
--- a/runtime/fieldmask.go
+++ b/runtime/fieldmask.go
@@ -88,7 +88,7 @@ func FieldMaskFromRequestBody(r io.Reader, msg proto.Message) (*field_mask.Field
 			fm.Paths = append(fm.Paths, item.path)
 		}
 	}
-	
+
 	return fm, nil
 }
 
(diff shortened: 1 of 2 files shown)


topherCantrell:
@googlebot I signed it!

nashikb:
Thanks @topherCantrell! This was still on my list, but hadn't yet made it to the top.

irridia:
Thanks @topherCantrell! This was still on my list, but hadn't yet made it to the top.

johanbrandhorst (changes requested):
Thanks for this! Could you please add a new test case that exhibits the new behaviour too?

topherCantrell:
There we go. A test for multiple repeats.

johanbrandhorst (approved):
Nice, thanks a lot for this!

Issue #1940, opened 2021-02-04
Non official implementation


rodoufu:
## 📚 Documentation

Do you have a wiki page for non-official implementations in other languages? Such as:
- https://www.npmjs.com/package/rest2grpc

That's a Typescript / Javascript implementation.

johanbrandhorst:
Hi @rodoufu, we don't have anything like that in the documentation at the moment, but we could add something like "related projects" to the docs if you think it'd be helpful to users. I'd be happy to review such a PR 😄.

rodoufu:
> Hi @rodoufu, we don't have anything like that in the documentation at the moment, but we could add something like "related projects" to the docs if you think it'd be helpful to users. I'd be happy to review such a PR .

https://github.com/grpc-ecosystem/grpc-gateway/pull/1943

PR #1943, opened 2021-02-04
Adding related projects to the documentation

rodoufu:
Closes: #1940

### DIFF
--- a/README.md
+++ b/README.md
@@ -422,3 +422,9 @@ See [CONTRIBUTING.md](http://github.com/grpc-ecosystem/grpc-gateway/blob/master/
 
 gRPC-Gateway is licensed under the BSD 3-Clause License.
 See [LICENSE.txt](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/LICENSE.txt) for more details.
+
+## Related projects
+
+Non-official implementations for other languages and related projects from the community.
+
+- TypeScript / JavaScript - https://www.npmjs.com/package/rest2grpc


rodoufu:
@johanbrandhorst here it is.

achew22:
@rodoufu, thanks so much for sending in the reference. I think this would be better as an additional stanza in our [FAQ](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/docs/docs/faq.md#my-grpc-server-is-written-in-scala-or-c-or-ruby-or-haskell-etc-is-there-a-scala-or-c-or-ruby-or-haskell-etc-version-of-grpc-gateway) titled "My gRPC server is written in (Scala or C++ or Ruby or Haskell etc). Is there a (Scala or C++ or Ruby or Haskell etc) version of gRPC-Gateway." Let's go with something like this:

```
As of now, not in general. But it should not be a big issue because the
reverse-proxy which gRPC-Gateway generates usually works as an independent
process and communicates with your gRPC server over TCP or a Unix domain
sockets (Unix systems only).

If you do insist that your gateway should be written in the same language
as your server, there are a few projects that will help with this:

* NodeJS:
  * [grps-dynamic-gateway](https://github.com/konsumer/grpc-dynamic-gateway)
  * Not a direct analog, but [rest2grpc](https://www.npmjs.com/package/rest2grpc) is similar.
* Other languages
  * Please add any references to other language implementations here.

```

rodoufu:
> @rodoufu, thanks so much for sending in the reference. I think this would be better as an additional stanza in our [FAQ](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/docs/docs/faq.md#my-grpc-server-is-written-in-scala-or-c-or-ruby-or-haskell-etc-is-there-a-scala-or-c-or-ruby-or-haskell-etc-version-of-grpc-gateway) titled "My gRPC server is written in (Scala or C++ or Ruby or Haskell etc). Is there a (Scala or C++ or Ruby or Haskell etc) version of gRPC-Gateway." Let's go with something like this:
> 
> ```
> As of now, not in general. But it should not be a big issue because the
> reverse-proxy which gRPC-Gateway generates usually works as an independent
> process and communicates with your gRPC server over TCP or a Unix domain
> sockets (Unix systems only).
> 
> If you do insist that your gateway should be written in the same language
> as your server, there are a few projects that will help with this:
> 
> * NodeJS:
>   * [grps-dynamic-gateway](https://github.com/konsumer/grpc-dynamic-gateway)
>   * Not a direct analog, but [rest2grpc](https://www.npmjs.com/package/rest2grpc) is similar.
> * Other languages
>   * Please add any references to other language implementations here.
> ```

I'm closing this one and creating #1946

PR #1947, opened 2021-02-04
Fix verb parsing

jefferai:
This is a long description for a very small change, apologies.

According to http.proto
(https://github.com/grpc-ecosystem/grpc-gateway/blob/master/third_party/googleapis/google/api/http.proto#L232)
the path template consists of a slash, one or more segments, and a verb.
The verb is specified as ':' followed by a LITERAL that matches literal
text in the path.

In the upstream http.proto
(https://github.com/googleapis/googleapis/blob/master/google/api/http.proto#L242)
they expand on the meaning of this:

> The syntax `LITERAL` matches literal text in the URL path. If the
> `LITERAL` contains any reserved character, such characters should be
> percent-encoded before the matching.

Unfortunately it doesn't say which characters are reserved. In some cases
it follows RFC 6570 (which reserves colon in some instances) and in
other cases it doesn't.

The intent, however, is pretty clear: after any segments, which are
delimited with a slash, a colon followed by a string indicates a verb.

The problem is that currently in some cases this parsing fails within
the generator. This works, but if you examine the generated code you'll
see that the verb is considered to be "e" rather than "d:e":

> /a/b/c:d:e

However if there is a path variable:

> /a/b/{name}:d:e

it simply errors out during generation.

The problem appears to be that when looking for the verb, it looks for
the last index of a colon instead of the first. This changes the logic
to look for the _first_ index of a colon. As you can see from the
associated test cases, this results in the behavior that a verb that
itself contains a colon is parsed as a verb.

There is a potential drawback, which is that if we suppose that
according to the spec a LITERAL can contain a colon, it means a LITERAL
in a segment can also contain a colon. So in reality you can never know
whether parsing the first or last colon is actually the right thing to
do when trying to figure out the verb.

Arguably, the only way to truly fix this would be to implement a custom
option that indicates what the desired verb is.

### DIFF
--- a/internal/descriptor/registry_test.go
+++ b/internal/descriptor/registry_test.go
@@ -780,6 +780,9 @@ func assertStringSlice(t *testing.T, message string, got, want []string) {
 	}
 }
 
+// TestVariableVerbparsing is a higher-level test than at the parser level of
+// the special logic around ensuring verbs are correctly parsed if the last
+// segment is a variable.
 func TestVariableVerbParsing(t *testing.T) {
 	reg := NewRegistry()
 	fd := loadFile(t, reg, `
--- a/internal/descriptor/registry_test.go
+++ b/internal/descriptor/registry_test.go
@@ -779,36 +779,3 @@ func assertStringSlice(t *testing.T, message string, got, want []string) {
 		}
 	}
 }
-
-// TestVariableVerbparsing is a higher-level test than at the parser level of
-// the special logic around ensuring verbs are correctly parsed if the last
-// segment is a variable.
-func TestVariableVerbParsing(t *testing.T) {
-	reg := NewRegistry()
-	fd := loadFile(t, reg, `
-		name: 'example.proto'
-		package: 'example'
-		message_type <
-			name: 'B'
-			field <
-				name: 'name'
-				label: LABEL_OPTIONAL
-				type: TYPE_STRING
-				number: 1
-			>
-		>
-		service <
-			name: "BService"
-			method <
-				name: "Meth"
-				input_type: "B"
-				output_type: "B"
-				options <
-					[google.api.http] < post: "/v1/b/{name}:foo:bar" body: "*" >
-				>
-			>
-		>
-	`)
-
-	_ = fd
-}
--- a/internal/descriptor/registry_test.go
+++ b/internal/descriptor/registry_test.go
@@ -779,3 +779,33 @@ func assertStringSlice(t *testing.T, message string, got, want []string) {
 		}
 	}
 }
+
+func TestVariableVerbParsing(t *testing.T) {
+	reg := NewRegistry()
+	fd := loadFile(t, reg, `
+		name: 'example.proto'
+		package: 'example'
+		message_type <
+			name: 'B'
+			field <
+				name: 'name'
+				label: LABEL_OPTIONAL
+				type: TYPE_STRING
+				number: 1
+			>
+		>
+		service <
+			name: "BService"
+			method <
+				name: "Meth"
+				input_type: "B"
+				output_type: "B"
+				options <
+					[google.api.http] < post: "/v1/b/{name}:foo:bar" body: "*" >
+				>
+			>
+		>
+	`)
+
+	_ = fd
+}
--- a/internal/httprule/parse.go
+++ b/internal/httprule/parse.go
@@ -45,9 +45,7 @@ func tokenize(path string) (tokens []string, verb string) {
 		field
 		nested
 	)
-	var (
-		st = init
-	)
+	st := init
 	for path != "" {
 		var idx int
 		switch st {
@@ -80,8 +78,30 @@ func tokenize(path string) (tokens []string, verb string) {
 	}
 
 	l := len(tokens)
+	// See
+	// https://github.com/grpc-ecosystem/grpc-gateway/pull/1947#issuecomment-774523693 ;
+	// although normal and backwards-compat logic here is to use the last index
+	// of a colon, if the final segment is a variable followed by a colon, the
+	// part following the colon must be a verb. Hence if the previous token is
+	// an end var marker, we switch the index we're looking for to Index instead
+	// of LastIndex, so that we correctly grab the remaining part of the path as
+	// the verb.
+	var penultimateTokenIsEndVar bool
+	switch l {
+	case 0, 1:
+		// Not enough to be variable so skip this logic and don't result in an
+		// invalid index
+	default:
+		penultimateTokenIsEndVar = tokens[l-2] == "}"
+	}
 	t := tokens[l-1]
-	if idx := strings.LastIndex(t, ":"); idx == 0 {
+	var idx int
+	if penultimateTokenIsEndVar {
+		idx = strings.Index(t, ":")
+	} else {
+		idx = strings.LastIndex(t, ":")
+	}
+	if idx == 0 {
 		tokens, verb = tokens[:l-1], t[1:]
 	} else if idx > 0 {
 		tokens[l-1], verb = t[:idx], t[idx+1:]
--- a/internal/httprule/parse_test.go
+++ b/internal/httprule/parse_test.go
@@ -13,6 +13,7 @@ func TestTokenize(t *testing.T) {
 	for _, spec := range []struct {
 		src    string
 		tokens []string
+		verb   string
 	}{
 		{
 			src:    "",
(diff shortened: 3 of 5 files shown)


jefferai:
@googlebot I signed it!

johanbrandhorst:
Hi, please rebase on `master` to get the fix for the flaky test.

johanbrandhorst (approved):
This change makes me a little bit nervous because we've had various issue around the parsing of the verb in the path (https://github.com/grpc-ecosystem/grpc-gateway/issues/224), but seeing as all the tests pass, I'm happy to merge this now and hope it doesn't break someone.

Just before I merge this, on Slack we discussed attempting to confirm the behaviour with a third party http.proto parser, were you able to confirm that this behaviour works as expected with the GCP HTTP/JSON transcoder?

jefferai:
Rebased to fix tests.

@johanbrandhorst I haven't. I honestly have no idea where I'd even look for the code (or utility/tool? or API? or web page?) for the GCP transcoder. I'm happy to do it but I need some handholding.

jefferai:
@johanbrandhorst I traipsed through the issue you linked and, I gotta be honest, I'm not sure this is the right solution. Clearly _a_ solution needs to be found, but given some of the other use-cases in that issue this may well break some even if your built in tests are passing.

The problem is that the final segment can contain a verb, but it's not clear whether the segment itself can contain a colon -- colons are [not reserved within URIs](https://stackoverflow.com/questions/2053132/is-a-colon-safe-for-friendly-url-use/2053640) and I can't find anything in http.proto which expands on what "reserved" means so I have taken it to mean the same as the URI spec.

However, this would mean that it really is absolutely impossible to know where one should split a verb because you have a LITERAL on both sides separated by a colon, and it appears valid for each LITERAL to have one or more colons themselves. So this change is no more valid than the existing behavior, and both will fail in various ways.

I think the only way to actually address this would be to allow some kind of override for a given service definition that allows you to override the parser's assertion of what is or isn't a verb and instead specify it manually.

I would suggest not merging the code as-is; if we can work out an alternative implementation I will update this code to reflect that instead. (What originally drove this PR was the fact that `.../{foo}:some:verb` would throw an error, which this code does fix, but it's possible that's a separate issue within the parser, or that the parser ends up giving something unexpected to a later consumer.

johanbrandhorst:
Thanks for the update Jeff, really appreciate your thorough research. I wouldn't mind adding a knob to tweak the behaviour of the parser if we can clearly define the two different behaviors it would toggle. I admit to not having a complete grasp of the subtleties involved here but if we can help users make an informed choice for their application we might be able to make everyone happy?

jefferai:
Status update...

I originally thought about extending `google.api.Http` to allow an option to be specified, but it actually disallows extensions 😞 

So my current thinking is to allow specifying gateway options on a method itself, similar to how you can specify OpenAPI options. This will require getting a new extension number because both plugins would end up extending the same google.descriptor.MethodOptions proto. (We could add more fields into the OpenAPI version but that feels distinctly weird to mix the two like that.)

Then, similar to the OpenAPI options, I'd have to pass those options down through the various layers. I'm not entirely sure what those are yet; it could be down to the parser, or maybe somewhere along the way. I did notice that `runtime` contains code very similar to the parser for figuring out what the verb is -- I'm not sure why those two aren't in sync as I haven't dabbled in it yet, I would have figured that the point of the parsing and generating is for the runtime to have that information. Possibly that means having to add that to the generated template. Or just having yet another option to tell the runtime the same thing.

Is it worth me continuing? I'm interested in continuing if the current path sounds tenable, but at this point it looks like a decent amount of work and changed code and if it's unlikely to be accepted (or sit unmerged for ages, as I have a need for this in the nearish term) then I won't continue investing time into it.

jefferai:
Actually, I am coming back around to the idea that the original patch is more or less correct, with a specific modification. 😁 

Although it's true that for a given literal it may contain some unknown number of colons, what I didn't think about was the semantics around variables.

Circling back to `http.proto`:
```
```
In the template, we can see that a verb is a literal after a segment, separated by a colon. If the final path segment is itself a literal, then what I said above is true -- we can't know the true intent of the user. However, what started me down this path was actually trying to do the following:

`/foo/{name}:action:subaction`

If the last segment is a variable, rather than a literal, then according to the spec, a colon following the variable declaration in that last segment _must_ be a verb, including any colons contained within the verb!

Although this would not solve it for the general case, I believe a patch that ensures that if the last path segment is a variable followed by a colon that anything after that colon is the content of the verb would actually both make the parser more spec-compliant (according to the grammar above) and would solve my problem.

I would likely have to implement this fix both in `httprule` and `runtime` as there's essentially duplicated logic there, but I think it's provably correct, and the blast radius is way smaller than what I was thinking earlier.

Does that sound good to you?

johanbrandhorst:
> Actually, I am coming back around to the idea that the original patch is more or less correct, with a specific modification. grin
> 
> Although it's true that for a given literal it may contain some unknown number of colons, what I didn't think about was the semantics around variables.
> 
> Circling back to `http.proto`:
> 
> ```
> //     Template = "/" Segments [ Verb ] ;
> //     Segments = Segment { "/" Segment } ;
> //     Segment  = "*" | "**" | LITERAL | Variable ;
> //     Variable = "{" FieldPath [ "=" Segments ] "}" ;
> //     FieldPath = IDENT { "." IDENT } ;
> //     Verb     = ":" LITERAL ;
> ```
> 
> In the template, we can see that a verb is a literal after a segment, separated by a colon. If the final path segment is itself a literal, then what I said above is true -- we can't know the true intent of the user. However, what started me down this path was actually trying to do the following:
> 
> `/foo/{name}:action:subaction`
> 
> If the last segment is a variable, rather than a literal, then according to the spec, a colon following the variable declaration in that last segment _must_ be a verb, including any colons contained within the verb!
> 
> Although this would not solve it for the general case, I believe a patch that ensures that if the last path segment is a variable followed by a colon that anything after that colon is the content of the verb would actually both make the parser more spec-compliant (according to the grammar above) and would solve my problem.
> 
> I would likely have to implement this fix both in `httprule` and `runtime` as there's essentially duplicated logic there, but I think it's provably correct, and the blast radius is way smaller than what I was thinking earlier.
> 
> Does that sound good to you?

This analysis sounds sound to me, but we'll probably need to jot these thoughts down in the docs somewhere for others to find, how do you feel about adding that to the list 😁?

I'm glad we're not going down that path, but in case you run into any trouble and want to consider having this be configurable behaviour, note that we can add knobs both to the `runtime.ServeMux` or `protoc-gen-grpc-gateway` for runtime and generation time configuration respectively. We haven't had to add any protobuf options that affect the gateway so far (only the openapiv2 generator).

As usual, thanks for your thoughtful analysis. Always a pleasure.

jefferai:
Already having some success. Converting to draft while I work on it.

jefferai:
Alright, ready for a new review. This includes the discussed change to the parser, along with tests, and includes necessary changes to runtime. I think that this actually makes runtime more robust generally when dealing with verbs (as I note in the comment), although I'm not sure that with these changes we wouldn't want to re-derive the verb in the fallback mechanism as well -- happy to add that if needed. I've tested this with a third-party project in addition to the built-in tests and it seems to work well.

internal/descriptor/registry_test.go
```diff
@@ -779,3 +779,36 @@ func assertStringSlice(t *testing.T, message string, got, want []string) {
...
+				input_type: "B"
+				output_type: "B"
+				options <
+					[google.api.http] < post: "/v1/b/{name}:foo:bar" body: "*" >
+				>
+			>
+		>
+	`)
+
+	_ = fd
```
johanbrandhorst, line 813: Is the purpose of this test simply to check if the file loads, or do we want to make any assertions here?

johanbrandhorst (commented):
Small question surrounding the new registry test, but generally looks great. Thanks again for taking the take for the thorough research and testing.

internal/descriptor/registry_test.go
jefferai: In the prior behavior such a file would cause the parser to error. It's meant as a higher level check against the same tests I put in the parser test file. I'll remove it as there are now higher level tests in mux_test.go as well.

jefferai: Removed!

jefferai:
Thanks for merging!

johanbrandhorst:
Thanks for your contribution! I think we'll cut a new release with this in the bag, but feel free to try out master for your own needs in the meantime!

PR #1951, opened 2021-02-05
gen-grpc-gateway, gen-openapiv2: add support for proto3 optional

### LINKED ISSUE #1278: Support optional annotation in proto3 files in generators

adambabik:
Fixes #1278.

Even though, the main piece of code was added in `google.golang.org/protobuf/compiler/protogen` package, each plugin needs to explicitly enable support for proto3 optional fields. This change does exactly that. Note that `gen-openapiv2` still does not fully benefit from `protogen`, hence, there are two helper methods enabling the feature in two different ways.

Also, this change does not treat optional fields in any different way in neither `gen-grpc-gateway`, nor `gen-openapiv2` which actually might be required to deliver a valid experience.


adambabik:
@johanbrandhorst I will need some help with failing bazel job. It seems that `--experimental_allow_proto3_optional` needs to be provided somehow in `examples/internal/proto/examplepb/BUILD.bazel`. [This](https://github.com/bazelbuild/rules_proto/issues/71) might be related.

achew22:
@adambabik I think you might be blocked by https://github.com/bazelbuild/rules_go/pull/2788 on adding protocopts to rules_go

johanbrandhorst (changes requested):
This LGTM, but I'm still slightly worried that we should do something special here. Do you think you could add an integration test that uses this new optional string? The test would go here: https://github.com/grpc-ecosystem/grpc-gateway/blob/master/examples/internal/integration/integration_test.go.

adambabik:
@johanbrandhorst thanks for the suggestion. Adding a case to the integration tests revealed one more missing piece, that is optional fields require different converters. I used converters from proto2 because they return pointers. It would be nice to move them to more generic place I think. It's for another PR :) I didn't add so through testing for optional fields as it is for repeated. If it's not enough, let me know.

examples/internal/proto/examplepb/a_bit_of_everything.proto
```diff
@@ -458,7 +460,7 @@ service ABitOfEverythingService {
 	// This API creates a new ABitOfEverything
 	rpc Create(ABitOfEverything) returns (ABitOfEverything) {
 		option (google.api.http) = {
-			post: "/v1/example/a_bit_of_everything/{float_value}/{double_value}/{int64_value}/separator/{uint64_value}/{int32_value}/{fixed64_value}/{fixed32_value}/{bool_value}/{string_value=strprefix/*}/{uint32_value}/{sfixed32_value}/{sfixed64_value}/{sint32_value}/{sint64_value}/{nonConventionalNameValue}/{enum_value}/{path_enum_value}/{nested_path_enum_value}/{enum_value_annotation}"
+			post: "/v1/example/a_bit_of_everything/{float_value}/{double_value}/{int64_value}/separator/{uint64_value}/{int32_value}/{fixed64_value}/{fixed32_value}/{bool_value}/{string_value=strprefix/*}/{uint32_value}/{sfixed32_value}/{sfixed64_value}/{sint32_value}/{sint64_value}/{nonConventionalNameValue}/{enum_value}/{path_enum_value}/{nested_path_enum_value}/{enum_value_annotation}/{optional_string_value}"
```
johanbrandhorst, line 463: This is an interesting case, because if `optional_string_value` is truly optional, should it be possible to have it be a path parameter? Will our parser handle the case where this is unset? We may need to trigger a generation-time error when a user does this?

examples/internal/proto/examplepb/a_bit_of_everything.swagger.json
```diff
@@ -1711,6 +1735,12 @@
               "ZERO",
               "ONE"
             ]
+          },
+          {
+            "name": "optionalStringValue",
+            "in": "path",
+            "required": true,
```
johanbrandhorst, line 1742: This might break the promise of optional values

internal/descriptor/types.go
```diff
@@ -356,8 +366,10 @@ func (p FieldPath) AssignableExpr(msgExpr string) string {
 	var preparations []string
 	components := msgExpr
 	for i, c := range p {
-		// Check if it is a oneOf field.
-		if c.Target.OneofIndex != nil {
+		if c.Target.GetProto3Optional() {
+			// Continue as a proto3 optional field does not need a special treatment.
+			// This condition is required as otherwise it is matched by `c.Target.OneofIndex != nil`.
+		} else if c.Target.OneofIndex != nil { // check if it is a oneOf field
```
johanbrandhorst, line 372: Can we invert this to avoid an else?
```suggestion
		if !c.Target.GetProto3Optional() && c.Target.OneofIndex != nil { // check if it is a oneOf field
```

johanbrandhorst (changes requested):
Thanks for the new test Adam, I think it exposes an important question - we should probably disallow optional fields in path parameters, since they're required by definition.

examples/internal/proto/examplepb/a_bit_of_everything.proto
adambabik: That's a good idea! Is there a place where I could document how optional fields behave?

internal/descriptor/types.go
adambabik: I wanted to separate proto3 optional and oneof cases, hence, separate blocks but I can also join them. I will keep the explanation.

adambabik: Actually I read a bit more about it. It [does use oneof](https://github.com/protocolbuffers/protobuf-go/blob/ac374a2335472e7ef5b5c28e72c3060121b8d1ed/types/descriptorpb/descriptor.pb.go#L884-L905) under the hood.

johanbrandhorst (approved):
LGTM

johanbrandhorst:
Bazel failure seems genuine - we need to configure `protoc` to use the new flag:
```
examples/internal/proto/examplepb/a_bit_of_everything.proto: This file contains proto3 optional fields, but --experimental_allow_proto3_optional was not set.
```

johanbrandhorst:
Please rebase on `master` 😄

mvrhov:
protoc 3.15.... 

> Optional fields for proto3 are enabled by default, and no longer require the --experimental_allow_proto3_optional flag.

achew22:
Well... I don't think you're blocked by rules_go any more then 😉

glerchundi:
Whoop I didn't notice that this PR was created before creating a new one #1988 after we decided to close the attempt I made months ago #1834.

My approach in the new PR is to just make `protoc-gen-grpc-gateway` & `protoc-gen-openapiv2` compatible with the latest `protoc` >3.15 releases without changing the current generators behaviour.

Feel free to do whatever you want with this new PR.

PR #1955, opened 2021-02-09
Add documenation for standalone gateway generation

### LINKED ISSUE #1931: What is the recommend way to bind 3rd party APIs (e.g. Health/Check to /healthzf
## 📚 Documentation

There are docs for [implementing `grpc.health.v1.Health/Check`](https://grpc-ecosystem.github.io/grpc-gateway/docs/operations/health_check/), but I have not been able to find much documentation on how to map 3rd party gRPC services to http methods, especially for standalone servers.

When using Cloud Endpoints, simply adding annotations to copied protos and running `protoc --include_imports ...` worked alright, and likely YAML configuration would have worked as well. However, I have found no straightforward way to generate the 3rd party mappings when running a standalone instance, and have currently resorted to munging generated output into a separate package.

I don't know what other standardized protocols exist in the gRPC ecosystem other than `grpc.health.v1`, so maybe this is worth treating as a special case.

dgparker:
Fixes #1931 

The documentation is quite concise and I've tried to follow in the same vein. Let me know if anyone thinks we should be verbose in certain areas.

FAQ section:
- added a small blurb that points readers to the gRPC API Configuration page in the docs for certain use cases

gRPC API Configuration:
- added an excerpt on the `standalone=true` option for situations where users may reference externally defined and generated proto files and only require the reverse-proxy code.


johanbrandhorst (approved):
LGTM

johanbrandhorst:
Thanks for your contribution! Hopefully this will help others in the future 😄.

PR #1973, opened 2021-02-16
adding UnimplementedGreeterServer to server struct

nwandabridges:
The compiler produces an error without this. The error is:

```
cannot use &server literal (type *server) as type helloworld.GreeterServer in argument to helloworld.RegisterGreeterServer:
	*server does not implement helloworld.GreeterServer (missing helloworld.mustEmbedUnimplementedGreeterServer method)
```


johanbrandhorst (approved):
Nice find!

johanbrandhorst:
Thanks for your contribution!

PR #1266, opened 2020-05-04
protoc-gen-swagger, Fix for infinite loop on circular references in query parameters 

Romeren:
**Added function nestedQueryParams** in *proto-gen-swagger/genswagger/template.go* with *map[string]bool* parameter for keeping track of and detecting circular references.
**Added test TestMessageToQueryParametersRecursive** in *proto-gen-swagger/genswagger/template_test.go* for testing gracefully handling of circular references between messages.
See issue #1167

PR #2022, opened 2021-03-09
implements a max recursive depth check

### LINKED ISSUE #2011: Support recursive types
## 🚀 Feature

#1437 removed the ability to use recursive types. We used these in protoc-gen-gorm to created nested query structures. We parse SQL where clauses into an AST tree ie. `WHERE group=acme AND tag=recent`. The issue is that this syntax can have multiple layers of logical operators (AND OR).

```
WHERE (group='acme' AND tag  = 'recent') or tag = 'override')
```

Now that nested query parameters are specifically rejected, we are unable to parse these nested structures. I would like to be able to have recursive types to build trees. Here's an example proto definition. One of the possible parameters of LogicalOperator is LogicalOperator. As openapiv2 plugin inspects this, it returns the error: `panic: recursive types are not allowed for query parameters, cycle found on ".infoblox.api.LogicalOperator"`

https://github.com/infobloxopen/atlas-app-toolkit/blob/master/query/collection_operators.proto#L83
```
message LogicalOperator {
    oneof left {
        LogicalOperator left_operator = 1;
        StringCondition left_string_condition = 2;
        NumberCondition left_number_condition = 3;
        NullCondition left_null_condition = 4;
        StringArrayCondition left_string_array_condition = 11;
        NumberArrayCondition left_number_array_condition = 12;
    }
    oneof right {
        LogicalOperator right_operator = 5;
        StringCondition right_string_condition = 6;
        NumberCondition right_number_condition = 7;
        NullCondition right_null_condition = 8;
        StringArrayCondition right_string_array_condition = 13;
        NumberArrayCondition right_number_array_condition = 14;
    }
    enum Type {
        AND = 0;
        OR = 1;
    }
    Type type = 9;
    bool is_negative = 10;
}
```

drewwells:

fixes #2011 
relates #1437 #1266

This modifies the absolute limit on recursive type checks for a cli configurable depth check.

#### Other comments
I don't quite follow how the recursive unit tests work, but they hung my computer when I disabled the recursive depth check.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,7 @@ bazel-genfiles
 bazel-grpc-gateway
 bazel-out
 bazel-testlogs
-.bazelrc
+/.bazelrc
 
 # Go vendor directory
 vendor
--- a/internal/descriptor/registry.go
+++ b/internal/descriptor/registry.go
@@ -109,6 +109,9 @@ type Registry struct {
 
 	// omitPackageDoc, if false, causes a package comment to be included in the generated code.
 	omitPackageDoc bool
+
+	// recursiveDepth sets the maximum depth of a field parameter
+	recursiveDepth int
 }
 
 type repeatedFieldSeparator struct {
@@ -134,6 +137,7 @@ func NewRegistry() *Registry {
 		messageOptions: make(map[string]*options.Schema),
 		serviceOptions: make(map[string]*options.Tag),
 		fieldOptions:   make(map[string]*options.JSONSchema),
+		recursiveDepth: 1000,
 	}
 }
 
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -7,7 +7,7 @@ bazel-genfiles
 bazel-grpc-gateway
 bazel-out
 bazel-testlogs
-.bazelrc
+/.bazelrc
```
drewwells, line 10: I accidentally made .bazelrc files in subdirectories and this check is ignoring any .bazelrc even if its made in places other than the project root.

internal/descriptor/registry.go
```diff
@@ -356,6 +360,16 @@ func (r *Registry) SetStandalone(standalone bool) {
...
 
+// SetRecursiveDepth records the max recursion count
+func (r *Registry) SetRecursiveDepth(count int) {
+	r.recursiveDepth = count
+}
+
+// GetRecursiveDepth returns the max recursion count
+func (r *Registry) GetRecursiveDepth() int {
+	return r.recursiveDepth
+}
```
drewwells, line 371: These aren't thread safe

johanbrandhorst (changes requested):
Hey, this looks like a great start, thank you! I might be wrong but I think we should be able to simply remove any consideration for concurrent operations, such as the mutex and the immutable branching, which would simplify the code a bit. Let me know what you think.

johanbrandhorst (approved):
LGTM

johanbrandhorst:
Thanks for your contribution!

Issue #500, opened 2017-12-12
Handling Binary File Uploads


davejohnston:
Hi,

I have a protobuf that defines one field

message FileUpload {
    bytes fileContents = 1;
}

I want to be able to upload a file, and have the contents of that upload stored in fileContents.  When sending json data, I understand that the keys map to grpc message fields.  But in the case of a binary upload there is no field name, so how can I map the contents of the request body into a message?

    rpc UploadFile(FileUpload) returns (google.protobuf.Empty)) {
        option (google.api.http) = {
            post: "/v1/files"
            body: "*"
        };
    }

If I use curl to make a request like this:
curl -X POST --data-binary "@/tmp/test_file.txt" http://localhost:9090/v1/files

The HTTP request looks like this:

```
POST /v1/cmd/secretfiles HTTP/1.1
Host: localhost:9090
User-Agent: curl/7.54.0
Accept: */*
Content-Length: 1675
Content-Type: application/x-www-form-urlencoded
Expect: 100-continue

uEH2KxtiW5Tpuj7bU5hmPO1h4QZuCcrrNDA/OvNyYmw0U02Uy/mR5iWnS6tMdPmZ
iJBmN06MZ2Khx5rV+rqvwF9CUGYzja/dDCm+2wIDAQABAoIBAB050/iJ0YpyGtig
hRPQ7IetKx8HRfJYLSlYu9+eo4/e7EGAfm1dZWz9pJO0kcnUCn3iOKZUDqdGqIz0
9mXR0wK8DaWgYFf0Wx9d8/EEOdUUrQ9Eh82CCWk94fbCbC/b1NdZ90DaUhIZ1J7y
keCiyuPj5rClhrAdf/GYy8bEXp2W9+zi5vH2dTi3JDe2rImh+urEjCYThV3dxoms
HrwX+nWrwNVXfAIJ3R8ojRoiFQtckYgSytyyYdGaxvaZf4rw3CWK+4U1b6NiC4XP
IDVpBfWSXsbPvliA4A35F/15Let/ASw4YZiUryrsYBVNMBMjiydULFvNX0WQVVJE
```

achew22:
[...]

PR #2054, opened 2021-03-25
Add docs page for binary file uploads, for #500

jonathanbp:
This PR adds a new page to the docs dir giving an example of how to do binary file uploads from the discussion on #500.

### DIFF
--- a/docs/docs/mapping/binary_file_uploads.md
+++ b/docs/docs/mapping/binary_file_uploads.md
@@ -1,0 +1,53 @@
+---
+layout: default
+title: Binary file uploads
+nav_order: 2
+parent: Mapping
+---
+
+# Binary file uploads
+
+If you need to do a binary file upload, e.g. via;
+
+```sh
+curl -X POST -F "attachment=@/tmp/somefile.txt" http://localhost:9090/v1/files
+```
+
+then your request will contain the binary data directly and there is no way to model this using gRPC.
+
+What you can do instead is to add a custom route directly on the `mux` instance.
+
+## Custom route on a mux instance
+
+Here we'll setup a handler (`handleBinaryFileUpload`) for `POST` requests: 
+
+```go
+// Create a mux instance
+mux := runtime.NewServeMux()
+
+// Attachment upload from http/s handled manually
+mux.HandlePath("POST", "/v1/files", handleBinaryFileUpload)
+```
+
+And then in your handler you can do something like:
+
+```go
+func handleBinaryFileUpload(w http.ResponseWriter, rq *http.Request, params map[string]string) {
+	err := r.ParseForm()
+	if err != nil {
+		http.Error(w, fmt.Sprintf("failed to parse form: %s", err.Error()), http.StatusBadRequest)
+		return
+	}
+
+	f, header, err := rq.FormFile("attachment")
+	if err != nil {
+		http.Error(w, fmt.Sprintf("failed to get file 'attachment': %s", err.Error()), http.StatusBadRequest)
+	}
+	defer f.Close()
+
+	//
+	// Now do something with the io.Reader in `f`, i.e. read it into a buffer or stream it to a gRPC client side stream.
+	// Also `header` will contain the filename, size etc of the original file.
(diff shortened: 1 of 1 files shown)


docs/docs/mapping/binary_file_uploads.md
```diff
@@ -0,0 +1,62 @@
...
+
+// Attachment upload from http/s handled manually
+mux.HandlePath("POST", "/v1/files", handleBinaryFileUpload)
+```
+
+And then in your handler you can do something like:
+
+```go
+func handleBinaryFileUpload(w http.ResponseWriter, rq *http.Request, params map[string]string) {
+
```
johanbrandhorst, line 36: Remove this newline please

docs/docs/mapping/binary_file_uploads.md
```diff
@@ -0,0 +1,62 @@
...
+```
+
+And then in your handler you can do something like:
+
+```go
+func handleBinaryFileUpload(w http.ResponseWriter, rq *http.Request, params map[string]string) {
+  
+	err := r.ParseForm()
+	if err != nil {
+		panic(err)
```
johanbrandhorst, line 39: Lets make this write an error to the response rather than panic
```suggestion
		http.Error(w, fmt.Sprintf("failed to parse form: %s", err.Error()), http.StatusBadRequest)
		return
```

docs/docs/mapping/binary_file_uploads.md
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	f, header, err := rq.FormFile("attachment")
+	if err != nil {
+		panic(err)
```
johanbrandhorst, line 44: Ditto error response

docs/docs/mapping/binary_file_uploads.md
```diff
@@ -0,0 +1,62 @@
...
+	}
+	defer f.Close()
+
+	// header contains filename etc if you need this
+
+	var buffer bytes.Buffer
+	_, err = io.Copy(buffer, f)
+	if err != nil {
+		panic(err)
+	}
```
johanbrandhorst, line 54: I think we can remove this part, since it's not clear the user even necessarily wants to do this, maybe they want to stream from `f` directly rather than read it fully into memory.

docs/docs/mapping/binary_file_uploads.md
```diff
@@ -0,0 +1,62 @@
...
+
+	//
+	// Now do something with the bytes in the `buffer`
```
johanbrandhorst, line 57: Given the above, we can reword this
```suggestion
	// Now do something with the io.Reader in `f`, i.e. read it into a buffer or stream it to a gRPC client side stream.
```

docs/docs/mapping/binary_file_uploads.md
```diff
@@ -0,0 +1,62 @@
...
+	//
+
+	w.WriteHeader(http.StatusOK)
```
johanbrandhorst, line 60: I think we can omit this, the Go HTTP server does it automatically if no other header is written.

johanbrandhorst (changes requested):
This looks great, thanks @jonathanbp 😄. A few comments on the example.

jonathanbp:
@googlebot I fixed it.

jonathanbp:
@googlebot I consent.

jonathanbp:
Alright, I've updated the PR according to your suggestions, @johanbrandhorst .

johanbrandhorst:
Great, thanks for the updates. Unfortunately the CLA bot is confused because some of the commits are under your current github user and some are under @purebadger. Could you try and squash the commits to be all under the same user, or log into @purebadger and sign the CLA, either will work.

purebadger:
Ah, yes. To add to the confusion, I signed the CLA using the wrong google account.

purebadger:
@googlebot I consent.

purebadger:
Can't get the CLA to work. I'm scratching this PR and doing a new one without involving @purebadger this time.

PR #527, opened 2018-01-17
Fix decoding empty request body

syhpoon:
Fix for this issue: https://github.com/grpc-ecosystem/grpc-gateway/issues/464

PR #2078, opened 2021-04-05
Fix path params being in the body

### LINKED ISSUE #201: Fields bound by path template also in body (Swagger)

stijndehaes:

Fixes #201

I have read it, and have regenerated the files with the first and second docker run command. 

I toke the previous PR here: https://github.com/grpc-ecosystem/grpc-gateway/pull/1145

And migrated it to the current version of the code. I also tested this out on a project of mine, and the result looks good.

#### Other comments

~I am not sure how to unit test this. If somebody could give me some hints that would be wonderful.~
I added one unit test, not sure if I added it to the right place though. I would like feedback on this :)


johanbrandhorst (commented):
Thanks for picking this up @stijndehaes. @jgiles, since this is based on your original contribution, could you please explicitly give consent to have it be merged here?

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -384,7 +375,92 @@ func skipRenderingRef(refName string) bool {
...
+		}
+
+		kv := keyVal{Value: fieldValue}
+		kv.Key = reg.FieldName(f)
+		if schema.Properties == nil {
+			schema.Properties = &openapiSchemaObjectProperties{}
+		}
+		*schema.Properties = append(*schema.Properties, kv)
+	}
+	return schema
```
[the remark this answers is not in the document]
stijndehaes: I quickly tried this. But for example in the `flow_combination.proto` there are empty responses, by making this method return an error that is a problem. I added the check where the method is used to generate the body instead. I hope this makes sense, if you would like to do this another way I can change it

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -986,14 +986,17 @@ func renderServices(services []*descriptor.Service, paths openapiPathsObject, re
 								}
 							}
 							if len(bodyExcludedFields) != 0 {
-								schema = renderMessageAsDefinition(meth.RequestType, reg, customRefs, bodyExcludedFields)
+								msgDef := renderMessageAsDefinition(meth.RequestType, reg, customRefs, bodyExcludedFields)
+								if msgDef.Properties == nil || len(*msgDef.Properties) == 0 {
+									return fmt.Errorf("not possible to create a schema for object with 0 properties for message: %s", *meth.RequestType)
+								}
```
johanbrandhorst, line 992: Thanks for this change, though I just had another thought about this. I would see this as a bug fix, but it's also probably a breaking change for users, so maybe we should just log a warning instead. See `glog.Errorf` uses in this file.

examples/internal/proto/examplepb/a_bit_of_everything.proto
```diff
@@ -641,8 +641,7 @@ service ABitOfEverythingService {
 	}
 	rpc PostWithEmptyBody(Body) returns (google.protobuf.Empty) {
 		option (google.api.http) = {
-			post: "/v2/example/postwithemptybody/{name}",
-			body: "*"
```
johanbrandhorst, line 645: Lets revert this change too, apparently this is behaviour we want to keep working, judging by https://github.com/grpc-ecosystem/grpc-gateway/pull/527. Sorry for the confusion.

johanbrandhorst (commented):
I changed my mind, it would be a breaking change to make this error, so lets just leave it as it was, sorry for the confusion.

examples/internal/proto/examplepb/a_bit_of_everything.swagger.json
```diff
@@ -1906,7 +1906,202 @@
             "in": "body",
             "required": true,
             "schema": {
-              "$ref": "#/definitions/examplepbABitOfEverything"
+              "type": "object",
+              "example": {
+                "uuid": "0cf361e1-4b44-483d-a159-54dabdf7e814"
+              },
```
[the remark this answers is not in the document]
stijndehaes: I changed the example to include `int64_value` and `double_value`

examples/internal/proto/examplepb/a_bit_of_everything.proto
stijndehaes: No problem, I'll change it

johanbrandhorst (approved):
LGTM, however I'd like to wait for @jgiles to have a chance to give consent to this being merged since it contains their original work.

stijndehaes:
@jgiles A friendly reminder to give your consent to merge this as I based this work on your original PR :)

jgiles:
I consent to merging the code. Thank you for picking this up, and apologies for abandoning the original PR :-)

jgiles:
@stijndehaes thank you!

johanbrandhorst:
Thanks for your contributions, @stijndehaes and @jgiles :)!

stijndehaes:
@jgiles without your original code I could have never landed this so easily so thank you for starting and giving your consent ;)

PR #2085, opened 2021-04-13
fix(trailers): add trailer header before Write/WriteHeader

j2gg0s:
Thanks for this great project.

We should add header before http Write/WriteHeader

#### Other comments
Do we need other pr for v2? I current use v2.3.0.


johanbrandhorst:
HI @j2gg0s, thanks for your pull request. I'm confused, I think we have existing tests that should cover this functionality. Do you have a failing test that this PR fixes? Otherwise I'm not sure it's the right fix.

j2gg0s:
@joelclouddistrict it's my wrong, trailer is different from header. No problem.

PR #2101, opened 2021-04-21
Handle custom status code mappings for routing errors

electrofelix:

https://gophers.slack.com/archives/CBATURP1D/p1618936187093600

yes

Not all gRPC error codes and http status codes can have a one to one
mapping. Http can have both a path and a method component, while gRPC
will just have a function that is specific to the combination of method
and path when translating between the two. That means that for gRPC a
missing function simply means "unimplemented", while for the equivalent
REST API it is common to distinguish an unhandled method request on a
known resource as "Method Not Allowed" with the response code 405.

This may not be the only place where expectations of response codes
needed may not align with what is provided by the current mapping
chain http.Status* -> gRPC code.Codes -> http.Status*, where some
information may be lost in the double conversion.

As the error handler accepts an 'error' interface, it should be possible
to provide a custom error structure that can contain the http Status
code to wrap the original error to allow for the expected response code
to be provided without impacting the current function signatures.

#### Other comments

This is intended initially to help have a conversation over whether
it is possible to solve in a reasonable fashion without requiring
copying of the default error handler and helper functions into
projects to be able to customize the status code response.

### DIFF
--- a/docs/docs/mapping/customizing_your_gateway.md
+++ b/docs/docs/mapping/customizing_your_gateway.md
@@ -358,3 +358,34 @@ HTTP statuses and their mappings to gRPC statuses:
 - HTTP `400 Bad Request` -> gRPC `3 INVALID_ARGUMENT`
 
 This method is not used outside of the initial routing.
+
+### Customizing Routing Errors
+
+If you want to retain HTTP `405 Method Not Allowed` instead of allowing it to be converted to the equivalent of the gRPC `12 UNIMPLEMENTED`, which is  HTTP `501 Not Implmented` you can use the following example:
+
+```go
+func handleRoutingError(ctx context.Context, mux *ServeMux, marshaler Marshaler, w http.ResponseWriter, r *http.Request, httpStatus int) {
+	var stderr error
+
+	switch httpStatus {
+	case http.StatusMethodNotAllowed:
+		sterr = HTTPStatusError{
+			HTTPStatus: httpStatus
+			Err:        status.Error(codes.Unimplemented, http.StatusText(httpStatus))
+		}
+	default:
+		runtime.DefaultRoutingErrorHandler(ctx, mux, marshaler, writer, request, httpStatus)
+
+		return
+	}
+
+	runtime.DefaultHTTPErrorHandler(ctx, mux, marshaler, w , r, stderr)
+}
+```
+
+To use this routing error handler, construct the mux as follows:
+```go
+mux := runtime.NewServeMux(
+	runtime.WithRoutingErrorHandler(handleRoutingError),
+)
+```
--- a/docs/docs/mapping/customizing_your_gateway.md
+++ b/docs/docs/mapping/customizing_your_gateway.md
@@ -367,7 +367,6 @@ If you want to retain HTTP `405 Method Not Allowed` instead of allowing it to be
 func handleRoutingError(ctx context.Context, mux *ServeMux, marshaler Marshaler, w http.ResponseWriter, r *http.Request, httpStatus int) {
 	if httpStatus != http.StatusMethodNotAllowed {
 		runtime.DefaultRoutingErrorHandler(ctx, mux, marshaler, writer, request, httpStatus)
-
 		return
 	}
 
--- a/docs/docs/mapping/customizing_your_gateway.md
+++ b/docs/docs/mapping/customizing_your_gateway.md
@@ -365,21 +365,19 @@ If you want to retain HTTP `405 Method Not Allowed` instead of allowing it to be
 
 ```go
 func handleRoutingError(ctx context.Context, mux *ServeMux, marshaler Marshaler, w http.ResponseWriter, r *http.Request, httpStatus int) {
-	var stderr error
-
-	switch httpStatus {
-	case http.StatusMethodNotAllowed:
-		sterr = HTTPStatusError{
-			HTTPStatus: httpStatus
-			Err:        status.Error(codes.Unimplemented, http.StatusText(httpStatus))
-		}
-	default:
+	if httpStatus != http.StatusMethodNotAllowed {
 		runtime.DefaultRoutingErrorHandler(ctx, mux, marshaler, writer, request, httpStatus)
 
 		return
 	}
 
-	runtime.DefaultHTTPErrorHandler(ctx, mux, marshaler, w , r, stderr)
+	// Use HTTPStatusError to customize the DefaultHTTPErrorHandler status code
+	err := &HTTPStatusError{
+		HTTPStatus: httpStatus
+		Err:        status.Error(codes.Unimplemented, http.StatusText(httpStatus))
+	}
+
+	runtime.DefaultHTTPErrorHandler(ctx, mux, marshaler, w , r, err)
 }
 ```
 
--- a/examples/internal/integration/integration_test.go
+++ b/examples/internal/integration/integration_test.go
@@ -1528,7 +1528,7 @@ func TestNotImplemented(t *testing.T) {
 		return
 	}
 
-	if got, want := resp.StatusCode, http.StatusMethodNotAllowed; got != want {
+	if got, want := resp.StatusCode, http.StatusNotImplemented; got != want {
 		t.Errorf("resp.StatusCode = %d; want %d", got, want)
 		t.Logf("%s", buf)
 	}
--- a/examples/internal/integration/integration_test.go
+++ b/examples/internal/integration/integration_test.go
@@ -1528,7 +1528,7 @@ func TestNotImplemented(t *testing.T) {
 		return
 	}
 
-	if got, want := resp.StatusCode, http.StatusNotImplemented; got != want {
+	if got, want := resp.StatusCode, http.StatusMethodNotAllowed; got != want {
 		t.Errorf("resp.StatusCode = %d; want %d", got, want)
 		t.Logf("%s", buf)
 	}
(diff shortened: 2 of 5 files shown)


examples/internal/integration/integration_test.go
```diff
@@ -1528,7 +1528,7 @@ func TestNotImplemented(t *testing.T) {
 		return
 	}
 
-	if got, want := resp.StatusCode, http.StatusNotImplemented; got != want {
+	if got, want := resp.StatusCode, http.StatusMethodNotAllowed; got != want {
```
johanbrandhorst, line 1531: The fact that you needed to change this tells us that this is a potentially breaking change in behaviour. I think we need to make this behaviour configurable, maybe we only expose the error and let the user create a custom routing error handler?

johanbrandhorst (changes requested):
I don't think we can merge this as is as it's a change of behaviour. I like that we can let the user create a custom error that we can have special handling for though. What do you think?

electrofelix (commented):
Thanks, I'll look to update to be more reflective of the final desired behaviour.

runtime/errors.go
```diff
@@ -20,6 +21,17 @@ type StreamErrorHandlerFunc func(context.Context, error) *status.Status
 // RoutingErrorHandlerFunc is the signature used to configure error handling for routing errors.
 type RoutingErrorHandlerFunc func(context.Context, *ServeMux, Marshaler, http.ResponseWriter, *http.Request, int)
 
+// StatusHTTPError is the error to use when needing to provide a different status code for an error
+// passed to the DefaultRoutingErrorHandler
+type StatusHTTPError struct {
+	Status int
```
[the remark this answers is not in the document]
electrofelix: Done, I think, did you just mean the leading part to keep the trailing `Error` or rename the full, I realized after updating that it could mean either?

runtime/errors.go
```diff
@@ -79,6 +91,12 @@ func DefaultHTTPErrorHandler(ctx context.Context, mux *ServeMux, marshaler Marsh
 	// return Internal when Marshal failed
 	const fallback = `{"code": 13, "message": "failed to marshal error message"}`
 
+	var customStatus *StatusHTTPError
+
+	if errors.As(err, &customStatus) {
+		err = customStatus.Err
+	}
```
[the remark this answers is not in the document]
electrofelix: Done, I think

runtime/errors.go
```diff
@@ -20,6 +21,17 @@ type StreamErrorHandlerFunc func(context.Context, error) *status.Status
 // RoutingErrorHandlerFunc is the signature used to configure error handling for routing errors.
 type RoutingErrorHandlerFunc func(context.Context, *ServeMux, Marshaler, http.ResponseWriter, *http.Request, int)
 
+// StatusHTTPError is the error to use when needing to provide a different status code for an error
```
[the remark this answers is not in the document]
electrofelix: Should be fixed now.

runtime/errors.go
```diff
@@ -20,6 +21,17 @@ type StreamErrorHandlerFunc func(context.Context, error) *status.Status
...
+// passed to the DefaultRoutingErrorHandler
```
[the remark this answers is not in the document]
electrofelix: Should be fixed now.

electrofelix:
So I think this now contains the correct behaviour, but I'm a little unsure about documentation, is it update https://github.com/grpc-ecosystem/grpc-gateway/blob/master/docs/docs/development/grpc-gateway_v2_migration_guide.md to explain about the HTTPStatus type for handling the edge cases?

Is there somewhere that I should be putting an example of providing a custom routing error handler function?

johanbrandhorst:
> So I think this now contains the correct behaviour, but I'm a little unsure about documentation, is it update https://github.com/grpc-ecosystem/grpc-gateway/blob/master/docs/docs/development/grpc-gateway_v2_migration_guide.md to explain about the HTTPStatus type for handling the edge cases?
> 
> Is there somewhere that I should be putting an example of providing a custom routing error handler function?

Good idea, I don't think we need to update the migration guide, but it would be cool to add an example of this new feature to https://github.com/grpc-ecosystem/grpc-gateway/blob/master/docs/docs/mapping/customizing_your_gateway.md?

runtime/errors.go
```diff
@@ -72,13 +84,23 @@ func HTTPError(ctx context.Context, mux *ServeMux, marshaler Marshaler, w http.R
 
 // DefaultHTTPErrorHandler is the default error handler.
 // If "err" is a gRPC Status, the function replies with the status code mapped by HTTPStatusFromCode.
+// If "err" is a HTTPStatus, the function replies with the status code provide by that struct. This is
+// intended to allow passing through of specific statuses via the function set via WithRoutingErrorHandler
+// for the ServeMux constructor to handle edge cases which the standard mappings in HTTPStatusFromCode
+// are insufficient.
```
johanbrandhorst, line 90: Had to read this sentence three times, but I think we accidentally a word here.
```suggestion
// are insufficient for.
```

runtime/errors.go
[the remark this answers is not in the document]
johanbrandhorst: Haha, sorry for the confusion, I _did_ mean the field, and I do think we should keep the `Error` suffix in this type. How about we call the type `HTTPError` and call the field `HTTPStatus`?

runtime/errors.go
```diff
@@ -72,13 +84,23 @@ func HTTPError(ctx context.Context, mux *ServeMux, marshaler Marshaler, w http.R
...
 // If otherwise, it replies with http.StatusInternalServerError.
 //
 // The response body written by this function is a Status message marshaled by the Marshaler.
 func DefaultHTTPErrorHandler(ctx context.Context, mux *ServeMux, marshaler Marshaler, w http.ResponseWriter, r *http.Request, err error) {
 	// return Internal when Marshal failed
 	const fallback = `{"code": 13, "message": "failed to marshal error message"}`
 
+	var customStatus *HTTPStatus
+
```
johanbrandhorst, line 99: Remove this newline please, these lines belong together

johanbrandhorst (commented):
Almost there, and lets add that docs example too!

runtime/errors.go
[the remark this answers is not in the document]
electrofelix: As there is a function in the packa called `HTTPError`, it appears that causes confusion for the compiler as how to distinguish between them. Switched it to HTTPStatusError for now.

runtime/errors.go
electrofelix: Habit due to some linters

docs/docs/mapping/customizing_your_gateway.md
```diff
@@ -358,3 +358,34 @@ HTTP statuses and their mappings to gRPC statuses:
...
+			HTTPStatus: httpStatus
+			Err:        status.Error(codes.Unimplemented, http.StatusText(httpStatus))
+		}
+	default:
+		runtime.DefaultRoutingErrorHandler(ctx, mux, marshaler, writer, request, httpStatus)
+
+		return
+	}
+
+	runtime.DefaultHTTPErrorHandler(ctx, mux, marshaler, w , r, stderr)
```
johanbrandhorst, line 382: This might just be me but this seems unnecessarily elaborate. How about this:
```suggestion
	if httpStatus != http.StatusMethodNotAllowed {
		runtime.DefaultRoutingErrorHandler(ctx, mux, marshaler, writer, request, httpStatus)
	}
	
	// Use HTTPStatusError to customize the DefaultHTTPErrorHandler status code
	err := &HTTPStatusError{
		HTTPStatus: httpStatus
		Err:        status.Error(codes.Unimplemented, http.StatusText(httpStatus))
	}
	runtime.DefaultHTTPErrorHandler(ctx, mux, marshaler, w , r, err)
```

johanbrandhorst (changes requested):
Thanks for adding the docs - I think we can simplify the example a bit.

docs/docs/mapping/customizing_your_gateway.md
electrofelix: I was assuming they might want to handle additional statuses, but a straightforward example should be sufficient for anyone to convert to handling multiple.

docs/docs/mapping/customizing_your_gateway.md
```diff
@@ -358,3 +358,32 @@ HTTP statuses and their mappings to gRPC statuses:
...
+
+### Customizing Routing Errors
+
+If you want to retain HTTP `405 Method Not Allowed` instead of allowing it to be converted to the equivalent of the gRPC `12 UNIMPLEMENTED`, which is  HTTP `501 Not Implmented` you can use the following example:
+
+```go
+func handleRoutingError(ctx context.Context, mux *ServeMux, marshaler Marshaler, w http.ResponseWriter, r *http.Request, httpStatus int) {
+	if httpStatus != http.StatusMethodNotAllowed {
+		runtime.DefaultRoutingErrorHandler(ctx, mux, marshaler, writer, request, httpStatus)
+
```
johanbrandhorst, line 370: Rogue newline 😂.

electrofelix: Linter habits, https://github.com/ssgreg/nlreturn is enabled for local projects and it's become part of my muscle memory now... ;-)

johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution!

PR #2103, opened 2021-04-21
Patcher for fields of type google.protobuf.Any

veith:
Add google.protobuf.Any type handling for the "Patch feature"

#### Other comments
The returned fieldmask is the key of the field, not the subfields. This is because a field of the type google.protobuf.Any will not work without the type_url and is useless without the value.


runtime/fieldmask_test.go
```diff
@@ -159,6 +159,13 @@ func TestFieldMaskFromRequestBody(t *testing.T) {
 				"nested",
 			),
 		},
+
+		{
+			name:     "protobuf-any",
+			msg:      &examplepb.ABitOfEverything{},
+			input:    `{"anytype":{"@type": "xx.xx/examplepb.NestedOuter", "one":{"two":{"three":{"a":true, "b":false}}}}}`,
+			expected: newFieldMask("anytype"), //going deeper makes no sense
```
johanbrandhorst, line 167: Is this even useful? What was the previous behaviour?

veith: The previous behaviour was a code 3 with the message that the field `@type` does not exist in the type `google.protobuf.Any`. It was not possible to patch a field of type `google.protobuf.Any` .

Maybe i should add an error for the case that `@type` was not set in the request, because this field is "required" to unmarshal.

johanbrandhorst:
@googlebot I consent.

johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution!

PR #2112, opened 2021-04-28
unmarshal: avoid dereferencing nil pointers

torkelrogstad:
N/A

Yes

Avoid dereferencing nil points when unmarshaling JSON. This might be the result of me doing something strange, but I'm using a custom JSON marshaler with gRPC-Gateway, where some fields (for example enums) end up as `null` in JSON if they are the zero value. When I then do a roundtrip JSON marshal/unmarshal the unmarshaler panics on a nil `*json.RawMessage`.

#### Other comments

I'm not sure if this needs a unit test, and in that case what that should look like. Any pointers are appreciated!

### DIFF
--- a/runtime/marshal_jsonpb.go
+++ b/runtime/marshal_jsonpb.go
@@ -237,6 +237,10 @@ func decodeNonProtoField(d *json.Decoder, unmarshaler protojson.UnmarshalOptions
 			}
 			bk := result[0]
 			bv := reflect.New(rv.Type().Elem())
+			if v == nil {
+				null := json.RawMessage("null")
+				v = &null
+			}
 			if err := unmarshalJSONPb([]byte(*v), unmarshaler, bv.Interface()); err != nil {
 				return err
 			}
--- a/runtime/marshal_jsonpb_test.go
+++ b/runtime/marshal_jsonpb_test.go
@@ -648,6 +648,24 @@ var (
 	}
 )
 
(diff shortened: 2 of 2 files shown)


johanbrandhorst:
Hi Torkel, thanks for your PR! This looks reasonable, but we definitely should try and squeeze in a test that will exercise this. Could you have a look around https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/marshal_json_test.go and see if one of those can be adapted for your new case? It may even mean creating a new marshaler with some custom behaviour, but since you managed to hit this in your own code it should be possible to reproduce as a test too.

Also, I think you need to rebase on latest `master` for CI to pass.

torkelrogstad:
After looking more at this it turned out that it wasn't even that hard to reproduce this, we only needed to try and use `runtime.JSONPb` to unmarshal a JSON blob with `null` as one of it's keys into `map[string]interface{}`. Added a test, please let me know if there's anything else I should do.

torkelrogstad:
I tried rebasing on master, but still seeing the error from `gorelease`...

johanbrandhorst:
> I tried rebasing on master, but still seeing the error from `gorelease`...

Hm, that's unfortunate, I'll have to have a look at why that's happening 🤔. I'm happy to merge this as is.

johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution!

torkelrogstad:
Thanks for merging!

Issue #2115, opened 2021-04-29
Custom Inbound Marshaler not being used for transcoding


0xKiwi:
## 🐛 Bug Report
After setting a custom marshaler for encoding/decoding within the grpc-gateway, I've noticed the inbound marshaler is not used at all within the compiled gateway files. Not sure if I'm missing something or if this is an oversight.

The inbound marshaler is not used at all for `pathParams` here: https://github.com/prysmaticlabs/ethereumapis/blob/gen-go-cast/eth/v1/beacon_chain_service.pb.gw.go#L57

Yet it is used as part of `runtime.ForwardResponseMessage` here: https://github.com/prysmaticlabs/ethereumapis/blob/gen-go-cast/eth/v1/beacon_chain_service.pb.gw.go#L2289

It is also unused for endpoints that take in a protobuf message struct here: https://github.com/prysmaticlabs/ethereumapis/blob/gen-go-cast/eth/v1/beacon_chain_service.pb.gw.go#L861

But also used for the return.

## To Reproduce
1. Set custom marshaler using gwmux.
2. Query API with a message that depends on marshaler.
3. Query will fail because the input is not being marshaled, only the output.

## Expected behavior
One would expect the marshaler to be used, with the input/output both being transcoded through the assigned marshaler.

## Actual Behavior
The marshaler is only used for the outbound messages, not the inbound, see here:
The marshaler (which changes base64 to hex) output here:
```
 kiwi@kiwi   ~/Work/prysm     change-to-gocast    15  curl -H 'Accept: application/json' localhost:3500/eth/v1/beacon/genesis
{
"data": {
"genesis_fork_version": "0x00002009",
"genesis_time": "2020-11-18T12:00:07Z",
"genesis_validators_root": "0x9436e8a630e3162b7ed4f449b12b8a5a368a4b95bc46b941ae65c11613bfa4c1"
}
}%   
```

Yet, when I try passing a field body or struct to a GET:
```
curl -H 'Accept: application/json' localhost:3500/eth/v1/beacon/states/0x0780527577054a5323d530d8c4216efdd28e9ffed39bc309214c6d75702515a1/root
> {"code":3,"details":[],"message":"type mismatch, parameter: state_id, error: illegal base64 data at input byte 64"}%  
```

## Your Environment

Ubuntu Linux, using a fork of grpc-gateway from latest master.

johanbrandhorst:
[...]

PR #2117, opened 2021-04-29
[WIP] Add implement custom parsing for path parameters

0xKiwi:
Discussion for this change began in #2115 

Yes.

We use protobufs and grpc-gateway for our APIs at Prysmatic Labs, and required hex encoded strings on all inputs/ouputs from the Ethereum Foundation.
After a very long battle of trying to get hex encoded strings working instead of base 64, I am very happy to contribute to grpc-gateway to save the trouble for others. 

This PR is currently incomplete, and very crude. I would appreciate help where commented. All the go-template is confusing and difficult to translate. I mainly just copied the QueryParser and adapted it for `map[string]string`.


protoc-gen-grpc-gateway/internal/gengateway/template.go
```diff
@@ -237,6 +238,9 @@ var _ status.Status
 var _ = runtime.String
 var _ = utilities.NewDoubleArray
 var _ = metadata.Join
+var _ = github_com_prysmaticlabs_eth2_types.Epoch(0)
+var _ = emptypb.Empty{}
+var _ = empty.Empty{}
```
0xKiwi: I added these to prevent errors on compilation. I'm not sure why but sometimes the compiled output imports 1 name of emptypb, but uses another different (unimported) name.

protoc-gen-grpc-gateway/internal/gengateway/template.go
```diff
@@ -348,30 +352,15 @@ var (
...
-	e{{if $param.IsRepeated}}s{{end}}, err = {{$param.ConvertFuncExpr}}(val{{if $param.IsRepeated}}, {{$binding.Registry.GetRepeatedPathParamSeparator | printf "%c" | printf "%q"}}{{end}}, {{$enum.GoType $param.Method.Service.File.GoPkg.Path}}_value)
-	if err != nil {
-		return nil, metadata, status.Errorf(codes.InvalidArgument, "type mismatch, parameter: %s, error: %v", {{$param | printf "%q"}}, err)
-	}
+	_, _ = val, ok
 {{else}}
-	{{$param.AssignableExpr "protoReq"}}, err = {{$param.ConvertFuncExpr}}(val{{if $param.IsRepeated}}, {{$binding.Registry.GetRepeatedPathParamSeparator | printf "%c" | printf "%q"}}{{end}})
-	if err != nil {
-		return nil, metadata, status.Errorf(codes.InvalidArgument, "type mismatch, parameter: %s, error: %v", {{$param | printf "%q"}}, err)
+	if err := runtime.PopulatePathParameters(&protoReq, pathParams, &utilities.DoubleArray{}); err != nil {
```
0xKiwi, line 360: Need help here. How much of this is still needed? The go-template is pretty complex so would appreciate any guidance on properly adding this branch.

runtime/query.go
```diff
@@ -51,7 +51,7 @@ func (*defaultQueryParser) Parse(msg proto.Message, values url.Values, filter *u
 		if filter.HasCommonPrefix(fieldPath) {
 			continue
 		}
-		if err := populateFieldValueFromPath(msg.ProtoReflect(), fieldPath, values); err != nil {
+		if err := populateFieldValueFromFieldPath(msg.ProtoReflect(), fieldPath, values); err != nil {
```
0xKiwi, line 54: Renamed to provide more context and prevent conflcit.

runtime/path.go
```diff
@@ -0,0 +1,329 @@
+package runtime
```
johanbrandhorst, line 1: Is this just a copy of the `query.go` file?

protoc-gen-grpc-gateway/internal/gengateway/template.go
johanbrandhorst: I would assume this is all still going to be required, I think you'll need an understanding of what each case does before we can consider replacing it. Remember that we want to have two way backward compatibility as much as possible, a user should be able to upgrade either the runtime or the generator without things breaking, ideally. If we're adding a new `runtime.ParsePathParameters` then it means new generated files can't keep using old versions of the library, which isn't great, but may be unavoidable for a feature like this.

The way to probably make this work is to copy whatever logic we have for filtering query parameters into `*utilities.DoubleArray` and reuse it for figuring out what parameters are path parameters, then use that as a parameter to `ParsePathParameters`, much like we do for query parameters today.

runtime/path.go
0xKiwi: Yes, pretty much, I just modified it to accept `map[string]string` instead of `http.Request`. Is that fine? lol

protoc-gen-grpc-gateway/internal/gengateway/template.go
0xKiwi: Got it. Can you provide some hints on what the cases here mean? Theres a lot of general terms like `IsNestedProto3` used in the other if statements.

johanbrandhorst: I'm frankly not all that familiar with the deeper workings of this template either - I'd recommend diving into where it's [being instantiated and executed](https://github.com/grpc-ecosystem/grpc-gateway/blob/74ecd1deffacf97bcbee90e81c631ef6a3c275f2/protoc-gen-grpc-gateway/internal/gengateway/template.go#L151-L214) to understand the parameters involved. I think a useful initial test would be to just add a flag that can replace the path parameter parsing with a call to runtime.ParsePathParameter.

https://github.com/grpc-ecosystem/grpc-gateway/blob/master/protoc-gen-grpc-gateway/internal/gengateway/template.go#L131-L143 is where the query parameter doubleArray is created (I don't know how that works tbh). We'll need somethin similar for paths.

runtime/path.go
johanbrandhorst: Sure, but we don't necessarily want to parse all the same values as we do in query parameters - for example maps and repeated types cannot be set in path parameters.

0xKiwi: I think I disabled those already, cool, will make sure, thanks!

protoc-gen-grpc-gateway/internal/gengateway/template.go
0xKiwi: Got it. I'm not sure if we need the filter so I've removed it. Studying all the code branches now.

runtime/path.go
0xKiwi: It seems repeated enums can be set yes? I'll add those into the parser.

runtime/path.go
```diff
@@ -0,0 +1,294 @@
...
+	if err != nil {
+		return fmt.Errorf("parsing field %q: %w", fieldDescriptor.FullName().Name(), err)
+	}
+
+	msgValue.Set(fieldDescriptor, v)
+	return nil
+}
+
+func populateRepeatedPathField(fieldDescriptor protoreflect.FieldDescriptor, list protoreflect.List, value string) error {
+	for _, value := range strings.Split(value, ",") {
```
0xKiwi, line 122: This is how repeated enum params were handled, so mirroring that here.

0xKiwi:
Hello @johanbrandhorst , I've made some changes to what I think is suitable for this feature. Please look them over and let me know. I studied each branch of the templating and figured this could easily replace the existing code branches minus passing the query message into the query parser.

0xKiwi:
Also, I'm not entirely sure whats failing, mind providing some guidance? All I see is git diff --exit-code failing.

johanbrandhorst:
> Also, I'm not entirely sure whats failing, mind providing some guidance? All I see is git diff --exit-code failing.

This is likely because you need to regenerate the files after changing the templates: https://github.com/grpc-ecosystem/grpc-gateway/blob/master/CONTRIBUTING.md#i-want-to-regenerate-the-files-after-making-changes.

johanbrandhorst (changes requested):
Ivan, I want to be realistic about our expectations here - I think best case this will come in as a generation time flag, because it will otherwise break a lot of our users. As it looks right now you've still remove a lot of code that is doing to remain essential for the old workflow. I think there is still a significant amount of work to do to figure out how feasible this is at all and I don't want you to waste your time if you can't commit to it, which I completely understand. What do you think?

0xKiwi:
Is the flag the only requirement needed from this? I've mirrored it towards the query parser in expectation that that would be fine, is there more besides breaking changes that you're concerned about? What else is needed? Would appreciate more guidance as it appears I'm lacking context. Thanks.

runtime/path.go
0xKiwi: Care to provide additional information? I don't know what you're referring to. The original grpc-gateway template code supports repeated enums as a path parameter from what I could tell.

0xKiwi:
I've gone down the rabbit hole of grpc and protobuf long enough and I would eagerly like to make a contribution of a feature that my team requires. I don't know what you're referring to about how "feasible" this feature is or how much I need to "commit to it". I would really appreciate more insight on what you mean and how this PR could be improved.

protoc-gen-grpc-gateway/internal/gengateway/template.go
```diff
@@ -505,54 +476,26 @@ func local_request_{{.Method.Service.GetName}}_{{.Method.GetName}}_{{.Index}}(ct
...
-{{end}}
-
-{{if and $enum $param.IsRepeated}}
-	s := make([]{{$enum.GoType $param.Method.Service.File.GoPkg.Path}}, len(es))
-	for i, v := range es {
-		s[i] = {{$enum.GoType $param.Method.Service.File.GoPkg.Path}}(v)
-	}
-	{{$param.AssignableExpr "protoReq"}} = s
-{{else if $enum}}
-	{{$param.AssignableExpr "protoReq"}} = {{$enum.GoType $param.Method.Service.File.GoPkg.Path}}(e)
```
johanbrandhorst, line 555: Undo all the removals in this code. Anything you remove here is going to break existing users who aren't opting into your new feature.

runtime/path.go
johanbrandhorst: I mean, I just don't understand how one could add repeated enums into a path. Do you know? What would it look like?

johanbrandhorst (changes requested):
Since it wasn't clear from my original comment, I've marked the code you need to add back. You'll need to add a generation time parameter to enable the use of your new feature. See https://github.com/grpc-ecosystem/grpc-gateway/blob/5ed4a057c46146f7c8160dcb073db09b6e926b77/protoc-gen-grpc-gateway/main.go#L31 for an example of this.

I still haven't seen any work started on the whole "figure out what part of the request message is being used by the path". That's currently done for query parameters, and used in the query parameter parser, but you'll need to add logic for doing that parsing and conditionally add that to the generated files too.

What I wanted to say with my original comment was that, though it appears simple, this is not a simple copy paste from the query parameters job.

0xKiwi:
Got it, thank you for being concise. I will look into making the changes you have mentioned. Thank you for providing an example of the generation time flag.

runtime/path.go
0xKiwi: I could be missing context but I see support for it here: https://github.com/grpc-ecosystem/grpc-gateway/blob/master/protoc-gen-grpc-gateway/internal/gengateway/template.go#L367

Maybe this is for general params instead of pathParams. Its not helpful that the source of all values here is coming from a map named `pathParams` so maybe I misunderstood how this is used.

0xKiwi:
Hello @johanbrandhorst, thank you for your assistance on this. I will close this PR for the time being and open another one when I'm able to complete this feature.

PR #2121, opened 2021-05-03
Move to temporary directory to fetch tools

### LINKED ISSUE #2120: Dependencies introduces CVE false positives
## 🐛 Bug Report

When importing `github.com/grpc-ecosystem/grpc-gateway/v2` a project will fail go nancy vulnerability scans.

## To Reproduce

Run the following in grpc-gateway or a project that imports it:

```bash
$ go list -json -m all | docker run --rm -i sonatypecommunity/nancy:latest sleuth
```

## Expected behavior

No vulnerabilities.

## Actual Behavior

```bash
$ go list -json -m all | docker run --rm -i sonatypecommunity/nancy:latest sleuth |grep CVE |cut -c 1-100
┃ [CVE-2020-15114] In etcd before versions 3.3.23 and 3.4.10, the etcd gateway is a simple TC
┃ [CVE-2020-15136] In ectd before versions 3.4.10 and 3.3.23, gateway TLS authentication is o
┃ [CVE-2020-15115] etcd before versions 3.3.23 and 3.4.10 does not perform any password lengt
┃ [CVE-2020-26160] jwt-go before 4.0.0-preview1 allows attackers to bypass intended access re
┃ [CVE-2021-3121] An issue was discovered in GoGo Protobuf before 1.3.2. plugin/unmarshal/u
```

## Your Environment

OS: Ubuntu 21.04
go version go1.16.3 linux/amd64
github.com/grpc-ecosystem/grpc-gateway/v2@v2.4.0

## Suspected Cause

`tools.go` includes `_ "github.com/bufbuild/buf/cmd/buf"`,  
`_ "github.com/bufbuild/buf/cmd/buf"` includes `https://github.com/spf13/viper` that has been slow releasing updates for security.

This code probably won't get used as it will shake out the tree at build time, but causes headaches for CI and security scans. Is there a more elegant way to handle things like `tools.go`?

physcat:
fixes #2120 

    Move to tmp directory to fetch tools
    
The tools.go file can cause fake dependencies in go.sum.
This will move to a temporary directory and fetch the tools so that it does not add to the go.mod and go.sum.

## Result
```bash
$ go list -json -m all | docker run --rm -i sonatypecommunity/nancy:latest sleuth 
Checking for updates...
Already up-to-date.
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┓
┃ Summary                      ┃
┣━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━┫
┃ Audited Dependencies    ┃ 75 ┃
┣━━━━━━━━━━━━━━━━━━━━━━━━━╋━━━━┫
┃ Vulnerable Dependencies ┃ 0  ┃
┗━━━━━━━━━━━━━━━━━━━━━━━━━┻━━━━┛
```


johanbrandhorst:
Hi Mitch,

This is not the solution I think we should take. We do not want to create a separate module within this module as it could confuse users. I would sooner that `make install` jumped into a temporary directory and installed versions before jumping back.

physcat:
@johanbrandhorst thanks for the reviews. What do you think of this approach?

johanbrandhorst:
> @johanbrandhorst thanks for the reviews. What do you think of this approach?

This is much better, thanks for changing it.

Makefile
```diff
@@ -81,11 +81,18 @@ $(GENERATE_UNBOUND_METHODS_EXAMPLE_SRCS): $(GENERATE_UNBOUND_METHODS_EXAMPLE_SPE
...
+TMP_INSTALL_DIR := $(shell mktemp -d)
 install:
+
+install:
+	@mkdir -p ${TMP_INSTALL_DIR}
+	cd ${TMP_INSTALL_DIR} && go get \
+		google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.0.0 \
+		google.golang.org/protobuf/cmd/protoc-gen-go@v1.26.0 \
+		github.com/bufbuild/buf/cmd/buf@v0.41.0
+	@rmdir ${TMP_INSTALL_DIR}
```
johanbrandhorst, line 93: Does this command still run if the previous one failed?

physcat: If the `mkdir` fails then the make process will exit with an error code. So the `rmdir` won't try to run. Actually if any of the 3 commands fail it will terminate the make process with an error code.

johanbrandhorst: That's what I mean, can we "trap" the rmdir so that it doesn't stick around if the installation fails?

physcat: Ah but the temp dir will be something like this `/tmp/tmp.4PHFyK4afw` so if it does not get cleaned up it is not too bad. Most systems will clean that up (eventually). Not sure if it is worth adding the complexity. Do you have an elegant suggestion?

johanbrandhorst: Nah it's not the end of the world, you're right.

johanbrandhorst (changes requested):
Please see steps to run to regenerate files: https://github.com/grpc-ecosystem/grpc-gateway/blob/master/CONTRIBUTING.md#i-want-to-regenerate-the-files-after-making-changes.

I think you may also need a `go mod tidy`.

physcat:
> Please see steps to run to regenerate files: https://github.com/grpc-ecosystem/grpc-gateway/blob/master/CONTRIBUTING.md#i-want-to-regenerate-the-files-after-making-changes.
> 
> I think you may also need a `go mod tidy`.

Done.

johanbrandhorst:
Sorry, looks like you need to rebase this on master

physcat:
> Sorry, looks like you need to rebase this on master

Obviously the renovate bot would sneak in while I'm not looking...
Done.

Makefile
```diff
@@ -81,11 +81,18 @@ $(GENERATE_UNBOUND_METHODS_EXAMPLE_SRCS): $(GENERATE_UNBOUND_METHODS_EXAMPLE_SPE
 	@rm -f $(EXAMPLE_CLIENT_DIR)/generateunboundmethods/README.md \
 		$(EXAMPLE_CLIENT_DIR)/generateunboundmethods/git_push.sh
 
+TMP_INSTALL_DIR := $(shell mktemp -d)
 install:
+
+install:
```
johanbrandhorst, line 87: Wait, one of these should go right?

johanbrandhorst (changes requested):
Surprised by the amount of changes here but

Makefile
physcat: Makefile fixed.

I've tried to check the dependencies but no obvious link to the large amount of changes, must be a dependency difference between building protoc-gen-go and protoc-gen-go-grpc with the local `go.mod` vs building it from their tree.

As a side effect I've also bumped the version of `google.golang.org/grpc/cmd/protoc-gen-go-grpc` v1.0.0 -> v1.1.0

johanbrandhorst:
Thanks for your contribution!

physcat:
> Thanks for your contribution!

Thanks for all the quick feedback!

PR #2134, opened 2021-05-07
Support new naming convention for go_library targets

tanyabouman:
Yes

By default, [Gazelle now uses](https://github.com/bazelbuild/bazel-gazelle/pull/863) the last component of the Bazel package, rather than `go_default_library` as the `go_library name`.

This causes problems when the naming conventions do not match.
https://github.com/bazelbuild/bazel-gazelle/issues/980
https://github.com/bazelbuild/bazel-gazelle/issues/890

Adding support for the new naming convention should not break anything that depends on using the old naming convention, because the Gazelle option `import_alias` makes aliases to the libraries and this allows the old naming convention to still work.


achew22:
Good fixup! Thanks for the change

PR #2180, opened 2021-06-09
fix(openapi): generate correct schema definition for google.protobuf.Any

### LINKED ISSUE #2177: protoc-gen-openapiv2 creates incorrect schema for google.protobuf.Any
## 🐛 Bug Report

I have a message containing a `google.protobuf.Any`, and the resultant `swagger.json` file contains a definition for `protobufAny` with a property `typeUrl` instead of `@type`.

If I use the swagger-ui to make a request following the schema, my grpc-gateway proxy ends up failing because the `Any` field is omitted from the request. If I manually change the request body so that `@type` is defined instead of `typeUrl`, everything works.

## To Reproduce

Create a message like so:

```proto
import "google/protobuf/any.proto";

message MyMessage {
  string id = 1;
  google.protobuf.Any config = 2;
}
```

Running `protoc-gen-openapiv2` using buf (following the instructions from the `README.md`), and I end up with a field in the `definitions` object of the swagger.json:

```json
{
  ...
  "protobufAny": {
      "type": "object",
      "properties": {
        "typeUrl": {
          "type": "string",
          "description": "A URL/resource name that uniquely identifies the type of the serialized\nprotocol buffer message. This string must contain at least\none \"/\" character. The last segment of the URL's path must represent\nthe fully qualified name of the type (as in\n`path/google.protobuf.Duration`). The name should be in a canonical form\n(e.g., leading \".\" is not accepted).\n\nIn practice, teams usually precompile into the binary all types that they\nexpect it to use in the context of Any. However, for URLs which use the\nscheme `http`, `https`, or no scheme, one can optionally set up a type\nserver that maps type URLs to message definitions as follows:\n\n* If no scheme is provided, `https` is assumed.\n* An HTTP GET on the URL must yield a [google.protobuf.Type][]\n  value in binary format, or produce an error.\n* Applications are allowed to cache lookup results based on the\n  URL, or have them precompiled into a binary to avoid any\n  lookup. Therefore, binary compatibility needs to be preserved\n  on changes to types. (Use versioned type names to manage\n  breaking changes.)\n\nNote: this functionality is not currently available in the official\nprotobuf release, and it is not used for type URLs beginning with\ntype.googleapis.com.\n\nSchemes other than `http`, `https` (or the empty scheme) might be\nused with implementation specific semantics."
        },
        "value": {
          "type": "string",
          "format": "byte",
          "description": "Must be a valid serialized protocol buffer of the above specified type."
        }
      },
      "description": "`Any` contains an arbitrary serialized protocol buffer message along with a\nURL that describes the type of the serialized message.\n\nProtobuf library provides support to pack/unpack Any values in the form\nof utility functions or additional generated methods of the Any type.\n\nExample 1: Pack and unpack a message in C++.\n\n    Foo foo = ...;\n    Any any;\n    any.PackFrom(foo);\n    ...\n    if (any.UnpackTo(\u0026foo)) {\n      ...\n    }\n\nExample 2: Pack and unpack a message in Java.\n\n    Foo foo = ...;\n    Any any = Any.pack(foo);\n    ...\n    if (any.is(Foo.class)) {\n      foo = any.unpack(Foo.class);\n    }\n\n Example 3: Pack and unpack a message in Python.\n\n    foo = Foo(...)\n    any = Any()\n    any.Pack(foo)\n    ...\n    if any.Is(Foo.DESCRIPTOR):\n      any.Unpack(foo)\n      ...\n\n Example 4: Pack and unpack a message in Go\n\n     foo := \u0026pb.Foo{...}\n     any, err := anypb.New(foo)\n     if err != nil {\n       ...\n     }\n     ...\n     foo := \u0026pb.Foo{}\n     if err := any.UnmarshalTo(foo); err != nil {\n       ...\n     }\n\nThe pack methods provided by protobuf library will by default use\n'type.googleapis.com/full.type.name' as the type URL and the unpack\nmethods only use the fully qualified type name after the last '/'\nin the type URL, for example \"foo.bar.com/x/y.z\" will yield type\nname \"y.z\".\n\n\nJSON\n====\nThe JSON representation of an `Any` value uses the regular\nrepresentation of the deserialized, embedded message, with an\nadditional field `@type` which contains the type URL. Example:\n\n    package google.profile;\n    message Person {\n      string first_name = 1;\n      string last_name = 2;\n    }\n\n    {\n      \"@type\": \"type.googleapis.com/google.profile.Person\",\n      \"firstName\": \u003cstring\u003e,\n      \"lastName\": \u003cstring\u003e\n    }\n\nIf the embedded message type is well-known and has a custom JSON\nrepresentation, that representation will be embedded adding a field\n`value` which holds the custom JSON in addition to the `@type`\nfield. Example (for message [google.protobuf.Duration][]):\n\n    {\n      \"@type\": \"type.googleapis.com/google.protobuf.Duration\",\n      \"value\": \"1.212s\"\n    }"
    }
...
}
```

## Expected behavior

The `typeUrl` field within `protobufAny` should be `@type`.

## Actual Behavior

The `@type` field is missing.

## Your Environment

Go 1.16
Windows 10
I'm not sure how to find out the version of protoc-gen-openapiv2 I'm using, but this is a new project today so I'm guessing up-to-date.

dan-j:

Fixes #2177, but still needs some tests.

See comments in #2177, but this essentially adds a new entry to the well-known types which is a ref to the type which will be generated by `customRefs`. We then do some edge-case handling at the end of `renderMessageAsDefinition()` when processing the `Any` type to make sure the field is `@type` and not `typeUrl/type_url`.


protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -525,6 +551,11 @@ func schemaOfField(f *descriptor.Field, reg *descriptor.Registry, refs refMap) o
...
 			core = wktSchema
 
+			// if wfkSchema is a reference to another known type, add it to refs to be processed later
+			if core.Ref != "" {
+				refs[fd.GetTypeName()] = struct{}{}
+			}
+
 			if fd.GetTypeName() == ".google.protobuf.Empty" {
 				props = &openapiSchemaObjectProperties{}
 			}
```
dan-j, line 561: This can be tidied up now, and `props` for `.google.protobuf.Empty` be defined on the `wktSchemas` map. Can address this later (I need to log off for a few hours)

protoc-gen-openapiv2/internal/genopenapi/types.go
```diff
@@ -157,7 +157,10 @@ type schemaCore struct {
 	Ref     string          `json:"$ref,omitempty"`
 	Example json.RawMessage `json:"example,omitempty"`
 
-	Items *openapiItemsObject `json:"items,omitempty"`
+	// Properties can be recursively defined
+	Properties *openapiSchemaObjectProperties `json:"properties,omitempty"`
```
dan-j, line 161: Moved this from the `openapiSchemaObject`

dan-j:
So I've just pushed another commit with a different approach which seems to do what it's supposed to. However now we have a problem that the generated go code from swagger-codegen doesn't actually match what it's meant to be. Will comment inline on what I mean

examples/internal/clients/generateunboundmethods/model_protobuf_any.go
```diff
@@ -10,6 +10,5 @@
 package generateunboundmethods
 
 type ProtobufAny struct {
-	TypeUrl string `json:"typeUrl,omitempty"`
-	Value string `json:"value,omitempty"`
+	Type_ string `json:"@type,omitempty"`
```
dan-j, line 13: `swagger-codegen` now generates go code which I wouldn't expect to work. Really I'd expect this type to not be generated, and any usages of `ProtobufAny` actually be of type `interface{}`, and say it's up to the caller to provide a value which marshals to a JSON object which is valid under the rules set out by the protojson package. 

I guess this would be a swagger-codegen problem, and one which most likely exists across most programming languages. Not sure how to approach this.

dan-j:
FYI, this isn't working yet. Just tried to use my locally built version on my project and there's an issue where swagger-ui won't allow additional properties to the `@type` field. Will update the MR by tomorrow

dan-j:
Ok, this seems to be working for now, but going to spend a couple more days finishing my tasks with work using my fork and if all goes well I'll add some tests and hopefully have this mergeable by the end of next week.

johanbrandhorst:
> Ok, this seems to be working for now, but going to spend a couple more days finishing my tasks with work using my fork and if all goes well I'll add some tests and hopefully have this mergeable by the end of next week.

Sounds great, thank you!

dan-j:
Apologies I've not had chance to pick this up. I was doing this for work and my personal life got in the way.

I'm on holiday now but will be able to finish this hopefully next week.

johanbrandhorst:
> Apologies I've not had chance to pick this up. I was doing this for work and my personal life got in the way.
> 
> I'm on holiday now but will be able to finish this hopefully next week.

No pressure buddy, I'm here to review whenever you need it :).

protoc-gen-openapiv2/internal/genopenapi/types.go
dan-j: I've just been familiarising myself with the changes I made before adding tests and I'm wondering whether or not this field should be moved back to `openapiSchemaObject`, or whether the change I've made is OK.

If we keep this change, we can also remove the workarounds on lines 1041 and 1108.

Thoughts?

dan-j (commented):
I've just opened this back up, rebased on top of the latest `master` and got familiar with my changes before adding tests.

I've added a comment to `types.go`, I'd like a response from somebody with more experience in this project.

I'm aiming to add some tests tomorrow.

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -76,7 +76,9 @@ var wktSchemas = map[string]schemaCore{
 	".google.protobuf.BoolValue": {
 		Type: "boolean",
 	},
-	".google.protobuf.Empty": {},
+	".google.protobuf.Empty": {
+		Properties: &openapiSchemaObjectProperties{},
+	},
```
johanbrandhorst, line 81: What is the consequence of this change?

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -526,8 +559,9 @@ func schemaOfField(f *descriptor.Field, reg *descriptor.Registry, refs refMap) o
 		if wktSchema, ok := wktSchemas[fd.GetTypeName()]; ok {
 			core = wktSchema
 
-			if fd.GetTypeName() == ".google.protobuf.Empty" {
-				props = &openapiSchemaObjectProperties{}
+			// if wfkSchema is a reference to another known type, add it to refs to be processed later
+			if core.Ref != "" {
+				refs[fd.GetTypeName()] = struct{}{}
```
johanbrandhorst, line 564: I'm not sure I understand this change, could you explain?

protoc-gen-openapiv2/internal/genopenapi/types.go
johanbrandhorst: I'm a little confused, could you elaborate on the reason for moving it? Does it make more sense for the openapiv2 output we're trying to get?

johanbrandhorst (commented):
This looks good, but I'd like some more background on some of the changes before we merge.

protoc-gen-openapiv2/internal/genopenapi/types.go
dan-j: So it makes the code easier to handle the edge-cases for the `.google.protobuf.Empty` type, at the minute there are if statements dotted around on lines 1041 and 1108 which could be removed if this change is kept.

However, as a more general point, I don't really understand why `schemaCore` and `openapiSchemaObject` are different types. The only usage of `schemaCore` in a type is in `openapiSchemaObject` and `openapiItemsObject`. As far as the JSON Schema spec goes, `"items"` is another schema (I'm not certain if this holds true for OpenAPI - I know there are a few differences).

If we wanted to keep things really simple, I'd do away with `schemaCore` completely, move the properties into `openapiSchemaObject` and refactor usages as necessary.

protoc-gen-openapiv2/internal/genopenapi/template.go
dan-j: I think I did this to remove the random if statements on lines 1041 and 1108, but forgot to actually remove them. However still happy to discuss what to do around moving the `Properties` field.

protoc-gen-openapiv2/internal/genopenapi/template.go
dan-j: I can't remember exactly, however I believe this was the previous behaviour:

`rpcStatus` (which references `protobufAny`) is hardcoded to always output in the `definitions` object, and `rpcStatus.details` is an array of `"$ref": "#/definitions/protobufAny"`.

If you define a type in your API with a `google.protobuf.Any` field, the generated schema outputted the schema as an explicit schema under that properties object, and not as a `$ref`. This meant if your schema referenced `Any` multiple times, your schema would define the schema for `Any` in every usage, growing your schema.

---

Because our `wktSchemas` maps now includes `.google.protobuf.Any` as a known-type with the ref field populated, any references to `Any` in a custom type will be declared using said `$ref`. However we need to make sure that the type is still outputted to `definitions`, so we add it to the `refs` map so that it is added during the call to `addCustomRefs`.

dan-j:
I've found a better way to fix this and added tests, going to close this and reopen a new PR to make reviewing easier

PR #2195, opened 2021-06-18
Add http path defined in proto to annotated context

### LINKED ISSUE #2005: Get http path defined in proto
## 🚀 Feature

We'd like to take the mapping http path defined in proto as the label value of request metrics. E.g. proto below:
```
	rpc GetMessageWithBody(MessageWithBody) returns (google.protobuf.Empty) {
		option (google.api.http) = {
			post: "/v2/example/withbody/{id}",
			body: "data"
		};
	}
```
We need `/v2/example/withbody/{id}` as the label value of request metrics.

Can we put the mapping http path into context or http headers ?

0daryo:

Fixes #2005 
Yes.
I found issue #2005 is in stale,  setting URL path param defined in proto to context.
I think backwards compatibility for user is not broken, with this options pattern, and easy to add another option same way.

### DIFF
--- a/runtime/context.go
+++ b/runtime/context.go
@@ -334,8 +334,8 @@ func withRPCMethod(ctx context.Context, rpcMethodName string) context.Context {
 	return context.WithValue(ctx, rpcMethodKey{}, rpcMethodName)
 }
 
-// HTTPPathPattern returns the http path pattern string for the server context. The returned
-// string is defined in api path template google.api.http option
+// HTTPPathPattern returns the HTTP path pattern string relating to the HTTP handler, if one exists.
+// The format of the returned string is defined by the google.api.http path template type.
 func HTTPPathPattern(ctx context.Context) (string, bool) {
 	m := ctx.Value(httpPathPatternKey{})
 	if m == nil {
--- a/runtime/context.go
+++ b/runtime/context.go
@@ -52,7 +52,7 @@ type (
 	AnnotateOption func(ao *annotateOptions)
 )
 
-func WithPattern(pattern string) AnnotateOption {
+func WithHTTPPathPattern(pattern string) AnnotateOption {
 	return func(ao *annotateOptions) {
 		ao.httpPath = pattern
 	}
--- a/runtime/context.go
+++ b/runtime/context.go
@@ -41,7 +41,22 @@ var (
 	DefaultContextTimeout = 0 * time.Second
 )
 
-type rpcMethodKey struct{}
+type (
+	rpcMethodKey       struct{}
+	httpPathPatternKey struct{}
+
+	annotateOptions struct {
+		httpPath string
+	}
+
+	AnnotateOption func(ao *annotateOptions)
+)
+
+func WithPattern(pattern string) AnnotateOption {
+	return func(ao *annotateOptions) {
+		ao.httpPath = pattern
+	}
+}
 
 func decodeBinHeader(v string) ([]byte, error) {
 	if len(v)%4 == 0 {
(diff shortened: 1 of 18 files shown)


runtime/context.go
```diff
@@ -72,8 +87,8 @@ func AnnotateContext(ctx context.Context, mux *ServeMux, req *http.Request, rpcM
 
 // AnnotateIncomingContext adds context information such as metadata from the request.
 // Attach metadata as incoming context.
-func AnnotateIncomingContext(ctx context.Context, mux *ServeMux, req *http.Request, rpcMethodName string) (context.Context, error) {
-	ctx, md, err := annotateContext(ctx, mux, req, rpcMethodName)
+func AnnotateIncomingContext(ctx context.Context, mux *ServeMux, req *http.Request, rpcMethodName string, options ...AnnotateOption) (context.Context, error) {
```
0daryo, line 90: If any concern about interface compatibility, I'll define another function like below.
```
func AnnotateIncomingContextWithOption(ctx context.Context, mux *ServeMux, req *http.Request, rpcMethodName string, options ...AnnotateOption) (context.Context, error) {
```

johanbrandhorst: I think this is technically a breaking change because someone could be saving the old function signature as a variable and it would change type, but that should be a vanishingly rare use case so we can probably just add this variadic parameter without practically breaking anyone.

runtime/context.go
```diff
@@ -41,7 +41,22 @@ var (
...
-type rpcMethodKey struct{}
+type (
+	rpcMethodKey       struct{}
+	httpPathPatternKey struct{}
+
+	annotateOptions struct {
+		httpPath string
+	}
+
+	AnnotateOption func(ao *annotateOptions)
```
johanbrandhorst, line 52: Since we're only using this in methods that mutate the `context.Context`, maybe it would make more sense for this to be of type

```suggestion
	AnnotateOption func(ctx context.Context) context.Context
```

Then it could just add incoming/outgoing context information directly.

runtime/context.go
```diff
@@ -311,3 +333,21 @@ func RPCMethod(ctx context.Context) (string, bool) {
 func withRPCMethod(ctx context.Context, rpcMethodName string) context.Context {
 	return context.WithValue(ctx, rpcMethodKey{}, rpcMethodName)
 }
+
+// HTTPPathPattern returns the http path pattern string for the server context. The returned
+// string is defined in api path template google.api.http option
```
johanbrandhorst, line 338:
```suggestion
// HTTPPathPattern returns the HTTP path pattern string relating to the HTTP handler, if one exists.
// The format of the returned string is defined by the google.api.http path template type.
```

johanbrandhorst (changes requested):
Thanks for this PR, great job! I had a slight idea on how the design could be simplified a bit, let me know what you think.

runtime/context.go
0daryo: Thanks!
I thiink this idea makes options simply.

Just considering if there may be usecase that not mutate context In future changes. ( e.g. add data not to ctx but to medatada.MD)

johanbrandhorst: I think we should be safe, the job of this function is to annotate the context, nothing more. Changing the metadata.MD would also require mutating the context.

johanbrandhorst (approved):
LGTM! This change is going to break users who upgrade their generator without upgrading the runtime, but I think that case is much more acceptable than the other way around, as most people will probably be managing their generator version less strictly. Any objections @achew22 @adambabik to merging this?

johanbrandhorst:
Nice to see `gorelease` failing as intended here - it's up to us to judge whether we think it's a bad enough breakage. I think after this merge any `gorelease` jobs rebased on this merge will fail, so we will probably want to make a new release after merging this.

achew22 (approved):
It feels fine to me to update the generator/runtime interface with the expectation that they'll be updated in lockstep.

johanbrandhorst:
Thanks for your contribution! I just realised something immediately after merging - it would be nice to add something to the documentation to explain how to use this new feature. Would you be willing to contribute a quick update to the docs? Maybe a brief new document in the operations section (https://github.com/grpc-ecosystem/grpc-gateway/tree/master/docs/docs/operations) explaining how to use this to enhance your metrics or something like that.

0daryo:
Thanks for merging.
I will work on documentation.

PR #2202, opened 2021-06-23
fix: adding root '/' path in custom route handler

### LINKED ISSUE #1909: Adding custom routes to the mux. Cannot add root route. ("/")

Shivam010:

Fixes #1909 

Yes

As the root route does not have any path, the parser was trying to parse the following tokens: `[eof]`, which was causing an error in `parser.topLevelSegments`. Therefore, I added a condition for `eof` in it and returns an `eof` segment, which is checked in `template.Compile` to obtain the desired result.

#### Other comments

A more direct approach would be the one in #1916 but I think this approach would be better. 
Open for suggestions.

### DIFF
--- a/internal/httprule/compile.go
+++ b/internal/httprule/compile.go
@@ -96,6 +96,10 @@ func (t template) Compile() Template {
 		if op.str == "" {
 			ops = append(ops, op.num)
 		} else {
+			// eof segment literal represents the "/" path pattern
+			if op.str == eof {
+				op.str = ""
+			}
 			if _, ok := consts[op.str]; !ok {
 				consts[op.str] = len(pool)
 				pool = append(pool, op.str)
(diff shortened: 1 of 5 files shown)


johanbrandhorst (approved):
Nice, this is a great fix!

PR #2215, opened 2021-07-01
 Add optional_nullable option (default: off) to mark Proto3 Optional fields as x-nullable

### LINKED ISSUE #669: protoc-gen-swagger: should well known types be nullable

irridia:
See #669 for more background.

This change adds an optional (default: off) `proto3_optional_nullable` flag to `protoc-gen-openapiv2` that will mark Proto3 fields marked with the `optional` label as `x-nullable`. The OAS specification calls this out as `nullable`, but this behavior has been "back-ported" to OAS2 using the System Extension `x-nullable`.

The effect of the Proto3 Optional flag on the GRPC is exactly this—it sets primitive types to pointers. It's certainly open to debate why we have Proto2 `optional`, the `optional` field descriptor, and Proto3 `optional`, but fundamentally Proto3 Optional is designed to solve the "true presence" issue as discussed in #669, at least IMHO. Whereas the other options never felt specific to this nullable behavior. Proto3 optional field presence discussion is [here](https://github.com/protocolbuffers/protobuf/blob/master/docs/field_presence.md).

It could be a valid approach to make this behavior the default, but for now it felt best to me to retain backwards compatibility until the semantics are well understood.

Potentially:
Fixes #669

Adds:
 * `proto3_optional_nullable` command line flag.
 * If `proto3_optional_nullable` is set, and the proto field has `proto3optional` set, the schema XNullable state is set to true, which emits the OAPI `x-nullable` tag.

Comments and feedback welcome! @johanbrandhorst

### DIFF
--- a/internal/descriptor/registry.go
+++ b/internal/descriptor/registry.go
@@ -87,8 +87,8 @@ type Registry struct {
 	// has no HttpRule annotation.
 	warnOnUnboundMethods bool
 
-	// optionalNullable specifies whether Proto3 Optional fields should be marked as x-nullable.
-	optionalNullable bool
+	// proto3OptionalNullable specifies whether Proto3 Optional fields should be marked as x-nullable.
+	proto3OptionalNullable bool
 
 	// fileOptions is a mapping of file name to additional OpenAPI file options
 	fileOptions map[string]*options.Swagger
@@ -571,14 +571,14 @@ func (r *Registry) GetOmitPackageDoc() bool {
 	return r.omitPackageDoc
 }
 
-// SetOptionalNullable set optionalNullable
-func (r *Registry) SetOptionalNullable(optionalNullable bool) {
-	r.optionalNullable = optionalNullable
+// SetProto3OptionalNullable set proto3OtionalNullable
+func (r *Registry) SetProto3OptionalNullable(proto3OtionalNullable bool) {
+	r.proto3OptionalNullable = proto3OtionalNullable
 }
 
-// GetOptionalNullable returns optionalNullable
-func (r *Registry) GetOptionalNullable() bool {
-	return r.optionalNullable
+// GetProto3OptionalNullable returns proto3OtionalNullable
+func (r *Registry) GetProto3OptionalNullable() bool {
+	return r.proto3OptionalNullable
 }
 
 // RegisterOpenAPIOptions registers OpenAPI options
--- a/internal/descriptor/registry.go
+++ b/internal/descriptor/registry.go
@@ -87,6 +87,9 @@ type Registry struct {
 	// has no HttpRule annotation.
 	warnOnUnboundMethods bool
 
+	// optionalNullable specifies whether Proto3 Optional fields should be marked as x-nullable.
+	optionalNullable bool
+
 	// fileOptions is a mapping of file name to additional OpenAPI file options
 	fileOptions map[string]*options.Swagger
 
@@ -566,6 +569,16 @@ func (r *Registry) SetOmitPackageDoc(omit bool) {
 // GetOmitPackageDoc returns whether a package comment will be omitted from the generated code
 func (r *Registry) GetOmitPackageDoc() bool {
 	return r.omitPackageDoc
+}
+
+// SetOptionalNullable set optionalNullable
+func (r *Registry) SetOptionalNullable(optionalNullable bool) {
+	r.optionalNullable = optionalNullable
+}
+
+// GetOptionalNullable returns optionalNullable
+func (r *Registry) GetOptionalNullable() bool {
+	return r.optionalNullable
 }
 
 // RegisterOpenAPIOptions registers OpenAPI options
--- a/protoc-gen-openapiv2/defs.bzl
+++ b/protoc-gen-openapiv2/defs.bzl
@@ -62,6 +62,7 @@ def _run_proto_gen_openapi(
         disable_default_errors,
         enums_as_ints,
         simple_operation_ids,
+        optional_nullable,
         openapi_configuration,
         generate_unbound_methods):
     args = actions.args()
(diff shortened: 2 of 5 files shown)


protoc-gen-openapiv2/defs.bzl
```diff
@@ -278,6 +283,11 @@ protoc_gen_openapiv2 = rule(
             doc = "whether to remove the service prefix in the operationID" +
                   " generation. Can introduce duplicate operationIDs, use with caution.",
         ),
+        "optional_nullable": attr.bool(
```
johanbrandhorst, line 286: Re: naming

internal/descriptor/registry.go
```diff
@@ -87,6 +87,9 @@ type Registry struct {
 	// has no HttpRule annotation.
 	warnOnUnboundMethods bool
 
+	// optionalNullable specifies whether Proto3 Optional fields should be marked as x-nullable.
+	optionalNullable bool
```
johanbrandhorst, line 91: Can we name this field and related helpers `proto3OptionalNullable` to make it more explicit?

protoc-gen-openapiv2/internal/genopenapi/types.go
```diff
@@ -152,10 +152,11 @@ type openapiParameterObject struct {
...
 // http://swagger.io/specification/#itemsObject
 type schemaCore struct {
-	Type    string          `json:"type,omitempty"`
-	Format  string          `json:"format,omitempty"`
-	Ref     string          `json:"$ref,omitempty"`
-	Example json.RawMessage `json:"example,omitempty"`
+	Type     string          `json:"type,omitempty"`
+	Format   string          `json:"format,omitempty"`
+	Ref      string          `json:"$ref,omitempty"`
+	Nullable bool            `json:"x-nullable,omitempty"`
```
johanbrandhorst, line 158: `x-nullable` isn't defined in the referenced docs (http://swagger.io/specification/#itemsObject), could you please add a reference here?

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -581,6 +582,10 @@ func schemaOfField(f *descriptor.Field, reg *descriptor.Registry, refs refMap) o
 		updateSwaggerObjectFromFieldBehavior(&ret, j, f)
 	}
 
+	if reg.GetOptionalNullable() && f.Proto3Optional != nil && *f.Proto3Optional {
```
johanbrandhorst, line 585: Can we just use the getter?
```suggestion
	if reg.GetOptionalNullable() && f.GetProto3Optional() {
```

johanbrandhorst (commented):
This looks great, just some nits on the naming and style!

johanbrandhorst:
Looks like there's a syntax error in the bazel definitions?
```
./protoc-gen-openapiv2/defs.bzl:291:32: syntax error near "openapi_configuration"
./protoc-gen-openapiv2/defs.bzl # reformat
```

protoc-gen-openapiv2/defs.bzl
nashikb: Updated!

internal/descriptor/registry.go
nashikb: Yep, done. I was debating this myself.

protoc-gen-openapiv2/internal/genopenapi/template.go
nashikb: Yep, I was using completion and missed the presence of the getters. Updated.

nashikb:
> Looks like there's a syntax error in the bazel definitions?
> 
> ```
> ./protoc-gen-openapiv2/defs.bzl:291:32: syntax error near "openapi_configuration"
> ./protoc-gen-openapiv2/defs.bzl # reformat
> ```

Fixed, I had a missing comma. I didn't have access to the CircleCI to see the details of why the test failed.

protoc-gen-openapiv2/internal/genopenapi/types.go
nashikb: I updated that URL to point to the v2 specification, and added a reference to the v3 spec that calls out `nullable`. `x-nullable` is a "back-ported" extension to emulate this v3 behavior in v2, but it's more of a widely used convention than part of any formal spec. I've added what I think is helpful; let me know if it's too much or missing something you'd want to see/know.

protoc-gen-openapiv2/defs.bzl
irridia: Updated!

internal/descriptor/registry.go
irridia: Yep, done. I was debating this myself.

protoc-gen-openapiv2/internal/genopenapi/types.go
irridia: I updated that URL to point to the v2 specification, and added a reference to the v3 spec that calls out nullable. x-nullable is a "back-ported" extension to emulate this v3 behavior in v2, but it's more of a widely used convention than part of any formal spec. I've added what I think is helpful; let me know if it's too much or missing something you'd want to see/know.

protoc-gen-openapiv2/internal/genopenapi/template.go
irridia: Yep, I was using completion and missed the presence of the getters. Updated.

irridia:
> Looks like there's a syntax error in the bazel definitions?
> 
> ```
> ./protoc-gen-openapiv2/defs.bzl:291:32: syntax error near "openapi_configuration"
> ./protoc-gen-openapiv2/defs.bzl # reformat
> ```

Fixed, I had a missing comma. I didn't have access to the CircleCI to see the details of why the test failed.

johanbrandhorst (approved):
Thanks for the changes, LGTM!

irridia:
Thanks so much for the time, @johanbrandhorst !

PR #2219, opened 2021-07-04
Supporting a scenario where proto_library is being used withour srcs...

### LINKED ISSUE #2208: Proto_library deps are being ignored
## 🐛 Bug Report

`protoc_gen_openapiv2` won't generate OpenAPI specification file for `proto_library` targets without `srcs` attribute. The official `proto_library` documentation mention:

> One usecase where srcs can be empty is an alias-library. This is a proto_library rule having one or more other proto_library in deps. This pattern can be used to e.g. export a public api under a persistent name. 

## To Reproduce
Consider the following BUILD file. I have 3 files, of whom 2 are public API to which I want to generate OpenAPIv2 specification for.

```
proto_library(
  name = "public_api1_proto",
  srcs = ["public_api1.proto"],
)

proto_library(
  name = "public_api2_proto",
  srcs = ["public_api2.proto"],
)
#
proto_library(
  name = "private_api_proto",
  srcs = ["private_api.proto"],
)

proto_library(
  name = "service_api_proto"
  deps = [":public_api1_proto", ":public_api2_proto"],
)
```

Now adding `protoc_gen_openapiv2` to to the public api target:
```
protoc_gen_openapiv2(
  name = "service_openapiv2_specification"
  proto = ":service_api_proto"
```

## Expected behavior
OpenAPI specification files get created for `public_api1.proto` and `public_api2.proto`. 

## Actual Behavior
Nothing is being generated. That's probably due to that:

https://github.com/grpc-ecosystem/grpc-gateway/blob/4bb7a7e2aae2e86f7a95b29288ab1232597b35a7/protoc-gen-openapiv2/defs.bzl#L30

Only direct sources are being collected. I quickly hack it locally using `check_deps_sources.to_list()]` to pick dependencies instead, and that solve the problem. Perhaps I can share a PR after you guys decide how exactly you want to handle that (if there's `srcs` use `srcs` and if not use `deps`?). 

## Your Environment
MacOS, Bazel 3.4.1

amitz:
#### References to other Issues or PRs ####
Fixes #2208

Yes

Instead of using `proto_source_root`, `check_deps_sources.to_list()`, which return the source if exists, and if not - support the other scenario of a `proto_library` without srcs and only deps.

### DIFF
--- a/protoc-gen-openapiv2/defs.bzl
+++ b/protoc-gen-openapiv2/defs.bzl
@@ -27,12 +27,12 @@ def _direct_source_infos(proto_info, provided_sources = []):
 
     source_root = proto_info.proto_source_root
     if "." == source_root:
-        return [struct(file = src, import_path = src.path) for src in proto_info.direct_sources]
+        return [struct(file = src, import_path = src.path) for src in proto_info.check_deps_sources.to_list()]
 
     offset = len(source_root) + 1  # + '/'.
 
     infos = []
-    for src in proto_info.direct_sources:
+    for src in proto_info.check_deps_sources.to_list():
         # TODO(yannic): Remove this hack when we drop support for Bazel < 1.0.
         local_offset = offset
         if src.root.path and not source_root.startswith(src.root.path):


achew22:
Your commit message says that you've added an example, but I can't see it in the diff. Can you help me understand what's going on?

amitz:
Your right, sorry. I originally added an example of the "Public API" scenario (detailed [here](https://github.com/grpc-ecosystem/grpc-gateway/issues/2208)) into `examples/internal/proto/examplepb/BUILD.bazel`, but I removed it during the re-base and forgot to edit the description. 

The reason I removed was that in after thought - it made the Example file more confusing. I was trying to emphasis a `proto_library` concept (of using just `deps`) with the current protobufs available at the example folder, and it didn't make a lot of sense. I ended up decided using `deps` instead of just `srcs` is more of a `proto_library` logic, and doesn't directly intertwined into how to use `protoc_gen_openapiv2`. 

Let me know if you think otherwise. 
Thanks.

achew22:
I agree that 1.0 compatibility is overrated and not worth persisting. Just wanted to check that there wasn't something missing.

PR #2221, opened 2021-07-06
No need to block on first handleSend

riarheos:
Hi, I want to discuss this a little.

When we handle a bidi streaming there are situations when the first event is actually coming from the server. In the current code however a client **must** send some event before we get into a send-receive loop.

The PR basically just skips that client-to-server block making it possible to do a completely independent bidi-communication.

### DIFF
--- a/examples/internal/proto/examplepb/flow_combination.pb.gw.go
+++ b/examples/internal/proto/examplepb/flow_combination.pb.gw.go
@@ -134,15 +134,6 @@ func request_FlowCombination_StreamEmptyStream_0(ctx context.Context, marshaler 
 		}
 		return nil
 	}
-	if err := handleSend(); err != nil {
-		if cerr := stream.CloseSend(); cerr != nil {
-			grpclog.Infof("Failed to terminate client stream: %v", cerr)
-		}
-		if err == io.EOF {
-			return stream, metadata, nil
-		}
-		return nil, metadata, err
-	}
 	go func() {
 		for {
 			if err := handleSend(); err != nil {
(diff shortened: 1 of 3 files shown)


riarheos:
@googlebot I signed it!

johanbrandhorst (commented):
This is hard for me to review as I'm not sure I understand the effects - how have you tested this change? The only way I know of doing bi-directional streaming with the gateway is using https://github.com/tmc/grpc-websocket-proxy. Have you verified that after this change the normal server streaming method still works?

riarheos:
I have tested it in bidi-mode with the websocket adapter you mentioned. The other modes seem unaffected because I've touched only the bidi template.

It might be hard to test this one, I have created a trivial all-in-one server here:
https://github.com/riarheos/grpc-test 

You can build it with `./generate.sh; go build` (you do require the patched generator in your PATH of course) and try curls / websocats:
```
$ curl -XPOST -d '{"name": "foo"}' http://localhost:11112/Greet/GreetMe
{"message":"Hello, foo"}

$ curl -XPOST -d '{"name": "foo"}' http://localhost:11112/Greet/GreetTimer
{"result":{"message":"Hello, foo #0"}}
{"result":{"message":"Hello, foo #1"}}
{"result":{"message":"Hello, foo #2"}}
...

$ curl -XPOST -d '{"name": "foo"}' http://localhost:11112/Greet/GreetUltra
{"result":{"message":"Hello, foo"}}

$ websocat ws://localhost:11113/Greet/GreetUltra
{"name": "foo"}
{"result":{"message":"Hello, foo"}}
{"name": "bar"}
{"result":{"message":"Hello, bar"}}
```

johanbrandhorst (approved):
OK, thanks for the compelling testing. I am also slightly more comfortable since you explained that this only affects bi-directional streams. I assume users of the grpc-gateway who rely on it for bi-directional streaming methods are going to be vanishingly rare.

johanbrandhorst:
Thanks for your contribution!

riarheos:
Great, thanks!

PR #2224, opened 2021-07-07
Merged same types of function parameters

iamrajiv:
## Description

If parameters of the same type lie consecutively, mention their type once at the end of the last parameter. Unlike other languages, like C, where all parameters must be specified with types explicitly, it is not required to do so in Go. Combining the types is usually recommended for the sake of brevity.

### DIFF
--- a/examples/internal/clients/abe/api_a_bit_of_everything_service.go
+++ b/examples/internal/clients/abe/api_a_bit_of_everything_service.go
@@ -12,11 +12,12 @@ package abe
 
 import (
 	"context"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"net/url"
 	"strings"
-	"fmt"
+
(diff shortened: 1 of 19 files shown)


johanbrandhorst (commented):
Hi Rajiv. Thanks for your contribution. Unfortunately, there are a few problems with this contribution:

1. You've manually changed some of the generated files, which is why the CI job is failing. You'll need to regenerate them and ignore those files in your editing.
2. I'm not sure I actually agree with this change. I've heard a compelling argument _against_ using these sorts of signatures, which is that it's easier to add a parameter with the wrong type by mistake, because you don't need to add the type. As such, I'm not convinced there is merit in this change in general, sorry.

iamrajiv:
Hi, @johanbrandhorst. Thanks for the review. Yeah, I am convinced by the points.

PR #2241, opened 2021-07-17
Tag support in comments

whockey:
Not sure if others would support this, but this enables you to add in OpenAPI tags in comments currently you have to add them as an annotation which is a bit cluttering. 

```proto
  // Gets a thing
  //
  // Retrieves a single legal entity by its ID
  //
  // Tags: Thing
  rpc GetThing(GetThingRequest) returns (Thing) {
    option (google.api.http) = {
      get: "/api/thing/{thing_id}"
    };
  }
```
vs
```proto
  // Gets a thing
  //
  // Retrieves a single legal entity by its ID
  rpc GetThing(GetThingRequest) returns (Thing) {
    option (google.api.http) = {
      get: "/api/thing/{thing_id}"
    };
      tags: "Thing"
    };
  }
```

### DIFF
--- a/protoc-gen-openapiv2/internal/genopenapi/template.go
+++ b/protoc-gen-openapiv2/internal/genopenapi/template.go
@@ -1758,6 +1758,9 @@ func processHeaders(inputHdrs map[string]*openapi_options.Header) (openapiHeader
 //
 // If there is no 'Summary', the same behavior will be attempted on 'Title',
 // but only if the last character is not a period.
+//
+// If there is a comment line that starts with 'Tags:' it will it will insert that
+// as the OpenAPI Tags field
 func updateOpenAPIDataFromComments(reg *descriptor.Registry, swaggerObject interface{}, data interface{}, comment string, isPackageObject bool) error {
 	if len(comment) == 0 {
 		return nil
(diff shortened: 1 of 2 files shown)


whockey:
@googlebot I fixed it.

johanbrandhorst:
Hi Will, thanks for your PR. I'm not sure I support this TBH, it seems a bit too magical to me, especially when we do have the option to specify tags within the annotations already. Sorry that you spent time on this already. If you want to avoid it in the future, please raise an issue first.

whockey:
> Hi Will, thanks for your PR. I'm not sure I support this TBH, it seems a bit too magical to me, especially when we do have the option to specify tags within the annotations already. Sorry that you spent time on this already. If you want to avoid it in the future, please raise an issue first.

Totally up to you all. However, from our experience documenting examples, tags, etc. all as an `grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation` gets very unwieldy. The openapi proto definitions end up taking up a majority of the code. If they could be moved to the comments it makes things way more readable - jut like description and titles current are. It makes the boundary between code and documentation much clearer.

johanbrandhorst:
Thanks for the feedback. An alternative solution is to use an external configuration file: https://grpc-ecosystem.github.io/grpc-gateway/docs/mapping/grpc_api_configuration/#using-an-external-configuration-file. That should let you clean up your protobuf files.

whockey:
Fair enough - not the approach I'd go with, but ya'lls project! Will close :)

PR #2247, opened 2021-07-24
fix(integ-test): timeout test is now consistent

### LINKED ISSUE #1527: Flaky test: examples/internal/integration/integration_test.go:TestTimeout
## 🐛 Bug Report

The `TestTimeout` test is sometimes failing randomly.

## To Reproduce

```shell
$ go test -run TestTimeout -count 100 ./examples/internal/integration
```

## Expected behavior

Test not to be flaky

## Actual Behavior

The tests fails with an incorrect status code error:
```
integration_test.go:1395: resp.StatusCode = 500; want 504
```

See https://app.circleci.com/pipelines/github/grpc-ecosystem/grpc-gateway/993/workflows/f7328257-19fe-436c-8cd4-8f2274bcb535/jobs/13566.

Tamiraltshuler:
Fixes #1527

The timeout test on the server side returned a regular error and not a grpc error. replaced this error with a grpc error using `grpc/status`. 

The problem caused by a race condition in the grpc client of the gateway not returning an error before the server made a response. 
(Race condition of the grpc client)

#### Other comments
By making the the server return a grpc error its not truly checks a timeout. the other option is to put a sleep after the context is done to make sure the client have enough time to return a timeout . so the question is if the second option is preferred? (sleep in tests are not best practice thats why its not in this pr)

### DIFF
--- a/examples/internal/server/a_bit_of_everything.go
+++ b/examples/internal/server/a_bit_of_everything.go
@@ -308,7 +308,7 @@ func (s *_ABitOfEverythingServer) NoBindings(ctx context.Context, msg *durationp
 
 func (s *_ABitOfEverythingServer) Timeout(ctx context.Context, msg *emptypb.Empty) (*emptypb.Empty, error) {
 	<-ctx.Done()
-	return nil, ctx.Err()
+	return nil, status.FromContextError(ctx.Err()).Err()
 }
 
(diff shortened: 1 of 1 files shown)


johanbrandhorst (approved):
This is super clever. Thanks for the fix!

PR #2259, opened 2021-07-30
fix(runtime): fix path parsing to use URL.EscapedPath() instead of Path and unescape individual components

remyleone:
Basically, we want to route for instance http://example.com/resources/some%2Fpath/something from this HTTP annotation :

    get: /resources/{key}/something

We want the  key field in gRPC to be able to contain 0 or many slashes

https://github.com/grpc-ecosystem/grpc-gateway/pull/660


johanbrandhorst (commented):
I'd like to rope in both @achew22 and @ivucica who were part of the discussion last time this came up. It seems to me this is a bug fix, and even though it may break users it would be fixing broken behavior. Would you agree that this change is OK?

PR #2279, opened 2021-08-11
fix: path template to valid openapi v2 parameters

### LINKED ISSUE #2084: Generated OpenAPI document fails validation on parameters in path
## 🐛 Bug Report

Take the example [a little bit of everything](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/examples/internal/proto/examplepb/a_bit_of_everything.swagger.json) swagger file and pop it into swagger editor/UI - it will fail validation checks

This feels like the issues raised in #720 and #1015 though I didn't want to bump an older issue or hijack another issue that seemed more specific.

Sometimes the path annotations generate fine in the OpenAPI document and other times not. For example, the following annotrations seem to work:

```proto
option (google.api.http) = {
      get: "/v1alpha/{user=users/*}/noun:verb"
    };

option (google.api.http) = {
      post: "/v1alpha/{user=users/*}/noun2:verb"
      body: "*"
    };
```

Generates:

```yaml
"/v1alpha/{user}/noun:verb": {
      "get": {
        "parameters": [
          {
            "name": "user",
            "in": "path",
            "required": true,
            "type": "string"
          }
        ],
    }
}
"/v1alpha/{user}/advice:generate": {
      "post": {
        "parameters": [
          {
            "name": "user",
            "in": "path",
            "required": true,
            "type": "string"
          },
          {
            "name": "body",
            "in": "body",
            "required": true,
            "schema": {
              "$ref": "#/definitions/v1alphaVerbRequest"
            }
          }
        ],
    }
}
```

However, the following does not:

```proto
    option (google.api.http) = {
      get: "/v1alpha/{parent=users/*}/accounts"
    };
    option (google.api.http) = {
      get: "/v1alpha/{name=users/*/accounts/*}"
    };
```

Generates:

```yaml
 "/v1alpha/{parent=users/*}/accounts": {
      "get": {
        "parameters": [
          {
            "name": "parent",
            "in": "path",
            "required": true,
            "type": "string"
          }
        ],
    }
}
 "/v1alpha/{name=users/*/accounts/*}": {
      "get": {
        "parameters": [
          {
            "name": "name",
            "in": "path",
            "required": true,
            "type": "string"
          }
        ],
    }
}
```

Interestingly, in the example above that doesn't' work, renaming `parent=` to `user=` generates properly!

```proto
    option (google.api.http) = {
      get: "/v1alpha/{user=users/*}/accounts"
    };
```
```yaml
  "/v1alpha/{user}/accounts": {
      "get": {
        "parameters": [
          {
            "name": "user",
            "in": "path",
            "required": true,
            "type": "string"
          }
        ],
   }
}
```

Does the generator expect specific values in which it will trigger the correct generation in the OpenAPI spec for the path annotations? Switching from `parent` to `user` to fix one such case of issues sounds buggy to me. I'm not certain how the second example `/v1alpha/{name=users/*/accounts/*}` should map to query params in the OpenAPI spec, but the URL should probably look like `/v1alpha/users/*/accounts/*` - I think the gateway itself processes all this okay, its just the generated OpenAPI spec that's falling short.

Generated w/ v2.3.0

tenevdev:

 Fixes #2084

yes

Still very much a draft, but the idea is to support some basic path templates as described in the referenced issue - by re-writing the path to be valid openapi v2 syntax and adding separate parameters if needed.

I intend to handle plural nouns in the path to make things nicer - e.g. `name=/orgs/*/projects/*` creates parameters `org` and `project` and path `/orgs/{org}/projects/{project}`.

#### Other comments

The code that I changed seemed quite complex in the sense that it may have been written to handle some cases that were not covered in unit tests and that I wasn't aware of. We just need to be careful about not breaking anyone with this change like it happened in the past when others tried to fix similar issues.

### DIFF
--- a/protoc-gen-openapiv2/internal/genopenapi/template.go
+++ b/protoc-gen-openapiv2/internal/genopenapi/template.go
@@ -18,14 +18,15 @@ import (
 	"time"
 
 	"github.com/golang/glog"
-	"github.com/grpc-ecosystem/grpc-gateway/v2/internal/casing"
-	"github.com/grpc-ecosystem/grpc-gateway/v2/internal/descriptor"
-	openapi_options "github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2/options"
 	"google.golang.org/genproto/googleapis/api/annotations"
 	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/descriptorpb"
 	"google.golang.org/protobuf/types/known/structpb"
+
+	"github.com/grpc-ecosystem/grpc-gateway/v2/internal/casing"
+	"github.com/grpc-ecosystem/grpc-gateway/v2/internal/descriptor"
+	openapi_options "github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2/options"
 )
 
 // wktSchemas are the schemas of well-known-types.
@@ -765,7 +766,32 @@ func resolveFullyQualifiedNameToOpenAPINames(messages []string, useFQNForOpenAPI
 	return uniqueNames
 }
 
-var canRegexp = regexp.MustCompile("{([a-zA-Z][a-zA-Z0-9_.]*).*}")
+var pathParamTemplateRegex = regexp.MustCompile("{([a-zA-Z][a-zA-Z0-9_.]*)=([a-zA-Z][a-zA-Z0-9_./\\*]*)}(.*)")
+
+func pathTemplateToOpenAPIParameterNames(path string, parameterName string) []string {
+	parameterRegex := regexp.MustCompile(fmt.Sprintf("{%s=([a-zA-Z][a-zA-Z0-9_./\\*]*)}.*", parameterName))
+	if !parameterRegex.MatchString(path) {
+		return []string{parameterName}
+	}
+
+	submatch := parameterRegex.FindStringSubmatch(path)
+	paramTemplate := submatch[1]
+	if strings.Count(paramTemplate, "*") <= 1 {
+		return []string{parameterName}
+	}
+
+	templateParts := strings.Split(paramTemplate, "/")
+	var parameterNames []string
+	var previousPart string
+	for _, templatePart := range templateParts {
+		if templatePart == "*" {
+			parameterNames = append(parameterNames, fmt.Sprintf("%s.%s", parameterName, previousPart))
+		} else {
+			previousPart = templatePart
+		}
+	}
+	return parameterNames
+}
 
 // OpenAPI expects paths of the form /path/{string_value} but gRPC-Gateway paths are expected to be of the form /path/{string_value=strprefix/*}. This should reformat it correctly.
 func templateToOpenAPIPath(path string, reg *descriptor.Registry, fields []*descriptor.Field, msgs []*descriptor.Message) string {
@@ -824,24 +850,36 @@ func templateToOpenAPIPath(path string, reg *descriptor.Registry, fields []*desc
 	// syntax for this subsection CAN be handled by a regexp since it has no
 	// memory.
 	for index, part := range parts {
-		// If part is a resource name such as "parent", "name", "user.name", the format info must be retained.
-		prefix := canRegexp.ReplaceAllString(part, "$1")
-		if isResourceName(prefix) {
+		if !pathParamTemplateRegex.MatchString(part) {
 			continue
 		}
-		parts[index] = canRegexp.ReplaceAllString(part, "{$1}")
+
+		if !(strings.HasPrefix(part, "{") && strings.Contains(part, "=") && strings.Contains(part, "}")) {
+			continue
+		}
+
+		submatch := pathParamTemplateRegex.FindStringSubmatch(part)
+		paramName := submatch[1]
+		paramTemplate := submatch[2]
+		rest := submatch[3]
+
+		if strings.Count(paramTemplate, "*") <= 1 {
+			parts[index] = strings.Replace(paramTemplate, "*", fmt.Sprintf("{%s}", paramName), 1) + rest
+		} else {
+			templateParts := strings.Split(paramTemplate, "/")
+			var previousPart string
+			for i, templatePart := range templateParts {
+				if templatePart == "*" {
+					templateParts[i] = fmt.Sprintf("{%s.%s}", paramName, previousPart)
+				} else {
+					previousPart = templatePart
+				}
+			}
+			parts[index] = strings.Join(templateParts, "/") + rest
+		}
 	}
 
 	return strings.Join(parts, "/")
-}
-
-func isResourceName(prefix string) bool {
-	words := strings.Split(prefix, ".")
-	l := len(words)
-	field := words[l-1]
-	words = strings.Split(field, ":")
-	field = words[0]
-	return field == "parent" || field == "name"
 }
 
 func renderServiceTags(services []*descriptor.Service) []openapiTagObject {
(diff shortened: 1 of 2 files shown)


protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -824,26 +850,38 @@ func templateToOpenAPIPath(path string, reg *descriptor.Registry, fields []*desc
...
 	return strings.Join(parts, "/")
 }
 
-func isResourceName(prefix string) bool {
-	words := strings.Split(prefix, ".")
-	l := len(words)
-	field := words[l-1]
-	words = strings.Split(field, ":")
-	field = words[0]
-	return field == "parent" || field == "name"
```
johanbrandhorst, line 844: Is it safe to remove this?

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -18,14 +18,15 @@ import (
...
-	openapi_options "github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2/options"
 	"google.golang.org/genproto/googleapis/api/annotations"
 	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/descriptorpb"
 	"google.golang.org/protobuf/types/known/structpb"
+
+	"github.com/grpc-ecosystem/grpc-gateway/v2/internal/casing"
+	"github.com/grpc-ecosystem/grpc-gateway/v2/internal/descriptor"
+	openapi_options "github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2/options"
```
johanbrandhorst, line 29: Why did you change this import grouping?

protoc-gen-openapiv2/internal/genopenapi/template_test.go
```diff
@@ -2616,7 +2617,7 @@ func TestApplyTemplateRequestWithBodyQueryParameters(t *testing.T) {
 								PathTmpl: httprule.Template{
 									Version:  1,
 									OpCodes:  []int{0, 0},
-									Template: "/v1/{parent=publishers/*}/books",
+									Template: "/v1/publishers/{parent}/books",
```
johanbrandhorst, line 2620: Why did you change this test?

protoc-gen-openapiv2/internal/genopenapi/template_test.go
```diff
@@ -2798,20 +2799,20 @@ func TestTemplateToOpenAPIPath(t *testing.T) {
...
+		{"/{name=prefix/*}", "/prefix/{name}"},
+		{"/{name=prefix1/*/prefix2/*}", "/prefix1/{name.prefix1}/prefix2/{name.prefix2}"},
+		{"/{user.name=prefix/*}", "/prefix/{user.name}"},
+		{"/{user.name=prefix1/*/prefix2/*}", "/prefix1/{user.name.prefix1}/prefix2/{user.name.prefix2}"},
+		{"/{parent=prefix/*}/children", "/prefix/{parent}/children"},
+		{"/{name=prefix/*}:customMethod", "/prefix/{name}:customMethod"},
+		{"/{name=prefix1/*/prefix2/*}:customMethod", "/prefix1/{name.prefix1}/prefix2/{name.prefix2}:customMethod"},
+		{"/{user.name=prefix/*}:customMethod", "/prefix/{user.name}:customMethod"},
+		{"/{user.name=prefix1/*/prefix2/*}:customMethod", "/prefix1/{user.name.prefix1}/prefix2/{user.name.prefix2}:customMethod"},
+		{"/{parent=prefix/*}/children:customMethod", "/prefix/{parent}/children:customMethod"},
```
johanbrandhorst, line 2815: I don't understand these test changes. We want to include the `prefix` part in the variables, do we not?

johanbrandhorst (commented):
I'm confused by this change. There is lots of changed logic, and also lots of changed tests to boot. I would have expected new logic, but I don't understand why the tests have changed. Could you please explain?

protoc-gen-openapiv2/internal/genopenapi/template_test.go
tenevdev: this is my mistake, ran a "replace all" operation but I only meant to change the result paths below, not this template

protoc-gen-openapiv2/internal/genopenapi/template_test.go
tenevdev: Thanks for your review @johanbrandhorst. I'll skip over the fact that the implementation is still rough for now, but of course it's something that I'm going to clean up respecting your comments above. I think we should try to agree on the expected behaviour before discussing the implementation details, so I really appreciate your comments on the unit tests.

For the specification, I followed https://swagger.io/docs/specification/2-0/describing-parameters/ and the rough suggestion from https://github.com/grpc-ecosystem/grpc-gateway/issues/2084#issuecomment-818286883. All I care about is that for all inputs that were previously supported, a valid openapi spec is still created but it may not be exactly the same one as before, and a valid openapi spec is created for more inputs than previously supported (the more complex path template cases). Based on the specification link above, my understanding is that the syntax for the path parameter in openapi v2 is `{[a-zA-Z][a-zA-Z0-9_.]*}` (roughly), so no `=` and all the template syntax after that which is common in protobuf path templates. So before I changed it, this unit test contained cases which produced paths with invalid syntax for openapi v2 parameter names. I hope we can easily agree on that, but looking at the git blame I don't understand how the `name` and `parent` thing and the way it worked before could have been useful for anyone. My use case also has these resource names as is best practice but the currently produced openapi v2 spec is invalid and can't be used with tools such as go-swagger without manual edits post-generation.

So if you would like a different style in the openapi v2 paths, please suggest something. What I did here was the "shortest path" to get a proof of concept working end-to-end. We can talk about making singular parameter names from the plural nouns before them, whether there should be a special behaviour for when there's only one `*` in the template vs multiple, etc.

protoc-gen-openapiv2/internal/genopenapi/template.go
tenevdev: Even after reading the issues and PRs related to this, I still have no idea why it was implemented this way. What's so special about `parent` and `name` other than being the recommended resource names to use by Google for writing APIs and how does that relate to openapi v2? Of course, if there's some convention to follow when creating the openapi v2 parameters based on these resource names, that would be nice to know and probably expected by users.

protoc-gen-openapiv2/internal/genopenapi/template.go
johanbrandhorst: No worries, I think I'm just worried about consistency with the rest of the codebase, I think we do two import groups generally (stdlib and rest). Please revert.

protoc-gen-openapiv2/internal/genopenapi/template_test.go
johanbrandhorst: OK! All good.

protoc-gen-openapiv2/internal/genopenapi/template.go
johanbrandhorst: Yea I don't know either. @ivucica do you know?

protoc-gen-openapiv2/internal/genopenapi/template_test.go
johanbrandhorst: I frankly do not know this part of the code base or behavior well enough to give this a thorough review, which is why I'm scared of changing tests that may have been guarding expected behavior. I agree that we should produce valid openapiv2 output for all valid inputs 😁, so that's a good start! Beyond that, I think I'm open to things that make sense and are well covered with understandable test cases. My concern with these changes is that it is not clearly communicated to the user that the `prefix` part of `"/{name=prefix/*}"` becomes part of the variable (i.e., for the input `/prefix/banana` the value of the `name` variable would be `prefix/banana`. With that in mind, shouldn't `"/{name=prefix/*}"` become `/{name}`? Or how do we indicate to the user that the first part must be a specific value, but the WHOLE part becomes part of the variable value?

PR #2307, opened 2021-08-26
add marshaller for application/x-www-form-urlencoded

tuhao1020:
#7


johanbrandhorst (commented):
I'm confused by this. From what I can tell, this is a copy paste of the JSON encoder? I'm really confused. Even if this was a complete implementation of a url-encoded form marshaler, I'm not convinced it needs to be part of this repository. If you need a marshaler like this today, you can just implement the Marshaler interface and use a `runtime.WithMarshaler` option.

PR #2313, opened 2021-08-30
fix: Set Content-Type from marshaler on stream error

vegather:

--

Yes

The marshaler was used to marshal the body of a stream error, but the content type was not used, leaving it as `text/plain; charset=utf-8`. This uses the content type as defined by the marshaler if the headers have not already been written.

#### Other comments

--

### DIFF
--- a/runtime/handler.go
+++ b/runtime/handler.go
@@ -202,10 +202,12 @@ func handleForwardResponseOptions(ctx context.Context, w http.ResponseWriter, re
 
 func handleForwardResponseStreamError(ctx context.Context, wroteHeader bool, marshaler Marshaler, w http.ResponseWriter, req *http.Request, mux *ServeMux, err error) {
 	st := mux.streamErrorHandler(ctx, err)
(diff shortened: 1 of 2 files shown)


johanbrandhorst (changes requested):
This looks reasonable to me, but it'd be great to have a test case that was broken before and is fixed by this change, so that we can protect ourselves against losing this again in the future. Could you see if you can add a test to [handler_test.go](https://github.com/grpc-ecosystem/grpc-gateway/blob/f81c435e72054a3fd141c7c9a9d64de291f363de/runtime/handler_test.go#L28)?

vegather:
Updated the tests to check the `Content-Type` if there is a body. The tests now fail without the fix and pass with the fix.

Good call updating the tests! Turns out I had `w.Header().Set...` and `w.WriteHeader...` in the wrong order.

johanbrandhorst (approved):
LGTM

johanbrandhorst:
Thanks for your contribution!

PR #2314, opened 2021-08-30
Add www-Authenticate header

### LINKED ISSUE #917: 401 Unauthorized response doesn't contain WWW-Authenticate header
I asked in the official gRPC Gitter channel, but didn't get a response so far: https://gitter.im/grpc/grpc?at=5ca0e2c1016a930a453cdc6d

Working on https://github.com/lightningnetwork/lnd/issues/2332, where this question came up.

Maybe related to grpc-gateway issues #4, #221 and #309.

When returning a `code.Unauthenticated` gRPC error, grpc-gateway maps this to a response with a 401 Unauthorized HTTP status code.

The response doesn't contain a WWW-Authenticate error though.

According to the HTTP specification https://tools.ietf.org/html/rfc7235#section-3.1, a 401 Unauthorized response "MUST" be accompanied by a WWW-Authenticate header.

I understand that no static value like `WWW-Authenticate: Basic realm="simple"` can be used, as it completely depends on the used authentication scheme, but maybe there can be a way to _easily_ configure this when setting up the gateway?

I know that error responses can be customized (apparently in two different ways: via [response forwarder](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/docs/_docs/customizingyourgateway.md#replace-a-response-forwarder-per-method) and via [custom error handler](https://mycodesmells.com/post/grpc-gateway-error-handler)), but shouldn't there be a way for grpc-gateway to adhere to the HTTP specification by default?

Also, from my understanding you can't just add a header in those response customizations, but instead you have to create a complete response, which could lead to inconsistent responses or breaking changes, but I'm not sure about this.

I think setting the WWW-Authenticate header is just not implemented and I didn't find an easy way to set one (other than using one of the above mentioned response customizations).

I think the implementation is this one: https://github.com/grpc-ecosystem/grpc-gateway/blob/7aca14d0a16247151f6e423f4cbfbd75a1bd4776/runtime/errors.go#L88

momom-i:
Fixes #917 

Yes

It's still a draft. johanbrandhorst showed me very kind instruction in the issue and I intend to follow that. 
I wrote proto, did `make generate` and wrote test code.

#### Other comments
I've not written handler yet.


examples/internal/proto/examplepb/echo_service.proto
```diff
@@ -90,4 +90,10 @@ service EchoService {
 			body: "body"
 		};
 	}
+	// EchoUnauthorized method receives a simple message and returns it.
```
johanbrandhorst, line 93: Could we update this comment to explain the purpose more clearly?
```suggestion
	// EchoUnauthorized method receives a simple message and returns it. It must
	// always return a google.rpc.Code of `UNAUTHENTICATED` and a HTTP Status code
	// of 401.
```

johanbrandhorst (changes requested):
This is a great start, well done! Now that we have a failing test, you'll want to implement the logic required to add the expected header. Since this is making us compliant with the spec, I don't think we need to add an option to turn off the behavior from the start. That makes things easier. I think it should just be a matter of checking for the relevant `google.rpc.Code` in [here](https://github.com/grpc-ecosystem/grpc-gateway/blob/f81c435e72054a3fd141c7c9a9d64de291f363de/runtime/handler.go#L203-L217) and adding the header. That should hopefully make the tests pass.

examples/internal/proto/examplepb/echo_service.proto
momom-i: I understand, thank you.

momom-i:
@johanbrandhorst 
Thank you for your comment.

> I think it should just be a matter of checking for the relevant google.rpc.Code in here and adding the header. That should hopefully make the tests pass.

I think we need to implement this on `DefaultHTTPErrorHandler` but please correct me if I'm wrong. 
I'm not sure about value of www-authenticate header. For now I set `status.Status.String()` as value.

johanbrandhorst:
> I think we need to implement this on `DefaultHTTPErrorHandler` but please correct me if I'm wrong.

You are right, that seems to be the right place.

> I'm not sure about value of www-authenticate header. For now I set `status.Status.String()` as value.
> I'd appreciate if you'd tell me what we should set here.

This is a great question, and I was looking back a bit at the original issue (#917). It's not entirely clear what we should set by default here, to be spec compliant it has to include a scheme and maybe a realm (from https://datatracker.ietf.org/doc/html/rfc7235#section-4.1). Scheme appears to be entirely up to the server, with common examples being `Basic`, `Bearer`, etc (https://www.iana.org/assignments/http-authschemes/http-authschemes.xhtml is a good list). I think maybe using `status.Message()` (instead of `status.String()`) would suffice, as it would allow users to customize the header value using the status message, though it would also mean that most users would still be non-compliant when we merge this.

I would like to hear some thoughts from other interested parties like @jefferai, @philippgille, what do you think the default behavior here should be? Would using `status.Message()`, even though it most likely won't be formatted correctly for the header, make sense for your use cases?

runtime/errors.go
```diff
@@ -108,6 +108,10 @@ func DefaultHTTPErrorHandler(ctx context.Context, mux *ServeMux, marshaler Marsh
 	contentType := marshaler.ContentType(pb)
 	w.Header().Set("Content-Type", contentType)
 
+	if s.Code() == codes.Unauthenticated {
+		w.Header().Set("WWW-Authenticate", s.String())
```
johanbrandhorst, line 112: I would use `s.Message()`, not the whole string.

momom-i:
@googlebot I signed it!

johanbrandhorst (changes requested):
Looks like the files need to be regenerated after updating the comment

johanbrandhorst (approved):
This LGTM but I'm wondering if this new implicit behavior is going to annoy users. I really don't know what a sensible default here would be. I will leave this open for feedback for another week or so and then we can merge it. I'm guessing we'll get more feedback after merging 😁.

momom-i:
I understand, thank you for your LGTM! I'm waiting for merging 👍

jefferai:
I checked what we're doing. I had thought we'd been manually adding www-authenticate to comply with RFC requirements but...we are not, at least according to `curl -vvv` :-)

I say this without knowing what status.Status.Message() actually outputs, but my overall $0.02 is that even common authentication schemes have a lot of wiggle room in them (e.g. Bearer is basically opaque other than the header starting with Bearer) so my main concern would be the ability to override the default -- either by turning this off, or providing some other values within the response.

For instance, the RFC mandates that WWW-Authenticate must be provided, but the Bearer RFC (https://www.rfc-editor.org/rfc/rfc6750.html) mandates that in addition to Bearer you MUST provide one or more auth-param values (realm, scope, error), and then for the most part helpfully nopes out on defining what those should look like. I have no idea what we'd put in realm that is meaningful in any way. We could put anything we wanted in scope, so it's likely also not meaningful, but would satisfy the "must contain one or more" condition. Or we could just put something in error like "not authenticated". And more importantly, I don't know the behavior of browsers or clients based on what we'd return here, although my suspicion is "basically nothing because they all likely ignore it" (which makes me wonder if this PR is motivated simply to follow RFC or because of an actual problem that's being seen).

So my overall point is: it's nice that grpc-gateway is getting support for returning _something_ in this header, but please be very wary of mandating _what_ is returned in the header.

johanbrandhorst:
Thanks a lot for the input Jeff, I would add one data point - I know that the docker engine client (not the CLI client) reads and parses the WWW-Authenticate response from registries it talks to, and uses it to determine how to proceed with authentication. I've gotten it to do what I want by sending back `Www-Authenticate: Basic`. 🤷🏻.

johanbrandhorst:
I'm leaning towards just shipping this to see how users react. `status.Message()` will be equal to whatever the `google.rpc.Status` message is that was encoded in the error received from the backend. Presumably if the backend can send a `codes.Unauthenticated` error code, they can also change the message. Ergo, we can assume that something about the message will be relevant to the user, and also that the wrapped server has some degree of control over the message, and can thus change the contents of the `Www-Authenticate` header by modifying the message. I think we can merge this, ship v2.6.0 and hopefully see what our users make of it, potentially adding some way to override the use of `status.Message` via some well-known header or something.

johanbrandhorst:
Thanks for your contribution :)!

PR #2332, opened 2021-09-09
Decode path-encoded URL components

v3n:
This change causes pct-encoded characters passed via path parameters to
be correctly decoded as described in google.api.http (see: path template
syntax) and as implemented in grpc-http-json-transcoding.

A new configuration option is introduced, `WithDecodeMode()`, which
understands several modes. Backwards compatibility is maintained, with
the hope of UnescapingModeAllExceptReserved becoming the default mode in
V3.

Fixes #660

Similiar to #2259, with a few significant differences: preserves backwards compatibility and strictly implements the spec describe in google.api.http and implemented in grpc-http-transcoding.

Yes.

- Backwards compatibility is maintained via `UnescapingModeLegacy`, which short-circuits the path encoding flow.
- Strictly implements the spec as described in the google.api.http, if the template is multi-segment, %2F _should not_  decoded into `/`.
- Follows the reference implementation at [grpc-http-json-transcoding](https://github.com/grpc-ecosystem/grpc-httpjson-transcoding/blob/master/src/include/grpc_transcoding/path_matcher.h)
    - Provides multiple user-configurable decoding modes  (`UnescapingModeAllExceptReserved`, `UnescapingModeAllExceptSlash`, `UnescapingModeDefault`, and `UnescapingModeLegacy`)
    - Decoding occurs only for variable binding, routing does not use decoding.

#### Other comments

n/a

### DIFF
--- a/runtime/mux.go
+++ b/runtime/mux.go
@@ -286,7 +286,6 @@ func (s *ServeMux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 		pathParams, err := h.pat.MatchAndEscape(components, verb, s.unescapingMode)
 		if err != nil {
-			// malformed escape sequence
 			if err == ErrMalformedSequence {
 				_, outboundMarshaler := MarshalerForRequest(s, r)
 				s.routingErrorHandler(ctx, s, outboundMarshaler, w, r, http.StatusBadRequest)
@@ -306,7 +305,6 @@ func (s *ServeMux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		for _, h := range handlers {
 			pathParams, err := h.pat.MatchAndEscape(components, verb, s.unescapingMode)
 			if err != nil {
-				// malformed escape sequence
 				if err == ErrMalformedSequence {
 					_, outboundMarshaler := MarshalerForRequest(s, r)
 					s.routingErrorHandler(ctx, s, outboundMarshaler, w, r, http.StatusBadRequest)
--- a/runtime/mux.go
+++ b/runtime/mux.go
@@ -14,6 +14,31 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
+// UnescapingMode defines the behavior of grpc-gateway for URL escaping.
+type UnescapingMode int
+
+const (
+	// UnescapingModeLegacy is the default V2 behavior, which escapes the entire
+	// path string before doing any routing.
+	UnescapingModeLegacy UnescapingMode = iota
+
+	// EscapingTypeExceptReserved unescapes all path parameters except RFC 6570
+	// reserved characters.
+	UnescapingModeAllExceptReserved
+
+	// EscapingTypeExceptSlash unescapes URL path parameters except path
+	// seperators, which will be left as "%2F".
+	UnescapingModeAllExceptSlash
+
+	// URL path parameters will be fully decoded.
+	UnescapingModeAllCharacters
+
+	// UnescapingModeDefault is the default escaping type.
+	// TODO(v3): default this to UnescapingModeAllExceptReserved per grpc-httpjson-transcoding's
+	// reference implementation
+	UnescapingModeDefault = UnescapingModeLegacy
+)
+
 // A HandlerFunc handles a specific pair of path pattern and HTTP method.
 type HandlerFunc func(w http.ResponseWriter, r *http.Request, pathParams map[string]string)
 
@@ -31,6 +56,7 @@ type ServeMux struct {
 	streamErrorHandler        StreamErrorHandlerFunc
 	routingErrorHandler       RoutingErrorHandlerFunc
 	disablePathLengthFallback bool
+	unescapingMode            UnescapingMode
 }
 
 // ServeMuxOption is an option that can be given to a ServeMux on construction.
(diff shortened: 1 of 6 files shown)


runtime/mux.go
```diff
@@ -244,8 +284,13 @@ func (s *ServeMux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			components[l-1], verb = lastComponent[:idx], lastComponent[idx+1:]
 		}
 
-		pathParams, err := h.pat.Match(components, verb)
+		pathParams, err := h.pat.MatchAndEscape(components, verb, s.unescapingMode)
 		if err != nil {
+			// malformed escape sequence
```
adambabik, line 289: Nitpick: likely can be omitted as the error name is pretty clear.

adambabik (commented):
Thank you for the contribution, it looks really solid!

The only thing I can comment on is to add some test cases. For the changes in `runtime/mux.go`, you can extend `TestMuxServeHTTP`. For the `runtime/pattern.go` there is `TestMatch`.

v3n:
Added tests, I think ready for merge unless there are more follow-ups?

runtime/mux.go
```diff
@@ -14,6 +14,31 @@ import (
 	"google.golang.org/protobuf/proto"
 )
 
+// UnescapingMode defines the behavior of grpc-gateway for URL escaping.
```
johanbrandhorst, line 17: Nit:
```suggestion
// UnescapingMode defines the behavior of the grpc-gateway for URL escaping.
```

runtime/mux.go
```diff
@@ -244,8 +284,12 @@ func (s *ServeMux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			components[l-1], verb = lastComponent[:idx], lastComponent[idx+1:]
 		}
 
-		pathParams, err := h.pat.Match(components, verb)
+		pathParams, err := h.pat.MatchAndEscape(components, verb, s.unescapingMode)
 		if err != nil {
+			if err == ErrMalformedSequence {
```
johanbrandhorst, line 289: What do you think about using `errors.Is` instead?
```suggestion
			if errors.Is(err,  ErrMalformedSequence) {
```

runtime/mux.go
```diff
@@ -259,8 +303,12 @@ func (s *ServeMux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 			continue
 		}
 		for _, h := range handlers {
-			pathParams, err := h.pat.Match(components, verb)
+			pathParams, err := h.pat.MatchAndEscape(components, verb, s.unescapingMode)
 			if err != nil {
+				if err == ErrMalformedSequence {
```
johanbrandhorst, line 308: Ditto errors.Is.

runtime/pattern.go
```diff
@@ -140,10 +142,11 @@ func MustPattern(p Pattern, err error) Pattern {
...
 }
 
-// Match examines components if it matches to the Pattern.
-// If it matches, the function returns a mapping from field paths to their captured values.
-// If otherwise, the function returns an error.
-func (p Pattern) Match(components []string, verb string) (map[string]string, error) {
+// MatchAndEscape examines components if it matches to the Pattern. If it matches,
+// the function returns a mapping from field paths to their captured values while
+// applying the provided unescaping mode, returning an error if the URL encoding
+// is malformed. Otherwise, the function returns an error.
```
johanbrandhorst, line 148: This comment made me think we sometimes return both an error and a map.
```suggestion
// MatchAndEscape examines components to determine if it matches to the Pattern.
// If it matches, the function returns a mapping from field paths to their captured
// values while applying the provided unescaping mode, returning an error if the
// URL encoding is malformed.
```

runtime/pattern.go
```diff
@@ -204,6 +217,15 @@ func (p Pattern) Match(components []string, verb string) (map[string]string, err
 	return bindings, nil
 }
 
+// Match examines components if it matches to the Pattern.
+// If it matches, the function returns a mapping from field paths to their captured values.
+// If otherwise, the function returns an error.
+//
+// Deprecated: Use MatchAndEscape.
```
johanbrandhorst, line 224: I know this was the existing comment, but lets reword it.
```suggestion
// Match examines components to determine if it matches to the Pattern.
// If it matches, the function returns a mapping from field paths to their captured values.
// Otherwise, the function returns an error.
//
// Deprecated: Use MatchAndEscape.
```

johanbrandhorst (changes requested):
This looks great, a few style comments is all I've got. I'm wondering if you could add a short blurb to our documentation to explain the nuances of these new settings as well? Maybe a section in https://github.com/grpc-ecosystem/grpc-gateway/blob/master/docs/docs/mapping/customizing_your_gateway.md?

runtime/mux.go
v3n: 100% for it, was just matching the rest of the source. Will update it and propagate a descriptive error message.

runtime/mux.go
v3n: I'll change grpc-gateway to ServeMux, since we're talking about a ServeMux option.

runtime/pattern.go
v3n: Happy to, how about:

```
// MatchAndEscape examines components to determine if they match to a Pattern. It will
// never perform path unescaping (UnescapeModeLegacy). MatchAndEscape will return an error
// if no Patterns matched. If successful, the function returns a mapping from field paths to their 
// captured values.
//
// Deprecated: Use MatchAndEscape.
```

runtime/pattern.go
v3n: How about:

```
// MatchAndEscape examines components to determine if they match to a Pattern.
// MatchAndEscape will return an error if no Patterns matched or if a pattern matched
// but contained malformed escape sequences. If successful, the function returns a 
// mapping from field paths to their captured values.
```

runtime/mux.go
v3n: I changed to errors.As and propagating the real error (which indicates the bad escape sequence), let me know if that looks okay.

johanbrandhorst (approved):
Fantastic, this looks good to me!

johanbrandhorst:
Looks like one of the files still needs to be `gofmt`ed

adambabik (approved):
👏

v3n:
Whoops! Should be fixed now.

johanbrandhorst:

PR #2356, opened 2021-09-27
populate updateMask field: init (using reflect) then set paths

dungnx:

Yes

Support [Gogo Protobuf](https://github.com/gogo/protobuf) field_mask.
If GogoProtobuf is used, our code gen `.gw.go` will not be compiled due to different type: `runtime.FieldMaskFromRequestBody` returns `google.protobuf.FieldMask` while the field type is [GogoProto type](https://github.com/gogo/protobuf/blob/master/types/field_mask.pb.go). 
This is done by initializing empty pointer of the field (using reflect) then set `Paths`.


dungnx-teko:
@googlebot I consent

johanbrandhorst (commented):
Hi, this is a really neat fix, but I'm confused, I wouldn't think that it would be possible to use gogo/protobuf with v2 of the gateway. Can you explain? Or maybe you would like to make this pull request against the v1 branch?

dungnx:
> Hi, this is a really neat fix, but I'm confused, I wouldn't think that it would be possible to use gogo/protobuf with v2 of the gateway. Can you explain? Or maybe you would like to make this pull request against the v1 branch?

Oops, I hadn't known that it's not possible to use gogo/protobuf with grpc-gateway v2. Can you explain why? I will change the PR against the v1 branch instead.

johanbrandhorst:
> > Hi, this is a really neat fix, but I'm confused, I wouldn't think that it would be possible to use gogo/protobuf with v2 of the gateway. Can you explain? Or maybe you would like to make this pull request against the v1 branch?
> 
> Oops, I hadn't known that it's not possible to use gogo/protobuf with grpc-gateway v2. Can you explain why? I will change the PR against the v1 branch instead.

gogo/protobuf is not compatible with the Go protobuf API v2, which the `master` branch uses.The `v1` branch uses the v1 API which is mostly compatible with gogo/protobuf.

PR #2367, opened 2021-10-01
Fix timestamp string conversion

### LINKED ISSUE #1683: Using timestamps in URL path parameters
## 🐛 Bug Report

When passing a timestamp (`google/protobuf/timestamp.proto`) as a path parameter in the URL, the gateway is not able to correctly deserialize the value.

## To Reproduce

1. Generate an api-gateway wrapping the following protobuf service:
```
message GoogleTypes {
    google.protobuf.Timestamp time = 1;
}
service TestAPI {
    rpc GooglePathParameter (GoogleTypes) returns (google.protobuf.Empty) {
        option (google.api.http) = {
            post: "/test/parameter/path/google/{time}"
            body: "*"
        };
    }
}
```
2. Perform a POST-request passing a timestamp within the URL.
Note: To correctly path-encode the timestamp, the following go code can be used:
```go
ts,_ := time.Now()
formatted := ts.Format(time.RFC3339Nano)
url := fmt.Sprintf("/test/parameter/path/google/%s", url.PathEscape(formatted))
```
This results in the following URL:
`.../test/parameter/path/google/2020-09-22T15:00:00Z`

3. When performing the call, I get a protobuf conversion error.
This is the response body:
`{"code":3,"message":"type mismatch, parameter: time, error: proto: syntax error (line 1:1): invalid value 2020-09-22T15","details":[]}`

Even when replacing the colon ':' with `%3A` (which should not be neccessary), the same error happens.
I assume the URL is split by colons somewhere within the gateway, causing only half of the timestamp to be forwarded to the protobuf unmarshaller?

Note that passing timestmaps in the request body works correctly.

## Your Environment

Im using `github.com/grpc-ecosystem/grpc-gateway/v2 v2.0.0-beta.4`

momom-i:
Fixes #1683

Yes
Now request contains timestamp causes error when convert timestamp string is passed
because a byte string without double quotations can't pass `unmarshaler.Unmarshal()`.


runtime/convert.go
```diff
@@ -207,6 +207,7 @@ func BytesSlice(val, sep string) ([][]byte, error) {
 // Timestamp converts the given RFC3339 formatted string into a timestamp.Timestamp.
 func Timestamp(val string) (*timestamppb.Timestamp, error) {
 	var r timestamppb.Timestamp
+	val = fmt.Sprintf(`"%s"`, strings.Trim(val, `"`))
```
johanbrandhorst, line 210: We could use `strconv.Quote` here instead of `fmt.Sprintf`.

runtime/convert.go
```diff
@@ -218,6 +219,7 @@ func Timestamp(val string) (*timestamppb.Timestamp, error) {
 // Duration converts the given string into a timestamp.Duration.
 func Duration(val string) (*durationpb.Duration, error) {
 	var r durationpb.Duration
+	val = fmt.Sprintf(`"%s"`, strings.Trim(val, `"`))
```
johanbrandhorst, line 222: Same here

johanbrandhorst (changes requested):
Is there any way we could more directly test this behaviour than via the runtime conversion tests? Could we add some tests to https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/mux_test.go#L15?

runtime/convert.go
momom-i: I understand, thank you!

momom-i:
Thank you for your review. I've tried some but I can't find a way to go through functions of convert.go in `mux_test.go`. I'd happy if you'd show me how to do that. 🙇

momom-i:
@googlebot I fixed it.

johanbrandhorst:
> Thank you for your review. I've tried some but I can't find a way to go through functions of convert.go in `mux_test.go`. I'd happy if you'd show me how to do that. bow

I had a good look and I think you're right, that's probably not possible. So this PR looks good to me! The last thing we need to do is make the CLA bot happy. Unfortunately it looks like the first commit you submitted couldn't be identified by GitHub, so the CLA bot is confused about who the author is. If you squash your commits and force push it should fix it again.

momom-i:
@googlebot I fixed it.

momom-i:
Thank you for your kind instruction! google cla finally said to me "yes". 😄

johanbrandhorst:
Not sure what's up with the lint job, but this looks fine!

PR #2422, opened 2021-11-15
chore(examples): some modifications

alissa-tung:
The current version of `BulkEcho` server is not streaming enough, namely, it can not do `Send` data once `Recv`ed.


alissa-tung:
@googlebot I signed it!

examples/internal/server/a_bit_of_everything.go
```diff
@@ -260,37 +260,19 @@ func (s *_ABitOfEverythingServer) Echo(ctx context.Context, msg *sub.StringMessa
...
 		if err != nil {
-			return err
-		}
-		msgs = append(msgs, msg)
-	}
-
-	hmd := metadata.New(map[string]string{
-		"foo": "foo1",
-		"bar": "bar1",
-	})
```
johanbrandhorst, line 278: Looks like the new version of this function isn't setting this header.

examples/internal/server/a_bit_of_everything.go
```diff
@@ -260,37 +260,19 @@ func (s *_ABitOfEverythingServer) Echo(ctx context.Context, msg *sub.StringMessa
...
-		if err := stream.Send(msg); err != nil {
-			return err
+			return status.Errorf(codes.Internal, "unexpected error sending to stream: %s", err)
 		}
 	}
-
-	stream.SetTrailer(metadata.New(map[string]string{
-		"foo": "foo2",
-		"bar": "bar2",
-	}))
```
johanbrandhorst, line 293: Or this trailer.

johanbrandhorst (changes requested):
I pushed some changes that were causing the tests and generation to fail. Still a bit confused about the rationale for this change. What made you want to change this?

alissa-tung:
> I pushed some changes that were causing the tests and generation to fail. Still a bit confused about the rationale for this change. What made you want to change this?

Oh sorry, my original intention is to make the semantic changed first and then fix the metadata if they are not going wrong.
Thanks for working on this. If i am not misunderstanding, the original impl do something like collection all msgs til EOF occur, and range them and send back.

A simple test is like
```sh
#!/usr/bin/sh
for _ in $(seq 10)
do xs="{\"value\": \"$(date +%s)\"}"
   echo "$xs"
   sleep 1
done | cat |
strace -t                                 \
  -e sendto,read -o /tmp/streaming.log    \
    curl --trace-ascii - -k               \
      -X POST                             \
      -H "Content-Type: application/json" \
      -H "Transfer-Encoding: chunked"     \
      -T -                                \
        'http://127.0.0.1:6580/echo'
```

which got the result like

```
==>
==>
==>
==>
... ...
<==
<==
<==
<==
```

instead of return immediately

```
==>
<==
==>
<==
... ...
==>
<==
```

I am not sure about which is the properly streaming model it would be demonstrate there.

alissa-tung:
Seems that grpc-go put the recv-send procedure in one for too
https://github.com/grpc/grpc-go/blob/b2317c762757d277fea21e68c714c6b423675570/examples/features/cancellation/server/main.go#L40

johanbrandhorst:
I don't think we need to change this. It's being used in some tests. It's not really supposed to show off anything. I'm leaning towards rejecting this PR if there isn't a bug here.

alissa-tung:
Okay, thx.

PR #2428, opened 2021-11-16
fix(tutorials): remove usage of `WithBlock` in examples

### LINKED ISSUE #2425: examples: remove usage of `WithBlock` in examples
## 📚 Documentation

ref: https://github.com/grpc/grpc-go/pull/4858

alissa-tung:
close #2425

### DIFF
--- a/docs/docs/tutorials/adding_annotations.md
+++ b/docs/docs/tutorials/adding_annotations.md
@@ -163,7 +163,6 @@ func main() {
 	conn, err := grpc.DialContext(
 		context.Background(),
 		"0.0.0.0:8080",
-		grpc.WithBlock(),
 		grpc.WithInsecure(),
 	)
 	if err != nil {


johanbrandhorst:
Seeing this in context, I'm not sure we should be making this change. `WithBlock` makes sense here because we want to make sure the gRPC connection has been established when we start the gateway. With that in mind, I'm going to close this and the issue. Please let me know if you disagree.

alissa-tung:
> Seeing this in context, I'm not sure we should be making this change. `WithBlock` makes sense here because we want to make sure the gRPC connection has been established when we start the gateway. With that in mind, I'm going to close this and the issue. Please let me know if you disagree.

That's okay. The first time it confused me so I file the issue for is that I haven't see the log in the code appear (which is `"Serving gRPC-Gateway on http://0.0.0.0:8090"`). At least, theses issues may help ppl who also search about it.

Issue #2447, opened 2021-12-01
Forwarding Connection header breaks after updating grpc/grpc-go to 1.42.0


MakDon:
## 🐛 Bug Report

grpc-gateway breaks after updating `grpc/grpc-go` to 1.42.0

After the update introduced by [this PR](https://github.com/grpc/grpc-go/pull/4803),  it will report `stream terminated by RST_STREAM with error code: PROTOCOL_ERROR` if `connection` in headers, which breaks down all of the http connections that comes with `connect:keep-alive`

## To Reproduce

Step 1:
    Update `google.golang.org/grpc` to `v1.42.0` in go mod

Step 2：
    invoke http request with header `connect:keep-alive`

## Expected behavior

grpc-gateway works well or reports warning, instead of returning `PROTOCOL_ERROR`

## Actual Behavior

it returns `stream terminated by RST_STREAM with error code: PROTOCOL_ERROR`

## Your Environment

go.mod: 
`google.golang.org/grpc v1.42.0`, `github.com/grpc-ecosystem/grpc-gateway/v2 v2.7.0`

johanbrandhorst:
Hm, this is curious, what should the behavior be here? Should the grpc-gateway strip these headers? I'm surprised we're forwarding this header, it should only forward explicitly added headers or permanent IANA headers (listed here: https://github.com/grpc-ecosystem/grpc-gateway/blob/094a6fe78b3ca888297d090185cdf30f0e42e157/runtime/context.go#L261). Would you be willing to submit a test and a fix for this?

MakDon:
> Hm, this is curious, what should the behavior be here? Should the grpc-gateway strip these headers? I'm surprised we're forwarding this header, it should only forward explicitly added headers or permanent IANA headers (listed here:
> 
> https://github.com/grpc-ecosystem/grpc-gateway/blob/094a6fe78b3ca888297d090185cdf30f0e42e157/runtime/context.go#L261
> 
> ). Would you be willing to submit a test and a fix for this?

ok, I'd like to submit a PR for it later.

MakDon:
> Hm, this is curious, what should the behavior be here? Should the grpc-gateway strip these headers? I'm surprised we're forwarding this header, it should only forward explicitly added headers or permanent IANA headers (listed here:
> 
> https://github.com/grpc-ecosystem/grpc-gateway/blob/094a6fe78b3ca888297d090185cdf30f0e42e157/runtime/context.go#L261
> 
[...]

PR #2455, opened 2021-12-04
Log warning if HeaderMatcherFunc passes malformed HTTP headers to grpc server

MakDon:

Solves #2447 

Log warning if the custom header matcher function returns "malformed" headers that would be rejected be grpc server.    
It will help others who is using passthrough HeaderMatcherFunc avoid breaking their server after they updating to grpc-gateway 2.7.x.

To re-produce this warning, use a custom HeaderMatcherFunc like following in example-gateway-server/main.go:
```go
func main() {
	flag.Parse()
	defer glog.Flush()

	ctx := context.Background()
	opts := gateway.Options{
		Addr: ":8080",
		GRPCServer: gateway.Endpoint{
			Network: *network,
			Addr:    *endpoint,
		},
		OpenAPIDir: *openAPIDir,
		Mux: []runtime.ServeMuxOption{
			// a custom passthroughs HeaderMatcher
			runtime.WithIncomingHeaderMatcher(func(key string) (string, bool) {return key, true}),
		},
	}
	if err := gateway.Run(ctx, opts); err != nil {
		glog.Fatal(err)
	}
}
```

I suppose this pr should be backport to 2.6.x or earlier.

And i suggest that we can provide a passthrough HeaderMatcherFunc, which passes the valid headers to grpc metadata as-is. It could help backward compatibility because devs can transform from http restful api to grpc without updating their headers.

### DIFF
--- a/runtime/context.go
+++ b/runtime/context.go
@@ -41,6 +41,13 @@ var (
 	DefaultContextTimeout = 0 * time.Second
 )
 
+// malformedHTTPHeaders are headers would be considered malformed and be rejected by grpc server
+// Connection: https://github.com/grpc/grpc-go/pull/4803
+var malformedHTTPHeaders = map[string]struct{}{
+	"Connection": {},
+	"connection": {},
+}
+
 type (
 	rpcMethodKey       struct{}
 	httpPathPatternKey struct{}
@@ -306,6 +313,13 @@ func isPermanentHTTPHeader(hdr string) bool {
 		return true
 	}
 	return false
+}
+
+// isMalformedHTTPHeaders checks whether hdr belongs to the list of
+// "malformed headers" and would be rejected by grpc server
+func isMalformedHTTPHeaders(hdr string) bool {
+	_, isMalFormed := malformedHTTPHeaders[hdr]
+	return isMalFormed
 }
 
 // RPCMethod returns the method string for the server context. The returned
--- a/runtime/context.go
+++ b/runtime/context.go
@@ -41,10 +41,9 @@ var (
 	DefaultContextTimeout = 0 * time.Second
 )
 
-// malformedHTTPHeaders are headers would be considered malformed and be rejected by grpc server
-// Connection: https://github.com/grpc/grpc-go/pull/4803
+// malformedHTTPHeaders lists the headers that the gRPC server may reject outright as malformed.
+// See https://github.com/grpc/grpc-go/pull/4803#issuecomment-986093310 for more context.
 var malformedHTTPHeaders = map[string]struct{}{
-	"Connection": {},
 	"connection": {},
 }
 
@@ -315,11 +314,11 @@ func isPermanentHTTPHeader(hdr string) bool {
 	return false
 }
 
-// isMalformedHTTPHeaders checks whether hdr belongs to the list of
-// "malformed headers" and would be rejected by grpc server
-func isMalformedHTTPHeaders(hdr string) bool {
-	_, isMalFormed := malformedHTTPHeaders[hdr]
-	return isMalFormed
+// isMalformedHTTPHeader checks whether header belongs to the list of
+// "malformed headers" and would be rejected by the gRPC server.
+func isMalformedHTTPHeader(header string) bool {
+	_, isMalformed := malformedHTTPHeaders[strings.ToLower(header)]
+	return isMalformed
 }
 
 // RPCMethod returns the method string for the server context. The returned
--- a/runtime/mux.go
+++ b/runtime/mux.go
@@ -115,16 +115,28 @@ func DefaultHeaderMatcher(key string) (string, bool) {
 // passed to gRPC context. To transform the header before passing to gRPC context, matcher should return modified header.
 func WithIncomingHeaderMatcher(fn HeaderMatcherFunc) ServeMuxOption {
 
-	for hdr := range malformedHTTPHeaders {
-		out, accept := fn(hdr)
-		if accept && isMalformedHTTPHeaders(out) {
-			grpclog.Warningf("malformed HTTP header %s would be sent to grpc server", out)
-		}
+	for _, hdr := range fn.forwardingMalformedHdr() {
+		grpclog.Warningf("malformed HTTP header %s would be sent to grpc server", hdr)
 	}
 
 	return func(mux *ServeMux) {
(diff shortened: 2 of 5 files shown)


MakDon:
I am stuck with testing cause it is a private function and can not be called from runtime_test package, and not sure if i need to create a test file in package `runtime`

johanbrandhorst:
> I am stuck with testing cause it is a private function and can not be called from runtime_test package, and not sure if i need to create a test file in package `runtime`

Feel free to create a new runtime_internal_test.go to test these private methods 👍🏻.

runtime/context.go
```diff
@@ -41,6 +41,13 @@ var (
 	DefaultContextTimeout = 0 * time.Second
 )
 
+// malformedHTTPHeaders are headers would be considered malformed and be rejected by grpc server
+// Connection: https://github.com/grpc/grpc-go/pull/4803
+var malformedHTTPHeaders = map[string]struct{}{
```
johanbrandhorst, line 46:
```suggestion
// malformedHTTPHeaders lists the headers that the gRPC server may reject outright as malformed.
// See https://github.com/grpc/grpc-go/pull/4803#issuecomment-986093310 for more context.
var malformedHTTPHeaders = map[string]struct{}{
```

runtime/context.go
```diff
@@ -41,6 +41,13 @@ var (
...
+	"Connection": {},
+	"connection": {},
```
johanbrandhorst, line 48: HTTP headers are not case sensitive, so we can just transform them to lower case when making the check.
```suggestion
	"connection": {},
```

runtime/context.go
```diff
@@ -308,6 +315,13 @@ func isPermanentHTTPHeader(hdr string) bool {
 	return false
 }
 
+// isMalformedHTTPHeaders checks whether hdr belongs to the list of
+// "malformed headers" and would be rejected by grpc server
+func isMalformedHTTPHeaders(hdr string) bool {
+	_, isMalFormed := malformedHTTPHeaders[hdr]
+	return isMalFormed
+}
```
johanbrandhorst, line 323:
```suggestion
// isMalformedHTTPHeader checks whether header belongs to the list of
// "malformed headers" and would be rejected by the gRPC server.
func isMalformedHTTPHeader(header string) bool {
	_, isMalformed := malformedHTTPHeaders[strings.ToLower(header)]
	return isMalformed
}
```

runtime/mux.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"google.golang.org/grpc/grpclog"
```
johanbrandhorst, line 7: This import belongs with the other third party imports

runtime/mux.go
```diff
@@ -113,11 +114,31 @@ func DefaultHeaderMatcher(key string) (string, bool) {
 // This matcher will be called with each header in http.Request. If matcher returns true, that header will be
 // passed to gRPC context. To transform the header before passing to gRPC context, matcher should return modified header.
 func WithIncomingHeaderMatcher(fn HeaderMatcherFunc) ServeMuxOption {
+
```
johanbrandhorst, line 117: Remove this whitespace please
```suggestion
```

runtime/mux.go
```diff
@@ -113,11 +114,31 @@ func DefaultHeaderMatcher(key string) (string, bool) {
...
+	for _, hdr := range fn.forwardingMalformedHdr() {
+		grpclog.Warningf("malformed HTTP header %s would be sent to grpc server", hdr)
+	}
```
johanbrandhorst, line 120:
```suggestion
	for _, header := range fn.forwardingMalformedHdr() {
		grpclog.Warningf("The configured forwarding filter would allow %q to be sent to the gRPC server, which will likely cause errors. See https://github.com/grpc/grpc-go/pull/4803#issuecomment-986093310 for more information.", header)
	}
```

runtime/mux.go
```diff
@@ -113,11 +114,31 @@ func DefaultHeaderMatcher(key string) (string, bool) {
...
+
 	return func(mux *ServeMux) {
 		mux.incomingHeaderMatcher = fn
 	}
 }
 
+// forwardingMalformedHdr returns malformed headers that would be forward to grpc server
+func (fn HeaderMatcherFunc) forwardingMalformedHdr() []string {
```
johanbrandhorst, line 128:
```suggestion
// matchedMalformedHeaders returns the malformed headers that would be forwarded to gRPC server.
func (fn HeaderMatcherFunc) matchedMalformedHeaders() []string {
```

runtime/mux.go
```diff
@@ -113,11 +114,31 @@ func DefaultHeaderMatcher(key string) (string, bool) {
...
+		return nil
+	}
+	hdrs := make([]string, 0)
+	for hdr := range malformedHTTPHeaders {
+		out, accept := fn(hdr)
+		if accept && isMalformedHTTPHeaders(out) {
+			hdrs = append(hdrs, out)
+		}
+	}
+	return hdrs
```
johanbrandhorst, line 139:
```suggestion
	headers := make([]string, 0)
	for header := range malformedHTTPHeaders {
		out, accept := fn(header)
		if accept && isMalformedHTTPHeader(out) {
			headers = append(headers, out)
		}
	}
	return headers
```

johanbrandhorst (commented):
This is a great start, thank you! I left some minor style suggestions.

runtime/mux.go
MakDon: My bad. I've move it to third party imports groupy

MakDon:
Hi @johanbrandhorst , i've made changes following your suggestion. And tests are added also. Could you review again? Thanks very much. :D

MakDon:
Ci failed cause Bazel BUILD files are not up-to-date. I'am working on it.

johanbrandhorst:
https://github.com/grpc-ecosystem/grpc-gateway/blob/master/CONTRIBUTING.md#i-want-to-regenerate-the-files-after-making-changes might help wit the bazel issue.

MakDon:
I stash my local changes and regenerate the files again. It seems that the CI is not complaining now.

johanbrandhorst (approved):
LGTM

johanbrandhorst:
Thanks for your contribution!

PR #2461, opened 2021-12-09
Fixes #407 and #1700 - google.api.http path parameter constraints and multiple HTTP path/method endpoint support

### LINKED ISSUE #407: Path variables with segments may produce invalid Swagger file
For example:

```
service YourService {
  rpc Foo(StringMessage) returns (StringMessage) {
    option (google.api.http) = {
      post: "/v1/{name=examples/*/foos/*}"
      body: "*"
    };
  rpc Bar(StringMessage) returns (StringMessage) {
    option (google.api.http) = {
      post: "/v1/{name=examples/*/bars/*}"
      body: "*"
    };
  }
}
```

This produces conflicting paths in the YAML file: `/v1/{name}` twice; the pattern of the variable is not captured in the swagger spec.

### LINKED ISSUE #1700: [Docs/Examples] No docs/examples for path parameter curly bracket syntax
## 📚 Documentation

Looking at https://github.com/grpc-ecosystem/grpc-gateway/issues/720
it seems the path patterns support some sort of `{foo=bar}` templating, but I can't seem to find any docs about it, what it does, how it works, or it's use in any examples.

Furthermore, there is no indication of where users can ask questions (whether it's the issue tracker, forum, mailing list etc).

Hence I am dumping it here:

I'd like to implement a rpc method with optional parameters for paging.
Namely:
```
GET /books = returns all books
GET /books/{page}/{per_page} = returns paginated books
```

I assume I might be able to do something like:

```
get: "/books/{page=0}/{per_page=0}
```

But it seems the swagger spec generated has both parameters as required and not optional and default value is nowhere to be seen, so I assume this doesn't do what I want.

Does this really need to be two separate grpc methods?
I guess I could have `additional_bindings` with the path parameters, but then I get an ugly GetBooks, GetBooks2 operations.
Is there a way to control the operation name for the additional binding, so I could call it GetBooksPaginated or something?

betmix-matt:
Fixes #407
Fixes #1700

This PR changes the behavior of the protoc-gen-openapiv2 plugin when multiple gRPC operations match the same HTTP path and method and include a path parameter.  Previously, all path parameters were converted from their google.api.http syntax to OpenAPI syntax (i.e. `/v1/{bar=foos/*}` to `/v1/{bar}`) unless the parameter was named `name` or `parent`.  This caused invalid OpenAPI specifications when `name` or `parent` path params existed with constraints as OpenAPI does not permit that syntax.

See https://github.com/googleapis/googleapis/blob/master/google/api/http.proto

There is a small "breaking change" in that with previous versions of the plugin, if multiple gRPC service methods had the same HTTP path and method, the last one specified was the result (i.e. last in wins).  This PR changes that behavior to now log an error if duplicate mappings exist but maintains the legacy behavior of "last in wins".

With this PR, when a given gRPC service method has an identical OpenAPI HTTP endpoint to another service, the plugin will now rename the path parameter on the 2nd service adding a _1 suffix (or _2 etc until it doesn't match an existing endpoint).   This will not change the actual contents of requests being sent to the endpoint as path parameter names are replaced with the values when making a request.

This PR also adds automatic "pattern" generation on path parameters that use the google.api.http path parameter constraints like `/v1/{name=foos/*}/bar` so that the OpenAPI specification properly matches the proto definition's requirements. It also fixes a bug where the "pattern" from proto query parameters was not being copied to the OpenAPI parameter.

Tests added for all changes.


johanbrandhorst:
Thanks for your contribution! I will review this in a bit, but note that you'll probably need to sign the Google CLA if you haven't already (see https://github.com/grpc-ecosystem/grpc-gateway/pull/2307#issuecomment-906183149 for instructions - not sure why the bot isn't replying to the PR directly anymore). It may be that you need to sign it and the bot still doesn't reply, in that case can you please share a screenshot of your registered signing of the CLA from the CLA website? I will get to looking into this soon!

betmix-matt:
All set now.

betmix-matt:
Screen Shot 2021-12-08 at 8 42 27 PM

betmix-matt:
I wanted to note I have used the updated version in my company's repository and it is now able to properly route the RFC 6570 format and enforced that the path parameter on the swagger page matched the given pattern.

protoc-gen-openapiv2/internal/genopenapi/template_test.go
```diff
@@ -30,6 +30,12 @@ import (
 
 var marshaler = &runtime.JSONPb{}
 
+func assertEqual(t *testing.T, testName string, a interface{}, b interface{}) {
```
johanbrandhorst, line 33: I appreciate the intention here, but this pattern is inconsistent with the rest of the code, and I don't think we should introduce it here. If you want to do this, we'll need one function for each type you want to test:
```go
func assertEqualInt(t *testing.T, testName string, a int, b int) {
```

protoc-gen-openapiv2/internal/genopenapi/template_test.go
```diff
@@ -4779,6 +4818,401 @@ func TestTemplateWithoutErrorDefinition(t *testing.T) {
...
+		t.Errorf("failed to reg.Load(): %v", err)
+		return
+	}
+	result, err := applyTemplate(param{File: crossLinkFixture(&file), reg: reg})
+	if err != nil {
+		t.Errorf("applyTemplate(%#v) failed with %v; want success", file, err)
+		return
+	}
+
+	assertEqual(t, "Results paths length", len(result.Paths), 4)
```
johanbrandhorst, line 5179: I would sooner just make this explicit than use assertion functions
```suggestion
	if got, want := len(result.Paths), 4; got != want {
		t.Fatalf("Results paths length differed, got %d want %d", got, want)
	}
```

runtime/mux.go
```diff
@@ -41,6 +42,10 @@ const (
 	UnescapingModeDefault = UnescapingModeLegacy
 )
 
+var (
+	EncodedPathSplitter = regexp.MustCompile("(/|%2F)")
```
johanbrandhorst, line 46: Does this need to be exported?

runtime/mux.go
```diff
@@ -265,7 +270,12 @@ func (s *ServeMux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		path = r.URL.RawPath
 	}
 
-	components := strings.Split(path[1:], "/")
+	var components []string
+	if s.unescapingMode == UnescapingModeAllCharacters || s.unescapingMode == UnescapingModeLegacy {
+		components = EncodedPathSplitter.Split(path[1:], -1)
+	} else {
+		components = strings.Split(path[1:], "/")
+	}
```
johanbrandhorst, line 278: What, if any, are the routing consequences of this change? Is this a change in behavior?

runtime/mux_test.go
```diff
@@ -408,15 +455,15 @@ func TestMuxServeHTTP(t *testing.T) {
 						t.Fatalf("runtime.NewPattern(1, %#v, %#v, %q) failed with %v; want success", p.ops, p.pool, p.verb, err)
 					}
 					mux.Handle(p.method, pat, func(w http.ResponseWriter, r *http.Request, pathParams map[string]string) {
-						fmt.Fprintf(w, "%s %s", p.method, pat.String())
+						_, _ = fmt.Fprintf(w, "%s %s", p.method, pat.String())
```
johanbrandhorst, line 458: Thanks for tidying up ❤️

johanbrandhorst (changes requested):
This is a really amazing contribution, thank you so much! I would ideally like to preserve the existing behavior here and add a warning log in the case of two matching routes.

I would like to see us remove the `assert` test function that was added as it's very at odds with the existing conventions of the test code, even if that makes the test verbose.

I'm also a little worried about the mux change. Could you please explain what the change is for?

betmix-matt:
We should be all set now! All requested changes have been incorporated.

protoc-gen-openapiv2/internal/genopenapi/template_test.go
```diff
@@ -58,6 +61,33 @@ func reqFromFile(f *descriptor.File) *pluginpb.CodeGeneratorRequest {
...
 
+// captureStderr executes the given function and returns the full string of what
+// was written to os.Stderr during execution and any error the function may have returned
+func captureStderr(f func() error) (string, error) {
+	old := os.Stderr // keep backup of the real stderr
+	r, w, err := os.Pipe()
+	if err != nil {
+		return "", err
+	}
+	os.Stderr = w
```
johanbrandhorst, line 72: This is one of the craziest things I've seen but I think I will allow it 😅.

runtime/mux.go
```diff
@@ -24,15 +25,15 @@ const (
...
 
-	// EscapingTypeExceptSlash unescapes URL path parameters except path
-	// seperators, which will be left as "%2F".
+	// UnescapingModeAllExceptSlash unescapes URL path parameters except path
+	// separators, which will be left as "%2F".
 	UnescapingModeAllExceptSlash
 
-	// URL path parameters will be fully decoded.
+	// UnescapingModeAllCharacters unescapes all URL path parameters.
 	UnescapingModeAllCharacters
```
johanbrandhorst, line 37: 👏🏻 thank you for this

johanbrandhorst (approved):
This is hands down one of the best contributions I've ever reviewed - very thoughtfully researched, implemented and documented. On behalf of the maintainers and the community - thank you!

betmix-matt:
Thank you for putting your time into making this project function!

Users of `grpc-gateway` can now use the Google style API syntax for the gRPC endpoints and still have a RESTful JSON HTTP 1.0 API with a valid OpenAPI specification!

Could I possibly ask for a release?  I want to use this immediately in our organization and I can't merge if it isn't a public version :-)

johanbrandhorst:
I'll make a patch release this evening, as this should probably just be considered a bug fix.

johanbrandhorst:
v2.7.2 is out!

dgparker:
This is huge! thank you so much @betmix-matt

johanbrandhorst:
It appears this may have introduced a regression: https://github.com/grpc-ecosystem/grpc-gateway/issues/2489. We should be able to add a test case mimicking the behaviour reported to help track down the issue.

PR #2464, opened 2021-12-13
Replace all deprecated ioutil with io and os

ahrtr:
The io/ioutil package is deprecated in golang 1.16, please see [go1.16#ioutil](https://go.dev/doc/go1.16#ioutil). Although it is not removed yet, it'd be better to replace all the deprecated io/iouitl with io and os; and it's exactly this PR tries to resolve.


ahrtr:
Signed the cla/google.

PR #2477, opened 2021-12-24
docs: specify invalid cases for FieldMask

hhhapz:
Specify that FieldMask is treated as a regular field in the scenario where the gateway' HTTP method is not PATCH, or if Body is `"*"`.

N/A

Clarifies the invalid cases for the `FieldMask` field.

#### Other comments

Initially when trying this field, I was using the PUT method instead of the PATCH method, and I was unsure why it wasn't working. The docs do indeed mention that this is a feature for PATCH methods, however, it still did not click to me that it was _exclusively_ for PATCH requests. Mentioning this can be helpful for  those in the future who might make the same (incorrect) assumption I did.

### DIFF
--- a/docs/docs/mapping/patch_feature.md
+++ b/docs/docs/mapping/patch_feature.md
@@ -9,7 +9,7 @@ parent: Mapping
 
 The HTTP PATCH method allows a resource to be partially updated.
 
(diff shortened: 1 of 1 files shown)


johanbrandhorst (approved):
LGTM

PR #2505, opened 2022-01-19
remove pattern_ and forward_ globals

CAFxX:

Yup

Move some variables out of the global scope and into the `Register*` functions. This minimizes the amount of live memory in case a package is imported but then the `Register*` functions are not actually called at runtime based on configuration.

#### Other comments

Tested in codespaces. No functionality changes intended.


CAFxX:
ping just to remove the label

johanbrandhorst:
Hi, do you need any assistance with getting this to a good place for review?

PR #2518, opened 2022-01-27
Include Package in Service Tags when Option is Enabled

dkiswanto:

Nope

Absolutely Yes

Proto service name isn't generated by full package when `include_package_in_tags` is true.

And it should be, because as the comment said `If set and the package directive is shown in the proto file, the package name will be prepended to the service name`

#### Other comments

before
```
{
  "swagger": "2.0",
  "info": {
    "title": "foo.example",
  },
  "tags": [
    {
      "name": "ExampleService"
    }
  ],
}
```

after
```

  "swagger": "2.0",
  "info": {
    "title": "foo.example",
  },
  "tags": [
    {
      // service tags will be generated in full package (prepended) if `include_package_in_tags` is true
      "name": "foo.example.ExampleService"
    }
  ],
}
```

### DIFF
--- a/protoc-gen-openapiv2/internal/genopenapi/template.go
+++ b/protoc-gen-openapiv2/internal/genopenapi/template.go
@@ -870,11 +870,16 @@ func partsToRegexpMap(parts []string) map[string]string {
 	return regExps
 }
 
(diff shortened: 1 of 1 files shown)


dkiswanto:
@googlebot I signed it!

PR #2553, opened 2022-02-25
protoc-gen-openapiv2: remove path parameters from body parameters

### LINKED ISSUE #1670: swagger.json: Generated examples for PUT requests don't omit properties provided by path-parameters
## 🐛 Bug Report

When generating the swagger.json from the proto definition, the generated examples for PUT request bodies that use path parameters seem incorrect.
Path parameters that are mapped to fields in the proto-request are not omitted from the generated swagger examples.

## To Reproduce/Example
(from the ABitOfEverything service definition):

```
rpc Update(ABitOfEverything) returns (google.protobuf.Empty) {
		option (google.api.http) = {
			put: "/v1/example/a_bit_of_everything/{uuid}"
			body: "*"
		};l
	}
```

Here, we have an update request that attempts to update the resource at /v1/example/a_bit_of_everything/{uuid}
On the server gateway side, the UUID from the path is correctly mapped to ABitOfEverything.uuid

In the generated swagger.json, we get this PUT definition (omitted unnecessary parts):

```
"put": {
        "operationId": "ABitOfEverythingService_Update",
        "parameters": [
                 {
                   "name": "uuid",
                   "in": "path",
                   "required": true,
                   "type": "string"
                 },
                 {
                   "name": "body",
                   "in": "body",
                   "required": true,
                   "schema": {
                     "$ref": "#/definitions/examplepbABitOfEverything"
                   }
                 }
               ]
        }
```

## Expected Behavior

The uuid field in the example-update-body should be omitted, because it is provided by the path parameter.

```
"examplepbUpdateABitOfEverything": {
      "type": "object",
      "properties": {
         (... lots of properties but no uuid ...)
      }
}
```

## Actual Behavior

However, the actually generated #/definitions/examplepbABitOfEverything is represented by this generated swagger object:

```
"examplepbABitOfEverything": {
      "type": "object",
      "properties": {
        "uuid": {
          "type": "string",
          "minLength": 1,
          "pattern": "[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}"
        }
        (... lots of properties ...)
      }
}
```

As you can see, the example body contains the uuid property AGAIN, even though it's already part of the path parameters.
I would expect this property to be omitted in the example put-request.

The reason for this seems to be excessive code reuse in the generated swagger json - it uses the same "examplepbABitOfEverything" for every endpoint based on the "ABitOfEverything" proto definition, even though the way that the properties are aggregated might differ (some might come from path or query parameters while the remaining ones are provided via request body).

My issue with this is that from an API-consumers point of view, the generated swagger doc implies that you should provide a UUID as part of the body, even though this field will be ignored by the gateway (it's actually overridden by the path-UUID). This is confusing and misleading.

In my opinion, the correct solution to this would be to remove path-based properties from the generated examples. Of course you would then need to generate different variations of "examplepbABitOfEverything" for different API-endpoints (or at least limit the reuse of these examples to endpoints that actually use the same path/query/body configurations).

Alternatively, give the user a way to define which properties to omit. So instead of using _body: "*"_ to include the whole request body, add a more powerful syntax that allows you to omit certain properties.

oyvindwe:

Fixes #1670

Yes.

I've changed the `renderServices` function to recursively render fields as inline definitions as long as they contain path parameters.

I've also changed to use the body field name from gRPC if set in the HTTP binding.

#### Other comments

This PR also fixes some build issues:
- 404 not found when downloading `rules_proto`
- `protoc` segfaults on macOS 12 without full XCode installation: https://github.com/protocolbuffers/protobuf/issues/8884

Note: regenerating files using Docker fails for `swagger-codegen` in the first step from https://github.com/grpc-ecosystem/grpc-gateway/blob/master/CONTRIBUTING.md#using-docker, e.g.:
```
swagger-codegen generate -i examples/internal/proto/examplepb/generate_unbound_methods.swagger.json \
    -l go -o examples/internal/clients/generateunboundmethods --additional-properties packageName=generateunboundmethods
/usr/local/bin/swagger-codegen: 1: -e: not found
[main] INFO io.swagger.parser.Swagger20Parser - reading from examples/internal/proto/examplepb/generate_unbound_methods.swagger.json
[main] WARN io.swagger.codegen.DefaultGenerator - 'host' not defined in the spec. Default to 'localhost'.
[main] ERROR io.swagger.codegen.DefaultCodegen - No Type defined for Additional Property io.swagger.models.properties.UntypedProperty@0
        In Property: io.swagger.models.properties.MapProperty@54a688ff
```

If a _message_ contains an example, that example is still used, even if it contains path parameters. However, generated examples (based on examples from the individual fields in a message) will omit path parameters.

I've introduced two TODOs that I intend do follow up next week.

As this is my first Go project, I'm very happy to get feedback on the code.

I'd also like feedback on test cases that should be added.


examples/internal/proto/examplepb/a_bit_of_everything.swagger.json
```diff
@@ -2670,12 +2872,26 @@
             "pattern": "publishers/[^/]+/books/[^/]+"
           },
           {
-            "name": "body",
-            "description": "The book to update.\n\nThe book's `name` field is used to identify the book to be updated.\nFormat: publishers/{publisher}/books/{book}",
+            "name": "book",
+            "description": "An example resource type from AIP-123 used to test the behavior described in\nthe CreateBookRequest message.\n\nSee: https://google.aip.dev/123",
```
johanbrandhorst, line 2876: I think this is using the wrong description - this is the comment above the `Book` message, but it should probably be the comment above the `UpdateBook.Book` field (as it was before).

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -522,22 +537,22 @@ func renderMessagesAsDefinition(messages messageMap, d openapiDefinitionsObject,
...
+		d[swgName] = renderMessageAsDefinition(msg, reg, customRefs, pathParams)
 	}
 }
 
-func shouldExcludeField(name string, excluded []*descriptor.Field, reg *descriptor.Registry) bool {
-	for _, f := range excluded {
-		if name == reg.FieldName(f) {
+func shouldExcludeField(name string, excluded []descriptor.Parameter) bool {
+	for _, p := range excluded {
+		if len(p.FieldPath) == 1 && name == p.FieldPath[0].Name {
```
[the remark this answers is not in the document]
oyvindwe: A path parameter may be any number of levels down in a nested message structure, e.g `/v1/example/really_deep_path/{abe.nested.name}` and `body="*"`.

For this binding, we must inline body (`ABitOfEverything`) and `nested`, and skip the field `name` from `Nested`.
However, field `single_nested` can refer to the schema of `Nested`.

https://github.com/googleapis/googleapis/blob/master/google/api/http.proto#L224-L231

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -486,6 +488,19 @@ func renderMessageAsDefinition(msg *descriptor.Message, reg *descriptor.Registry
...
+func renderFieldAsDefinition(f *descriptor.Field, reg *descriptor.Registry, refs refMap, pathParams []descriptor.Parameter) openapiSchemaObject {
+	if len(pathParams) == 0 {
+		return schemaOfField(f, reg, refs)
+	}
+	if msg, err := reg.LookupMsg("", f.GetTypeName()); err == nil {
+		// TODO(wergeland): Add description from field.
+		schema := renderMessageAsDefinition(msg, reg, refs, pathParams)
+		return schema
+	} else {
+		panic(err)
```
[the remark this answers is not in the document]
oyvindwe: I wasn't exactly sure what to do here; if the error is returned via `renderServices`, `applyTemplate` will also just call `panic`. I'm happy to return it instead.

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -486,6 +488,19 @@ func renderMessageAsDefinition(msg *descriptor.Message, reg *descriptor.Registry
...
 }
 
+func renderFieldAsDefinition(f *descriptor.Field, reg *descriptor.Registry, refs refMap, pathParams []descriptor.Parameter) openapiSchemaObject {
+	if len(pathParams) == 0 {
+		return schemaOfField(f, reg, refs)
+	}
+	if msg, err := reg.LookupMsg("", f.GetTypeName()); err == nil {
+		// TODO(wergeland): Add description from field.
+		schema := renderMessageAsDefinition(msg, reg, refs, pathParams)
+		return schema
```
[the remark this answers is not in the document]
oyvindwe: I plan to do it in this PR.

examples/internal/proto/examplepb/a_bit_of_everything.swagger.json
```diff
@@ -904,7 +904,7 @@
             "type": "string"
           },
           {
-            "name": "body",
+            "name": "single_nested",
```
oyvindwe, line 907: This should be in JSON form.

examples/internal/proto/examplepb/a_bit_of_everything.swagger.json
```diff
@@ -2611,7 +2611,209 @@
...
+                    "$ref": "#/definitions/ABitOfEverythingNested"
+                  }
+                },
+                "floatValue": {
+                  "type": "number",
+                  "format": "float",
+                  "default": "0.2",
+                  "description": "Float value field",
+                  "required": [
+                    "float_value"
```
oyvindwe, line 2640: This should be in JSON form.

examples/internal/proto/examplepb/a_bit_of_everything.swagger.json
```diff
@@ -3456,7 +4066,219 @@
...
+                  "type": "string",
+                  "description": "The paths to update."
+                }
+              },
+              "title": "UpdateV2Request request for update includes the message and the update mask",
+              "required": [
+                "abe",
+                "abe",
+                "abe",
+                "abe"
```
oyvindwe, line 4280: Should only generate one instance of `"abe"`

examples/internal/clients/abe/api/swagger.yaml
```diff
@@ -2053,7 +2052,7 @@ paths:
         name: "body"
         required: true
         schema:
-          $ref: "#/definitions/examplepbABitOfEverything"
+          $ref: "#/definitions/A bit of everything_1"
```
oyvindwe, line 2055: Title/summary instead of name!

examples/internal/proto/examplepb/a_bit_of_everything.swagger.json
oyvindwe: Fixed.

examples/internal/clients/abe/api/swagger.yaml
```diff
@@ -766,11 +766,11 @@ paths:
         type: "string"
         x-exportParamName: "StringValue"
       - in: "body"
-        name: "body"
+        name: "single_nested"
```
oyvindwe: snake_case should be camelCase here

examples/internal/clients/abe/api/swagger.yaml
oyvindwe: How is this generated? `a_bit_of_everything.swagger.json` lists the properties. Is it an assumption in json -> yaml that body parameters always has a schema ref?

examples/internal/proto/examplepb/a_bit_of_everything.swagger.json
```diff
@@ -2670,12 +2886,27 @@
             "pattern": "publishers/[^/]+/books/[^/]+"
           },
           {
-            "name": "body",
-            "description": "The book to update.\n\nThe book's `name` field is used to identify the book to be updated.\nFormat: publishers/{publisher}/books/{book}",
+            "name": "book",
+            "description": "The book's `name` field is used to identify the book to be updated.\nFormat: publishers/{publisher}/books/{book}",
```
oyvindwe, line 2890: Need to preserve the title/summary paragraph here.

protoc-gen-openapiv2/internal/genopenapi/template.go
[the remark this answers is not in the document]
oyvindwe: Fixed this, and I also removed some more low level `panic(err)` at the same time.

examples/internal/proto/examplepb/a_bit_of_everything.swagger.json
```diff
@@ -2611,7 +2611,209 @@
...
+                },
+                "int64OverrideType": {
+                  "type": "integer",
+                  "format": "int64"
+                },
+                "requiredStringViaFieldBehaviorAnnotation": {
+                  "type": "string",
+                  "title": "mark a field as required in Open API definition",
+                  "required": [
+                    "required_string_via_field_behavior_annotation"
```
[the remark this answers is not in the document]
oyvindwe: I'm working on this, but note that this was not introduced in this PR - the `"required"` arrays in master all have `"snake_case"` values. This issue appears to be fixed, but that is an illusion, as I need to check out the use of `json_name` fields.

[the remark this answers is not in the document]
oyvindwe: Correction - _some_ `"required"` arrays are `"snake_case"`, while others are `"camelCase"`

examples/internal/proto/examplepb/a_bit_of_everything.swagger.json
oyvindwe: Working on this (see above)

examples/internal/clients/abe/api/swagger.yaml
oyvindwe: I was looking at a diff against an earlier commit. This is fixed now.

examples/internal/clients/abe/api/swagger.yaml
johanbrandhorst: I think this yaml file is generated from the swagger.json source of truth.

oyvindwe: This is generated by `swagger-codegen`. I'll investigate if this is a bug with that program or if the generated swagger.json contains wrong.

examples/internal/proto/examplepb/a_bit_of_everything.swagger.json
oyvindwe: This is fixed now. Note that the old code had some inconsistency where some places the comment string was used directly, while in another place it would be split on `\n\n` and joined with `. `. I made this consistent to join again with `\n\n`.

examples/internal/clients/abe/api/swagger.yaml
oyvindwe: Aha - `swagger-codegen` moves the inline schema to the `definitions:` section, and uses the title to generate the reference. All good then!

protoc-gen-openapiv2/internal/genopenapi/template.go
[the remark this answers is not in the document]
oyvindwe: Done.

examples/internal/proto/examplepb/a_bit_of_everything.swagger.json
oyvindwe: Done.

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -483,7 +480,33 @@ func renderMessageAsDefinition(msg *descriptor.Message, reg *descriptor.Registry
...
+		return openapiSchemaObject{}, err
+	}
+	schema, err := renderMessageAsDefinition(msg, reg, refs, pathParams)
+	if err != nil {
+		return openapiSchemaObject{}, err
+	}
+	comments := fieldProtoComments(reg, f.Message, f)
+	if len(comments) > 0 {
+		// Use title and description from field instead of nested message if present.
+		paragraphs := strings.Split(comments, "\n\n")
```
johanbrandhorst, line 505: I'm a little confused, where does this logic come from? Are we doing this because we've already joined the title and description comments into one, and we need to separate them again? If so, is there no way to retrieve the original data so we don't have to depend on behavior like "title and descriptions are merged by double newlines"?

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -2328,6 +2385,12 @@ func updateswaggerObjectFromJSONSchema(s *openapiSchemaObject, j *openapi_option
 	s.MaxProperties = j.GetMaxProperties()
 	s.MinProperties = j.GetMinProperties()
 	s.Required = j.GetRequired()
+	if reg.GetUseJSONNamesForFields() {
+		for i, r := range s.Required {
+			// TODO(oyvindwe): Look up field and use field.GetJsonName()?
+			s.Required[i] = doCamelCase(r)
```
johanbrandhorst, line 2391: This makes me a little nervous, if we could look up the field and use the canonical field.GetJsonName that would be much preferred, but if it's too much trouble we can leave it for a follow up.

protoc-gen-openapiv2/internal/genopenapi/template.go
oyvindwe: This took me a while to figure out as well (my first attempt to use the description with the field comment string instead of the message comment string missed it). I generalised it from https://github.com/grpc-ecosystem/grpc-gateway/blob/09a279d441d5819789b08c0ae17544045601ca67/protoc-gen-openapiv2/internal/genopenapi/template.go#L243-L246 as it was needed in 2 places

`schema` has fields `title` and `description`, where the first paragraph of a comment string is put into the title and the rest into the description. However, `parameter` only has `description`, so we need to merge the string back.

It is split here: https://github.com/grpc-ecosystem/grpc-gateway/blob/09a279d441d5819789b08c0ae17544045601ca67/protoc-gen-openapiv2/internal/genopenapi/template.go#L1890

I changed the join from `. ` to `\n\n` for consistency, as some places the comment string is used directly, e.g. here: https://github.com/grpc-ecosystem/grpc-gateway/blob/09a279d441d5819789b08c0ae17544045601ca67/protoc-gen-openapiv2/internal/genopenapi/template.go#L986-L988

protoc-gen-openapiv2/internal/genopenapi/template.go
oyvindwe: It is a bug in the the existing code that just passes the `required` array on without transforming to the JSON format as you can see here: https://github.com/grpc-ecosystem/grpc-gateway/blob/09a279d441d5819789b08c0ae17544045601ca67/examples/internal/proto/examplepb/a_bit_of_everything.swagger.json#L2091 and here: https://github.com/grpc-ecosystem/grpc-gateway/blob/09a279d441d5819789b08c0ae17544045601ca67/examples/internal/proto/examplepb/a_bit_of_everything.swagger.json#L3744

This also broke my new check to set the boolean `required` flag in this PR.

I investigated fixing this properly, but a lot of context is missing here to look up field name. I considered transforming the `Required` slice to JSON format earlier, but that would require different check vs. field names. I think the best approach is to represent `Required` as fields rather than strings, and look them up when retrieving the options from the proto file (and discard/warn about undefined fields), as with this annotation it's possible to generate illegal an OpenAPIv2 file. I'd consider this outside the scope of this PR.

johanbrandhorst: That is fair, lets leave this for now.

protoc-gen-openapiv2/internal/genopenapi/template.go
oyvindwe: Good idea. Done.

johanbrandhorst (approved):
LGTM

oyvindwe:
Thanks for swift and helpful reviews!

johanbrandhorst:
This was released in https://github.com/grpc-ecosystem/grpc-gateway/releases/tag/v2.8.0!

PR #2574, opened 2022-03-09
Annotate incoming and outgoing context

### LINKED ISSUE #2571: Forward Response Option - Context contains no request anotations
## 🐛 Bug Report

The context that is passed to `WithForwardResponseOption` function does not include any of the Annotation Options and therefore `runtime.HTTPPathPattern(ctx)` always returns empty.

The annotated context is created here: https://github.com/grpc-ecosystem/grpc-gateway/blob/master/protoc-gen-grpc-gateway/internal/gengateway/template.go#L599 with variable `rctx`
```go
rctx, err := runtime.AnnotateIncomingContext(ctx, mux, req, "/{{$svc.File.GetPackage}}.{{$svc.GetName}}/{{$m.GetName}}", runtime.WithHTTPPathPattern("{{$b.PathTmpl.Template}}"))
```

However a new context is created here but it's parent is the original `ctx` rather than `rctx` https://github.com/grpc-ecosystem/grpc-gateway/blob/master/protoc-gen-grpc-gateway/internal/gengateway/template.go#L609 
```go
ctx = runtime.NewServerMetadataContext(ctx, md)
forward_{{$svc.GetName}}_{{$m.GetName}}_{{$b.Index}}(ctx, mux, outboundMarshaler, w, req, response_{{$svc.GetName}}_{{$m.GetName}}_{{$b.Index}}{resp}, mux.GetForwardResponseOptions()...)
```

A potential (easy) fix is to change the parent context:
```diff
--- ctx = runtime.NewServerMetadataContext(ctx, md)
+++ ctx = runtime.NewServerMetadataContext(rctx, md) 
```
But I'm unfamiliar if this would cause unwanted side effects... ie there must be a reason why we are creating a new metadata context.

Another option would be to re-hydrate the new context with the HTTP Path Pattern context key in the template Eg:
```go
ctx = runtime.NewServerMetadataContext(ctx, md)
ctx = runtime.WithHTTPPathPattern("/example/foo/bar")(ctx)
```

Happy to raise a PR, but would like some guidance on the correct approach.

## To Reproduce

```go
runtime.WithForwardResponseOption(func(ctx context.Context, w http.ResponseWriter, resp proto.Message) error {
    	if pattern, _ := runtime.HTTPPathPattern(ctx); pattern == "/example/foo/bar" {
                // DO WORK
		return nil
	}
        return nil
}))
```

## Expected behavior

Call stack should branch into the `if` statement when the HTTP path pattern matches `"/example/foo/bar"` and execute `DO WORK`

## Actual Behavior

`DO WORK` is never called, even when the pattern matches.

## Your Environment

N/A

rogchap:

Fixes #2571

Yes.

Ensures that the incoming and outgoing context has been annotated, specifically making sure that the HTTP Pattern
Context value is passed to Forward Response Options.

#### Other comments

N/A

### DIFF
--- a/examples/internal/helloworld/helloworld.pb.gw.go
+++ b/examples/internal/helloworld/helloworld.pb.gw.go
@@ -743,12 +743,13 @@ func RegisterGreeterHandlerServer(ctx context.Context, mux *runtime.ServeMux, se
 		var stream runtime.ServerTransportStream
 		ctx = grpc.NewContextWithServerTransportStream(ctx, &stream)
 		inboundMarshaler, outboundMarshaler := runtime.MarshalerForRequest(mux, req)
-		rctx, err := runtime.AnnotateIncomingContext(ctx, mux, req, "/grpc.gateway.examples.internal.helloworld.Greeter/SayHello", runtime.WithHTTPPathPattern("/say/{name}"))
+		var err error
+		ctx, err = runtime.AnnotateIncomingContext(ctx, mux, req, "/grpc.gateway.examples.internal.helloworld.Greeter/SayHello", runtime.WithHTTPPathPattern("/say/{name}"))
 		if err != nil {
 			runtime.HTTPError(ctx, mux, outboundMarshaler, w, req, err)
 			return
 		}
-		resp, md, err := local_request_Greeter_SayHello_0(rctx, inboundMarshaler, server, req, pathParams)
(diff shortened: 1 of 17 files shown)


johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution!

Issue #1791, opened 2020-11-05
Can protoc-gen-swagger hide the specified fields?


yunqiang-sun:
## 🚀 Feature

I want to hide certain fields in the json generated by protoc-gen-swagger

johanbrandhorst:
I don't think we support this today. If you'd like to contribute this I would consider it if you can come up with a good design.

ewhauser:
I think extensions are the best way to handle hidden/private operations, parameters, fields, etc. There's an existing open source tool that allows you to filter your generated schema: https://github.com/Mermade/openapi-filter

I don't believe `protoc-gen-swagger` lets you define `extensions` on fields at the moment, but that could be added.

santhosh-v:
Field level support is there, I think with this we could use this filter. Is it possible to add soon ?

johanbrandhorst:
This isn't going to be added unless someone has the time to contribute it.

james-o-johnstone:
Added a PR https://github.com/grpc-ecosystem/grpc-gateway/pull/2418 to define `extensions` on fields, so fields could be marked with an `x-internal` extension and filtered using https://github.com/Mermade/openapi-filter

johanbrandhorst:
There is now an alternative implementation to hiding fields/RPCs in https://github.com/grpc-ecosystem/grpc-gateway/pull/2578. We probably only want to merge one, so would interested parties weigh in on which they would prefer?

BCook98:
I'm slightly biased towards https://github.com/grpc-ecosystem/grpc-gateway/pull/2578 😉. 

Both are nice solutions towards hiding fields/RPCs, but `google.api.VisibilityRules` provides the ability to completely hide fields from the output. That way consumers never know a field exists, instead of having to rely on the Swagger renderer/ codegen to do that for you.

That being said it looks like https://github.com/grpc-ecosystem/grpc-gateway/pull/2418 is still useful in other scenarios where you want to apply other swagger extensions to fields and schemas.

ewhauser:
I concur with @BCook98. First class support for `google.api.VisbilityRules` seems like the way to go - I'm not sure this was around (or at least I didn't know about it) when I suggested extensions.

I also agree that these two implementations are not mutually exclusive.

johanbrandhorst:
Fair points, lets consider each on its own merits, thank you both!

PR #2578, opened 2022-03-10
Add support for `google.api.VisibilityRule`s annotations to hide APIs and fields

BCook98:

https://github.com/grpc-ecosystem/grpc-gateway/issues/1791

Yes

This PR adds support for annotations to set the visibility restrictions of fields, services, methods and enum values. These rules can be used in combination with a new option `visibility_restriction_selectors` to hide fields and APIs from the generated API output. 

Google's [`google.api.VisibilityRule`](https://github.com/googleapis/googleapis/blob/master/google/api/visibility.proto#L89)s are used as a standard annotation.

This is useful when generating documentation where protobuf APIs are used both internally and externally, where some fields may need to be hidden. Or if an API field isn't fully baked yet it can be tagged with `BETA` or `PREVIEW` and only generated to internal or private documentation. 

##### Example 

`opt: visibility_restriction_selectors=PREVIEW` will result in:

Input Example:
```protobuf
service Echo {
    rpc EchoInternal(VisibilityRuleSimpleMessage) returns (VisibilityRuleSimpleMessage) {
        option (google.api.method_visibility).restriction = "INTERNAL";
        option (google.api.http) = {
            get: "/v1/example/echo_internal"
        };
    }
    rpc EchoInternalAndPreview(VisibilityRuleSimpleMessage) returns (VisibilityRuleSimpleMessage) {
        option (google.api.method_visibility).restriction = "INTERNAL,PREVIEW";
        option (google.api.http) = {
            get: "/v1/example/echo_internal_and_preview"
        };
    }
}

message VisibilityRuleSimpleMessage {
     enum VisibilityEnum {
          UNSPECIFIED = 0;
          VISIBLE = 1;
          INTERNAL = 2 [(google.api.value_visibility).restriction = "INTERNAL"];
          PREVIEW = 3 [(google.api.value_visibility).restriction = "INTERNAL,PREVIEW"];
     }
     
     string internal_field = 1 [(google.api.field_visibility).restriction = "INTERNAL"];
     string preview_field = 2 [(google.api.field_visibility).restriction = "INTERNAL,PREVIEW"];
     VisibilityEnum an_enum = 3;
}
```

Output json:
```json
{
    "paths": {
        "/v1/example/echo_internal_and_preview": {
            "get": {
                "summary": "EchoInternalAndPreview is a internal and preview API that should be visibile in the OpenAPI spec.",
                "operationId": "VisibilityRuleEchoService_EchoInternalAndPreview",
                "responses": {
                    "200": {
                        "schema": {
                        "$ref": "#/definitions/examplepbVisibilityRuleSimpleMessage"
                        }
                    },
                    "default": {
                        "description": "An unexpected error response.",
                        "schema": {
                            "$ref": "#/definitions/rpcStatus"
                        }
                    }
                },
                "parameters": [
                    {
                        "name": "previewField",
                        "in": "query",
                        "required": false,
                        "type": "string"
                    },
                    {
                        "name": "anEnum",
                        "in": "query",
                        "required": false,
                        "type": "string",
                        "enum": [
                            "UNSPECIFIED",
                            "VISIBLE",
                            "PREVIEW"
                        ],
                        "default": "UNSPECIFIED"
                    }
                ],
                "tags": [
                    "VisibilityRuleEchoService"
                ]
            }
        }
    }
}
```

#### Other comments

This PR doesn't add support for selectors in visibility rules, just using them as annotations. See https://github.com/googleapis/googleapis/blob/master/google/api/visibility.proto#L71 for more detail. This is useful if you want to provide an external list of rules.

Support is also missing for adding annotations to messages and enums, this was excluded due to complexity and potential edge cases when excluding a type that is required somewhere else in the OpenAPI Schema.


johanbrandhorst:
This is very cool, I will need some time to review, but this is a commonly requested feature. Thanks!

johanbrandhorst:
Closing an reopening to trigger CI runs

johanbrandhorst:
Closing and reopening to trigger CI... not sure why this is happening 😬.

docs/docs/mapping/customizing_openapi_output.md
```diff
@@ -284,3 +284,96 @@ Output json:
 },
 ```
 {% endraw %}
+
+### Hiding fields, methods, services and enum values
+
+If you require internal or unreleased fields and APIs to be hidden from your API documentation, [`google.api.VisibilityRule`](https://github.com/googleapis/googleapis/blob/master/google/api/visibility.proto#L89) annotations can be added customize where they are generated. Combind with the option `visibility_restriction_selectors` overlapping rules will appear in the OpenAPI output. 
+
+Fields and APIs without `google.api.VisibilityRule` annotations will always appear in the generated output.
```
johanbrandhorst, line 292: I like the wording in the original source better: https://github.com/googleapis/googleapis/blob/9916192ab15e3507e41ba2c5165182fec06120d0/google/api/visibility.proto#L66-L67. 
```suggestion
Elements without `google.api.VisibilityRule` annotations will appear as usual in the generated output.
```

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -554,6 +577,26 @@ func renderMessagesAsDefinition(messages messageMap, d openapiDefinitionsObject,
 	return nil
 }
 
+func checkVisibility(r *visibility.VisibilityRule, reg *descriptor.Registry) bool {
```
johanbrandhorst, line 580: Please add a comment to this method. The name is a bit confusing to me, I'd prefer it to be something like `isVisible` since it returns a bool, what do you think?

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -554,6 +577,26 @@ func renderMessagesAsDefinition(messages messageMap, d openapiDefinitionsObject,
...
+	}
+
+	restrictions := strings.Split(strings.TrimSpace(r.Restriction), ",")
+	// No restrictions results in the element always being visible
+	if len(restrictions) == 0 {
+		return true
+	}
+
+	for _, restriction := range restrictions {
+		if reg.GetVisibilityRestrictionSelectorsMap()[strings.TrimSpace(restriction)] {
```
johanbrandhorst, line 592: I don't see us doing any case folding anywhere, which means that this matching is case sensitive, which I think makes sense, just wanted to check that this was intentional?

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -2268,6 +2326,66 @@ func extractFieldBehaviorFromFieldDescriptor(fd *descriptorpb.FieldDescriptorPro
 	return opts, nil
 }
 
+func getFieldVisibilityOption(fd *descriptor.Field) (*visibility.VisibilityRule, error) {
```
johanbrandhorst, line 2329: I think we can make this signature simpler by just returning a visibility rule that may or may not be nil, then we can skip the error check that we're not doing anything with anyway. What do you think?

johanbrandhorst (commented):
This is a really top notch contribution, even including docs 😍. I have some initial questions:

* It appears that the visibility labels themselves can be arbitrary strings, both in these examples and in the Google protobuf sources. Because of this, I think we need to make it clear that there is no implicit hierarchy in the labels (which it seemed like to me initially), and the visibility selector will only match those rules that are provided exactly.
* Another consequence of this is that unless you intend to invoke the plugin with different options, you only really need the INTERNAL label, probably. Maybe we should have the example include an invocation without a selector, with INTERNAL and with PREVIEW respectively. We might also need an example with more than one selector. Maybe only one with more than one selector actually. What do you think?

protoc-gen-openapiv2/internal/genopenapi/template.go
BCook98: Yep this is intentionally case sensitive. Though there is no real justification for it being that way, I can't someone wanting `internal` and `INTERNAL` to mean different things.

But something doesn't feel right about making this case insensitive.

BCook98:
Thanks for your feedback, much appreciated 🙏 

Let me take some time to revisit the docs and make it clear that these selectors are completely arbitrary and add some more examples around it!

---

I may need your help with Bazel and those imports... From what I can see I don't think I am doing the wrong thing with https://github.com/grpc-ecosystem/grpc-gateway/pull/2578/files#diff-c5fc7051bb345b50fbe0e92d76c0eadf61359cb387e689308ecc1ade85cf96c7R106. In fact `bazel test //...` requires them in that format over what it is suggesting, otherwise it fails.

johanbrandhorst:
It should be possible to fix the bazel errors by running the command in CONTRIBUTING.md

BCook98:
It looks like `buf` has a bug and can't correctly parse 
```yaml
version: v1
plugins:
  - name: openapiv2
    out: .
    opt:
     - visibility_restriction_selectors=INTERNAL,PREVIEW
```

Results in: `F0311 17:34:09.453974   92854 main.go:80] Error parsing flags: cannot set flag PREVIEW: no such flag -PREVIEW` as it appears to just be [blindly splitting opts on `,`](https://github.com/bufbuild/buf/blob/main/private/buf/cmd/buf/command/alpha/protoc/flags.go#L379)...

I'll raise an issue against buf, but for now it appears we will have to leave out a combined `visibility_restriction_selectors` example. Alternatively we could use a different delimiter for `visibility_restriction_selectors`, if you have a better suggestion?

BCook98:
Running

```sh
$ bazel run :gazelle -- update-repos -from_file=go.mod -to_macro=repositories.bzl%go_repositories && \
    bazel run :gazelle && \
    bazel run :buildifier
```

results in 

```diff
@@ -59,6 +59,7 @@ proto_library(
         "//examples/internal/proto/pathenum:pathenum_proto",
         "//examples/internal/proto/sub:sub_proto",
         "//examples/internal/proto/sub2:sub2_proto",
+        "//google/api:api_proto",
         "//protoc-gen-openapiv2/options:options_proto",
         "@com_google_protobuf//:duration_proto",
         "@com_google_protobuf//:empty_proto",
@@ -68,7 +69,6 @@ proto_library(
         "@com_google_protobuf//:wrappers_proto",
         "@go_googleapis//google/api:annotations_proto",
         "@go_googleapis//google/api:httpbody_proto",
-        "@go_googleapis//google/api:visibility_proto",
         "@go_googleapis//google/rpc:status_proto",
     ],
 )
@@ -99,11 +99,11 @@ go_proto_library(
         "//examples/internal/proto/pathenum",
         "//examples/internal/proto/sub",
         "//examples/internal/proto/sub2",
+        "//google/api:visibility_proto",
         "//protoc-gen-openapiv2/options",
         "@com_github_golang_protobuf//descriptor:go_default_library_gen",  # keep
         "@go_googleapis//google/api:annotations_go_proto",
         "@go_googleapis//google/api:httpbody_go_proto",
-        "@go_googleapis//google/api:visibility_go_proto",
         "@go_googleapis//google/rpc:status_go_proto",
         "@org_golang_google_protobuf//proto:go_default_library",  # keep
     ],
```

However subsequently running `bazel test //...`

I get the following error
```
$ bazel test //...
ERROR: /workspaces/grpc-gateway/examples/internal/proto/examplepb/BUILD.bazel:89:17: no such package 'google/api': BUILD file not found in any of the following directories. Add a BUILD file to a directory to mark it as a package.
 - /workspaces/grpc-gateway/google/api and referenced by '//examples/internal/proto/examplepb:examplepb_go_proto'
ERROR: Analysis of target '//examples/internal/proto/examplepb:examplepb_go_proto' failed; build aborted: no such package 'google/api': BUILD file not found in any of the following directories. Add a BUILD file to a directory to mark it as a package.
 - /workspaces/grpc-gateway/google/api
INFO: Elapsed time: 0.437s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (1 packages loaded, 0 targets configured)
FAILED: Build did NOT complete successfully (1 packages loaded, 0 targets configured)
```

In the current state the tests pass with
```
$ bazel test //...
INFO: Analyzed 100 targets (1 packages loaded, 43 targets configured).
INFO: Found 90 targets and 10 test targets...
INFO: From Action examples/internal/proto/examplepb/_virtual_imports/examplepb_protoc_gen_openapiv2/examples/internal/proto/examplepb/a_bit_of_everything.swagger.json:
W0311 07:01:40.237947   49249 template.go:1136] created a body with 0 properties in the message, this might be unintended: name:"Body"  field:{name:"name"  number:1  label:LABEL_OPTIONAL  type:TYPE_STRING  json_name:"name"}
INFO: From Action examples/internal/proto/examplepb/examplepb_protoc_gen_openapiv2_merged.swagger.json:
W0311 07:01:40.326045   49270 template.go:1136] created a body with 0 properties in the message, this might be unintended: name:"Body"  field:{name:"name"  number:1  label:LABEL_OPTIONAL  type:TYPE_STRING  json_name:"name"}
INFO: Elapsed time: 11.598s, Critical Path: 10.68s
INFO: 32 processes: 2 internal, 30 processwrapper-sandbox.
INFO: Build completed successfully, 32 total actions
//internal/codegenerator:codegenerator_test                     (cached) PASSED in 0.1s
//internal/descriptor:descriptor_test                           (cached) PASSED in 0.2s
//internal/httprule:httprule_test                               (cached) PASSED in 0.1s
//protoc-gen-grpc-gateway:protoc-gen-grpc-gateway_test          (cached) PASSED in 0.5s
//protoc-gen-grpc-gateway/internal/gengateway:gengateway_test   (cached) PASSED in 0.3s
//protoc-gen-openapiv2:protoc-gen-openapiv2_test                (cached) PASSED in 0.1s
//protoc-gen-openapiv2/internal/genopenapi:genopenapi_test      (cached) PASSED in 0.2s
//runtime:runtime_test                                          (cached) PASSED in 0.3s
//utilities:utilities_test                                      (cached) PASSED in 0.2s
//examples/internal/integration:integration_test                         PASSED in 0.3s

Executed 1 out of 10 tests: 10 tests pass.
INFO: Build completed successfully, 32 total actions
```

This is stretching my very limited Bazel knowledge, it's not clear if I am missing something. Or need to declare `@go_googleapis//google/api:visibility_go_proto` somewhere?

johanbrandhorst:
> It looks like `buf` has a bug and can't correctly parse
> 
> ```yaml
> version: v1
> plugins:
>   - name: openapiv2
>     out: .
>     opt:
>      - visibility_restriction_selectors=INTERNAL,PREVIEW
> ```
> 
> Results in: `F0311 17:34:09.453974 92854 main.go:80] Error parsing flags: cannot set flag PREVIEW: no such flag -PREVIEW` as it appears to just be [blindly splitting opts on `,`](https://github.com/bufbuild/buf/blob/main/private/buf/cmd/buf/command/alpha/protoc/flags.go#L379)...
> 
> I'll raise an issue against buf, but for now it appears we will have to leave out a combined `visibility_restriction_selectors` example. Alternatively we could use a different delimiter for `visibility_restriction_selectors`, if you have a better suggestion?

I'm glad we tried this, because I'm not sure I'd even consider this a bug, I think `protoc` would tell you the same thing. We might not be able to use `,` in our option. Can you try the following:

 ```yaml
 version: v1
 plugins:
   - name: openapiv2
     out: .
     opt:
      - visibility_restriction_selectors=INTERNAL
      - visibility_restriction_selectors=PREVIEW
 ```

I'm not sure what that will actually look like inside the generator, but that may be how we have to handle multiple labels. We also should consider not using `,` as a separator. Maybe we want `:`? There is precedent for using `:` for mapping, I need to see if there's any precedent for inputting lists.

Re: the bazel issues, I invoke the power of @achew22, may he have mercy on your `BUILD`. I'll try to take a look myself but I'm also not a bazel expert.

johanbrandhorst:
Here's `protoc` unconditionally splitting on `,`: https://github.com/protocolbuffers/protobuf/blob/5dab09408cd186a5235ace943d9c22936e52f376/src/google/protobuf/compiler/code_generator.cc#L110.

johanbrandhorst:
It appears protoc may have special handling for both `:` and `;` depending on the platform: https://github.com/protocolbuffers/protobuf/blob/5dab09408cd186a5235ace943d9c22936e52f376/src/google/protobuf/compiler/command_line_interface.cc#L894-L898.

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -554,6 +577,30 @@ func renderMessagesAsDefinition(messages messageMap, d openapiDefinitionsObject,
 	return nil
 }
 
+// isVisible checks if a field/RPC is visible based on the visibility restriction
+// combined with the `visibility_restriction_selectors`.
+// Elements with an overlap on `visibility_restriction_selectors`, those without are not visible.
```
johanbrandhorst, line 582: This sentence feels like it accidentally

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -2268,6 +2330,70 @@ func extractFieldBehaviorFromFieldDescriptor(fd *descriptorpb.FieldDescriptorPro
...
+	if fd.Options == nil {
+		return nil
+	}
+	if !proto.HasExtension(fd.Options, visibility.E_FieldVisibility) {
+		return nil
+	}
+	ext := proto.GetExtension(fd.Options, visibility.E_FieldVisibility)
+	opts, ok := ext.(*visibility.VisibilityRule)
+	if !ok {
+		fmt.Fprintf(os.Stderr, "%s: extension is %T; want a *VisibilityRule object", fd.FQFN(), ext)
```
johanbrandhorst, line 2343: I don't know if we even need to log this, this is a fundamental assumption error somewhere, no? Our tests should start failing if this fails, I think we can just ignore it?

protoc-gen-openapiv2/internal/genopenapi/BUILD.bazel
```diff
@@ -24,6 +24,7 @@ go_library(
         "@go_googleapis//google/api:annotations_go_proto",
         "@go_googleapis//google/rpc:status_go_proto",
         "@io_bazel_rules_go//proto/wkt:any_go_proto",
+        "@org_golang_google_genproto//googleapis/api/visibility",
```
johanbrandhorst, line 27: These references also look wrong compared to the `annotations_go_proto` on line 24

johanbrandhorst:
I've been playing around a bit with bazel on this branch and I've concluded that it's `bazel run :gazelle` which is adding the spurious reference to `"//google/api:api_proto"`. I agree that this seems wrong. I'm not sure if this is a gazelle bug, but I have raised https://github.com/bazelbuild/bazel-gazelle/issues/1193 to get some thoughts on this.

johanbrandhorst:
@BCook98 leaving the bazel issues aside for a moment, lets try my suggestion for multiple labels? I think the conclusion from the initial test is that we can't use comma separation at all, and if specifying the option multiple times doesn't do what we want, we may need to use a separator like `;` or something.

johanbrandhorst:
I submitted a fix to bazel-gazelle which I have confirmed fixes the bazel issues: https://github.com/bazelbuild/bazel-gazelle/pull/1194

johanbrandhorst:
I tried pushing to this branch but I believe you have not ticked "allow pushes from contributors" on this PR, so I couldn't. Here's the diff that fixes the bazel issues, until the aforementioned PR is merged:

```diff
diff --git a/WORKSPACE b/WORKSPACE
index df6b45f..3eaeaa8 100644
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -1,6 +1,7 @@
 workspace(name = "grpc_ecosystem_grpc_gateway")
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 
 # Define before rules_proto, otherwise we receive the version of com_google_protobuf from there
 http_archive(
@@ -48,13 +49,11 @@ http_archive(
     ],
 )
 
-http_archive(
+git_repository(
     name = "bazel_gazelle",
-    sha256 = "de69a09dc70417580aabf20a28619bb3ef60d038470c7cf8442fafcf627c21cb",
-    urls = [
-        "https://storage.googleapis.com/bazel-mirror/github.com/bazelbuild/bazel-gazelle/releases/download/v0.24.0/bazel-gazelle-v0.24.0.tar.gz",
-        "https://github.com/bazelbuild/bazel-gazelle/releases/download/v0.24.0/bazel-gazelle-v0.24.0.tar.gz",
-    ],
+    commit = "4a1aeae7cab962fd8088f42038d3a477cdca91a5",
+    remote = "https://github.com/johanbrandhorst/bazel-gazelle",
+    shallow_since = "1647116890 +0000",
 )
 
 load("@io_bazel_rules_go//go:deps.bzl", "go_register_toolchains", "go_rules_dependencies")
diff --git a/protoc-gen-openapiv2/internal/genopenapi/BUILD.bazel b/protoc-gen-openapiv2/internal/genopenapi/BUILD.bazel
index 93de3bd..b684277 100644
--- a/protoc-gen-openapiv2/internal/genopenapi/BUILD.bazel
+++ b/protoc-gen-openapiv2/internal/genopenapi/BUILD.bazel
@@ -22,9 +22,9 @@ go_library(
         "@com_github_golang_glog//:glog",
         "@com_github_golang_protobuf//descriptor:go_default_library_gen",
         "@go_googleapis//google/api:annotations_go_proto",
+        "@go_googleapis//google/api:visibility_go_proto",
         "@go_googleapis//google/rpc:status_go_proto",
         "@io_bazel_rules_go//proto/wkt:any_go_proto",
-        "@org_golang_google_genproto//googleapis/api/visibility",
         "@org_golang_google_protobuf//encoding/protojson",
         "@org_golang_google_protobuf//proto",
         "@org_golang_google_protobuf//types/descriptorpb",
@@ -50,8 +50,8 @@ go_test(
         "//runtime",
         "@com_github_google_go_cmp//cmp",
         "@go_googleapis//google/api:annotations_go_proto",
+        "@go_googleapis//google/api:visibility_go_proto",
         "@io_bazel_rules_go//proto/wkt:field_mask_go_proto",
-        "@org_golang_google_genproto//googleapis/api/visibility",
         "@org_golang_google_protobuf//proto",
         "@org_golang_google_protobuf//reflect/protodesc",
         "@org_golang_google_protobuf//types/descriptorpb",
```

BCook98:
Great catch on protoc not supporting `,` either. I'm not sure how I didn't test this 🤦‍♂️...

I like the idea of using multiple options but IIRC `flag` doesn't support that out of the box. Let me do some testing and verify it. 

I've applied your patch, thank you very much for identifying the issue!

BCook98:
`flag` doesn't work, the last supplied flag is the one that is used. But using `StringArray` from https://github.com/spf13/pflag will allow for multiple flags to be supplied. 

Happy to also use `;` as a delimiter instead if you'd like to avoid adding additional dependencies.

```diff
diff --git a/examples/internal/proto/examplepb/visibility_rule_preview_and_internal_echo_service.buf.gen.yaml b/examples/internal/proto/examplepb/visibility_rule_preview_and_internal_echo_service.buf.gen.yaml
new file mode 100644
index 0000000..0a2a741
--- /dev/null
+++ b/examples/internal/proto/examplepb/visibility_rule_preview_and_internal_echo_service.buf.gen.yaml
@@ -0,0 +1,7 @@
+version: v1
+plugins:
+  - name: openapiv2
+    out: .
+    opt:
+      - visibility_restriction_selectors=PREVIEW
+      - visibility_restriction_selectors=INTERNAL
diff --git a/go.mod b/go.mod
index afa6531..1fc926f 100644
--- a/go.mod
+++ b/go.mod
@@ -16,6 +16,7 @@ require (
 )
 
 require (
+	github.com/spf13/pflag v1.0.5 // indirect
 	golang.org/x/net v0.0.0-20220127200216-cd36cc0744dd // indirect
 	golang.org/x/sys v0.0.0-20211216021012-1d35b9e2eb4e // indirect
 	golang.org/x/text v0.3.7 // indirect
diff --git a/go.sum b/go.sum
index 71b3f14..881fa88 100644
--- a/go.sum
+++ b/go.sum
@@ -133,6 +133,8 @@ github.com/prometheus/client_model v0.0.0-20190812154241-14fe0d1b01d4/go.mod h1:
 github.com/rogpeppe/fastuuid v1.2.0 h1:Ppwyp6VYCF1nvBTXL3trRso7mXMlRrw9ooo375wvi2s=
 github.com/rogpeppe/fastuuid v1.2.0/go.mod h1:jVj6XXZzXRy/MSR5jhDC/2q6DgLz+nrA6LYCDYWNEvQ=
 github.com/rogpeppe/go-internal v1.3.0/go.mod h1:M8bDsm7K2OlrFYOpmOWEs/qY81heoFRclV5y23lUDJ4=
+github.com/spf13/pflag v1.0.5 h1:iy+VFUOCP1a+8yFto/drg2CJ5u0yRoB7fZw3DKv/JXA=
+github.com/spf13/pflag v1.0.5/go.mod h1:McXfInJRrz4CZXVZOBLb0bTZqETkiAhM9Iw0y3An2Bg=
 github.com/stretchr/objx v0.1.0/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
 github.com/stretchr/testify v1.4.0/go.mod h1:j7eGeouHqKxXV5pUuKE4zz7dFj8WfuZ+81PSLYec5m4=
 github.com/stretchr/testify v1.5.1/go.mod h1:5W2xD1RspED5o8YsWQXVCued0rvSQ+mT+I5cxcmMvtA=
diff --git a/internal/descriptor/registry.go b/internal/descriptor/registry.go
index 87a2cfe..7be4576 100644
--- a/internal/descriptor/registry.go
+++ b/internal/descriptor/registry.go
@@ -576,9 +576,12 @@ func (r *Registry) GetOmitEnumDefaultValue() bool {
 }
 
 // SetVisibilityRestrictionSelectors sets the visibility restriction selectors.
-func (r *Registry) SetVisibilityRestrictionSelectors(selectors string) {
+func (r *Registry) SetVisibilityRestrictionSelectors(selectors *[]string) {
 	r.visibilityRestrictionSelectors = make(map[string]bool)
-	for _, selector := range strings.Split(selectors, ",") {
+	if selectors == nil {
+		return
+	}
+	for _, selector := range *selectors {
 		r.visibilityRestrictionSelectors[strings.TrimSpace(selector)] = true
 	}
 }
diff --git a/protoc-gen-openapiv2/main.go b/protoc-gen-openapiv2/main.go
index 5a290f3..801c59a 100644
--- a/protoc-gen-openapiv2/main.go
+++ b/protoc-gen-openapiv2/main.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"flag"
 	"fmt"
 	"os"
 	"strings"
@@ -10,6 +9,7 @@ import (
 	"github.com/grpc-ecosystem/grpc-gateway/v2/internal/codegenerator"
 	"github.com/grpc-ecosystem/grpc-gateway/v2/internal/descriptor"
 	"github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2/internal/genopenapi"
+	flag "github.com/spf13/pflag"
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/pluginpb"
 )
@@ -37,7 +37,7 @@ var (
 	generateUnboundMethods         = flag.Bool("generate_unbound_methods", false, "generate swagger metadata even for RPC methods that have no HttpRule annotation")
 	recursiveDepth                 = flag.Int("recursive-depth", 1000, "maximum recursion count allowed for a field type")
 	omitEnumDefaultValue           = flag.Bool("omit_enum_default_value", false, "if set, omit default enum value")
-	visibilityRestrictionSelectors = flag.String("visibility_restriction_selectors", "", "comma separated list of `google.api.VisibilityRule` visibility labels to include in the generated output when a visibility annotation is defined. Elements without visibility annotations are unaffected by this setting.")
+	visibilityRestrictionSelectors = flag.StringArray("visibility_restriction_selectors", []string{}, "comma separated list of `google.api.VisibilityRule` visibility labels to include in the generated output when a visibility annotation is defined. Elements without visibility annotations are unaffected by this setting.")
 )
 
 // Variables set by goreleaser at build time
@@ -115,7 +115,7 @@ func main() {
 	reg.SetGenerateUnboundMethods(*generateUnboundMethods)
 	reg.SetRecursiveDepth(*recursiveDepth)
 	reg.SetOmitEnumDefaultValue(*omitEnumDefaultValue)
-	reg.SetVisibilityRestrictionSelectors(*visibilityRestrictionSelectors)
+	reg.SetVisibilityRestrictionSelectors(visibilityRestrictionSelectors)
 	if err := reg.SetRepeatedPathParamSeparator(*repeatedPathParamSeparator); err != nil {
 		emitError(err)
 		return
```

protoc-gen-openapiv2/internal/genopenapi/template.go
BCook98: Great call. I have removed these.

johanbrandhorst:
Instead of adding a dependency, what about a custom flag variable like this: https://stackoverflow.com/a/28323276? I think it'll be important for us to emphasize how to use multiple values correctly (remove any mention of comma separation from all the docs).

docs/docs/mapping/customizing_openapi_output.md
```diff
@@ -284,3 +284,108 @@ Output json:
 },
 ```
 {% endraw %}
+
+### Hiding fields, methods, services and enum values
+
+If you require internal or unreleased fields and APIs to be hidden from your API documentation, [`google.api.VisibilityRule`](https://github.com/googleapis/googleapis/blob/9916192ab15e3507e41ba2c5165182fec06120d0/google/api/visibility.proto#L89) annotations can be added customize where they are generated. Combined with the option `visibility_restriction_selectors` overlapping rules will appear in the OpenAPI output. 
+
+`visibility_restriction_selectors` can be declared multiple times as an option to include multiple visibility restrictions in the output. 
+e.g. `opt: visibility_restriction_selectors=PREVIEW,visibility_restriction_selectors=INTERNAL`.
```
johanbrandhorst, line 293: What do you think about adding a `protoc` example here too?
```suggestion
e.g. if you are using `buf`:

```yaml
version: v1
plugins:
  - name: openapiv2
    out: .
    opt:
      - visibility_restriction_selectors=PREVIEW
      - visibility_restriction_selectors=INTERNAL
```
or, with `protoc`:

```shell
$ protoc --openapiv2_out=. --openapiv2_opt=visibility_restriction_selectors=PREVIEW --openapiv2_opt=visibility_restriction_selectors=INTERNAL ./path/to/file.proto
``` 

```

File: docs/docs/mapping/customizing_openapi_output.md [hunk 9]
```diff
@@ -284,3 +284,108 @@ Output json:
...
+
+Elements without `google.api.VisibilityRule` annotations will appear as usual in the generated output.
+
+These restrictions and selectors are completely arbitrary and you can define whatever values or hierarchies you want. In the example `INTERNAL`, `PREVIEW` are used, but `INTERNAL`, `ALPHA`, `BETA`, `RELEASED`, or anything else could be used if you wish.
```
johanbrandhorst, line 297:
```suggestion
These restrictions and selectors are completely arbitrary and you can define whatever values or hierarchies you want. In this example we use `INTERNAL` and `PREVIEW`, but `INTERNAL`, `ALPHA`, `BETA`, `RELEASED`, or anything else could be used if you wish.
```

protoc-gen-openapiv2/defs.bzl
```diff
@@ -119,6 +120,9 @@ def _run_proto_gen_openapi(
     if proto3_optional_nullable:
         args.add("--openapiv2_opt", "proto3_optional_nullable=true")
 
+    if visibility_restriction_selectors:
+        args.add("--openapiv2_opt", "visibility_restriction_selectors=%s" % visibility_restriction_selectors)
```
johanbrandhorst, line 124: I assume we actually need to loop over `visibility_restriction_selectors` now and add a new argument for each element, instead of this string interpolation?

protoc-gen-openapiv2/defs.bzl
```diff
@@ -323,6 +328,13 @@ protoc_gen_openapiv2 = rule(
             doc = "generate swagger metadata even for RPC methods that have" +
                   " no HttpRule annotation",
         ),
+        "visibility_restriction_selectors": attr.string(
```
johanbrandhorst, line 331: This needs to be a string array, I hope there is such a type 😄.

utilities/string_array_flag.go
```diff
@@ -0,0 +1,35 @@
...
+type StringArrayFlags []string
+
+// String returns a string representation of `StringArrayFlags`
+func (i *StringArrayFlags) String() string {
+	return strings.Join(*i, ",")
+}
+
+// Set appends a value to `StringArrayFlags`
+func (i *StringArrayFlags) Set(value string) error {
+	fmt.Println(value)
```
johanbrandhorst, line 32: I think we want to drop this?

johanbrandhorst (commented):
Almost there, just a few nits left!

johanbrandhorst (approved):
LGTM 🎉

BCook98:
Thanks for the feedback! I've addressed all of those comments. 

Just running a quick manual test on our API protos to make sure it's all still working as expected.

johanbrandhorst:
Looks like we need another generation after the `master` merge

BCook98:
I believe I have fixed all of the test/lint failures.

Also a manual test across our APIs gives us the expected output 🎉.

johanbrandhorst:
🤞🏻

johanbrandhorst:
Thank you for this incredible contribution ❤️

PR #2596, opened 2022-03-19
Add delimiter after response stream error message (#2591)

### LINKED ISSUE #2591: Error message on server streaming endpoint does not get delimiter appended
## 🐛 Bug Report

When calling a Server Streaming grpc endpoint through grpc-gateway, when an error is returned, it does not have a delimiter appended (by default, `\n`) unlike other streaming messages.

## To Reproduce

1. Call an endpoint that maps onto a Server Streaming grpc endpoint, that returns an error.

## Expected behavior

grpc-gateway returns `{"error":....}\n` (or other configured delimiter)

## Actual Behavior

grpc-gateway returns `{"error":....}` (without the configured delimiter)

## Your Environment

Mac OS 12.2.1, Go 1.17.6, grpc-gateway v2.8.0, google.golang.org/grpc v1.45.0

### Note

Am aware this might not be a "bug" - the behaviour is always the same when passing errors, so it might be as designed. If that's the case, please don't hesitate to tell me so :D

Should this be considered a bug, am happy to try to get a PR in as well.

stelcodes:
Fixes #2591 

Yes

I made forwardHandleResponseStreamError append a delimiter to the http.ResponseWriter after an error message. I also fixed the function's test by adding the delimiter to the expected output.

#### Other comments
I did not regenerate any code because I don't think that's necessary but I could be wrong. This is my first PR not only for grpc-gateway but for any Golang open source project so please let me know how I can improve the patch. Cheers.

### DIFF
--- a/runtime/handler_test.go
+++ b/runtime/handler_test.go
@@ -120,6 +120,7 @@ func TestForwardResponseStream(t *testing.T) {
 						// Skip non-stream errors
 						t.Skip("checking error encodings")
 					}
+					delimiter := marshaler.Delimiter()
 					st := status.Convert(msg.err)
 					b, err := marshaler.Marshal(map[string]proto.Message{
 						"error": st.Proto(),
@@ -128,7 +129,7 @@ func TestForwardResponseStream(t *testing.T) {
 						t.Errorf("marshaler.Marshal() failed %v", err)
 					}
 					errBytes := body[len(want):]
-					if string(errBytes) != string(b) {
+					if string(errBytes) != string(b)+string(delimiter) {
 						t.Errorf("ForwardResponseStream() = \"%s\" want \"%s\"", errBytes, b)
 					}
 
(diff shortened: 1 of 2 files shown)


johanbrandhorst (approved):
LGTM!

runtime/handler_test.go
```diff
@@ -128,7 +129,7 @@ func TestForwardResponseStream(t *testing.T) {
 						t.Errorf("marshaler.Marshal() failed %v", err)
 					}
 					errBytes := body[len(want):]
-					if string(errBytes) != string(b) {
+					if string(errBytes) != string(b) + string(delimiter) {
```
johanbrandhorst, line 132: This should be
```suggestion
					if string(errBytes) != string(b)+string(delimiter) {
```
to satisfy the linter

johanbrandhorst (approved):

PR #2599, opened 2022-03-21
protoc-gen-openapiv2: Use the canonical camelCase converter for protobuf

### LINKED ISSUE #2363: protoc-gen-openapiv2: json_names_for_fields=true can generate invalid OpenAPI output
## 🐛 Bug Report

With `json_names_for_fields=true` (the default), some parameter names cause the generated OpenAPI JSON to contain parameter name mismatches.

## To Reproduce

Create an `example.proto` file with the following content:

```
syntax = "proto3";
package example;
option go_package = ".;example";

message Get1Request {
	string FooBar = 1;
}

message Get1Response {
	string X = 1;
}

message Get2Request {
	string ID = 1;
}

message Get2Response {
	string Y = 1;
}

service Example {
	rpc Get1(Get1Request) returns (Get1Response) {
		option (google.api.http) = {
			get: "/v1/foo/{FooBar=*}"
		};
	}
	rpc Get2(Get2Request) returns (Get2Response) {
		option (google.api.http) = {
			get: "/v1/bar/{ID=*}"
		};
	}
}
```
Copy the `third_party` directory from the `github.com/grpc-ecosystem/grpc-gateway` repository into a local `third_party` directory.

Generate the OpenAPI JSON by running:

```
protoc -I. -I./third_party/googleapis --openapiv2_out=. example.proto
```

Run the generated `example.swagger.json` file through an OpenAPI verifier, such as https://editor.swagger.io/.

## Expected behavior

I'd expect the verification to succeed (i.e. the generated OpenAPI should be valid).

## Actual Behavior

There is an inconsistency between the parameter names in the URL and those in the `parameters` field.

For example:

 ```
    "/v1/bar/{iD}": {
      "get": {
...
        "parameters": [
          {
            "name": "ID",
```

OpenAPI parameters are case sensitive.

Here's an example of the reported errors:

```
Semantic error at paths./v1/bar/{iD}
Parameter names are case-sensitive. The parameter named "ID" does not match the case used in the path "/v1/bar/{iD}".
Jump to line 18
Semantic error at paths./v1/foo/{podName}
Parameter names are case-sensitive. The parameter named "PodName" does not match the case used in the path "/v1/foo/{podName}".
Jump to line 48
```

(as an aside: `iD` is not a great spelling for the JSON name).

## Your Environment

I'm using github.com/grpc-ecosystem/grpc-gateway/v2  at v2.5.0 under Ubuntu.

### PR DESCRIPTION
**Pull request by oyvindwe**

Fixes #2363

Yes.

Fixes the generation of camelCase names in generated OAS files from field names in proto files.

#### Other comments

Updated `internal/casing/LICENSE.md`.

### DIFF
--- a/internal/casing/camel.go
+++ b/internal/casing/camel.go
@@ -50,6 +50,24 @@ func Camel(s string) string {
 	return string(t)
 }
 
+// JSONCamelCase converts a snake_case identifier to a camelCase identifier,
+// according to the protobuf JSON specification.
+func JSONCamelCase(s string) string {
+	var b []byte
+	var wasUnderscore bool
+	for i := 0; i < len(s); i++ { // proto identifiers are always ASCII
+		c := s[i]
+		if c != '_' {
+			if wasUnderscore && isASCIILower(c) {
+				c -= 'a' - 'A' // convert to uppercase
+			}
+			b = append(b, c)
+		}
+		wasUnderscore = c == '_'
+	}
+	return string(b)
+}
+
 // And now lots of helper functions.
 
 // Is c an ASCII lower-case letter?
--- a/internal/casing/LICENSE.md
+++ b/internal/casing/LICENSE.md
@@ -1,4 +1,4 @@
-Copyright 2010 The Go Authors.  All rights reserved.
+Copyright 2010, 2019 The Go Authors.  All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
 modification, are permitted provided that the following conditions are
--- a/internal/casing/README.md
+++ b/internal/casing/README.md
@@ -1,5 +1,7 @@
 # Case conversion
 
-This package contains a single function, copied from the
-`github.com/golang/protobuf/protoc-gen-go/generator` package. That
-modules LICENSE is referenced in its entirety in this package.
+This package contains two functions:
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

**Comment by johanbrandhorst** [comment]
The `proto_lint` error should be fixed by merging in master.

File: internal/casing/camel.go [hunk 1]
```diff
@@ -50,6 +50,24 @@ func Camel(s string) string {
 	return string(t)
 }
```
johanbrandhorst, line 51: Hm, do we still need the old function? We should probably always use the new one?

johanbrandhorst: I expect their logic to be roughly the same, though it might be worth using the new one anyway. We should probably always be consistent with what APIv2 uses (the new repo).

johanbrandhorst (approved):

internal/casing/camel.go
johanbrandhorst: I don't think we need to make any changes here, but there may be a subtle bug hidden somewhere here that will eventually rear its head. By APIv2 I mean the Go protobuf stack in google.golang.org/protobuf. That's the "new" stack. See https://blog.golang.org/protobuf-apiv2 for more information. The grpc-gateway v2 is built on APIv2.

PR #2600, opened 2022-03-21
protoc-gen-openapiv2: Remove path parameters from body when body is a snake_case field

oyvindwe:

Fixes a case missed in PR #2553

Yes.

If the JSON representation of the body field name is different from the native proto field name, path parameters were not removed from the body.

#### Other comments

None.

### DIFF
--- a/protoc-gen-openapiv2/internal/genopenapi/template.go
+++ b/protoc-gen-openapiv2/internal/genopenapi/template.go
@@ -1171,7 +1171,7 @@ func renderServices(services []*descriptor.Service, paths openapiPathsObject, re
 							bodyFieldName = bodyField.Name
 						}
 						// Align pathParams with body field path.
-						pathParams := subPathParams(bodyFieldName, b.PathParams)
+						pathParams := subPathParams(bodyField.Name, b.PathParams)
 						var err error
 						schema, err = renderFieldAsDefinition(bodyField.Target, reg, customRefs, pathParams)
 						if err != nil {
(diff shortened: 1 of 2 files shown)


protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -1164,7 +1164,7 @@ func renderServices(services []*descriptor.Service, paths openapiPathsObject, re
 							bodyFieldName = bodyField.Name
 						}
 						// Align pathParams with body field path.
-						pathParams := subPathParams(bodyFieldName, b.PathParams)
+						pathParams := subPathParams(bodyField.Name, b.PathParams)
```
johanbrandhorst, line 1167: Are there no test changes required for this? Do we have any test cases that cover this?

oyvindwe: > Are there no test changes required for this? Do we have any test cases that cover this?

I will add a test case.

oyvindwe: @johanbrandhorst test case added!

johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution! I think this warrants a patch release!

oyvindwe:
> Thanks for your contribution! I think this warrants a patch release!

I'm happy to help out! Unfortunately this is a side project, so it took a while to get the unit test in place, but I should have made that as a part of PR #2553.

A patch release would be great!

PR #1978, opened 2021-02-17
Use base64.URLEncoding for []byte parameters in query

tvoll:
Changes encoding method to base64.URLEncoding for query parameters of type []byte.

PR #2603, opened 2022-03-22
Use Bytes from convert.go to unmarshal byte value

HubertZhang:
Yes 
`bytes` field in path can use `base64.StdEncoding` form also. 

In #1978 , the way how `bytes` field in path is unmarshalled is changed to `base64.URLEncoding`. However, the `bytes` field is marshalled using `base64.StdEncoding` in json. If both these encodings are accepted as a path components, user may use result in json output as path parameter directly.

### DIFF
--- a/runtime/query.go
+++ b/runtime/query.go
@@ -1,7 +1,6 @@
 package runtime
 
 import (
-	"encoding/base64"
 	"errors"
 	"fmt"
 	"net/url"
@@ -234,7 +233,7 @@ func parseField(fieldDescriptor protoreflect.FieldDescriptor, value string) (pro
 	case protoreflect.StringKind:
 		return protoreflect.ValueOfString(value), nil
 	case protoreflect.BytesKind:
-		v, err := base64.URLEncoding.DecodeString(value)
+		v, err := Bytes(value)
 		if err != nil {
 			return protoreflect.Value{}, err
 		}
@@ -312,7 +311,7 @@ func parseMessage(msgDescriptor protoreflect.MessageDescriptor, value string) (p
 	case "google.protobuf.StringValue":
 		msg = &wrapperspb.StringValue{Value: value}
 	case "google.protobuf.BytesValue":
-		v, err := base64.URLEncoding.DecodeString(value)
+		v, err := Bytes(value)
 		if err != nil {
 			return protoreflect.Value{}, err
 		}


johanbrandhorst:
Hi, I'm not sure I see the benefit or rationale here. We can't use `StdEncoding` in paths because it's not safe to store in URLs.

HubertZhang:
Hi, the scene I encounter is that I used bytes as an identifier in our API. Then the create API would return a json object with identifier encoded in `StdEncoding`. When it is used as parameter in query, even if it is escaped, it will be rejected by gateway.

Following is an example. An object may be created and has id `\xff\xff`, or encoded `//8=`. `GET /v1/resource/%2F%2F8%3D`would works since 2.7.0's `UnescapeMode` change, but `GET /v1/resources?begin=%2F%2F8%3D` does not since gateway rejects `StdEncoding` in params.

```protobuf
message Object {
  bytes id = 1;
  bytes some_other_data = 2;
}

message ScanRequest {
  bytes begin = 1;
  bytes end = 2;
}

message Objects {
  repeated Object objs = 1;
}

service A {
  rpc Create(Object) returns (Object) {
    option (google.api.http) = {
      post: "/v1/resource"
      body: "*"
    };
  }
  rpc Get(Object) returns (Object) {
    option (google.api.http) = {
      get: "/v1/resource/{id}"
    };
  }
  rpc Scan(ScanRequests) returns (Objects) {
    option (google.api.http) = {
      get: "/v1/resources"
    };
  }
}
```

johanbrandhorst:
> Hi, the scene I encounter is that I used bytes as an identifier in our API. Then the create API would return a json object with identifier encoded in `StdEncoding`. When it is used as parameter in query, even if it is escaped, it will be rejected by gateway.
> 
> Following is an example. An object may be created and has id `\xff\xff`, or encoded `//8=`. `GET /v1/resource/%2F%2F8%3D`would works since 2.7.0's `UnescapeMode` change, but `GET /v1/resources?begin=%2F%2F8%3D` does not since gateway rejects `StdEncoding` in params.
> 
> ```proto
> message Object {
>   bytes id = 1;
>   bytes some_other_data = 2;
> }
> 
> message ScanRequest {
>   bytes begin = 1;
>   bytes end = 2;
> }
> 
> message Objects {
>   repeated Object objs = 1;
> }
> 
> service A {
>   rpc Create(Object) returns (Object) {
>     option (google.api.http) = {
>       post: "/v1/resource"
>       body: "*"
>     };
>   }
>   rpc Get(Object) returns (Object) {
>     option (google.api.http) = {
>       get: "/v1/resource/{id}"
>     };
>   }
>   rpc Scan(ScanRequests) returns (Objects) {
>     option (google.api.http) = {
>       get: "/v1/resources"
>     };
>   }
> }
> ```

I see, this is about changing the parsing of query parameters. I think that might make sense, since they're escaped already, unlike URL path parameters.

johanbrandhorst (approved):

PR #2628, opened 2022-04-06
support google.protobuf.Value

wclssdn:

yes

Support `google.protobuf.Value` type in proto files, when request with x-www-form-urlencoded.


runtime/query.go
```diff
@@ -320,6 +322,17 @@ func parseMessage(msgDescriptor protoreflect.MessageDescriptor, value string) (p
...
+		var x interface{}
+		err := json.Unmarshal([]byte(value), &x)
+		if err != nil {
+			return protoreflect.Value{}, err
+		}
+		v, err := structpb.NewValue(x)
+		if err != nil {
+			return protoreflect.Value{}, err
+		}
+		msg = v
```
johanbrandhorst, line 335: I think we should use `protojson`, not the standard library marshaler. Also, can't we just do this in one fell swoop?
```suggestion
		var v structpb.Value
		err := protojson.Unmarshal([]byte(value), &v)
		if err != nil {
			return protoreflect.Value{}, err
		}
		msg = &v
```

johanbrandhorst (changes requested):
Please add a few new testcases to test the marshaling behavior. Should we add `google.protobuf.Struct` support at the same time?

wclssdn:
> Please add a few new testcases to test the marshaling behavior. Should we add `google.protobuf.Struct` support at the same time?

OK, i'm trying to do that.

wclssdn:
> Please add a few new testcases to test the marshaling behavior. Should we add `google.protobuf.Struct` support at the same time?

I have added some testcases to test `google.protobuf.Struct` and `google.protobuf.Value` support.

wclssdn:
> 

It's very common in different projects. Especially in PHP projects. If I rewrite PHP project to Go, I must implement the origin logic.

Even a Go project, that also need a unknown json object to deal with. e.g. `payload` field, it provide a method for client to save there personal data.

johanbrandhorst:
Please rebase on master for latest linter

wclssdn:
> Please rebase on master for latest linter

done~

johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution!

PR #2634, opened 2022-04-09
fix: Content-Type and Grpc-Metadata-Content-Type headers with the health endpoint

### LINKED ISSUE #2611: WithHealthzEndpoint should answer with application/json Content-Type
## 🐛 Bug Report

`WithHealthzEndpoint` didn't set Content-Type to anything but `text/plain; charset=utf-8`. Shouldn't it be application/json?

## To Reproduce

1. Enable WithHealthzEndpoint
2. GET to the /healthz endpoint

## Expected behavior

```
Content-Type: application/json
Grpc-Metadata-Content-Type: application/grpc
```

## Actual Behavior

```
Content-Type: text/plain; charset=utf-8
```

## Your Environment
Go 1.18 with grpc-gateway v2.10.0
Linux 5.17.0-arch1-1 #1 SMP PREEMPT Tue, 22 Mar 2022 23:36:09 +0000 x86_64 GNU/Linux

GreyXor:
Fixes #2611

Manually set the `Content-Type` and `Grpc-Metadata-Content-Type` headers to `application/json` and `application/grpc` respectively.
I also added a test to check the headers of the `healthCheckTests` set.

#### Other comments
Since we stay in the [WithHealthEndpointAt()](https://github.com/grpc-ecosystem/grpc-gateway/blob/d8e890717a850f4bcc23f192fb567176d7f92e1b/runtime/mux.go#L216) method and we use the [outgoing marshaler](https://github.com/grpc-ecosystem/grpc-gateway/blob/d8e890717a850f4bcc23f192fb567176d7f92e1b/runtime/mux.go#L222) directly in it, we can't take advantage of the Content-Type deduction (except in the [errorHandler](https://github.com/grpc-ecosystem/grpc-gateway/blob/d8e890717a850f4bcc23f192fb567176d7f92e1b/runtime/mux.go#L241)). In my opinion it is easier in this case to directly add the headers [before](https://github.com/grpc-ecosystem/grpc-gateway/blob/d8e890717a850f4bcc23f192fb567176d7f92e1b/runtime/mux.go#L231) writing the response writer. 

CC @brumhard @antonioiubatti93

### DIFF
--- a/runtime/mux.go
+++ b/runtime/mux.go
@@ -229,6 +229,9 @@ func WithHealthEndpointAt(healthCheckClient grpc_health_v1.HealthClient, endpoin
 					return
 				}
 
+				w.Header().Set("Content-Type", "application/json")
+				w.Header().Set("Grpc-Metadata-Content-Type", "application/grpc")
+
 				if resp.GetStatus() != grpc_health_v1.HealthCheckResponse_SERVING {
 					var err error
 					switch resp.GetStatus() {
--- a/runtime/mux_test.go
+++ b/runtime/mux_test.go
@@ -719,6 +719,34 @@ func TestWithHealthzEndpoint_serviceParam(t *testing.T) {
 	}
 }
 
+func TestWithHealthzEndpoint_header(t *testing.T) {
+	for _, tt := range healthCheckTests {
+		t.Run(tt.name, func(t *testing.T) {
+			mux := runtime.NewServeMux(runtime.WithHealthzEndpoint(&dummyHealthCheckClient{status: tt.status, code: tt.code}))
+
+			r := httptest.NewRequest(http.MethodGet, "/healthz", nil)
+			rr := httptest.NewRecorder()
+
+			mux.ServeHTTP(rr, r)
+
+			// Expected header values in the HTTP response
+			expectedHeaderValue := map[string]string{
+				"Content-Type": "application/json",
+				//"Grpc-Metadata-Content-Type": "application/grpc",
+			}
+
+			for header, value := range expectedHeaderValue {
+				if actualHeader := rr.Header().Get(header); actualHeader != value {
+					t.Errorf(
+						"result http header \"%s\" for grpc code %q and status %q should be %s, got %s",
+						header, tt.code, tt.status, value, actualHeader,
+					)
+				}
+			}
+		})
+	}
+}
+
 var _ grpc_health_v1.HealthClient = (*dummyHealthCheckClient)(nil)
 
 type dummyHealthCheckClient struct {


runtime/mux.go
```diff
@@ -229,6 +229,9 @@ func WithHealthEndpointAt(healthCheckClient grpc_health_v1.HealthClient, endpoin
 					return
 				}
 
+				w.Header().Set("Content-Type", "application/json")
+				w.Header().Set("Grpc-Metadata-Content-Type", "application/grpc")
```
johanbrandhorst, line 233: Could we add a comment to explain why we need both of these please?

runtime/mux_test.go
```diff
@@ -719,6 +719,34 @@ func TestWithHealthzEndpoint_serviceParam(t *testing.T) {
...
+
+			r := httptest.NewRequest(http.MethodGet, "/healthz", nil)
+			rr := httptest.NewRecorder()
+
+			mux.ServeHTTP(rr, r)
+
+			// Expected header values in the HTTP response
+			expectedHeaderValue := map[string]string{
+				"Content-Type": "application/json",
+				//"Grpc-Metadata-Content-Type": "application/grpc",
```
johanbrandhorst, line 735: Why is this commented out?

johanbrandhorst (changes requested):

runtime/mux.go
GreyXor: The gRPC RFC indicates that the header should be [application/grpc](https://datatracker.ietf.org/doc/html/draft-kumar-rtgwg-grpc-protocol-00). However when I call the other endpoints I get `application/json` and `Grpc-Metadata-Content-Type` `application/grpc`. So to being consistent with other endpoints I simply copied the headers received.

What do you want as commentary? The fact that i have to manually set the headers or should I talk about the RFC instead?
Thanks

runtime/mux_test.go
GreyXor: This was just an example to show that you can easily add more or less headers to test.
And `Grpc-Metadata-Content-Type` can't pass because it is deleted by the errorHandler (so the test wont pass).
If I uncomment the `Grpc-Metadata` line all tests pass except the 404 one

runtime/mux.go
johanbrandhorst: I don't think we need to include the translated grpc content type header tbh. It's there normally because we automatically translate those headers, but we don't need to set it here explicitly. If we remove it, there's little reason for a comment at all, I think it's pretty self explanatory why we need to set the JSON content type.

GreyXor: Thanks for your feedback, I have removed the grpc header and then no comments are needed

runtime/mux_test.go
GreyXor: While there is only one header left to test. Would you like that i change the map+loop to something lighter?

johanbrandhorst: Sure, lets just do a hardcoded check, thanks.

johanbrandhorst:
It looks like the CLA bot is unhappy, could you try following the instructions here?

> Thanks for your pull request! It looks like this may be your first contribution to a Google open source project. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> For more information, open the [CLA check for this pull request](https://github.com/grpc-ecosystem/grpc-gateway/pull/2634/checks?check_run_id=5957434240).

If you've done this and it's still not working, just include a screenshot of the signed confirmation on their website and I can force merge it.

GreyXor:
@johanbrandhorst Hello, now the CLA should be good, thanks

johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution!

PR #2650, opened 2022-04-16
Format protobuf files with buf

johanbrandhorst:
Format protobuf files with `buf`. I think this came out pretty nice!


achew22 (approved):

tmc:
We might want to consider `.git-blame-ignore-revs` to prevent making git blame history noisy from this change.

johanbrandhorst:
TIL, that's a great idea

PR #2651, opened 2022-04-17
Export `defaultQueryParser` struct for custom query parsers

### LINKED ISSUE #2632: Default query parameter parsing behaviour change
## 🐛 Bug Report

The behavior when parsing duplicate query parameters has changed between v1 and v2.

For example in v1, making the `GET` request `/v2/example/echo?value=foo&value=bar` will log an error `too many field values: name` but return a HTTP 200 status code and response `{ "value": "foo" }`.

However, in v2 when making this same request, it will return an HTTP 400 status code and the error with message `too many values for field "value": foo, bar`.

We expected v2 to behave like v1 after upgrading since this change was not documented in the [v2 Migration Guide](https://grpc-ecosystem.github.io/grpc-gateway/docs/development/grpc-gateway_v2_migration_guide/). We have a client using our v1 gRPC Gateway that cannot be changed easily and it is sending duplicate query parameters but it is OK that the duplicate query parameters are ignored/ discarded.

The only way we can get v2 to behave like v1 is to set our own custom query parser via `runtime.SetQueryParameterParser()` but this requires us to copy verbatim most of the code in `runtime/query.go` [minus these few lines of code](https://github.com/grpc-ecosystem/grpc-gateway/blob/v2.10.0/runtime/query.go#L119-L121)... Not ideal!

Pertinent [v1 lines](https://github.com/grpc-ecosystem/grpc-gateway/blob/v1/runtime/query.go#L123-L125) where it doesn't behave this way.

It would be nice if we could opt-out of this behavior. If this is a feature you're willing to support, I can make a PR implementing it.

## To Reproduce

1. Run the example gRPC server and gateway.
2. Make the `GET` request `localhost:8080/v2/example/echo?value=foo&value=bar`

## Expected behavior

See above.

## Actual Behavior

See above.

## Your Environment

N/A

MikeSouza:

Fixes #2632

Yes

- Updated the v2 migration guide to reflect that the default query parameter parsing behavior has changed.
- Exported the `defaultQueryParser` struct so custom query parsers can wrap it and pass to `SetQueryParameterParser`

### DIFF
--- a/docs/docs/development/grpc-gateway_v2_migration_guide.md
+++ b/docs/docs/development/grpc-gateway_v2_migration_guide.md
@@ -162,3 +162,7 @@ There is no workaround for this, as we considered it a correct interpretation of
 ## Error handling configuration has been overhauled
 
 `runtime.HTTPError`, `runtime.OtherErrorHandler`, `runtime.GlobalHTTPErrorHandler`, `runtime.WithProtoErrorHandler` are all gone. Error handling is rewritten around the use of gRPCs Status types. If you wish to configure how the gateway handles errors, please use `runtime.WithErrorHandler` and `runtime.WithStreamErrorHandler`. To handle routing errors (similar to the removed `runtime.OtherErrorHandler`) please use `runtime.WithRoutingErrorHandler`.
+
+## Default query parameter parsing behaviour change
+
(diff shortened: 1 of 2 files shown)


johanbrandhorst (approved):
LGTM

PR #2679, opened 2022-05-11
Fix example in openapiv2.proto

baryluk:
Makes it more copy-pastable


baryluk:
Wrong part about the schema. Ignore.

PR #2699, opened 2022-05-20
Added support for `(google.api.message_visibility).restriction`

elb3k:
#### Have you read the [Contributing Guidelines](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/CONTRIBUTING.md)? Yes

`openapiv2` had support for hiding methods, values, enums, fields, but not messages.
So I added support for `(google.api.message_visibility).restriction`, which simply hides message definitions if it is restricted.


examples/internal/proto/examplepb/visibility_rule_none_echo_service.swagger.json
```diff
@@ -112,6 +112,44 @@
...
+    },
+    "/v1/example/echo_hidden": {
+      "get": {
+        "summary": "EchoInternalMessage method is always visible",
+        "operationId": "VisibilityRuleEchoService_EchoInternalMessage",
+        "responses": {
+          "200": {
+            "description": "A successful response.",
+            "schema": {
+              "$ref": "#/definitions/examplepbHiddenMessage"
```
johanbrandhorst, line 124: This reference is still here, even though the definition is not. Isn't this an invalid OpenAPI spec now?

examples/internal/proto/examplepb/visibility_rule_none_echo_service.swagger.json
```diff
@@ -112,6 +112,44 @@
...
+            "type": "string"
+          },
+          {
+            "name": "num",
+            "in": "query",
+            "required": false,
+            "type": "string",
+            "format": "int64"
+          }
+        ],
```
johanbrandhorst, line 148: These parameters are inferred from the message... I guess since the fields themselves are not marked as INTERNAL this might be OK, but it still feels a bit weird.

johanbrandhorst (changes requested):
I think we have the remove references to hidden messages too, but it makes me think that this RPC shouldn't be visible at all, since we can't really show an RPC with no input and output. What do you think? CC @BCook98

examples/internal/proto/examplepb/visibility_rule_none_echo_service.swagger.json
elb3k: I fixed this

examples/internal/proto/examplepb/visibility_rule_none_echo_service.swagger.json
elb3k: Was not sure if parameters should be listed when request type is hidden.

johanbrandhorst (commented):
The more I think about this, the less I think it's going to be possible for us to implement easily like this. As we've seen in this simple example, hiding a message requires us to add logic for what to do when that message is used in a RPC request/response. It might be possible for us to do this if we also introduce the limitation that the message must NOT be used as a request/response. But what about when that request/response is in another protobuf file altogether? This just gets more and more complicated, and I think we need to go back to the drawing board and come up with a full solution before attempting this.

elb3k:
> The more I think about this, the less I think it's going to be possible for us to implement easily like this. As we've seen in this simple example, hiding a message requires us to add logic for what to do when that message is used in a RPC request/response. It might be possible for us to do this if we also introduce the limitation that the message must NOT be used as a request/response. But what about when that request/response is in another protobuf file altogether? This just gets more and more complicated, and I think we need to go back to the drawing board and come up with a full solution before attempting this.

fair enough.

fals:
Did we come to any conclusion or way to hide messages with this attribute? Can't find anything else related but the issues in the repo.

johanbrandhorst:
Not that I know

Issue #2691, opened 2022-05-18
protoc-gen-openapiv2: RPC visibility setting is not transitively applied to request and response messages


elb3k:
## 🐛 Bug Report

Definitions of some internal only grpc request/responses are still visible on generated spec.
Even if endpoints are specifically marked as `INTERNAL` using `option (google.api.method_visibility).restriction = "INTERNAL";`

## To Reproduce

Sample proto defintions:
```proto
syntax = "proto3";
option go_package = "sample.service";
package service;

import "protoc-gen-openapiv2/options/annotations.proto";
import "google/api/annotations.proto";
import "google/api/visibility.proto";

message HelloRequest {
    string name = 1;
}

message HelloReply {
    string message = 1;
}

message ByeRequest {
    string name = 1;
}

message ByeReply {
    string message = 1;
}

service Service {
    rpc Hello(HelloRequest) returns (HelloReply){
        option (google.api.method_visibility).restriction = "INTERNAL";
        option (google.api.http) = {
            post: "/hello"
            body: "*"
          };
          option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
            description: "Sample message",
          };
    }

    rpc Bye(ByeRequest) returns (ByeReply){
        option (google.api.http) = {
            post: "/bye"
            body: "*"
          };
          option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
            description: "Sample message",
          };
    } 
}
```
buf.yaml:
```yaml
version: v1
deps:
  - buf.build/googleapis/googleapis
  - buf.build/grpc-ecosystem/grpc-gateway
breaking:
  use:
    - FILE
```
buf.gen.yaml:
```yaml
version: v1
plugins:
  - name: openapiv2
    out: oapi
    opt:
      - json_names_for_fields=false
```

generated `sample.swagger.json`
```json
....
"definitions": {
    "protobufAny": {
      "type": "object",
      "properties": {
        "@type": {
          "type": "string"
        }
      },
      "additionalProperties": {}
    },
    "rpcStatus": {
      "type": "object",
      "properties": {
        "code": {
          "type": "integer",
          "format": "int32"
        },
        "message": {
          "type": "string"
        },
        "details": {
          "type": "array",
          "items": {
            "$ref": "#/definitions/protobufAny"
          }
        }
      }
    },
    "serviceByeReply": {
      "type": "object",
      "properties": {
        "message": {
[...]

PR #2715, opened 2022-05-25
Infer request/response visibility from the method

elb3k:

https://github.com/grpc-ecosystem/grpc-gateway/issues/2691

Visibility of message can be inferred from the visibility of method. 
Should solve the problem in https://github.com/grpc-ecosystem/grpc-gateway/issues/2691


johanbrandhorst (commented):
I'm not sure this is right either, what if I am reusing a message between RPCs and only one of the RPCs is hidden? We'd also need to do the inverse, that is, hide an RPC if a message is hidden. I'm still worried about hiding messages that are used in other packages too. I'm not sure there is a satisfying solution to this problem that isn't very complicated.

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -1009,10 +1016,34 @@ func renderServices(services []*descriptor.Service, paths openapiPathsObject, re
...
+
+			swgRspName, ok := fullyQualifiedNameToOpenAPIName(meth.ResponseType.FQMN(), reg)
+			if !ok && !skipRenderingRef(meth.ResponseType.FQMN()) {
+				glog.Errorf("couldn't resolve OpenAPI name for FQMN '%v'", meth.ResponseType.FQMN())
+				continue
+			}
+
 			if !isVisible(getMethodVisibilityOption(meth), reg) {
+				// Set request/response as invisible if not set
+				if reg.GetMessageVisibility(swgReqName) == descriptor.MessageVisibility_NOT_SET {
```
elb3k, line 1034: I one is reusing a message between RPCs and only one of the RPCs is hidden, message will not be hidden. It is handled here.

`RenderServices` looks up all RPCs, and marks message as visible if one or more of RPC uses it.

elb3k:
> I'm not sure this is right either, what if I am reusing a message between RPCs and only one of the RPCs is hidden? We'd also need to do the inverse, that is, hide an RPC if a message is hidden. I'm still worried about hiding messages that are used in other packages too. I'm not sure there is a satisfying solution to this problem that isn't very complicated.

I will test the case with different packages.

elb3k:
> I'm not sure this is right either, what if I am reusing a message between RPCs and only one of the RPCs is hidden? We'd also need to do the inverse, that is, hide an RPC if a message is hidden. I'm still worried about hiding messages that are used in other packages too. I'm not sure there is a satisfying solution to this problem that isn't very complicated.

johanbrandhorst:
@elb3k, I am grateful that you want to contribute to the repository, but I closed the last PR because we need to go back to the drawing board on this design before we can start implementing it. This PR has several issues and we should attempt to sort out everything as part of the design so we don't waste both of ours time reviewing things that are not going to work. Please do not open another PR, lets discuss options in #2691.

elb3k:
> @elb3k, I am grateful that you want to contribute to the repository, but I closed the last PR because we need to go back to the drawing board on this design before we can start implementing it. This PR has several issues and we should attempt to sort out everything as part of the design so we don't waste both of ours time reviewing things that are not going to work. Please do not open another PR, lets discuss options in #2691.

ok

PR #2727, opened 2022-05-30
Fix overriding path parameter with custom verbs

### LINKED ISSUE #2688: Overriding path parameter names does not work with path with colon `:`
## 🐛 Bug Report

**Plugin version: `buf.build/grpc-ecosystem/plugins/openapiv2:v2.10.0-1`** 

Google [AIP-136](https://google.aip.dev/136) suggests using colon `:` in custom methods.
>The HTTP URI must use a : character followed by the custom verb (:archive in the above example), and the verb in the URI must match the verb in the name of the RPC.

The path parameter is not correctly overwritten when the path contains `:`.
 
## To Reproduce

1. Given a proto
```
syntax = "proto3";

package endpoints.examples.bookstore;
import "protoc-gen-openapiv2/options/annotations.proto";

// A shelf resource.
message Shelf {
  // Shelf resource name.
  string name = 1;
}

// Request message for RenameShelfRequest method.
message RenameShelfRequest {
  string name = 1
      [ (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
        field_configuration : {path_param_name : "shelf.name"}
      } ];
  // New ID of this model
  string new_shelf_id = 2;
}

// The API manages shelves resources.
service Bookstore {
  // Rename a specific bookstore shelf.
  rpc RenameShelf(RenameShelfRequest) returns (Shelf) {
    option (google.api.http) = {
      post : "/{name=shelves/*}:rename"
      body : "*"
    };
  }
}

``` 
2. buf generate `openapiv2.swagger` file
```YAML
swagger: "2.0"
info:
  title: bookstore.proto
  version: version not set
tags:
- name: Bookstore
consumes:
- application/json
produces:
- application/json
paths:
  /{name}:rename:
    post:
      summary: Rename a specific bookstore shelf.
      operationId: Bookstore_RenameShelf
      responses:
        "200":
          description: A successful response.
          schema:
            $ref: '#/definitions/bookstoreShelf'
        default:
          description: An unexpected error response.
          schema:
            $ref: '#/definitions/rpcStatus'
      parameters:
      - name: shelf.name
        in: path
        required: true
        type: string
        pattern: shelves/[^/]+
      - name: body
        in: body
        required: true
        schema:
          type: object
          properties:
            new_shelf_id:
              type: string
              title: New ID of this model
          description: Request message for GetShelf method.
      tags:
      - Bookstore
definitions:
  bookstoreShelf:
    type: object
    properties:
      name:
        type: string
        description: Shelf resource name.
    description: A shelf resource.
  protobufAny:
    type: object
    properties:
      '@type':
        type: string
    additionalProperties: {}
  rpcStatus:
    type: object
    properties:
      code:
        type: integer
        format: int32
      message:
        type: string
      details:
        type: array
        items:
          $ref: '#/definitions/protobufAny'
```

## Expected behavior

Expect to overwrite the path from `{name}:rename` to `/{shelf.name}:rename` 

## Actual Behavior

the path is `{name}:rename`.
image

## Your Environment

- MacOS
- Buf
- Plugin version: `buf.build/grpc-ecosystem/plugins/openapiv2:v2.10.0-1`

oyvindwe:

Fixes #2688

Yes.

Parsing path templates with a custom verb (defined as `":" LITERAL`) at the end would include this substring in the last part,  which could include a path parameter. This in turn doesn't work with the regex used to substitute path parameter names.

I changed the parser so extract the verb into a separate part. An issue with this is that when the path is rebuilt from the parts, joining the verb needs a special case as it uses `:` as delimiter and not `/`.

#### Other comments

I'd rather do it this way than change `canRegexp`, as it also used in `partsToRegexpMap()`.


johanbrandhorst (commented):
Hi @oyvindwe, thanks for another quality PR. Do you think it's worth adding a new mapping in one of the examples that exercises this literal form?

oyvindwe:
> Do you think it's worth adding a new mapping in one of the examples that exercises this literal form?

Yes, that's a good idea. I'll look into it!

oyvindwe:
> > Do you think it's worth adding a new mapping in one of the examples that exercises this literal form?
> 
> Yes, that's a good idea. I'll look into it!

Done!

johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution!

johanbrandhorst:
https://github.com/grpc-ecosystem/grpc-gateway/releases/tag/v2.10.3

PR #634, opened 2018-04-30
[WIP] Support oneof fields in body

yugui:
* `oneof` fields have already been supported in path parameters (#621) and query parameters (#321).
* This PR replaces #416 and fixes #413.

* [ ] oneof fields defined in a message in another package.

PR #2739, opened 2022-06-02
feat: add support for oneof fields in request bodies

### LINKED ISSUE #413: oneof fields are not supported as the "body"?

aesadde:

This PR is a mashup of https://github.com/grpc-ecosystem/grpc-gateway/pull/634 and https://github.com/grpc-ecosystem/grpc-gateway/pull/570.

Fixes #413.

Yes.

The PR adds a template to unroll `oneofs` before applying request templates.

Added tests to the `EchoService`.

#### Other comments

First-time contributor here, any more guidance will be appreciated 😁.


examples/internal/proto/examplepb/echo_service.swagger.json
```diff
@@ -468,6 +468,171 @@
...
+        ]
+      },
+      "patch": {
+        "summary": "EchoBody method receives a simple message and returns it.",
+        "operationId": "EchoService_EchoBody3",
+        "responses": {
+          "200": {
+            "description": "A successful response.",
+            "schema": {
+              "$ref": "#/definitions/examplepbSimpleMessage"
```
johanbrandhorst, line 557: Where did this method came from? Shouldn't there just be a `put`?

internal/descriptor/types.go
```diff
@@ -363,42 +388,40 @@ func (p FieldPath) AssignableExpr(msgExpr string) string {
...
-
-			if c.Target.ForcePrefixedName {
-				oneofFieldName = msg.File.Pkg() + "." + oneofFieldName
+			var (
+				index = c.Target.OneofIndex
+				msg   = c.Target.Message
+				buf   bytes.Buffer
+			)
+
+			// TODO(yugui) Use the package which the caller tempalte is dealing with.
```
johanbrandhorst, line 403:
```suggestion
			// TODO(yugui) Use the package which the caller template is dealing with.
```

examples/internal/proto/examplepb/echo_service.pb.gw.go
```diff
@@ -231,12 +231,15 @@ func request_EchoService_Echo_2(ctx context.Context, marshaler runtime.Marshaler
...
-	protoReq.Code.(*SimpleMessage_Lang).Lang, err = runtime.String(val)
+	func() *SimpleMessage_Lang {
+		msg := &protoReq
+		oneof := msg.Code
+		if _, ok := oneof.(*SimpleMessage_Lang); oneof == nil || !ok {
+			oneof = new(SimpleMessage_Lang)
+			msg.Code = oneof
+		}
+		return oneof.(*SimpleMessage_Lang)
+	}().Lang, err = runtime.String(val)
```
johanbrandhorst, line 242: This looks super whack. Do we need to put this in its own little function like this? Can we not do this in the outer scope?

examples/internal/proto/examplepb/echo_service.swagger.json
aesadde: yep, sorry. Uploaded old version of the generated files.

examples/internal/proto/examplepb/echo_service.pb.gw.go
aesadde: Agreed. It was the simplest way I could find to cover all cases. 

Having this function wrapping everything makes the oneof template also work for cases like [` request_EchoService_EchoBody_1` below](https://github.com/grpc-ecosystem/grpc-gateway/pull/2739/files#diff-72aeb6791cd8a6135693f8392966b173654d28b572b94d68f84796046540dd2cR562).

Do you think a better solution would involve editing the templates directly? I've been playing around with such an approach but nothing that works quite yet with minimal impact on the current templates.

johanbrandhorst: I'm not sure what you mean by editing the templates directly, but we certainly should consider making more sweeping changes to fit in this behavior. Obviously the less we can touch the better.

It also looks like you need to format the protobuf files, you can just run `buf format -w` and that should do it.

johanbrandhorst: No worries. I think we have to go back to the drawing board and accept that there will be template changes to support this. I haven't really got the bandwidth to work on this right now but will be happy to review any ideas you might have.

aesadde: Ok cool. Let me dig a bit more into the code and see what I manage to come up with. I'll mark this as draft for now

aesadde: @johanbrandhorst ended up going on a different route (similar to the old #570). Looks much better now, I hope.

examples/internal/proto/examplepb/echo_service.pb.gw.go
```diff
@@ -289,11 +289,6 @@ func local_request_EchoService_Echo_2(ctx context.Context, marshaler runtime.Mar
 		return nil, metadata, status.Errorf(codes.InvalidArgument, "missing parameter %s", "lang")
 	}
 
-	if protoReq.Code == nil {
-		protoReq.Code = &SimpleMessage_Lang{}
-	} else if _, ok := protoReq.Code.(*SimpleMessage_Lang); !ok {
```
johanbrandhorst, line 294: Do we not need this `nil` check anymore? It seems to me like this could panic now.

examples/internal/proto/examplepb/a_bit_of_everything.pb.gw.go
```diff
@@ -491,6 +491,7 @@ func request_ABitOfEverythingService_CreateBody_0(ctx context.Context, marshaler
 	if berr != nil {
 		return nil, metadata, status.Errorf(codes.InvalidArgument, "%v", berr)
 	}
+
```
johanbrandhorst, line 494: While I think this is probably a bit nicer looking, I think we should try not to introduce a change like this that will affect all users upgrading their generator. It doesn't look trivial but it'd be nice if we can get it working!

johanbrandhorst (commented):
This looks much better, just a few questions for me.

examples/internal/proto/examplepb/echo_service.pb.gw.go
aesadde: Yep, should be fixed now.

examples/internal/proto/examplepb/a_bit_of_everything.pb.gw.go
aesadde: @johanbrandhorst I edited the template to check for `""` and avoid printing these unnecessary newlines.

Decided to do it directly inside the template as opposed to tweaking the output of the Prep functions.

johanbrandhorst (approved):
This looks much better 👍🏻. Thanks!

johanbrandhorst:
Thanks for your contribution!

aesadde:
Awesome, glad I could help!

PR #2800, opened 2022-07-19
Fix buf plugin's library version

AlmogBaku:
The buf `mod` currently returns a defunct version which cause go mod to complain.

See https://bufbuild.slack.com/archives/CRZ680FUH/p1658149356049699?thread_ts=1658096456.542309&cid=CRZ680FUH


johanbrandhorst:
Doh, great find!

PR #2813, opened 2022-07-24
Default allow_repeated_fields_in_body option to true and deprecate option

### LINKED ISSUE #2809: Support array-type response body by default
## 🚀 Feature

EDIT: I was able to resolve this by finding the `allow_repeated_fields_in_body=true` plugin option. However, I suggest changing that option default from `false` to `true`.

~~I would like the openapiv2 and grpc-gateway plugins to support mapping an HTTP response to a repeated field, so that we can have array-type response bodies.~~

~~For example, I want my response body to look like:~~

```json
[{"id":1},{"id":2}]
```

~~Instead of:~~

```json
{"pets":[{"id":1},{"id":2}]}
```

~~Example proto:~~

```proto
import "google/api/http.proto";

service PetService {
  rpc GetPets(GetPetsRequest) returns (GetPetsResponse) {
    option (google.api.http) = {
      get: "/pets"
      response_body: "pets"
    };
  }
}

message GetPetsRequest {}

message GetPetsResponse {
  repeated Pet pets = 1;
}

message Pet {
  uint64 id = 1;
}
```

~~Plugin output:~~

```
Failure: plugin openapiv2: repeated field not allowed in field path: pets in pets
Failure: plugin grpc-gateway: repeated field not allowed in field path: pets in pets
```

~~I read the previous discussion in #317; however there are situations where you need an array-type response body, like when you are implementing an existing REST API spec while migrating a codebase to gRPC.~~

~~[The documentation for the `google.api.http`](https://github.com/googleapis/googleapis/blob/de1ace47a265f06f6d7da31f10f1679d685812a3/google/api/http.proto#L310-L313) is clear that mapping a repeated field to `response_body` can be supported.~~

> If an API needs to use a JSON array for request or response body, it can map
> the request or response body to a repeated field. However, some gRPC
> Transcoding implementations may not support this feature.

~~Thank you 👐~~

armsnyder:

Closes #2809 

Historical context:

- Repeated fields disallowed (2015): https://github.com/grpc-ecosystem/grpc-gateway/commit/d3a16a26bfb883b499be122bff9e30e19efef858
- `allow_repeated_fields_in_body ` option added (2018): https://github.com/grpc-ecosystem/grpc-gateway/commit/ab0345bb328757bfef2f3d7d4e642e182eb985b9

Yes

The `allow_repeated_fields_in_body` option is now ignored and always behaves as if it is set to `true`. Users who configure the option will see a warning.

#### Other comments

👋

I went whole hog with this, and instead of changing the default from `false` to `true`, I deprecated the option. This is cleaning up after a successful feature flag experiment.

Note that on the `protoc-gen-openapiv2/main_test.go` test cases, I did not modify the input strings.


protoc-gen-grpc-gateway/main_test.go
```diff
@@ -6,24 +6,26 @@ import (
 	"github.com/grpc-ecosystem/grpc-gateway/v2/internal/descriptor"
 )
 
+// TODO(asnyder): These tests are covering a function parseFlags() which became unused after https://github.com/grpc-ecosystem/grpc-gateway/pull/1756. The function and tests may be deleted.
```
johanbrandhorst, line 9: Lets just have them out now if they're not adding any value!

johanbrandhorst (commented):
Looks like we have a test failure 😬

armsnyder:
Got the tests passing locally again. Give it another shot?

protoc-gen-grpc-gateway/main_test.go
armsnyder: And they're gone! 🔥

internal/descriptor/services.go
```diff
@@ -325,9 +325,6 @@ func (r *Registry) resolveFieldPath(msg *Message, path string, isPathParam bool)
 		if f == nil {
 			return nil, fmt.Errorf("no field %q found in %s", path, root.GetName())
 		}
-		if !(isPathParam || r.allowRepeatedFieldsInBody) && f.GetLabel() == descriptorpb.FieldDescriptorProto_LABEL_REPEATED {
-			return nil, fmt.Errorf("repeated field not allowed in field path: %s in %s", f.GetName(), path)
-		}
```
johanbrandhorst, line 330: Not sure this check is safe to remove, we still don't support repeated fields in paths, do we?

armsnyder: Hmm, there may have been a bug in the original code then. Here was my thinking for removing it:

```go
if !(isPathParam || r.allowRepeatedFieldsInBody) && f.GetLabel() == descriptorpb.FieldDescriptorProto_LABEL_REPEATED
```

simplifies to

```go
if !(isPathParam || true) && f.GetLabel() == descriptorpb.FieldDescriptorProto_LABEL_REPEATED
```

simplifies to

```go
if false && f.GetLabel() == descriptorpb.FieldDescriptorProto_LABEL_REPEATED
```

so it would never execute the body of the `if`, so I removed it.

I can change it to this if you like:

```go
if isPathParam && f.GetLabel() == descriptorpb.FieldDescriptorProto_LABEL_REPEATED
```

johanbrandhorst: Yeah I was trying to grok what the original code was doing too. I think your simplification is correct, can you try that and see if it breaks any tests 😂?

armsnyder: Heh, the tests passed with that change, so I wrote a new test based on `TestCauseErrorWithPathParam`.

johanbrandhorst (approved):
Very nice 😄

internal/descriptor/services.go
johanbrandhorst: Thank you!

armsnyder:
Crumbs, the generate test is failing now because it wants to be able to map a path param to a repeated field, as a comma-delimited value. I'll revert that last commit then.

johanbrandhorst:
TIL we support that

johanbrandhorst (approved):
LGTM

johanbrandhorst:
Thanks for your contribution!

armsnyder:
Thanks for all the timely feedback!

PR #2826, opened 2022-08-01
Fix identifiers generated from snake-cased enums not matching pb.go definitions

### LINKED ISSUE #2776: Snake case protobuf enums fail to compile.
## 🐛 Bug Report

When declaring snake cased enums in a proto file, grpc-gateway fails to render the correct go type name for it, writing the type name as-is instead of converting to camel case.

GRPC correctly uses property `GoIdent` of [`protogen.Enum`](https://pkg.go.dev/google.golang.org/protobuf/compiler/protogen#Enum) to produce the proper type name.

## To Reproduce

1. write a protobuf file with snake case enums
2. write a request message with a member `what` typed with said enum
3. add a service with an rpc that uses said request message
4. annotate rpc so that the `what` member is part of the REST URL (e.g.: `/api/{what}`)
5. generate grpc-gateway sources using the protobuf plugin
6. compile go code that uses the generated sources

A minimal repro is available as a docker image source in [repro.tar.gz](https://github.com/grpc-ecosystem/grpc-gateway/files/9011421/repro.tar.gz). To reproduce, run:
```
docker build -t bug-repro . && docker run -it --rm bug-repro
```

Contents of the attached archive:
```
-rw-r--r-- 1 mj mj  631 Jun 29 09:21 api.proto
-rw-r--r-- 1 mj mj 1703 Jun 29 09:18 Dockerfile
-rwxr-xr-x 1 mj mj  395 Jun 29 09:14 entrypoint.sh
-rw-r--r-- 1 mj mj   59 Jun 29 09:03 repro.go
```

## Expected behavior

The source file generated by the grpc-gateway plugin should compile successfully.

## Actual Behavior

Compilation fails:
```
# bug/report
../api/bug/report/api.pb.gw.go:109:29: undefined: snake_case_enum_value
../api/bug/report/api.pb.gw.go:114:18: undefined: snake_case_enum
../api/bug/report/api.pb.gw.go:138:29: undefined: snake_case_enum_value
../api/bug/report/api.pb.gw.go:143:18: undefined: snake_case_enum
```

## Your Environment

An isolated [Debian-based](https://hub.docker.com/r/bitnami/minideb/) container with go 1.18.

jbaxx:

Fixes #2776 

Yes.

First add snake-cased enums to `a_bit_of_everything.proto` to reproduce the issue, and to `path_enum.proto` to recreate the scenario when importing snake-cased enums from another package.
To fix, apply the CamelCasing function into the enum identifiers that reference those generated in the corresponding `.pb.go` file. 
As the enums are looped-through inside the template, the camel casing function needs to be registered into the template and be called there.
Given the template not only generates the enum identifiers but it matches it with the corresponding import package, the CamelCasing function must ensure to CamelCase the identifier name but leave the package name/path intact, that's why introduced the CamelIdentifier function wrapping the Camel function.


examples/internal/proto/examplepb/a_bit_of_everything.proto
```diff
@@ -781,3 +781,30 @@ service camelCaseServiceName {
 service AnotherServiceWithNoBindings {
   rpc NoBindings(google.protobuf.Empty) returns (google.protobuf.Empty) {}
 }
+
+service SnakeEnum {
```
johanbrandhorst, line 785: Nit: please suffix this with "Service".
```suggestion
service SnakeEnumService {
```

internal/casing/camel.go
```diff
@@ -50,6 +55,18 @@ func Camel(s string) string {
 	return string(t)
 }
 
+// CamelIdentifier returns the CamelCased identifier without affecting the package name/path if any.
+func CamelIdentifier(s string) string {
+	dot := "."
```
johanbrandhorst, line 60: Nit: could probably make this `const`.
```suggestion
	const dot = "."
```

johanbrandhorst (commented):
Hi, thanks for the PR! Only minor comments from me, looks great!

examples/internal/proto/examplepb/a_bit_of_everything.proto
```diff
@@ -781,3 +781,30 @@ service camelCaseServiceName {
...
+
```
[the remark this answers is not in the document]
jbaxx: Done

examples/internal/proto/examplepb/a_bit_of_everything.proto
jbaxx: Done

internal/casing/camel.go
jbaxx: Done

jbaxx:
Thanks for the comments! Done now.
Just one thing, it will not pass the proto_lint step from the workflow as it doesn't follow the suggested linting, which is the edge case this PR aims to solve.

johanbrandhorst:
> Thanks for the comments! Done now. Just one thing, it will not pass the proto_lint step from the workflow as it doesn't follow the suggested linting, which is the edge case this PR aims to solve.

I see, we'll need to add an ignore for that. I think in this case we probably want to take advantage of the ability to ignore a lint warning with a comment, as specified here: https://docs.buf.build/lint/configuration#allow_comment_ignores. Please add these comment ignores to the new enums.

johanbrandhorst:
Oh and you'll need to rebase/merge main into this branch

jbaxx:
Done rebasing and adding lint ignore through comments.

johanbrandhorst:
Did you try running `buf lint` locally? Looks like the linter is still failing.

jbaxx:
> Did you try running `buf lint` locally? Looks like the linter is still failing.

Yes, but my bad, `buf lint` is good, I didn't run `buf format` and spaces were off, fixed now.

johanbrandhorst:
Thanks for your contribution!

PR #2840, opened 2022-08-07
fix: unnecessary -e arg for echo command in dockerfile

### LINKED ISSUE #2839: Bug: dockerfile echo -e into /usr/local/bin/swagger-codegen
Referring to [this PR reply](https://github.com/grpc-ecosystem/grpc-gateway/pull/2829#issuecomment-1206699525), I found that there is a `-e` at the beginning of `/usr/local/bin/swagger-codegen`, which reports error while generating files.

The reason is, the `RUN` in the dockerfile would interpreted as `/bin/sh -c <command>`([Reference](https://docs.docker.com/engine/reference/builder/#run)):
> RUN \<command\> (shell form, the command is run in a shell, which by default is /bin/sh -c on Linux or cmd /S /C on Windows)

And `echo -e` would print `-e` into stdout directly when we are using /bin/sh:
```
makdon-System-Product-Name% /bin/sh -c "echo -e 'foo\nbar'" > foo
makdon-System-Product-Name% cat foo
-e foo
bar
makdon-System-Product-Name% 
```

We can just simply remove the `-e`, the `\n` could still be interpreted as a new line:
```
makdon-System-Product-Name% /bin/sh -c "echo 'foo\nbar'" > foo 
makdon-System-Product-Name% 
makdon-System-Product-Name% cat foo                           
foo
bar
makdon-System-Product-Name% 
```

At a nutshell, remove the `-e` in the dockerfile could avoid the errors occur.
I will make a PR for this.

MakDon:
fix #2839

### DIFF
--- a/.github/Dockerfile
+++ b/.github/Dockerfile
@@ -37,7 +37,7 @@ RUN apt-get update && \
 ENV SWAGGER_CODEGEN_VERSION=2.4.8
 RUN wget https://repo1.maven.org/maven2/io/swagger/swagger-codegen-cli/${SWAGGER_CODEGEN_VERSION}/swagger-codegen-cli-${SWAGGER_CODEGEN_VERSION}.jar \
     -O /usr/local/bin/swagger-codegen-cli.jar && \
-    echo -e '#!/bin/bash\njava -jar /usr/local/bin/swagger-codegen-cli.jar "$@"' > /usr/local/bin/swagger-codegen && \
+    echo '#!/bin/bash\njava -jar /usr/local/bin/swagger-codegen-cli.jar "$@"' > /usr/local/bin/swagger-codegen && \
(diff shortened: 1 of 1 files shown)


johanbrandhorst (approved):

johanbrandhorst:
I'm going to wait with merging this until I'm able to push a new version (I'm on low speed internet at the moment), but thanks a lot for this fix! We should consider doing another PR that just updates some of the versions, if you'd like it would be nice to experiment with that as a follow up.

MakDon:
I am glad to do that :D please let me know when the updates are done.

johanbrandhorst:
I have pushed `ghcr.io/grpc-ecosystem/grpc-gateway/build-env:1.19` from this branch. Merging.

PR #2858, opened 2022-08-21
fix: support service tags in OpenAPI config file (#2817)

### LINKED ISSUE #2817: Support tags in service annotations
It seems that the `service` option in OpenAPI configuration file is not working, or at least I am unable to get it to work. 

The `protoc` compiles and successfully generates the swagger but the `tags` is not getting updated with the description set by the `service` tag. Here is [OpenAPIServiceOption](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/internal/descriptor/openapiconfig/openapiconfig.proto#L27)

I also checked [unannotated_echo_service.swagger.yaml](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/examples/internal/proto/examplepb/unannotated_echo_service.swagger.yaml#L68) and could not find the corresponding service description in [unannotated_echo_service.swagger.json](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/examples/internal/proto/examplepb/unannotated_echo_service.swagger.json) (which im assuming is the generated swagger output of `unannotated_echo_service.swagger.yaml`)

However, in the [a_bit_of_everything.proto](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/examples/internal/proto/examplepb/a_bit_of_everything.proto#L501) where the annotations are in the proto file, it seems the tags for the service are properly generated in [a_bit_of_everything.swagger.json](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/examples/internal/proto/examplepb/a_bit_of_everything.swagger.json#L17)

y-takuya:
Fixes #2817 

YES

- Before: Service tags are extracted only from the proto file.
- After: Service tags are extracted from both the proto file and OpenAPI configuration file.
  - [`register.GetOpenAPIServiceOption`](https://github.com/grpc-ecosystem/grpc-gateway/blob/209f8e5908959d02d04558294bf06f0a3656d9e2/internal/descriptor/registry.go#L713): extract the service options defined in the OpenAPI configuration file
  - `extractTagOptionFromServiceDescriptor`: extract the service options defined in the proto file
  - `getServiceOpenAPIOption`: integrate above two functions to extract the service options defined in either the proto file or OpenAPI configuration file

#### Other comments
`examples/internal/proto/examplepb/unannotated_echo_service.swagger.json` and `examples/internal/clients/unannotatedecho/api/swagger.yaml` are generated automatically by fixing `template.go`.


protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -2488,6 +2504,21 @@ func getMessageOpenAPIOption(reg *descriptor.Registry, msg *descriptor.Message)
...
 }
 
+func getServiceOpenAPIOption(reg *descriptor.Registry, svc *descriptor.Service) (*openapi_options.Tag, error) {
+	opts, err := extractTagOptionFromServiceDescriptor(svc.ServiceDescriptorProto)
+	if err != nil {
+		return nil, err
+	}
+	if opts != nil {
+		return opts, nil
+	}
```
johanbrandhorst, line 2514: This gives the protobuf file annotations precedence over the openapi configuration file, is that what we want? I think users usually use the openapi configuration file when they can't edit the protobuf file for some reason, so maybe the precedence should be in the other direction, so that an openapi configuration file can overwrite what's set in the protobuf file?

examples/internal/proto/examplepb/unannotated_echo_service.swagger.json
```diff
@@ -17,7 +17,12 @@
   },
   "tags": [
     {
-      "name": "UnannotatedEchoService"
+      "name": "UnannotatedEchoService",
+      "description": "UnannotatedEchoService description -- which should not be used in place of the documentation comment!",
+      "externalDocs": {
+        "description": "Find out more about UnannotatedEchoService",
+        "url": "https://github.com/grpc-ecosystem/grpc-gateway"
+      }
```
johanbrandhorst, line 25: Interesting. I guess this is correct? 😂.

protoc-gen-openapiv2/internal/genopenapi/template.go
y-takuya: Ah Right. I'll fix it so that the options defined in the openapi configuration file take precedence.

examples/internal/proto/examplepb/unannotated_echo_service.swagger.json
y-takuya: Changes to this file have been automatically generated by modifying the template.go. I think the tags defined in the following opanapi configuration file of `UnannotatedEchoService`, which were previously unsupported, now simply output.
https://github.com/grpc-ecosystem/grpc-gateway/blob/c12472a456fecdd504ccbb8eb645776c3c05f073/examples/internal/proto/examplepb/unannotated_echo_service.swagger.yaml#L70-L74

This might be the same as the output of the following tags defined in the a_bit_of_everything.proto example.
https://github.com/grpc-ecosystem/grpc-gateway/blob/c12472a456fecdd504ccbb8eb645776c3c05f073/examples/internal/proto/examplepb/a_bit_of_everything.proto#L508-L514

Do you have concerns about tags not being the expected output?

johanbrandhorst: No, I think it's fine :), I think you're right.

protoc-gen-openapiv2/internal/genopenapi/template.go
y-takuya: I found that the options defined in the **proto** file always overwrite openapi configurations for the part of Method, Message, File and Field.

Example (for Method):
https://github.com/grpc-ecosystem/grpc-gateway/blob/c12472a456fecdd504ccbb8eb645776c3c05f073/protoc-gen-openapiv2/internal/genopenapi/template.go#L2465

I think we need to unify which configuration file has priority, but I feel that giving priority to openapi would have a not small impact on the existing output for Method, Message, File and Field.

How about creating a new issue for the fix to change the priority of the Proto and OpenAPI configuration files, and scope this fix to support the Service tag defined in the OpenAPI configuration file?

y-takuya: If it is better to change the priority of the Proto and OpenAPI configuration files only for `Service` in advance, I'll fix it as planned.

johanbrandhorst: Sounds good!

y-takuya: Thank you for your confirmation!

Just to be sure, do you want me to change the priority of configs only for Service in advance? Or do  we not change it in this PR, but handle it in conjunction with other changes?

johanbrandhorst: Handle the service here, and lets do the others in a separate PR.

y-takuya: Thanks. I've fixed the priority of config files only for Service.

johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution!

PR #2870, opened 2022-09-02
mux: calculate verb correctly for cases like DELETE /foo/bar:archive when user provided wrong method

jonathaningram:

In the test, the method the user meant was POST, which is supported.
DELETE is not supported, but I believe this should be NotImplemented, not NotFound.

#### Other comments

I think maybe the bug is because for this variable:

```
// Verb out here is to memoize for the fallback case below
var verb string
```

It's only ever set in the loop using `s.handlers[r.Method]` but the loop never enters because the request method is different to the registered route method. Hence `verb` is always an empty string in the later logic.

Of course, please challenge me on this being the desired behaviour or not.

### DIFF
--- a/runtime/mux.go
+++ b/runtime/mux.go
@@ -343,8 +343,7 @@ func (s *ServeMux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	// Verb out here is to memoize for the fallback case below
-	var verb string
+	lastComponent := components[len(components)-1]
 
 	for _, h := range s.handlers[r.Method] {
 		// If the pattern has a verb, explicitly look for a suffix in the last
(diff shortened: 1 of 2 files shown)


johanbrandhorst:
Hi, thanks for this test, this does seem like the wrong behavior, though I would suggest it should be `StatusMethodNotAllowed` (https://github.com/grpc-ecosystem/grpc-gateway/blob/660400cf774a51e6bf1969e7651adcf1ae1637bc/runtime/mux.go#L4210). Would you like to include a fix for the behavior?

jonathaningram:
Hey @johanbrandhorst, unfortunately that line of code you refer to isn't hit by this test case i.e. the error is not a `var mse MalformedSequenceError`. From the debugging frame below you can see that it's just a plain Go error `"not match to the path pattern"`. Also note the `verb` is an empty string in this frame.

image

I believe the `verb` needs to be calculated differently in this block https://github.com/grpc-ecosystem/grpc-gateway/blob/660400cf774a51e6bf1969e7651adcf1ae1637bc/runtime/mux.go#L349 which I've done in a follow up commit. There's a bit of code duplication in `idx` calculation but I believe it's fine. 

LMKWYT.

johanbrandhorst (approved):
LGTM! Thanks for the fix ❤️

PR #2872, opened 2022-09-05
protoc-gen-openapiv2: add support deprecated options

sirtakidancers:


sirtakidancers:
before -> after 
before
ng)

protoc-gen-openapiv2/internal/genopenapi/types.go
```diff
@@ -170,6 +170,7 @@ type schemaCore struct {
 	XNullable bool       `json:"x-nullable,omitempty" yaml:"x-nullable,omitempty"`
 	Example   RawExample `json:"example,omitempty" yaml:"example,omitempty"`
 
+	Deprecated bool `json:"deprecated,omitempty" yaml:"deprecated,omitempty"`
```
johanbrandhorst, line 173: I'm not sure this is a valid openapiv2 property of the items object. This is the definition of this object: https://swagger.io/specification/v2/#itemsObject. See how the `deprecated` field only exists on the OperationObject?

johanbrandhorst (commented):
I don't think this is a valid openapiv2 property, am I missing something?

protoc-gen-openapiv2/internal/genopenapi/types.go
johanbrandhorst: Thanks for your interest in creating a openapiv3 generator, this is something that is of great interest to many users. I will warn you that this has been attempted many times and abandoned just as many times, especially in the way that you suggested. I think any attempt at writing a openapiv3 generator would have to be a from-scratch effort and it would not be a small undertaking. Please feel free to try it regardless, in a separate PR please.

johanbrandhorst:
This is not supported in openapiv2

Issue #1937, opened 2021-02-02
[protoc-gen-openapiv2] [BUG] Incorrect handling of non-wildcard google.api.http.body when using field_behaviour annotation


gganley:
## 🐛 Bug Report

With the introduction of #1806 rpc definitions definitions that have a non-wildcard body property now fail validation. The generated code breaks because it produces a swagger document whose body parameter is a [reference](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#referenceObject) to the request definition but also has a `required` entry in the object.

Note that this file swagger file failed `npx @opernapitools/openapi-generator-cli generate` with error saying that `required` is unexpected.

### Proto File (./genproto/test/v1alpha/test.proto)

```proto
syntax = "proto3";

package example.test.v1alpha;

import "google/api/annotations.proto";
import "google/api/client.proto";
import "google/api/field_behavior.proto";
import "google/api/resource.proto";
import "google/protobuf/field_mask.proto";
import "protoc-gen-openapiv2/options/annotations.proto";

option go_package = "gitlab.com/example/example/genproto/test/v1alpha;test";
option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger) = {
  host: "api.example.com"
  schemes: HTTPS,
};

service Test {
  rpc UpdateFoo(UpdateFooRequest) returns (Foo) {
    option (google.api.http) = {
      patch: "/v1alpha/foo/{foo.id}",
      body: "foo"
    };
    option (google.api.method_signature) = "foo,update_mask";
  }
}

message UpdateFooRequest {
  Foo foo = 1 [(google.api.field_behavior) = REQUIRED];

  google.protobuf.FieldMask update_mask = 2;
}

// Foo test.
message Foo {
  option (google.api.resource) = {
    type: "api.example.com/Foo"
    pattern: "foo/{id}"
  };

  int64 id = 1;
}
```

### Run

```sh
 protoc -I. --openapiv2_out . --openapiv2_opt logtostderr=true ./genproto/test/v1alpha/test.proto
```

## Expected behavior

```json
{
  "swagger": "2.0",
  "info": {
    "title": "genproto/test/v1alpha/test.proto",
    "version": "version not set"
  },
  "tags": [
    {
      "name": "Test"
    }
  ],
  "host": "api.example.com",
  "schemes": [
    "https"
  ],
  "consumes": [
    "application/json"
  ],
  "produces": [
    "application/json"
[...]

PR #2904, opened 2022-09-26
fix: required properties of message type are required in OpenAPI

### LINKED ISSUE #2837: Required properties of message types get optional in OpenAPI
## 🐛 Bug Report

Message properties marked with `[(google.api.field_behavior) = REQUIRED]` get ignored when generating OpenAPI spec, if properties reference other messages (schemas).

## To Reproduce

Create a proto with the following definition:
```
message Vendor {
  uint32 id = 1 [(google.api.field_behavior) = REQUIRED];
}
message CreateVendorRequest {
  Vendor vendor = 1 [(google.api.field_behavior) = REQUIRED];
}
```

And generate anOpenApi spec. It will contain:

```json
    "CreateVendorRequest": {
      "type": "object",
      "properties": {
        "vendor": {
          "$ref": "#/definitions/Vendor"
        }
      }
    },
```

i.e. the `required: [ "vendor" ]` part is missed.

## Expected behavior

```json
    "CreateVendorRequest": {
      "type": "object",
      "properties": {
        "vendor": {
          "$ref": "#/definitions/Vendor"
        }
      },
      "required": [
        "vendor"
      ]
    },
```

## Actual Behavior

The required part is missed.

## Your Environment

Linux
2.11.1

gostajonasson:
Properties of message type having the annotation `[(google.api.field_behavior) = REQUIRED]` are not marked as required in the generated OpenAPI specification.
This PR fixes that.

A maintainer should sanity check this PR since it naively removes logic someone added on purpose linking to a "JSON Reference syntax".

Fixes #2837


keyz:
Would love to see this merged!

@gostajonasson -- I added a few tests (in my fork) to verify your change, and I *think* you have the right fix.

- https://github.com/keyz/grpc-gateway/commit/91060e6d6e21615dbe77c581d828e330a74b74e2: before applying the fix, `required` annotations are missing
- https://github.com/keyz/grpc-gateway/commit/deab66ec3f46144141824885744dc5af5d3cc043: after applying the fix, openapi files seems legit with no regressions

Feel free to copy anything from my fork if it helps to get your PR merged :)

johanbrandhorst:
Huh, this is a surprisingly small fix. I can't say I know why that check was there. @MakDon what do you think about the changes? You've contributed a fix in this space recently.

MakDon:
> Huh, this is a surprisingly small fix. I can't say I know why that check was there. @MakDon what do you think about the changes? You've contributed a fix in this space recently.

I will take a look at it ASAP :D

gostajonasson:
Thanks @keyz!
@johanbrandhorst @MakDon just tell me if you want me to add the tests from @keyz to this PR.

MakDon:
It seems that this solution would cause another issue, see #1937 AND #1944
I am trying to confirm if this PR would break the generation again.

IMO, it is better to add the tests from @keyz 
I am trying to confirm why the test `TestSchemaOfField` is not broken when running the CI

johanbrandhorst:
More tests shouldn't hurt, so please feel free to add them here. Thanks @MakDon for taking a look!

gostajonasson:
> It seems that this solution would cause another issue, see https://github.com/grpc-ecosystem/grpc-gateway/issues/1937 AND https://github.com/grpc-ecosystem/grpc-gateway/pull/1944

Yes you are correct @MakDon, this PR is naive and **I need help fixing it**.
This PR would cause message types marked with the `OUTPUT_ONLY` field behaviour as read only like this:
```
  Location pickup_location = 1 [(google.api.field_behavior) = OUTPUT_ONLY];
```
```
    properties:
      pickupLocation:
        $ref: '#/definitions/Location'
        readOnly: true
```
Which isn't recommended since other members than `$ref` should be ignored per [the JSON Reference spec](https://datatracker.ietf.org/doc/html/draft-pbryan-zyp-json-ref-03#section-3).

But I'm also seeing that the `description` field is added (somewhere else in the code) next to `$ref` members so that's an existing bug.

Personally I think that the best fix is **outside** of the [`schemaOfField` function](https://github.com/gostajonasson/grpc-gateway/blob/5870b530a6bfc1d7145d1704a21df258afb4b6fc/protoc-gen-openapiv2/internal/genopenapi/template.go#L648) and not output any members next to `$ref` (if `openapiSchemaObject.Ref` is set) in the final OpenAPI spec.
What I mean is that the `schemaOfField` function should not be responsible for filtering out fields just because it is of message type. That should a renderer do since it's OK to list the required fields outside of the property:
```
  Shipment:
    type: object
    properties:
      pickupLocation:
        $ref: '#/definitions/Location'
    description: A description.
    required:
      - pickupLocation
```

gostajonasson:
Pushed a fix for the bug that @MakDon mentioned my initial fix would cause.
Right now this PR fixes all bugs for my use case.
Will add the tests from @keyz later.

johanbrandhorst:
Hi @gostajonasson, thanks again for working on this, do you think you'll bring this over the line in the next week or so? I'm just planning out the next release timing. Thanks!

gostajonasson:
> Hi @gostajonasson, thanks again for working on this, do you think you'll bring this over the line in the next week or so? I'm just planning out the next release timing. Thanks!

Yes I hope so! Sorry for the delay.

keyz:
@gostajonasson you can:
1. start with copying what I did in `examples/internal/proto/examplepb/a_bit_of_everything.proto` (commit https://github.com/keyz/grpc-gateway/commit/91060e6d6e21615dbe77c581d828e330a74b74e2).
2. after that, follow [`CONTRIBUTING.md`](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/CONTRIBUTING.md) to regenerate the fixtures
3. you might need to tweak `examples/internal/clients/abe/BUILD.bazel`, but my commits (+ https://github.com/keyz/grpc-gateway/commit/deab66ec3f46144141824885744dc5af5d3cc043) should contain everything you need

gostajonasson:
@keyz Thanks for the instructions! I've followed them and pushed the changes. Happy for a review and a workflow run. :hugs:

examples/internal/proto/examplepb/a_bit_of_everything.proto
```diff
@@ -794,8 +794,16 @@ service ABitOfEverythingService {
...
+
+  rpc PostRequiredMessageType(RequiredMessageTypeRequest) returns (google.protobuf.Empty) {
+    option (google.api.http) = {
+      post: "/v1/example/requiredmessagetype"
+      body: "*"
+    };
+  }
 }
 
+
```
johanbrandhorst, line 806: Nit: remove this

examples/internal/proto/examplepb/a_bit_of_everything.proto
```diff
@@ -843,3 +851,18 @@ message SnakeEnumRequest {
...
+  Foo foo = 2 [(google.api.field_behavior) = REQUIRED];
+}
+
+message Foo {
+  Bar bar = 1 [(google.api.field_behavior) = REQUIRED];
+}
+
+message Bar {
+  string id = 1 [(google.api.field_behavior) = REQUIRED];
+}
```
johanbrandhorst: Nit: Please add a newline at the end of the file

examples/internal/proto/examplepb/a_bit_of_everything.swagger.json
```diff
@@ -2335,9 +2335,7 @@
                   "title": "Repeated numeric enum title"
                 },
                 "enumValueAnnotation": {
-                  "$ref": "#/definitions/examplepbNumericEnum",
-                  "description": "Numeric enum description.",
-                  "title": "Numeric enum title"
```
johanbrandhorst, line 2340: I might have asked this already, but do we understand why these were removed?

examples/internal/proto/examplepb/a_bit_of_everything.proto
gostajonasson: Done

examples/internal/proto/examplepb/a_bit_of_everything.swagger.json
gostajonasson: It's explained [in the code](https://github.com/grpc-ecosystem/grpc-gateway/pull/2904/files#diff-c904837bc1a47dae3fba9863430f2ab46fa8a1bd0c9e25475308ab426a0d4efeR520-R521) (even though I personally would like to have descriptions next to $ref:s):
```
// Per the JSON Reference syntax: Any members other than "$ref" in a JSON Reference object SHALL be ignored.
// https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03#section-3
```
More info here:
https://swagger.io/docs/specification/using-ref/

johanbrandhorst: Ah, so these appearing here was already a bug?

gostajonasson: Yes, or they are ignored so it's unnecessary to render them there I guess.

johanbrandhorst:
Thanks for your contribution!

johanbrandhorst:
https://github.com/grpc-ecosystem/grpc-gateway/releases/tag/v2.12.0 has been released

PR #2952, opened 2022-10-17
Adjust the return value "result"

Super-zhw:
fix: "result" is a redundant return value, which will lead to inconsistency with the grpc request result.


johanbrandhorst:
> fix: "result" is a redundant return value, which will lead to inconsistency with the grpc request result.

Hi, thanks for your PR. We cannot merge this change. The `result` envelope exists for a consistent message definition when using streaming responses that may error mid-stream. Not to mention that this would be a breaking change. Please open an issue before spending time like this working on something that will never be merged.

PR #2928, opened 2022-10-11
feat: add option to disable rendering of service tags

kkolur:
#### References to other Issues or PRs
[Issue #2917](https://github.com/grpc-ecosystem/grpc-gateway/issues/2917#issuecomment-1272779703)

#### Have you read the [Contributing Guidelines](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/CONTRIBUTING.md)?
YES

#### Brief description of what is fixed or changed
Add option to disable rendering of service tags for those who wish to not expose the name of their backend grpc services. This allows for greater customizability when generating tags and defining custom tags for the APIs with different grouping strategy.

Issue #2917, opened 2022-10-01
Add option to disable rendering service level tags


kkolur:
## 🚀 Feature

[here](https://github.com/grpc-ecosystem/grpc-gateway/blob/3849b8a0cf8a66fe2799ff9dc1857f0097d13411/protoc-gen-openapiv2/options/openapiv2.proto#L98) it looks like tags at the root level are still `TODO`. Is this planned to be prioritized?Right now, in the output, the tags are automatically generated with the service(s) names but I would prefer to override that behavior by defining my own set of tags at the root level. OR, we can make the rendering of service tags optional and defaulted to true through command line flags?

Also, according to OpenAPI docs, [extensions are supported for tags as well](https://swagger.io/docs/specification/2-0/swagger-extensions/). Could we add this as well?

How difficult would this be overall?

johanbrandhorst:
There are a few settings already for how the tags are generated I believe: https://github.com/grpc-ecosystem/grpc-gateway/blob/b2fed3ce9c921d47c82c6b888ebeae0469f6cb5d/protoc-gen-openapiv2/main.go#L31. I'm not sure exactly what you're asking for. Do you want the ability add tags? How do you want to map those tags to objects? What do you want to do with extensions in tags? I think the use case has to be compelling enough that I could see it being useful to many of our users. Notably, you are the first to ask for this, which is not a great indicator of the need.

kkolur:
@johanbrandhorst I want the option to not automatically render service tags. RIght now the service tags are automatically generated [here](https://github.com/grpc-ecosystem/grpc-gateway/blob/e78f570b573df28b85203f1242c949ade8e5bfc9/protoc-gen-openapiv2/internal/genopenapi/template.go#L1592). If we could have some sort of command line flag that could let us disable that is what I was asking for. 

Using ReDoc, after we added tags to each endpoint, the service level tags were still displayed in the left hand side nav bar which we dont want to display for the users of the docs. We could write a post process script that removes the root level tags object, but I figured it wouldn't hurt if we added a cli flag thats defaulted to true but have the option to set it to false

johanbrandhorst:
I see, thanks for clarifying. I could see a flag that disables tags altogether, but I'm not convinced the need is good enough yet. Maybe we can leave this issue open and see if there are more users who want this functionality before we decide to add it?

davix:
We have the same requirement.
[...]

PR #2960, opened 2022-10-18
fix: remove default service tag generation for methods

kkolur:
[Issue #2917](https://github.com/grpc-ecosystem/grpc-gateway/issues/2917)

As part of [Issue #2917](https://github.com/grpc-ecosystem/grpc-gateway/issues/2917), generation of service tags should be disabled if the flag is set to true as part of the corresponding [PR](https://github.com/grpc-ecosystem/grpc-gateway/pull/2928). However, although the tags are not generated at the service level, we missed a spot where service tags were still being generated on the `operation`/`method` level. 

For example, if a `method` did not explicitly have `tags` defined in its `options`, it would still generate the service name as a tag and apply it to the `method`. This behavior should not happen if the flag `disable_service_tags` is set to `true`.

This PR is to address that

#### Other comments

### DIFF
--- a/protoc-gen-openapiv2/internal/genopenapi/template.go
+++ b/protoc-gen-openapiv2/internal/genopenapi/template.go
@@ -1380,13 +1380,7 @@ func renderServices(services []*descriptor.Service, paths openapiPathsObject, re
 					responseSchema.Ref = ""
 				}
 
-				tag := svc.GetName()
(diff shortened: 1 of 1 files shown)


johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution :)

PR #2965, opened 2022-10-19
feat(gen-openapiv2): support trailing comments

### LINKED ISSUE #2929: Parse trailing comments in protoc-gen-openapiv2
https://github.com/grpc-ecosystem/grpc-gateway/blob/0443675f6acea5bddca361673c6f0bc241dcb644/protoc-gen-openapiv2/internal/genopenapi/template.go#L2179-L2191

The current `protoc-gen-openapiv2` only parse leading comments, no trailing.
And I think trailing is better for short comments, like this:

```proto
message SearchRequest {
  string query = 1;              // Query string
  int32 page_number = 2;         // Page number
  int32 result_per_page = 3;     // Per page result count
}
```

instead of:

```proto
message SearchRequest {
  // Query string
  string query = 1;
  // Page number
  int32 page_number = 2;
  // Per page result count
  int32 result_per_page = 3;
}
```

So how about add this feature?

ionling:

Fixes #2929

Support parsing trailing comments.


johanbrandhorst (changes requested):
This looks like a good start, but could you try adding some trailing comments to one of our example files so that we can see what this looks like in practice? Here's the file we usually use: https://github.com/grpc-ecosystem/grpc-gateway/blob/master/examples/internal/proto/examplepb/a_bit_of_everything.proto.

ionling:
After I synced the fork, the PR is automatically closed.
I added some trailing comments to examplepb just now.
So let's check it again.

examples/internal/proto/examplepb/a_bit_of_everything.proto
```diff
@@ -406,6 +406,15 @@ message ABitOfEverything {
...
   ];
   string required_field_schema_json_name = 46 [json_name = "required_field_schema_json_name_custom"];
+
+  string trailing_only = 47;     // Trailing only
+  string trailing_only_dot = 48; // Trailing only dot.
+  // Leading both
+  string trailing_both = 49;     // Trailing both.
+  // Leading multiline
+  //
+  // This is a example of multi-line comments.
```
johanbrandhorst, line 416: Nit:
```suggestion
  // This is an example of a multi-line comment.
```

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -2228,6 +2228,14 @@ func protoComments(reg *descriptor.Registry, file *descriptor.File, outers []str
 			// - join by \n
 			comments = strings.Replace(comments, "\n ", "\n", -1)
 		}
+		if loc.TrailingComments != nil {
+			trailing := strings.TrimSpace(*loc.TrailingComments)
+			if comments == "" {
+				comments = trailing
+			} else {
+				comments += "\n\n" + trailing
```
johanbrandhorst, line 2236: Is there a good reason to include double newlines between these? I kinda feel like a single newline might be enough?

johanbrandhorst (changes requested):
Thanks for the examples

protoc-gen-openapiv2/internal/genopenapi/template.go
ionling: There are two points:
1. To distinguish the title and the description, when there is a leading comment without dot and a trailing comment
2. To put the trailing comment in separate paragraph. This is work for [Redoc](https://github.com/Redocly/redoc) (I think it use markdown style rendering), but not [Swagger UI](https://swagger.io/tools/swagger-ui/) (it just outputs string in spec yaml)

johanbrandhorst:
Reopening to run CI

johanbrandhorst (approved):
LGTM

johanbrandhorst:
Looks like you need to regenerate the files again

johanbrandhorst (approved):
LGTM

johanbrandhorst:
Thanks for your contribution!

PR #2987, opened 2022-11-01
release: Update release.yml with option to workaround SLSA generator failure

asraa:
### Description

This updates the SLSA generator workflow with a workaround due to an issue. https://github.com/slsa-framework/slsa-github-generator/issues/1163

The rough context is that Sigstore made a final breaking change related to an online service that distributes its trust root material (a TUF repository) which was backwards incompatible with their older libraries. Thus, our older builders failed, and we are working on updates. 

Adding `compile-generator: true` means that the generator code is compiled from source rather than download and verified with Sigstore. The Sigstore verification was broken due to the above problem in the old builders verifier versions.

### Other

We are tracking stability improvements: https://github.com/slsa-framework/slsa-github-generator/milestone/9

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,6 +42,7 @@ jobs:
       contents: write # To add assets to a release.
     uses: slsa-framework/slsa-github-generator/.github/workflows/generator_generic_slsa3.yml@v1.2.1
     with:
+      compile-generator: true # Workaround for https://github.com/slsa-framework/slsa-github-generator/issues/1163
       base64-subjects: "${{ needs.goreleaser.outputs.hashes }}"
       upload-assets: true # upload to a new release
 


johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution :). Lets hope it works next time 😂.

asraa:
> Thanks for your contribution :). Lets hope it works next time joy.

Yes: I also wonder, is it possible to add a workflow_dispatch dry-run for the workflow? So no tags pushed, but a build, provenance generation and no upload.

I can look into it if you want!

johanbrandhorst:
Sure, that'd be great, please do!

johanbrandhorst:
If we could backfill slsa signatures for the last few releases that'd be awesome. Could we make the release a parameter?

asraa:
> If we could backfill slsa signatures for the last few releases that'd be awesome. Could we make the release a parameter?

Right! We had a few requests on that and are thinking of the best ways to recommend this, which I'm trying to write up here. https://github.com/slsa-framework/slsa-github-generator/issues/1190 
I can try it out for you here with a workflow dispatch with a release parameter. If the artifacts are reproducible then it's all good. Otherwise, there are more complex solutions like grabbing the artifact from a previous run. I'll see what's easiest.

laurentsimon:
>  Otherwise, there are more complex solutions like grabbing the artifact from a previous run

wdum?

asraa:
> wdum?

You could potentially download artifacts from previous workflow runs from the GitHub API, but I'm not sure if that can be tampered with. (The workflow run, not the release)

PR #2989, opened 2022-11-02
release: add a workflow_dispatch trigger for testing

asraa:
Signed-off-by: Asra Ali <asraa@google.com>'

Follow-up to https://github.com/grpc-ecosystem/grpc-gateway/pull/2987 discussed with @johanbrandhorst 

I decided against inputing the release tag in the workflow_dispatch since it's not reproducible anyway. Instead, this will just run on main as a dry-run before pushing any release tags.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,6 +2,10 @@ on:
   push:
     tags:
       - v2.[0-9]+.[0-9]+
+  # For testing the workflow before pushing a tag
+  # This will run goreleaser with --snapshot and test the
+  # SLSA generator.
+  workflow_dispatch:
 permissions:
   contents: read
 name: release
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -19,10 +23,19 @@ jobs:
...
           check-latest: true
+      - name: Generate goreleaser args
+        id: args
+        run: |
+          set -euo pipefail
+          args='release --rm-dist'
+          if [[ "$GITHUB_REF" != refs/tags/* ]]; then
+            args+=' --snapshot'
+          fi
+          echo "args=$args" >> $GITHUB_OUTPUT
```
johanbrandhorst, line 34: Thank you for using the new format. Would you mind creating a separate PR to update all our `set-output` instances the same way? It would be greatly appreciated ❤️.

johanbrandhorst (approved):

johanbrandhorst:
https://github.com/grpc-ecosystem/grpc-gateway/actions/runs/3382044252 👀

johanbrandhorst:
Looks like it worked 🎉

PR #2996, opened 2022-11-06
protoc gen oas v2 cleanup

sashamelentyev:


protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -1227,18 +1225,18 @@ func renderServices(services []*descriptor.Service, paths openapiPathsObject, re
 							return err
 						}
 						if schema.Title != "" {
-							desc = mergeDescription(schema)
+							desc.WriteString(mergeDescription(schema))
```
johanbrandhorst, line 1228: Isn't this actually incorrect? The old code overwrote `desc` but this appends to it.

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -2265,7 +2248,7 @@ var packageProtoPath = protoPathIndex(reflect.TypeOf((*descriptorpb.FileDescript
 var serviceProtoPath = protoPathIndex(reflect.TypeOf((*descriptorpb.FileDescriptorProto)(nil)), "Service")
 var methodProtoPath = protoPathIndex(reflect.TypeOf((*descriptorpb.ServiceDescriptorProto)(nil)), "Method")
 
-func isProtoPathMatches(paths []int32, outerPaths []int32, typeName string, typeIndex int32, fieldPaths []int32) bool {
+func isProtoPathMatches(paths, outerPaths []int32, typeName string, typeIndex int32, fieldPaths []int32) bool {
```
johanbrandhorst, line 2251: I think I'd rather we keep this as it is and make the parameters explicit.

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -2356,7 +2339,7 @@ func protoPathIndex(descriptorType reflect.Type, what string) int32 {
 	}
 	path, err := strconv.Atoi(strings.Split(pbtag, ",")[1])
 	if err != nil {
-		panic(fmt.Errorf("protobuf descriptor id for %s cannot be converted to a number: %s", what, err.Error()))
+		panic(fmt.Errorf("protobuf descriptor id for %s cannot be converted to a number: %v", what, err))
```
johanbrandhorst, line 2342: I also prefer the old style here. I avoid using `%v` so that we can use the warnings `go vet` produces when the wrong type is used with the wrong verb.

johanbrandhorst (changes requested):
Thanks for this, most of these look great!

protoc-gen-openapiv2/internal/genopenapi/template.go
sashamelentyev: %v for error is ok. And %s for error is ok too. But if error is nil verb %s text this %!s(<nil>), verb %s text this <nil>. But error is handling before

protoc-gen-openapiv2/internal/genopenapi/template.go
sashamelentyev: In this case desc with WriteString without allocate. If we use string - we have allocate in case like this s += "add text"

johanbrandhorst: I'm not saying I disagree with the use of WriteString, I'm just wondering if this is the equivalent logic, doesn't `WriteString` append, where this logic before would replace?

protoc-gen-openapiv2/internal/genopenapi/template.go
johanbrandhorst: I'm not sure I understand what you're saying, but yeah we don't need to worry about the error being `nil` here so we should be able to keep using `%s`.

protoc-gen-openapiv2/internal/genopenapi/template.go
sashamelentyev: Reverted this to source

protoc-gen-openapiv2/internal/genopenapi/template.go
sashamelentyev: Reverted this to source

protoc-gen-openapiv2/internal/genopenapi/template.go
sashamelentyev: Reverted this to source

johanbrandhorst:
Thanks for your contribution!

PR #3006, opened 2022-11-09
Support disabling default response rendering

### LINKED ISSUE #2979: Option to turn off default 200 response generation in protoc-gen-openapiv2
## 🚀 Feature

I was wondering if it is feasible to have a similar option to disableDefaultErrors to turn off the default generation of the 200 responses? Has this been considered and dismissed?

Some examples:

- I have a post request where I only want to return 201 Created (or error codes)
- I have a delete request where I only want to return 204 No Content (or error codes)
- I have a head request where I only want to return 204 No Content

This would marry nicely with the ability to inject a httpResponseModifier in the mux

krak3n:
Fixes #2979

# Changes

Addes a new `disable_default_responses` flag (default `false`) which disables the rendering of the default responses. This allows users to define different API responses for cases for where `200` is not the default response code from an existing API, for example it maybe `201`.

The side effect of this is all responses would need to be defined when this option is set `true`.

### DIFF
--- a/docs/docs/mapping/customizing_openapi_output.md
+++ b/docs/docs/mapping/customizing_openapi_output.md
@@ -631,14 +631,105 @@ info:
   title: helloproto/v1/example.proto
   version: version not set
 consumes:
-- application/json
+  - application/json
 produces:
-- application/json
+  - application/json
 paths:
   /api/hello:
     get:
       operationId: EchoService_Hello
-...
+```
+
+### Disable default responses
+
+By default a 200 OK response is rendered for each service operation. But it is possible to disable this and explicitly define your service's responses, using the `disable_default_responses` option. Allowed values are: `true`, `false`.
+
+For example, if you are using `buf`:
+
+```yaml
+version: v1
+plugins:
+  - name: openapiv2
+    out: .
+    opt:
+      - disable_default_responses=true
+```
+
+or with `protoc`
+
+```sh
+protoc --openapiv2_out=. --openapiv2_opt=disable_default_responses=true ./path/to/file.proto
+```
+
+Input example:
+
+```protobuf
+syntax = "proto3";
+
+package helloproto.v1;
+
+import "google/api/annotations.proto";
+import "protoc-gen-openapiv2/options/annotations.proto";
+
+option go_package = "helloproto/v1;helloproto";
+
+service EchoService {
+  rpc Hello(HelloReq) returns (HelloResp) {
+    option (google.api.http) = {get: "/api/hello"};
+    option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_operation) = {
+      responses: {
+        key: "201",
+        value: {
+          description: "Created";
+          schema: {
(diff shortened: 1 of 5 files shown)


protoc-gen-openapiv2/internal/genopenapi/template_test.go
```diff
@@ -3316,6 +3316,17 @@ func TestApplyTemplateProtobufAny(t *testing.T) {
...
+		{
+			// we have a protobufAny in a message but with automatic rendering of responses disabled
+			name: "protobufAny_referenced_in_message_with_default_responses_disabled",
+			args: args{
+				msgContainsAny: true,
+				regConfig: func(reg *descriptor.Registry) {
+					reg.SetDisableDefaultResponses(true)
+				},
+			},
+			wantNumDefinitions: 4,
```
krak3n, line 3328: Not exactly sure what this tests but it's here 😄

johanbrandhorst (changes requested):
LGTM! One thing, do you think you could add a small section to our docs about this new capability too? They live in https://github.com/grpc-ecosystem/grpc-gateway/blob/master/docs/docs/mapping/customizing_openapi_output.md. Thanks!

krak3n:
@johanbrandhorst add example output to the documentation 👍

docs/docs/mapping/customizing_openapi_output.md
```diff
@@ -631,14 +631,105 @@ info:
...
 paths:
   /api/hello:
     get:
       operationId: EchoService_Hello
-...
+```
+
+### Disable default responses
+
+By default a 200 OK response is rendered for each service operation. But it is possible to disable this and explicitly define your service's responses, using the `disable_default_responses` option. Allowed values are: `true`, `false`.
```
johanbrandhorst, line 645: Could you just add a note somewhere to say that this does _not_ change the behavior of the gateway itself, and must be coupled with a custom ForwardResponseWriter that adjusts the response to be an accurate reflection of the API behavior? Maybe link to the docs section that talks about changing the response code.

johanbrandhorst (commented):
Thanks for the docs, just a small idea.

docs/docs/mapping/customizing_openapi_output.md
krak3n: Added a note @johanbrandhorst

johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution!

zshainsky:
Hi all what would it take to add this to the bazel `protoc_gen_openapiv2` command? Would that be just a simple update to the `defs.bzl` file?

johanbrandhorst:
Yep, that should be it. We should've added it when we added this flag, oops 😬

zshainsky:
Thanks @johanbrandhorst want me to open an issue for this?

johanbrandhorst:
Seems like @rajukrishnamurthy beat you to it in #4123.

PR #3010, opened 2022-11-11
Support request header parameters

### LINKED ISSUE #2932: Add header annotations to OpenAPIv2 generator
I am generating openapi v2 documentation. using `buf generate`

```
service PageService {
  option (grpc.gateway.protoc_gen_swagger.options.openapiv2_tag) = {
    info: {
      title: "Create and view pages to group notes";
    };
  };

  rpc CreatePage (CreatePageRequest) returns (ApiResponse) {
    option(google.api.http) = {
      post: "/api/pages/{id}",
      body: "*"
    };
  }

  rpc GetPageDetails(GetPageRequest) returns (ApiResponse) {
    option(google.api.http) = {
      get: "/api/pages/{id}"
    };
  }
}

message CreatePageRequest {
  string id = 1;
  string description = 2;

  google.protobuf.Timestamp dated_at = 3;
}

message GetPageRequest {
  string id = 1;
  bool show_items = 2;
}

/* more code */
```
this is a part of how my proto file looks like. I was wondering if there is any way to use annotations to generate the swagger.json such that it shows up in headers like.

```
api_key:
  type: apiKey
  name: api_key
  in: header
```

krak3n:
Fixes #2932

# Feature 🚀 

Adds support for optional HTTP request header parameters to be defined on method operations, for example:

```protobuf
service FooService {
  rpc CreateFoo(CreateFooRequest) returns (CreateFooResponse) {
    option (google.api.http) = {
      post: "/foo"
      body: "*"
    };

      summary: "Create a new foo.";
      parameters: {
        headers: {
          name: "Authorization";
          description: "Basic Auth";
          required: true;
        };
      };
    };
  }
}
```

This should generate the following `yaml`:

```yaml
paths:
  /v2/foo:
    post:
      summary: Create a new foo.
      operationId: CreateFoo
      responses:
        "200":
          description: OK
          schema:
            $ref: '#/definitions/CreateFooResponse'
      parameters:
        - name: foo
          in: body
          required: true
          schema:
            $ref: '#/definitions/CreateFooRequest'
        - name: Authorization
          in: header
          required: false
```

# Changes

* New protocol buffer definitions for HTTP Header Parameters
* New loop in applyTemplate which loops over header parameters if defined and > 0 which builds and appends the header parameters. to the existing generated parameters.
* Test cases for the above


protoc-gen-openapiv2/options/openapiv2.proto
```diff
@@ -175,6 +175,41 @@ message Operation {
...
+  Parameters parameters = 14;
+}
+
+// `Parameters` is a representation of OpenAPI v2 specification's parameters object.
+message Parameters {
+  // `Headers` is one or more HTTP header parameter.
+  repeated HeaderParameter headers = 1;
+}
+
+// `HeaderParameter` a HTTP header parameter.
```
johanbrandhorst, line 188:
```suggestion
// `HeaderParameter` a HTTP header parameter.
// See: https://swagger.io/specification/v2/#parameter-object
```

protoc-gen-openapiv2/options/openapiv2.proto
```diff
@@ -175,6 +175,41 @@ message Operation {
   // extra functionality that is not covered by the standard OpenAPI Specification.
   // See: https://swagger.io/docs/specification/2-0/swagger-extensions/
   map<string, google.protobuf.Value> extensions = 13;
+  // Custom parameters such as HTTP request headers.
+  Parameters parameters = 14;
+}
+
+// `Parameters` is a representation of OpenAPI v2 specification's parameters object.
```
johanbrandhorst, line 182:
```suggestion
// `Parameters` is a representation of OpenAPI v2 specification's parameters object.
// See: https://swagger.io/specification/v2/#parameter-object
```

protoc-gen-openapiv2/options/openapiv2.proto
```diff
@@ -175,6 +175,41 @@ message Operation {
...
+message Parameters {
+  // `Headers` is one or more HTTP header parameter.
+  repeated HeaderParameter headers = 1;
```
johanbrandhorst, line 185: Note that I believe this is the first instance of these definitions breaking from the general OpenAPI 2 definition structure. I think it's okay in this case since we do want to limit what users can put in here. For example, I don't think we ever want to support users specifying custom  non-header parameters beyond those inferred from the Protobuf schema.

protoc-gen-openapiv2/options/openapiv2.proto
```diff
@@ -175,6 +175,41 @@ message Operation {
...
+  string description = 2;
+  // `Type` is the type of the object. The value MUST be one of "string", "number", "integer", or "boolean". The "array" type is not supported.
+  Type type = 3;
+  // `Format` The extending format for the previously mentioned type.
+  string format = 4;
+  // `Required` indicates if the header is optional
+  bool required = 5;
+  // field 6 is reserved for 'items', but in OpenAPI-specific way.
+  reserved 6;
+  // field 7 is reserved `Collection Format` Determines the format of the array if type array is used.
```
johanbrandhorst, line 211:
```suggestion
  // field 7 is reserved `Collection Format`. Determines the format of the array if type array is used.
```

protoc-gen-openapiv2/internal/genopenapi/template_test.go
```diff
@@ -7965,6 +7965,213 @@ func TestRenderServicesWithColonInSegment(t *testing.T) {
...
+							MethodDescriptorProto: meth,
+							RequestType:           msg,
+							ResponseType:          msg,
+							Bindings: []*descriptor.Binding{
+								{
+									HTTPMethod: "GET",
+									PathTmpl: httprule.Template{
+										Version:  1,
+										OpCodes:  []int{0, 0},
+										Template: "/v1/echo", // TODO(achew22): Figure out what this should really be
```
johanbrandhorst, line 8020: 😂. I think we can remove this comment, at least in this test.

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -1515,6 +1515,34 @@ func renderServices(services []*descriptor.Service, paths openapiPathsObject, re
...
+								param.Type = "string"
+							case openapi_options.HeaderParameter_NUMBER:
+								param.Type = "number"
+							case openapi_options.HeaderParameter_INTEGER:
+								param.Type = "integer"
+							case openapi_options.HeaderParameter_BOOLEAN:
+								param.Type = "boolean"
+							default:
+								// TODO(krak3n): Should we error or assume string?
+								param.Type = "string"
```
johanbrandhorst, line 1539: This should be an error per https://swagger.io/specification/v2/#parameterType which says this is required. I think we should make the user pick a type.

johanbrandhorst (changes requested):
Thanks a lot for looking at this! Some minor comments, but in addition to those please add a section to the docs about this ✨ new feature ✨. This sort of thing has been requested a lot, thanks for helping bring it over the line. Please ensure the docs make it clear that adding headers to your schema via these annotations does _nothing_ to change the behavior of the gateway and that these annotations should only be coupled with custom header parsing behavior in the application. Also that adding header parameters can chance the forward and backward compatibility guarantees of the schema. Thanks!

protoc-gen-openapiv2/options/openapiv2.proto
krak3n: I'll add a note underneath.

protoc-gen-openapiv2/internal/genopenapi/template.go
krak3n: Added an error, I wanted to test this but looks like we panic in `applyTemplate` if `renderServices` fails which makes it quite tricky to test the returned error since I'd need to defer & recover etc. Is it ok to not have a test case for this?

protoc-gen-openapiv2/options/openapiv2.proto
krak3n: I've added a little note about this 🙌

protoc-gen-openapiv2/internal/genopenapi/template.go
johanbrandhorst: Yeah this is fine 😁

johanbrandhorst (changes requested):
The code changes look fine but could you please add something to the docs as well? Also, it looks like you need to regenerate the files again 😁.

krak3n:
@johanbrandhorst added documentation & regenerated the go code. I noticed when I run `buf generate` a load of swagger json also gets generated, does that need to be committed too?

johanbrandhorst:
> @johanbrandhorst added documentation & regenerated the go code. I noticed when I run `buf generate` a load of swagger json also gets generated, does that need to be committed too?

Just running `buf generate` will not generate the files the way we want, you'll need the command in the `CONTRIBUTING.md` file, which it looks like you're aware of. If there were more files needed, our `generate` job should detect it. Looks good from this side!

johanbrandhorst (approved):
LGTM

johanbrandhorst:
I think this is one of the most user requested features, so thank you for this great contribution! I'll consider making a new release shortly.

amitavaghosh1:
Thanks @krak3n

imrenagi:
hi @amitavaghosh1 @johanbrandhorst if I have a HTTP header that is used across all API (not for security), is there any way to use `$ref` on the endpoint parameter? I cant find how to define the parameters definition yet. Please let me know if you have any pointers. thanks!

something like this in swagger.
```
parameters:
  shardId:  
    in: header
    name: x-something-something
    required: false
    type: string

paths:
  /users:
    get:
      summary: Gets a list of users.
      parameters:
        - $ref: '#/parameters/shardId'
```

PR #3023, opened 2022-11-21
Use `debug.ReadBuildInfo` for retrieving plugin version

### LINKED ISSUE #3022: Use `debug.ReadBuildInfo()` for retrieving a version
## 🚀 Feature

It's a common case when a developers just installs the utility from sources, without retrieving the binary that includes build info, which raises an issue of tracking the utility's version. [Tool dependency](https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module) has no help in here as mostly developers are installing it via running the command above:
```bash
$ go install \
    github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway \
    github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2 \
    google.golang.org/protobuf/cmd/protoc-gen-go \
    google.golang.org/grpc/cmd/protoc-gen-go-grpc
```

which just takes the latest versions of all tools without embedding any information about the utility's build.

I think it's a good idea to retrieve utility's version via `debug.ReadBuildInfo`, which was added in Go 1.18.

Semior001:
closes #3022 

Yes

Retrieve version of grpc-gateway and openapiv2 plugins via `debug.ReadBuildInfo`, upgraded go version to 1.18 (probably last part is not necessary)

### DIFF
--- a/protoc-gen-grpc-gateway/main.go
+++ b/protoc-gen-grpc-gateway/main.go
@@ -49,7 +49,7 @@ func main() {
 	defer glog.Flush()
 
 	if *versionFlag {
-		fmt.Printf("Version %v, commit %v, built at %v\n", version, commit, date)
+		fmt.Printf("Version %v, commit %v, built at %v\n", readVersion(), commit, date)
 		os.Exit(0)
 	}
 
--- a/protoc-gen-grpc-gateway/main.go
+++ b/protoc-gen-grpc-gateway/main.go
@@ -13,7 +13,6 @@ import (
 	"flag"
 	"fmt"
 	"os"
-	"runtime/debug"
 	"strings"
 
 	"github.com/golang/glog"
@@ -45,20 +44,12 @@ var (
 	date    = "unknown"
 )
 
-func buildVersion() string {
-	v, ok := debug.ReadBuildInfo()
-	if !ok {
-		return version
-	}
-	return v.Main.Version
-}
-
 func main() {
 	flag.Parse()
 	defer glog.Flush()
 
 	if *versionFlag {
-		fmt.Printf("Version %v, commit %v, built at %v\n", buildVersion(), commit, date)
+		fmt.Printf("Version %v, commit %v, built at %v\n", version, commit, date)
 		os.Exit(0)
 	}
 
--- a/protoc-gen-grpc-gateway/main.go
+++ b/protoc-gen-grpc-gateway/main.go
@@ -13,6 +13,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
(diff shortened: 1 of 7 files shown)


protoc-gen-grpc-gateway/main.go
```diff
@@ -44,12 +45,20 @@ var (
...
+	return v.Main.Version
+}
+
 func main() {
 	flag.Parse()
 	defer glog.Flush()
 
 	if *versionFlag {
-		fmt.Printf("Version %v, commit %v, built at %v\n", version, commit, date)
+		fmt.Printf("Version %v, commit %v, built at %v\n", buildVersion(), commit, date)
```
johanbrandhorst, line 61: Could you add the 1.18 specific stuff to a `main_go1.18` and support the old format for Go 1.17 and below? We try to support the last 3 releases I think.

Semior001: done, if i understood you correctly

johanbrandhorst (changes requested):
Sorry for the confusion, but is not what I meant. You've got the new files right, but I don't want them to be using an init function like this. Instead, please define a `readVersion` both in each of the go1.18 specific files and in some non-go1.18 specific files, so that the function is always available, then call the function from the main file like you did before. Does that make sense?

Semior001:
> Sorry for the confusion, but is not what I meant. You've got the new files right, but I don't want them to be using an init function like this. Instead, please define a readVersion both in each of the go1.18 specific files and in some non-go1.18 specific files, so that the function is always available, then call the function from the main file like you did before. Does that make sense?

Okay, got it. Please, check out the latest version of PR.

johanbrandhorst (approved):
Perfect, thank you!

johanbrandhorst:
The bazel build is failing because the new files have not been added to the bazel definitions. Please run the file regeneration commands: https://github.com/grpc-ecosystem/grpc-gateway/blob/master/CONTRIBUTING.md#i-want-to-regenerate-the-files-after-making-changes.

tikto29:
$ go install \
    github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway \
    github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2 \
    google.golang.org/protobuf/cmd/protoc-gen-go \
    google.golang.org/grpc/cmd/protoc-gen-go-grpc a

Semior001:
resolved by https://github.com/grpc-ecosystem/grpc-gateway/pull/4359

PR #3026, opened 2022-11-23
Feature: Items support on OpenAPI V2 Responses

krak3n:
# Feature 🚀 

Adds support for array `items` on Open API V2 response objects. This is currently not supported when `disable_default_responses` is `true`.

This allows users to define array response objects, for example:

```protobuf
message Foo {
  string bar = 1;
}

message ListFoosResponse {
  repeated Foo foos = 1;
}

service FooService {
  rpc ListFoos(google.protobuf.Empty) returns (ListFoosResponse) {
    option (google.api.http) = {
      get: "/foo";
      response_body: "foos"; // This will not be handled when disable_default_responses=true
    };

      responses: {
        key: "200",
        value: {
          description: "List Foos"
          schema: {
            json_schema: {
              type: ARRAY;
              items: {ref: ".some.package.v1.Foo"};
            }
          }
        };
      };
    };
  };
};
```

This will generate the following `yaml` when `disable_default_responses=true`

```yaml
  /foo:
    get:
      summary: List Foos.
      operationId: ListFoos
      responses:
        "200":
          description: List Foos
          schema:
            type: array
            items:
              $ref: '#/definitions/some.package.v1.Foo'
```

See examples [here](https://swagger.io/specification/v2/#response-object-examples) on the OpenAPI V2 spec page.

## TODO

* [ ] Documentation


krak3n:
Before I went any further adding tests / docs I wanted to get your thoughts on this one @johanbrandhorst 😄

protoc-gen-openapiv2/options/openapiv2.proto
```diff
@@ -578,6 +577,45 @@ message JSONSchema {
...
+  // }
+  //
+  // rpc ListFoos(google.protobuf.Empty) returns (ListFoosResponse) {
+  //   option (google.api.http) = {
+  //     get: "/foo";
+  //     response_body: "foos";
+  //   };
+  // };
+  // ```
+  JSONSchema items = 50;
```
johanbrandhorst, line 618: Any reason we shouldn't just make this 19 and hoist it up to the comment above?

johanbrandhorst (commented):
Hi Chris, thanks for the PR! I understand we already allow users to specify custom responses, but this now adds the ability to specify a custom response where the outermost type is an array, is that correct? If so, this isn't so much about how this feature interacts with the `disable_default_responses` feature, but rather just about adding this functionality, correct? If so, should we rewrite the proto file comment to focus more on that?

spatecon (approved):

PR #3040, opened 2022-12-04
chore(logspam): allow_repeated_fields_in_body is deprecated

alexeagle:
This flag causes logspam due to https://github.com/grpc-ecosystem/grpc-gateway/commit/672f079d1ec726166bbdf413b1fa24bf6da2cb35

It seems that change just neglected to remove the explicit flag settings for Bazel users.


johanbrandhorst (approved):
LGTM. Apologies for missing this in the original PR. Thanks for the fix!

Issue #3059, opened 2022-12-12
single file with grpc annotations for multiple services don't work


gvencadze:
## 🐛 Bug Report

If you have yaml file with grpc annotations which contains selectors for multiple services, it'll raise an error if you try to generate code.

#### File structure scheme:
```
project
└───proto
    └───api
        └───first_svc
        │     └───v1
        │          └───first_svc.proto
        └───second_svc
        │     └───v1
        │          └───second_svc.proto
        │
        │
        └─── grpc_annotations.yaml
```

#### grpc_annotations.yaml content:
```
type: google.api.Service
config_version: 3

http:
  rules:
      ### First service ###
    - selector: api.first_svc.v1.FirstService.Get
      get: /api/v1/first/get

      ### Second service ###
    - selector: api.second_svc.v1.SecondService.Get
      get: /api/v2/second/get
```

## To Reproduce

1. Create multiple proto services
2. Add yaml file with grpc configuration which contains selectors for this services
3. Try to generate (I'm using buf)

## Expected behavior

I'm expecting to see generated code without errors

## Actual Behavior

I'm getting error:
```
Failure: plugin grpc-gateway: HTTP rules without a matching selector:
```

But If there will be single service and file with annotations will be located at the same path - everything will be okay

## Your Environment

| instrument  |  version  | 
|---|---|
| OS  | macOS 12   | 
|  Go | 1.19.3 darwin/arm64  | 
| protoc-gen-grpc-gateway  | v2.16.0  | 
| protoc-gen-openapiv2  | v2.16.0  | 
| buf | v1.9.0 |

johanbrandhorst:
Thanks for the issue. As discussed on Slack, I would expect our generator to be able to find the messages you are trying to select, since the selectors are using the global namespace. Would you be interested in fixing this? I think the first step would be to create a reproduction scenario in our examples protos. We already have [an example file](https://github.com/grpc-ecosystem/grpc-gateway/blob/main/examples/internal/proto/examplepb/standalone_echo_service.yaml) for external API configuration, maybe we could extend that? Or maybe a whole new folder would make sense.

gvencadze:
@johanbrandhorst you mean to create new folder which will contain two proto services and global `grpc_configuration` for them as illustrated on my example project scheme?

Btw, I'm interested in contributing to open source :)

dlanov:
[...]

PR #3062, opened 2022-12-13
feat: add example of non-working global grpc_api_configuration file

gvencadze:
It's a reproducible example of issue https://github.com/grpc-ecosystem/grpc-gateway/issues/3059.

I've added two proto services, `grpc_configuration` and `globalconfig.buf.gen.yaml` files to `examples/internal/proto/globalconfig`


examples/internal/proto/globalconfig/grpc_annotations.yaml
```diff
@@ -0,0 +1,12 @@
+type: google.api.Service
+config_version: 3
+
+http:
+  rules:
+      ### First service ###
+    - selector: api.first_svc.v1.FirstService.Get
```
johanbrandhorst, line 7: This should match the package definition in the protobuf file
```suggestion
    - selector: grpc.gateway.examples.internal.proto.globalconfig.first_svc.v1.FirstService.Get
```

examples/internal/proto/globalconfig/grpc_annotations.yaml
```diff
@@ -0,0 +1,12 @@
...
+      get: /api/v1/first/get
+
+      ### Second service ###
+    - selector: api.second_svc.v1.SecondService.Get
```
johanbrandhorst, line 11: Same here

examples/internal/proto/globalconfig/globalconfig.buf.gen.yaml
```diff
@@ -0,0 +1,11 @@
+version: v1
+plugins:
+  - name: grpc-gateway
+    out: .
+    opt:
+      - paths=source_relative
+      - grpc_api_configuration=examples/internal/proto/globalconfig/grpc_annotations.yaml
```
johanbrandhorst, line 7: I don't think this option does anything to this generator?

johanbrandhorst (changes requested):
I think you need to run the generation commands? See CONTRIBUTING.md in the root of the repository for a guide.

examples/internal/proto/globalconfig/globalconfig.buf.gen.yaml
gvencadze: Which option? `-paths=source-relative`?

gvencadze:
@johanbrandhorst yeah, I've take a look at contributing guide, but I can't run regenerate command cause it raise following error:
```
Failure: plugin grpc-gateway: HTTP rules without a matching selector: .grpc.gateway.examples.internal.proto.globalconfig.second_svc.v1.SecondService.Get
HTTP rules without a matching selector: .grpc.gateway.examples.internal.proto.globalconfig.first_svc.v1.FirstService.Get
make: *** [Makefile:115: proto] Error 1
```

gvencadze:
Btw, what do you think about ability to specify multiple configuration files? If it's okay, I'll take a look how to implement it

examples/internal/proto/globalconfig/globalconfig.buf.gen.yaml
johanbrandhorst: Huh, nevermind, it's here: https://github.com/grpc-ecosystem/grpc-gateway/blob/9b09466d8afe9003306397242ba3d1e3fbcd4cae/protoc-gen-grpc-gateway/main.go#L29. I don't know why I left this comment.

johanbrandhorst:
> Btw, what do you think about ability to specify multiple configuration files? If it's okay, I'll take a look how to implement it

I really don't think that will be a good idea. I would sooner we allow users to specify multiple target files and directories in a single file. Specifying an external file is already a hack, unfortunately.

johanbrandhorst:
> @johanbrandhorst yeah, I've take a look at contributing guide, but I can't run regenerate command cause it raise following error:
> 
> ```
> Failure: plugin grpc-gateway: HTTP rules without a matching selector: .grpc.gateway.examples.internal.proto.globalconfig.second_svc.v1.SecondService.Get
> HTTP rules without a matching selector: .grpc.gateway.examples.internal.proto.globalconfig.first_svc.v1.FirstService.Get
> make: *** [Makefile:115: proto] Error 1
> ```

Yeah uhm I'm not entirely sure what the selector _should_ be. Try and play around with it a bit 😝

gvencadze:
@johanbrandhorst selector is right, If I delete second_svc rule and remove folder "second_svc" everything generates as expected, otherwise returns selector error. Also it's look kinda strange that in output with error selector is got dot at the start of selector e.g. instead of `app_selector` in output I see `.app_selector`

gvencadze:
> I really don't think that will be a good idea. I would sooner we allow users to specify multiple target files and directories in a single file. Specifying an external file is already a hack, unfortunately.

Oh, if it's already a hack, than adding new features to it will be strange, I agree

johanbrandhorst:
> @johanbrandhorst selector is right, If I delete second_svc rule and remove folder "second_svc" everything generates as expected, otherwise returns selector error. Also it's look kinda strange that in output with selector error we got dot at the start of selector e.g. instead of `app_selector` we got `.app_selector`

It does look strange, I agree. I think it has to do with how the protobuf internal representation looks. I'm not sure why you're getting that selector error, you might have to do some debugging in the code, sorry. This part of the generator is rarely used and could do with some love.

gvencadze:
> It does look strange, I agree. I think it has to do with how the protobuf internal representation looks. I'm not sure why you're getting that selector error, you might have to do some debugging in the code, sorry. This part of the generator is rarely used and could do with some love.

Maybe you can share any instruction how to debug such instruments? I tried run it as standard go application, but got error saying that it should be run using protoc

PR #3072, opened 2022-12-16
fix: include json name to query param filter

### LINKED ISSUE #3069: path parameters are overwritten by query parameters with the same name
## 🐛 Bug Report

if you have path parameter in your proto endpoint definition, you can overwrite the value by setting a query parameter with the same name. This is a security issue in my opinion since a gateway or routing rule maybe only checks the path but not the query parameters.

## To Reproduce

The following service definition given:
```
service MyService {
  rpc GetResource(GetResourceRequest) returns (GetResourceResponse) {
    option (google.api.http) = {
      get: "/admin/{resource_id}"
    };
  }
}
```
One can call the API with `GET <host>/admin/<some_resource_id>` and gets the resource back. But it is also possible to call the endpoint like this:  `GET <host>/admin/<my_resource_id>?resourceId=<some_other_resource_id>`

In the GRPC Handler the value of resource_id is set to <some_other_resource_id> instead of <my_resource_id>

## Expected behavior
It should not be possible to overwrite path parameters with query parameters.

## Actual Behavior
It is possible to overwrite path parameters with query parameters.

ljmsc:
Fixes #3069

Yes

The fix will include the JSON name of the proto field into the query param filter to prevent the overwrite of the path parameters with query parameters


protoc-gen-grpc-gateway/internal/gengateway/template.go
```diff
@@ -72,6 +72,9 @@ func (b binding) QueryParamFilter() queryParamFilter {
 	}
 	for _, p := range b.PathParams {
 		seqs = append(seqs, strings.Split(p.FieldPath.String(), "."))
+		if p.Target.JsonName != nil {
+			seqs = append(seqs, strings.Split(*p.Target.JsonName, "."))
+		}
```
johanbrandhorst, line 77: We probably also want to do this for the Body parameters as well, since the same thing could happen there, I think? Could you try and see what happens if you try to override a body parameter?

johanbrandhorst (changes requested):
I think we'd need to add some tests to make sure we don't regress on this - do you think you could add a test to the [integration tests](https://github.com/grpc-ecosystem/grpc-gateway/blob/main/examples/internal/integration/integration_test.go) that would have failed without this change? Also you'll need to regenerated the files after this update, see `CONTRIBUTING.md` in the root of the repository for a step by step guide.

examples/internal/integration/integration_test.go
```diff
@@ -347,6 +348,35 @@ func testEchoOneof2(t *testing.T, port int, apiPrefix string, contentType string
...
+
+	msg := new(examplepb.UnannotatedSimpleMessage)
+	if err := marshaler.Unmarshal(buf, msg); err != nil {
+		t.Errorf("marshaler.Unmarshal(%s, msg) failed with %v; want success", buf, err)
+		return
+	}
+	if got, want := msg.GetResourceId(), "my_resource_id"; got != want {
+		t.Errorf("msg.GetResourceId() = %q; want %q", got, want)
+	}
+}
```
johanbrandhorst, line 378: Thanks for this! Could you also add a test where the variable is set in the body and attempted overwritten in the query?

johanbrandhorst (changes requested):
The new test looks great, thanks, lets try using the body too?

johanbrandhorst:
Please rebase on `main`, I've fixed the Bazel error

johanbrandhorst (approved):
LGTM

johanbrandhorst:
Thanks for your contribution!

johanbrandhorst:
Hi, unfortunately I've had to revert this feature for the next release because it introduced a regression (not your fault, we just didn't see it when working on it). If you'd like to reintroduce this, please open a new PR and we can make sure we cover the case that was affected. Thanks! CC @ljmsc

ljmsc:
hey @johanbrandhorst thanks for letting me know about this change. I'm actually a bit confused about this since we talked about the breaking change in the original issue I created #3069 and you agreed that it is more important to fix a potential security issue. The revert of this change will introduce the security issue back into the system. :(

johanbrandhorst:
Yes, it does reintroduce the security issue, but as you can see in #3848 this fix actually made it impossible to set any values on the referenced message through the query. Ideally we'd be able to reintroduce this fix in a way that still allows other fields on the same message to be set via the query. Does that make sense?

ljmsc:
It makes sense to fix the bug, yeah. But for me it doesn't make sense to reintroduce a security issue to the library because of a bug which wasn't noticed over a year. IMHO a security issue is much worse than a bug which seems to be only a corner case.
I will try to provide a fix for the security issue again in the next days...

PR #3082, opened 2022-12-20
Make read_only work with references using AllOf

### LINKED ISSUE #1137: read_only not working with references.
So I want to be able to define messages in other messages as `read_only` fields when generating openapi doc, however, that does not seem possible at the moment.

Given example:

```

syntax = "proto3";

import "protoc-gen-swagger/options/annotations.proto";

service SomeService {
  // Create new Some
  rpc SomeService(SomeServiceRequest) returns (SomeMessage) {
    option (google.api.http) = {
      post: "/do-something"
      body: "body"
    };
  }
}

message SomeServiceRequest {
  SomeMessage body = 1;
}

message SomeMessage {
  // Readonly id, works fine here
  string id = 1 [(grpc.gateway.protoc_gen_swagger.options.openapiv2_field) = {read_only: true}];

  // OtherReadOnlyMessage readonly, does not work.
  OtherReadOnlyMessage field = 2 [(grpc.gateway.protoc_gen_swagger.options.openapiv2_field) = {read_only: true}];

  // OtherReadOnlyMessage not readonly
  OtherReadOnlyMessage other = 3;
}

// Message I want to use in multiple places
message OtherReadOnlyMessage {
  string whatever = 1;
}

```

It will output something like this:

```
{
  "swagger": "2.0",
  "info": {
    "title": "readonly.proto",
    "version": "version not set"
  },
  "consumes": [
    "application/json"
  ],
  "produces": [
    "application/json"
  ],
  "paths": {
    "/do-something": {
      "post": {
        "summary": "Create A new Some",
        "operationId": "SomeService",
        "responses": {
          "200": {
            "schema": {
              "$ref": "#/definitions/SomeMessage"
            }
          },
          "default": {
            "description": "An unexpected error response",
            "schema": {
              "$ref": "#/definitions/runtimeError"
            }
          }
        },
        "parameters": [
          {
            "name": "body",
            "in": "body",
            "required": true,
            "schema": {
              "$ref": "#/definitions/SomeMessage"
            }
          }
        ],
        "tags": [
          "SomeService"
        ]
      }
    }
  },
  "definitions": {
    "OtherReadOnlyMessage": {
      "type": "object",
      "properties": {
        "whatever": {
          "type": "string"
        }
      },
      "title": "Message I want to use in two places"
    },
    "SomeMessage": {
      "type": "object",
      "properties": {
        "id": {
          "type": "string",
          "title": "Readonly id, works",
          "readOnly": true
        },
        "field": {
          "$ref": "#/definitions/OtherReadOnlyMessage",
          "title": "OtherReadOnlyMessage readonly",
          "readOnly": true
        },
        "other": {
          "$ref": "#/definitions/OtherReadOnlyMessage",
          "title": "OtherReadOnlyMessage not readonly"
        }
      }
    },
...
  }
}
```

The `readOnly` field is set correctly but according to swagger docs:

> Any sibling elements of a $ref are ignored. This is because $ref works by replacing itself and everything on its level with the definition it is pointing at.

* https://swagger.io/docs/specification/using-ref/

So its not reading and listing this property as read only in the end.

It seems like common practice is to instead of directly using `$ref` is to use `allOf` of the `$ref` instead, then the readOnly property would be used.
* https://stackoverflow.com/questions/51402156/how-to-declare-a-ref-property-as-readonly-in-openapi-swagger
* https://github.com/OAI/OpenAPI-Specification/issues/1671
* https://github.com/swagger-api/swagger-ui/issues/3445#issuecomment-339649576

So this
```
...
"field": {
          "$ref": "#/definitions/OtherReadOnlyMessage",
          "title": "OtherReadOnlyMessage readonly",
          "readOnly": true
        },
...
```

Should in that case become this:
```
...
        "field": {
          "title": "OtherReadOnlyMessage readonly",
          "readOnly": true,
          "allOf": [
            { "$ref": "#/definitions/OtherReadOnlyMessage" }
          ]
        },
...
```

Thoughts?

same-id:

Fixes #1137

Yes

As described here:

https://github.com/grpc-ecosystem/grpc-gateway/issues/1137

This partially reverts the recent commit:

a0fe8d7c8676db9674ccbc4c4fc642790a8b8065

Which doesn't solve the problem for read_only being set on the field.

This feature will require setting the "use_allof_for_refs" to true since this breaks existing compatibility with swagger-codegen generated files from the AllOf'd schema - type is now "object" and not $ref.

Test added + ReDoc now works with read_only message/enum fields.

### DIFF
--- a/internal/descriptor/registry.go
+++ b/internal/descriptor/registry.go
@@ -136,6 +136,10 @@ type Registry struct {
 	// disableDefaultResponses disables the generation of default responses.
 	// Useful if you have to support custom response codes that are not 200.
 	disableDefaultResponses bool
+
+	// useAllOfForRefs, if set, will use allOf as container for $ref to preserve same-level
+	// properties
+	useAllOfForRefs bool
 }
 
 type repeatedFieldSeparator struct {
@@ -771,4 +775,14 @@ func (r *Registry) SetDisableDefaultResponses(use bool) {
 // GetDisableDefaultResponses returns disableDefaultResponses
 func (r *Registry) GetDisableDefaultResponses() bool {
 	return r.disableDefaultResponses
+}
+
+// SetUseAllOfForRefs sets useAllOfForRefs
+func (r *Registry) SetUseAllOfForRefs(use bool) {
+	r.useAllOfForRefs = use
+}
+
+// GetUseAllOfForRefs returns useAllOfForRefs
+func (r *Registry) GetUseAllOfForRefs() bool {
+	return r.useAllOfForRefs
 }
--- a/protoc-gen-openapiv2/defs.bzl
+++ b/protoc-gen-openapiv2/defs.bzl
@@ -356,6 +356,12 @@ protoc_gen_openapiv2 = rule(
                   " Repeat this option to supply multiple values. Elements without" +
                   " visibility annotations are unaffected by this setting.",
         ),
+        "use_allof_for_refs": attr.bool(
+            default = False,
+            mandatory = False,
+            doc = "if set, will use allOf as container for $ref to preserve" +
+                  " same-level properties.",
+        ),
         "_protoc": attr.label(
             default = "@com_google_protobuf//:protoc",
             executable = True,
--- a/protoc-gen-openapiv2/internal/genopenapi/template.go
+++ b/protoc-gen-openapiv2/internal/genopenapi/template.go
@@ -515,6 +515,24 @@ func renderMessageAsDefinition(msg *descriptor.Message, reg *descriptor.Registry
 			fieldSchema.Required = nil
 		}
 
+		if reg.GetUseAllOfForRefs() {
+			if fieldSchema.Ref != "" {
+				// Per the JSON Reference syntax: Any members other than "$ref" in a JSON Reference object SHALL be ignored.
+				// https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03#section-3
+				// However, use allOf to specify Title/Description/readOnly fields.
+				if fieldSchema.Title != "" || fieldSchema.Description != "" || fieldSchema.ReadOnly {
+					fieldSchema = openapiSchemaObject{
+						Title:       fieldSchema.Title,
+						Description: fieldSchema.Description,
+						ReadOnly:    fieldSchema.ReadOnly,
+						AllOf:       []allOfEntry{{Ref: fieldSchema.Ref}},
+					}
+				} else {
+					fieldSchema = openapiSchemaObject{schemaCore: schemaCore{Ref: fieldSchema.Ref}}
+				}
+			}
+		}
+
 		kv := keyVal{Value: fieldSchema}
 		kv.Key = reg.FieldName(f)
 		if schema.Properties == nil {
--- a/protoc-gen-openapiv2/internal/genopenapi/template_test.go
+++ b/protoc-gen-openapiv2/internal/genopenapi/template_test.go
@@ -4690,6 +4690,7 @@ func TestRenderMessagesAsDefinition(t *testing.T) {
 		openAPIOptions        *openapiconfig.OpenAPIOptions
 		pathParams            []descriptor.Parameter
 		UseJSONNamesForFields bool
+		UseAllOfForRefs       bool
 	}{
 		{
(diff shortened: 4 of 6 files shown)


johanbrandhorst:
Please rebase on `main`, I've fixed the Bazel error

johanbrandhorst (commented):
Hi, I'm a little confused about this, but before diving into the fix itself, I'm wondering if your need for read only fields can be satisfied by using the OUTPUT_ONLY field annotation? See https://github.com/grpc-ecosystem/grpc-gateway/blob/492703c51e9c90b84e404ff11562407573e405fb/examples/internal/proto/examplepb/a_bit_of_everything.proto#L386 for an example. We might be better off just removing this annotation from here and only supporting this other field annotation.

same-id:
> Please rebase on `main`, I've fixed the Bazel error

Done

> Hi, I'm a little confused about this, but before diving into the fix itself, I'm wondering if your need for read only fields can be satisfied by using the OUTPUT_ONLY field annotation? See
> 
> https://github.com/grpc-ecosystem/grpc-gateway/blob/492703c51e9c90b84e404ff11562407573e405fb/examples/internal/proto/examplepb/a_bit_of_everything.proto#L386
> 
> for an example. We might be better off just removing this annotation from here and only supporting this other field annotation.

Not sure I follow - this is exactly OUTPUT_ONLY.
This commit is about making OUTPUT_ONLY generate the correct Redoc documentation.

`OUTPUT_ONLY` is the protobuf annotation, `readOnly` is the OpenAPI parameter.

If I specify OUTPUT_ONLY on Enums/Sub-messages - this doesn't work in OpenAPI v2 since Enums/Sub-messages are translated to $ref, and according to OpenAPI v2 you can't have readOnly in the same level - so when I later use Redoc on the output of the swagger.yaml file, `readOnly` is not set correctly and ReDoc ignores it, resulting in a non readonly field in the documentation.

Here is another illustration of the problem and the solution.

https://stackoverflow.com/a/51402417

johanbrandhorst:
I see, apologies for the confusion. The issue you linked describes using the `read_only` field annotation, which is different from the OUTPUT_ONLY field annotation:

```protobuf
string id = 1 [(grpc.gateway.protoc_gen_swagger.options.openapiv2_field) = {read_only: true}];
```.

johanbrandhorst (changes requested):
This change looks simple enough, but it looks like it's missing the bazel changes required for new flags, please have a look at https://github.com/grpc-ecosystem/grpc-gateway/blob/main/protoc-gen-openapiv2/defs.bzl#L352 for an example.

same-id:
> I see, apologies for the confusion. The issue you linked describes using the `read_only` field annotation, which is different from the OUTPUT_ONLY field annotation:
> 
> ```proto
> string id = 1 [(grpc.gateway.protoc_gen_swagger.options.openapiv2_field) = {read_only: true}];
> ```

Ah, I see the confusion now.
However this change will work for both, OUTPUT_ONLY and read_only swagger proto annotation since I'm doing the change at a point where the logic is unified

same-id:
> This change looks simple enough, but it looks like it's missing the bazel changes required for new flags, please have a look at https://github.com/grpc-ecosystem/grpc-gateway/blob/main/protoc-gen-openapiv2/defs.bzl#L352 for an example.

Will do

same-id:
Done

johanbrandhorst (approved):
LGTM

johanbrandhorst:
Thanks for your contribution!

warrenb95:
Hi I've been following on with this MR and I was just wondering why `use_allof_for_refs` hasn't been added in these other places like what was done with `visibility_restriction_selector`:
* https://github.com/grpc-ecosystem/grpc-gateway/blob/main/protoc-gen-openapiv2/defs.bzl#L206
* https://github.com/grpc-ecosystem/grpc-gateway/blob/main/protoc-gen-openapiv2/defs.bzl#L130

---
I've tried using `use_allof_for_refs` but getting an error `no such flag` with version v2.15.0

johanbrandhorst:
Argh, I guess we missed some of the places when adding this option, @same-id would you mind tidying this up?

sam-utila:
Missed it, thanks for fixing :-)

PR #3084, opened 2022-12-21
Fix field order in yaml generated OAS file

same-id:

Yes

Today, the field order in swagger.json files is consistent with the actual order of fields in the .proto messages.
(i.e. not by protobuf index/tag and not alphabetically sorted, but actually in the same order inside the protobuf)

However, in swagger.yaml files the order becomes alphabetical due to automatic map[string]interface{} ordering in yaml marshaling.
This gives less control over generated documentation (Redoc) from those yaml files.

This PR uses yaml.Node introduces in yaml.v3 to keep the same field order in yaml generated files as well.


johanbrandhorst (approved):
Nice!

johanbrandhorst:
Thanks for your contribution!

PR #3087, opened 2022-12-21
Add optional support for writeOnly OASv3 property

same-id:

Yes

Only available in Open API v3, but some tools (like Redoc) support it for v2 as well.


protoc-gen-openapiv2/options/openapiv2.proto
```diff
@@ -534,6 +534,8 @@ message JSONSchema {
   reserved 30 to 33;
   // Items in 'array' must be unique.
   repeated string array = 34;
+  // Only available in Open API v3, but some tools (like Redoc) support it for v2.
+  bool write_only = 99;
```
same-id, line 538: I added it at number 99 since it's not in the openapi v2 spec, but can change that if needed

johanbrandhorst: I don't want to create this precedent, we don't support any openapiv3 properties in our openapiv2 generator. We support setting extensions, which have been used to backport some openapiv3 features. I think this is going too far.

johanbrandhorst:
I could see a case for accepting https://github.com/googleapis/googleapis/blob/master/google/api/field_behavior.proto#L72 and output some sort of extension based on it, but I don't want to output any properties that aren't defined in OpenAPIv2.

same-id:
But then we will have to patch Redoc to allow that extension as well which is basically writeOnly.

Isn't putting it behind a flag which is off by default a good enough compromise?

same-id:
We can also just change the protobuf annotations to be "x-writeOnly" but actually output "writeOnly" in the yaml.
This way it signifies in annotated protobufs that this is an extension (but it will have a non-standard b

same-id:
We can do something "interesting" like setting an "x-writeOnly" extension in the protobuf and let "allowWriteOnly" flag to set "writeOnly" in the actually outputted openapi yaml/json - this way we don't need to edit the `openapiv2.proto` and it will be more clear in annotated protobufs that this is an extension.

johanbrandhorst:
I'm sorry, this is one of the few lines I enforce with the project in general. We will not be adding any non-openapiv2 fields to the output. We have plenty of interest in an OpenAPIv3 generator, I suppose this is another reason. Fair warning, if you do want to do that, it is a large piece of work.

same-id:
Actually there is not need for x-writeOnly.

It can be done only by translating INPUT_ONLY directly into `writeOnly`.
(removing all .proto changes entirely)

But you're saying you don't want to do even that, even behind a default-false flag.

Even though the output is considered somehow valid in tools like Redoc.

same-id:
Regarding OpenAPI v3 - this seems indeed like a "too much" work.

I tried using gnostic which is OpenAPI v3, but there are other problems with that tool since it is less mature, so we decided to stay in v2 with grpc-gateway.

same-id:
To be honest, there are some discrepancies between protobuf annotations to OpenAPI, even in v3.

For example REQUIRED means only "Required" in requests, not in responses.

But there is no semantic for "RequiredInRequest" in OpenAPIv3:
https://github.com/OAI/OpenAPI-Specification/issues/1873#issuecomment-483229565

So maybe instead we can have a flag for grpc-gateway that if set, creates different schemas (recursively) for requests and responses if they have REQUIRED or INPUT_ONLY annotations.

And this way the OpenAPI semantics don't break.

johanbrandhorst:
I appreciate that you're trying to find a solution here, but it is simply unacceptable to me to output any fields that are not explicitly defined in the OpenAPIv2 spec. I hope you can appreciate my responsibility to all my other users, both current and future, that may be affected by a change like this, however small it may seem. If this is a required piece of functionality for you, you are free to fork the project or use another tool.

sam-utila:
Sure, I just testing it with readme.io - they fail to parse if it contains writeOnly, so it's probably only Redoc that does not validate it properly.

Anyway, we can always create different schemas for request/response, or even do it automatically in grpc-gateway

johanbrandhorst:
I'm going to go ahead and close this PR, thanks for being understanding. As you say, you can always post-process the generated swagger file and add custom handling to your gateway, worst case. Thanks for your contributions so far :).

same-id:
I'm thinking of automatically adding `x-writeOnly: true` extension when a field is flagged `INPUT_ONLY`.
Instead of writing the following on every INPUT_ONLY field.
```
          extensions: {
            key: "x-writeOnly";
            value {
              bool_value: true;
            }
          }
        }
```

Does it even make sense to upstream this?

johanbrandhorst:
Unless this extension is well-established in the OpenAPI community (like `x-nullable`), I doubt it.

same-id:
I decided to name this x-writeOnly for obvious reasons.

Seems that there are other people that do something similar in various projects:

https://github.com/search?p=3&q=%22x-writeOnly%22&type=Code

If I manage to upstream this to Redoc will this tip the scale in my favor?
:-)

johanbrandhorst:
I'm not gonna say it'll never happen but I think the current rubric for this sort of thing is:

1. Is the feature requested by at least 2 separate users?
2. Is the feature established within the OpenAPIv2 community?

The second one might be hard to answer but `x-nullable` is the only extension we've added so far, so anything of that or greater adoption would work.

Issue #2988, opened 2022-11-02
Partial resource update RPCs include the update mask in the swagger specification


peterwillis:
## 🐛 Bug Report

RPCs can perform partial resource updates through the use of a `patch` binding and `google.protobuf.FieldMask` as described [here](https://grpc-ecosystem.github.io/grpc-gateway/docs/mapping/patch_feature/). The swagger specification generated still includes an `updateMask` field which is ultimately ignored by the gRPC-Gateway.

## To Reproduce

Generate the swagger specification for the "a bit of everything" [example](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/examples/internal/proto/examplepb/a_bit_of_everything.proto).

## Expected behavior

The updateMask field is not included in the swagger specification

## Actual Behavior

The updateMask field is [included](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/examples/internal/proto/examplepb/a_bit_of_everything.swagger.json#L5952)

## Your Environment

v2.12.0

johanbrandhorst:
Hi Peter, thanks for your issue! I thought we already supported this but evidently we don't. Considering that the patch feature is the default behavior on the gateway I think we should definitely add this when detecting a field called `update_mask` of type `.google.protobuf.FieldMask` at the top level of a request (as explained in [the docs](https://grpc-ecosystem.github.io/grpc-gateway/docs/mapping/patch_feature/)). If this surprises users, we can introduce a flag to disable the behavior later. Would you be willing to help contribute this? I think we'd start by defining a failing test in https://github.com/grpc-ecosystem/grpc-gateway/blob/master/protoc-gen-openapiv2/internal/genopenapi/template_test.go. Feel free to copy one of the existing tests and adapting it to your use case. This file is a bit of a mess so be warned 😅. I then think we'd start at https://github.com/grpc-ecosystem/grpc-gateway/blob/master/protoc-gen-openapiv2/internal/genopenapi/template.go#L436 and add some special case logic for this scenario. What do you think?

peterwillis:
Thank you for the pointers, I would be very happy to help.

csh995426531:
[...]

PR #3093, opened 2022-12-24
fix: Partial resource update, no longer generate updateMask field

csh995426531:
**References to other Issues or PRs**
Fixes https://github.com/grpc-ecosystem/grpc-gateway/issues/2988

**Have you read the [Contributing Guidelines](https://github.com/grpc-ecosystem/grpc-gateway/blob/main/CONTRIBUTING.md)?**
Yes

**Brief description of what is fixed or changed**
When the request method is patch, and the http body is not "*", do not need to generate updateMask field to swagger document

### DIFF
--- a/docs/docs/mapping/patch_feature.md
+++ b/docs/docs/mapping/patch_feature.md
@@ -15,7 +15,9 @@ There are two scenarios:
 
 - The FieldMask is hidden from the REST request as per the
   [Google API design guide](https://cloud.google.com/apis/design/standard_methods#update) (as in the first additional binding in the
-  [UpdateV2](https://github.com/grpc-ecosystem/grpc-gateway/blob/370d869f65d1ffb3d07187fb0db238eca2371ce3/examples/internal/proto/examplepb/a_bit_of_everything.proto#L428-L431) example). In this case, the FieldMask is updated from the request body and set in the gRPC request message.
+  [UpdateV2](https://github.com/grpc-ecosystem/grpc-gateway/blob/370d869f65d1ffb3d07187fb0db238eca2371ce3/examples/internal/proto/examplepb/a_bit_of_everything.proto#L428-L431) example). In this case, the FieldMask is updated from the request body and set in the gRPC request message. 
+  - By default this feature is enabled, if you need to disable it, you can use the plugin option `allow_patch_feature=false`. 
+  - Note: The same option is supported by the `protoc-gen-openapiv2` plugin.
 - The FieldMask is exposed to the REST request (as in the second additional binding in the [UpdateV2](https://github.com/grpc-ecosystem/grpc-gateway/blob/370d869f65d1ffb3d07187fb0db238eca2371ce3/examples/internal/proto/examplepb/a_bit_of_everything.proto#L432-L435) example). For this case, the field mask is left untouched by the gateway.
 
 ## Example Usage
--- a/protoc-gen-openapiv2/internal/genopenapi/template.go
+++ b/protoc-gen-openapiv2/internal/genopenapi/template.go
@@ -145,9 +145,12 @@ func getEnumDefault(reg *descriptor.Registry, enum *descriptor.Enum) string {
 }
 
 // messageToQueryParameters converts a message to a list of OpenAPI query parameters.
-func messageToQueryParameters(message *descriptor.Message, reg *descriptor.Registry, pathParams []descriptor.Parameter, body *descriptor.Body) (params []openapiParameterObject, err error) {
+func messageToQueryParameters(message *descriptor.Message, reg *descriptor.Registry, pathParams []descriptor.Parameter, body *descriptor.Body, httpMethod string) (params []openapiParameterObject, err error) {
 	for _, field := range message.Fields {
 		if !isVisible(getFieldVisibilityOption(field), reg) {
+			continue
+		}
+		if reg.GetAllowPatchFeature() && field.GetTypeName() == ".google.protobuf.FieldMask" && field.GetName() == "update_mask" && httpMethod == "PATCH" && len(body.FieldPath) != 0 {
 			continue
 		}
 
@@ -1271,7 +1274,7 @@ func renderServices(services []*descriptor.Service, paths openapiPathsObject, re
 				}
 
 				// add the parameters to the query string
-				queryParams, err := messageToQueryParameters(meth.RequestType, reg, b.PathParams, b.Body)
+				queryParams, err := messageToQueryParameters(meth.RequestType, reg, b.PathParams, b.Body, b.HTTPMethod)
 				if err != nil {
 					return err
 				}
--- a/protoc-gen-openapiv2/main.go
+++ b/protoc-gen-openapiv2/main.go
@@ -43,6 +43,7 @@ var (
 	disableServiceTags             = flag.Bool("disable_service_tags", false, "if set, disables generation of service tags. This is useful if you do not want to expose the names of your backend grpc services.")
 	disableDefaultResponses        = flag.Bool("disable_default_responses", false, "if set, disables generation of default responses. Useful if you have to support custom response codes that are not 200.")
 	useAllOfForRefs                = flag.Bool("use_allof_for_refs", false, "if set, will use allOf as container for $ref to preserve same-level properties.")
+	allowPatchFeature              = flag.Bool("allow_patch_feature", true, "whether to hide update_mask fields in PATCH requests from the generated swagger file.")
 )
 
 // Variables set by goreleaser at build time
@@ -129,6 +130,7 @@ func main() {
 	reg.SetDisableServiceTags(*disableServiceTags)
 	reg.SetDisableDefaultResponses(*disableDefaultResponses)
 	reg.SetUseAllOfForRefs(*useAllOfForRefs)
+	reg.SetAllowPatchFeature(*allowPatchFeature)
 	if err := reg.SetRepeatedPathParamSeparator(*repeatedPathParamSeparator); err != nil {
 		emitError(err)
 		return
--- a/examples/internal/clients/abe/api/swagger.yaml
+++ b/examples/internal/clients/abe/api/swagger.yaml
@@ -4144,13 +4144,6 @@ paths:
         schema:
           $ref: "The book to update."
         x-exportParamName: "Book"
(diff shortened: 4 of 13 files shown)


johanbrandhorst (changes requested):
Hi, thanks for this PR! This looks great, but we're gonna need a few more things before we can merge it.

1. The `protoc-gen-grpc-gateway` must opt-in to the field mask behavior. We should add this opt in option here too, and keep it named consistently: `allow_patch_feature=true`. Please add a flag that enables this behavior and keep it disabled by default.
2. We have a great docs page for the patch feature: https://github.com/grpc-ecosystem/grpc-gateway/blob/main/docs/docs/mapping/patch_feature.md. Please add a section on how to use this new option to this page.

Thanks!

csh995426531:
> Hi, thanks for this PR! This looks great, but we're gonna need a few more things before we can merge it.
> 
> 1. The `protoc-gen-grpc-gateway` must opt-in to the field mask behavior. We should add this opt in option here too, and keep it named consistently: `allow_patch_feature=true`. Please add a flag that enables this behavior and keep it disabled by default.
> 2. We have a great docs page for the patch feature: https://github.com/grpc-ecosystem/grpc-gateway/blob/main/docs/docs/mapping/patch_feature.md. Please add a section on how to use this new option to this page.
> 
> Thanks!

Thank you for your guidance, I have revised these contents

docs/docs/mapping/patch_feature.md
```diff
@@ -15,7 +15,9 @@ There are two scenarios:
 
 - The FieldMask is hidden from the REST request as per the
   [Google API design guide](https://cloud.google.com/apis/design/standard_methods#update) (as in the first additional binding in the
-  [UpdateV2](https://github.com/grpc-ecosystem/grpc-gateway/blob/370d869f65d1ffb3d07187fb0db238eca2371ce3/examples/internal/proto/examplepb/a_bit_of_everything.proto#L428-L431) example). In this case, the FieldMask is updated from the request body and set in the gRPC request message.
+  [UpdateV2](https://github.com/grpc-ecosystem/grpc-gateway/blob/370d869f65d1ffb3d07187fb0db238eca2371ce3/examples/internal/proto/examplepb/a_bit_of_everything.proto#L428-L431) example). In this case, the FieldMask is updated from the request body and set in the gRPC request message. 
+  - By default this feature is enabled, if you need to disable it, you can use the plugin option `allow_patch_feature=false`. 
+  - Note: The same option is supported by the `gen-openapiv2` plugin.
```
johanbrandhorst, line 20: I think we usually refer to this by the full name or just the plugin.
```suggestion
  - Note: The same option is supported by the `protoc-gen-openapiv2` plugin.
```

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -1269,7 +1278,11 @@ func renderServices(services []*descriptor.Service, paths openapiPathsObject, re
 				if err != nil {
 					return err
 				}
-				parameters = append(parameters, queryParams...)
+				for _, queryParam := range queryParams {
+					if _, ok := notVisibleFields[queryParam.Name]; !ok {
```
johanbrandhorst, line 1282: Does the update_mask only appear in query parameters? Does it never appear in the body unintentionally?

protoc-gen-openapiv2/main.go
```diff
@@ -43,6 +43,7 @@ var (
 	disableServiceTags             = flag.Bool("disable_service_tags", false, "if set, disables generation of service tags. This is useful if you do not want to expose the names of your backend grpc services.")
 	disableDefaultResponses        = flag.Bool("disable_default_responses", false, "if set, disables generation of default responses. Useful if you have to support custom response codes that are not 200.")
 	useAllOfForRefs                = flag.Bool("use_allof_for_refs", false, "if set, will use allOf as container for $ref to preserve same-level properties.")
+	allowPatchFeature              = flag.Bool("allow_patch_feature", true, "determines whether to use PATCH feature involving update masks (using google.protobuf.FieldMask).")
```
johanbrandhorst, line 46: This help text is not great. We should tailor it to this generator. What do you think about
```suggestion
	allowPatchFeature              = flag.Bool("allow_patch_feature", true, "whether to hide update_mask fields from the generated swagger file.")
```

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -1250,6 +1252,13 @@ func renderServices(services []*descriptor.Service, paths openapiPathsObject, re
 						} else {
 							desc = fieldProtoComments(reg, bodyField.Target.Message, bodyField.Target)
 						}
+						if b.HTTPMethod == "PATCH" && reg.GetAllowPatchFeature() {
+							for _, requestTypeField := range meth.RequestType.Fields {
+								if requestTypeField.GetTypeName() == ".google.protobuf.FieldMask" {
+									notVisibleFields[requestTypeField.GetJsonName()] = nil
```
johanbrandhorst, line 1258: I think we should also only match fields that are named `update_mask`, right? So we don't even necessarily need this to be a map, it can juts be a bool, right?

johanbrandhorst (changes requested):

docs/docs/mapping/patch_feature.md
csh995426531: ok

protoc-gen-openapiv2/internal/genopenapi/template.go
csh995426531: Could indeed be more concise

protoc-gen-openapiv2/main.go
csh995426531: I mistakenly thought I needed to keep the same help text

protoc-gen-openapiv2/internal/genopenapi/template.go
csh995426531: I remember that the update_mask field does not appear in the body, I need to look at the code to confirm

protoc-gen-openapiv2/internal/genopenapi/template.go
csh995426531: > I think we should also only match fields that are named `update_mask`, right? So we don't even necessarily need this to be a map, it can juts be a bool, right?
Maybe I can modify the visibility of the fields to achieve this?

csh995426531: > I think we should also only match fields that are named `update_mask`, right? So we don't even necessarily need this to be a map, it can juts be a bool, right?

Maybe I can modify the visibility of the fields to achieve this?

johanbrandhorst: I'd rather we had logic to just omit the field than set a value that's not added by the user.

protoc-gen-openapiv2/internal/genopenapi/template.go
csh995426531: I didn't find that the body may appear to be filled with FieldMask

protoc-gen-openapiv2/internal/genopenapi/template.go
csh995426531: Yes, that would be bad

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -145,11 +145,14 @@ func getEnumDefault(reg *descriptor.Registry, enum *descriptor.Enum) string {
 }
 
 // messageToQueryParameters converts a message to a list of OpenAPI query parameters.
-func messageToQueryParameters(message *descriptor.Message, reg *descriptor.Registry, pathParams []descriptor.Parameter, body *descriptor.Body) (params []openapiParameterObject, err error) {
+func messageToQueryParameters(message *descriptor.Message, reg *descriptor.Registry, pathParams []descriptor.Parameter, body *descriptor.Body, httpMethod string) (params []openapiParameterObject, err error) {
 	for _, field := range message.Fields {
 		if !isVisible(getFieldVisibilityOption(field), reg) {
 			continue
 		}
+		if field.GetTypeName() == ".google.protobuf.FieldMask" && reg.GetAllowPatchFeature() && httpMethod == "PATCH" && len(body.FieldPath) != 0 {
```
johanbrandhorst, line 153: I think we also want to check that the field name is `update_mask` exactly. Also, we should check the flag first, since it invalidates everything else
```suggestion
		if reg.GetAllowPatchFeature() && field.GetTypeName() == ".google.protobuf.FieldMask" && httpMethod == "PATCH" && len(body.FieldPath) != 0 {
```

protoc-gen-openapiv2/main.go
```diff
@@ -43,6 +43,7 @@ var (
 	disableServiceTags             = flag.Bool("disable_service_tags", false, "if set, disables generation of service tags. This is useful if you do not want to expose the names of your backend grpc services.")
 	disableDefaultResponses        = flag.Bool("disable_default_responses", false, "if set, disables generation of default responses. Useful if you have to support custom response codes that are not 200.")
 	useAllOfForRefs                = flag.Bool("use_allof_for_refs", false, "if set, will use allOf as container for $ref to preserve same-level properties.")
+	allowPatchFeature              = flag.Bool("allow_patch_feature", true, "whether to hide update_mask fields from the generated swagger file.")
```
johanbrandhorst, line 46:
```suggestion
	allowPatchFeature              = flag.Bool("allow_patch_feature", true, "whether to hide update_mask fields in PATCH requests from the generated swagger file.")
```

johanbrandhorst (changes requested):
Thanks, this is almost there!

protoc-gen-openapiv2/internal/genopenapi/template.go
csh995426531: ok

csh995426531:
I didn't change anything else, why does the make generate command start with an error when I run it locally?
image

johanbrandhorst (approved):
LGTM

johanbrandhorst:
Thanks for your contribution!

johanbrandhorst:
That panic is interesting, but it's probably fixed in new versions of `buf`. If you want, we could update the `buf` version we use.

PR #3096, opened 2022-12-26
Update main.go

hamedprog:


johanbrandhorst:
Hi, thanks for your PR. Unfortunately, this is a breaking change, and while I'm sympathetic to making this consistent, I don't think we can do this without breaking existing users.

PR #3100, opened 2022-12-27
Support field extensions with use_allof_for_refs

same-id:

Yes

Allow extensions to be used with the recently added AllOf feature


johanbrandhorst (changes requested):
Hi, thank you for this, I think it's probably a good time to add an example protobuf file that is generated with this option turned on, so we can see the difference introduced by this sort of chance. Please add a small new file to the https://github.com/grpc-ecosystem/grpc-gateway/tree/main/examples/internal/proto/examplepb folder. You'll need a new `buf.gen.yaml` file which sets this option, see the existing files in the directory for an example. Thanks!

same-id:
Will do.

same-id:
Hey, still haven't got to creating a whole new example protobuf.

I see that some other PRs extending my work have also made their way there:

https://github.com/grpc-ecosystem/grpc-gateway/pull/3240

same-id:
Hi, @johanbrandhorst 

Been some time

Wanted to try and upstream one of our private changes, tended to your requested changes.

Not sure if I should also look at the v3 tree.

johanbrandhorst (approved):
Thanks!

PR #3114, opened 2023-01-09
feat(3102):update DefaultHeaderMatcher function comment docs

### LINKED ISSUE #3102: The http header mapping behavior is different from documentation
## 📚 Documentation

The documentation says:
> HTTP headers that start with 'Grpc-Metadata-' are mapped to gRPC metadata (prefixed with grpcgateway-).

I expected the metadata being prefixed with `grpcgateway-`. However they are not. After digging into the code, it turns out that the code doesn't add the prefix to the metadata:

```
// DefaultHeaderMatcher is used to pass http request headers to/from gRPC context. This adds permanent HTTP header
// keys (as specified by the IANA) to gRPC context with grpcgateway- prefix. HTTP headers that start with
// 'Grpc-Metadata-' are mapped to gRPC metadata after removing prefix 'Grpc-Metadata-'.
func DefaultHeaderMatcher(key string) (string, bool) {
	key = textproto.CanonicalMIMEHeaderKey(key)
	if isPermanentHTTPHeader(key) {
		return MetadataPrefix + key, true
	} else if strings.HasPrefix(key, MetadataHeaderPrefix) {
		return key[len(MetadataHeaderPrefix):], true
	}
	return "", false
}
```

Since our code base relies on this behavior, fixing the code may break existing users. So maybe it's good to just change the documentation to align with the code.

li31727:
Fixes #3102

yes
update  DefaultHeaderMatcher function comment docs

### DIFF
--- a/runtime/mux.go
+++ b/runtime/mux.go
@@ -101,8 +101,9 @@ func SetQueryParameterParser(queryParameterParser QueryParameterParser) ServeMux
 type HeaderMatcherFunc func(string) (string, bool)
 
 // DefaultHeaderMatcher is used to pass http request headers to/from gRPC context. This adds permanent HTTP header
-// keys (as specified by the IANA) to gRPC context with grpcgateway- prefix. HTTP headers that start with
-// 'Grpc-Metadata-' are mapped to gRPC metadata after removing prefix 'Grpc-Metadata-'.
+// keys (as specified by the IANA) to gRPC context with grpcgateway- prefix. e.g:Accept、Cookie、Host. If you want
+// to know support permanent HTTP header enum value,you can view isPermanentHTTPHeader function.
+// HTTP headers that start with 'Grpc-Metadata-' are mapped to gRPC metadata after removing prefix 'Grpc-Metadata-'.
 func DefaultHeaderMatcher(key string) (string, bool) {
 	switch key = textproto.CanonicalMIMEHeaderKey(key); {
 	case isPermanentHTTPHeader(key):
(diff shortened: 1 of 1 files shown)


runtime/mux.go
```diff
@@ -101,8 +101,9 @@ func SetQueryParameterParser(queryParameterParser QueryParameterParser) ServeMux
 type HeaderMatcherFunc func(string) (string, bool)
 
 // DefaultHeaderMatcher is used to pass http request headers to/from gRPC context. This adds permanent HTTP header
-// keys (as specified by the IANA) to gRPC context with grpcgateway- prefix. HTTP headers that start with
-// 'Grpc-Metadata-' are mapped to gRPC metadata after removing prefix 'Grpc-Metadata-'.
+// keys (as specified by the IANA) to gRPC context with grpcgateway- prefix. e.g:Accept、Cookie、Host. If you want
+// to know support permanent HTTP header enum value,you can view isPermanentHTTPHeader function.
+// HTTP headers that start with 'Grpc-Metadata-' are mapped to gRPC metadata after removing prefix 'Grpc-Metadata-'.
```
johanbrandhorst, line 106:
```suggestion
// keys (as specified by the IANA, e.g: Accept, Cookie, Host) to the gRPC metadata with the grpcgateway- prefix. // If you want to know which headers are considered permanent, you can view the isPermanentHTTPHeader function.
// HTTP headers that start with 'Grpc-Metadata-' are mapped to gRPC metadata after removing the prefix 'Grpc-Metadata-'.
// Other headers are not added to the gRPC metadata.
```

johanbrandhorst (changes requested):

johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution!

PR #3117, opened 2023-01-10
Rename root BUILD to BUILD.bazel for consistency

rsepassi:
In addition to the nicety of consistent naming, it also plays better with `bazel-gazelle`. Gazelle infers the build file name from the root `BUILD` and so ignores `BUILD.bazel` in the subdirs, which forces specifying `build_file_name = "BUILD.bazel",` in the `go_repository` rule. With consistent naming, `build_file_name` doesn't need to be specified, and heads need not be scratched when things don't work out of the box.


johanbrandhorst (approved):
Sounds reasonable to me!

johanbrandhorst:
Thanks for your contribution!

PR #3167, opened 2023-02-01
fix: enum should be int when enums_as_int = true

### LINKED ISSUE #3162: Swagger docs validate failed when enums_as_ints=true
## 🐛 Bug Report

(A clear and concise description of what the bug is.)

Related issue: https://github.com/go-swagger/go-swagger/issues/2890

## To Reproduce

(Write your steps here:)

Details can be found in this pr https://github.com/hnlq715/proto/pull/3/files

## Expected behavior

Pass validation.
(Write what you thought would happen.)

## Actual Behavior
```
The swagger spec at "gen/go/api/helloworld/v1/helloworld.swagger.json" is invalid against swagger specification 2.0.
See errors below:
- in operation "GreeterService_SayHello", default value in response 200 does not validate its schema
- 200.data.sex.default in body must be of type integer: "string"
- 200.data.sex.default in body should be one of [1 2]
- in operation "GreeterService_SayHello2", default value in response 200 does not validate its schema
- definitions.v1Sex.default in body must be of type integer: "string"
- definitions.v1Sex.default in body should be one of [1 2]
```
(Write what happened. Add screenshots, if applicable.)

## Your Environment

(Environment name, version and operating system.)

hnlq715:

Fixes #3162

### DIFF
--- a/protoc-gen-openapiv2/internal/genopenapi/template.go
+++ b/protoc-gen-openapiv2/internal/genopenapi/template.go
@@ -1168,7 +1168,6 @@ func renderServices(services []*descriptor.Service, paths openapiPathsObject, re
 					if parameter.IsRepeated() {
 						core := schemaCore{Type: paramType, Format: paramFormat}
 						if parameter.IsEnum() {
-							var s interface{}
 							core.Enum = enumNames
 							enumNames = nil
 						}
--- a/protoc-gen-openapiv2/internal/genopenapi/template.go
+++ b/protoc-gen-openapiv2/internal/genopenapi/template.go
@@ -117,10 +117,15 @@ func listEnumNames(reg *descriptor.Registry, enum *descriptor.Enum) (names []str
 		}
 		names = append(names, value.GetName())
 	}
-	return names
+
+	if len(names) > 0 {
+		return names
+	}
+
+	return nil
 }
 
-func listEnumNumbers(reg *descriptor.Registry, enum *descriptor.Enum) (numbers []string) {
+func listEnumNumbers(reg *descriptor.Registry, enum *descriptor.Enum) (numbers []int) {
 	for _, value := range enum.GetValue() {
 		if reg.GetOmitEnumDefaultValue() && value.GetNumber() == 0 {
 			continue
@@ -128,12 +133,17 @@ func listEnumNumbers(reg *descriptor.Registry, enum *descriptor.Enum) (numbers [
 		if !isVisible(getEnumValueVisibilityOption(value), reg) {
 			continue
 		}
-		numbers = append(numbers, strconv.Itoa(int(value.GetNumber())))
+		numbers = append(numbers, int(value.GetNumber()))
 	}
-	return
+
+	if len(numbers) > 0 {
+		return numbers
+	}
+
+	return nil
 }
 
-func getEnumDefault(reg *descriptor.Registry, enum *descriptor.Enum) string {
+func getEnumDefault(reg *descriptor.Registry, enum *descriptor.Enum) interface{} {
 	if !reg.GetOmitEnumDefaultValue() {
 		for _, value := range enum.GetValue() {
 			if value.GetNumber() == 0 {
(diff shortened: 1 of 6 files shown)


hnlq715:
With this change, everything works fine.

https://github.com/hnlq715/proto/tree/feat/enums-as-ints

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -117,31 +117,52 @@ func listEnumNames(reg *descriptor.Registry, enum *descriptor.Enum) (names []str
 		}
 		names = append(names, value.GetName())
 	}
-	return names
+
+	if len(names) > 0 {
+		return names
+	}
+
+	return nil
```
johanbrandhorst, line 125: I don't think this needs to change

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -117,31 +117,52 @@ func listEnumNames(reg *descriptor.Registry, enum *descriptor.Enum) (names []str
...
-		numbers = append(numbers, strconv.Itoa(int(value.GetNumber())))
+		numbers = append(numbers, int(value.GetNumber()))
 	}
-	return
+
+	if len(numbers) > 0 {
+		return numbers
+	}
+
+	return nil
```
johanbrandhorst, line 143: Please use the same format as the other function
```suggestion
	return numbers
```

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -1147,7 +1174,7 @@ func renderServices(services []*descriptor.Service, paths openapiPathsObject, re
 					if parameter.IsRepeated() {
 						core := schemaCore{Type: paramType, Format: paramFormat}
 						if parameter.IsEnum() {
-							var s []string
+							var s interface{}
 							core.Enum = enumNames
 							enumNames = s
```
johanbrandhorst, line 1179: I'm confused by this code, but I think we can just do this assignment directly instead
```suggestion
							enumNames = nil
```

protoc-gen-openapiv2/internal/genopenapi/template_test.go
```diff
@@ -8583,6 +8598,10 @@ func TestArrayMessageItemsType(t *testing.T) {
 
 		t.Errorf("applyTemplate(%#v).%s = %v want to be %v", file, name, is, want)
 	}
+	d1, _ := json.Marshal(result.Definitions)
+	d2, _ := json.Marshal(expect)
+	fmt.Println(string(d1))
+	fmt.Println(string(d2))
```
johanbrandhorst, line 8604: I think these are for debugging? Can we remove them?

johanbrandhorst (changes requested):
I think you'll need to regenerate the files, please check https://github.com/grpc-ecosystem/grpc-gateway/blob/main/CONTRIBUTING.md#i-want-to-regenerate-the-files-after-making-changes

protoc-gen-openapiv2/internal/genopenapi/template.go
hnlq715: Generated openapi docs would contains `"enum": []` when remove this change, which is not the expected behavior before.

johanbrandhorst: 🤔 that's wild, but okay.

protoc-gen-openapiv2/internal/genopenapi/template_test.go
hnlq715: done

protoc-gen-openapiv2/internal/genopenapi/template.go
hnlq715: Nice catch, let's fix this.

hnlq715:
> I think you'll need to regenerate the files, please check https://github.com/grpc-ecosystem/grpc-gateway/blob/main/CONTRIBUTING.md#i-want-to-regenerate-the-files-after-making-changes

I regenerate the files, but nothing changed, seems like there is no use case about `enums_as_ints` in examples, Maybe we should add a case for this?

johanbrandhorst:
> > I think you'll need to regenerate the files, please check https://github.com/grpc-ecosystem/grpc-gateway/blob/main/CONTRIBUTING.md#i-want-to-regenerate-the-files-after-making-changes
> 
> I regenerate the files, but nothing changed, seems like there is no use case about `enums_as_ints` in examples, Maybe we should add a case for this?

That's a great idea, I think you could try picking one of the existing files and adding a new `buf.gen.yaml` file for it.

spatecon (approved):

hnlq715:
@johanbrandhorst Sorry for the late! 
I have created a new `buf.gen.yaml` for the `enums_as_ints` case in exmaples.

Let me know if anything I can do, thank you in advance :-)

johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution!

PR #3187, opened 2023-02-14
feat(openapiv2): add include_without_visibility flag

black-06:
I have a lot of proto files, but I only want  to generate api for some methods.
So it would be useful if user could control the behavior of the elements without visibility annotations.

YES

- add `include_without_visibility` flag, defaule `true` which will include elements without visibility annotations, just like now.


johanbrandhorst:
Hi, thanks for your PR! I generally prefer to have a discussion about any new features before implementing them. I'm not sure I understand your use case. Is it correct to say that you have a lot of protobuf files, within which you only want some messages/rpcs etc to be rendered, and while it could be solved by annotating them as internal, etc, you'd like that to be implicit by their lack of annotation, to avoid having to annotate all the messages and RPCs?

black-06:
> Hi, thanks for your PR! I generally prefer to have a discussion about any new features before implementing them. I'm not sure I understand your use case. Is it correct to say that you have a lot of protobuf files, within which you only want some messages/rpcs etc to be rendered, and while it could be solved by annotating them as internal, etc, you'd like that to be implicit by their lack of annotation, to avoid having to annotate all the messages and RPCs?

Yes, but in fact it doesn't work well., so I marked it as a draft.
It is more complicated than I think, so you are right, we need to discuss it.

johanbrandhorst:
If we agree on the use case this solves, then we have to decide whether the problem this solves is worth the cost of this feature, which I'm not yet convinced of. Every new flag has to be considered together with all existing flags, so it gets more complex the more we add. This is solving a scalability problem with the visibility options, but the workaround is to actually annotate your messages and rpcs. I like to ask that at least two independent users request the same feature before we consider implementing it. With that in mind, would you mind just creating an issue instead, and we can have the discussion about the _merit_ of this feature before we start worrying about the implementation?

black-06:
> If we agree on the use case this solves, then we have to decide whether the problem this solves is worth the cost of this feature, which I'm not yet convinced of. Every new flag has to be considered together with all existing flags, so it gets more complex the more we add. This is solving a scalability problem with the visibility options, but the workaround is to actually annotate your messages and rpcs. I like to ask that at least two independent users request the same feature before we consider implementing it. With that in mind, would you mind just creating an issue instead, and we can have the discussion about the _merit_ of this feature before we start worrying about the implementation?

of course not, I will create an issue instead.

PR #3190, opened 2023-02-14
gateway: Dial -> DialContext

torkelrogstad:

Yes

Use `DialContext` instead of `Dial` in the gateway template. This allows service registration from endpoints to work properly with context cancellations if the endpoint can't be reached.


johanbrandhorst (approved):
Can't believe we didn't already do this 🤔. Thank you!

PR #3206, opened 2023-02-23
Add correct comment, remove redundant vars and error, use short form

sashamelentyev:


johanbrandhorst (approved):
LGTM, though technically removing the exported error is a breaking change, I think we already silently broke any users of this error when moving to `MalformedSequenceError`.

Issue #1680, opened 2020-09-21
Unsupported HEAD call to known URL does not return error response body.


maja42:
## 🐛 Bug Report

When performing a HTTP call using an unsupported method, for example when making a POST request to a URL that only supports GET, I get a http-501-error-response from the gateway. The response body contains the grpc error code 12 ("Unimplemented") with the message "Method Not Allowed".

When performing a HEAD-call to the same get-only-URL, I expect the same error response. 
What happens is that the http request fails with 501 (which is correct), but the response body stays empty. 
According to the documentation, the HEAD call should not have any special behaviour, so I expect the body to contain the same error details. An error-response-body should never be empty.
Docs: https://cloud.google.com/service-infrastructure/docs/service-management/reference/rpc/google.api#special-notes

## To Reproduce

1. Define a gRPC endpoint with only a GET-http-mapping (or any other method).
2. Generate and start the grpc-gateway
3. Perform a http HEAD-call to the same URL.
4. The response (status 501) will have an empty body.

## Your Environment

Im using `github.com/grpc-ecosystem/grpc-gateway/v2 v2.0.0-beta.4`

johanbrandhorst:
Hi again Maja, thanks for the bug report! I think the logic for this is [here](https://github.com/grpc-ecosystem/grpc-gateway/blob/v2/runtime/mux.go#L262), [here](https://github.com/grpc-ecosystem/grpc-gateway/blob/v2/runtime/errors.go#L148) and [here](https://github.com/grpc-ecosystem/grpc-gateway/blob/v2/runtime/errors.go#L78). I don't see any special handling for POST vs HEAD though, so I'm not sure what's going on here. Would you be interested in contributing a fix for this?

maja42:
I can take a look at it, but since I'm pretty busy right now I'm not sure if and when I have spare time for it.

adambabik:
This behaviour might be included deeper (in the HTTP server implementation, for instance) or the client you use discards body. Discarding body in HEAD requests is actually correct:

> A response to a HEAD method should not have a body. If it has one anyway, that body must be ignored: any entity headers that might describe the erroneous body are instead assumed to describe the response which a similar GET request would have received.

From [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD).

PR #3228, opened 2023-03-07
Support for HTTP HEAD and OPTIONS

emilhem:
Adds logic for handling HEAD and OPTIONS requests in the mux

To get responses for HEAD requests
```go
runtime.WithRespondOnHttpHead()
```

To get responses for OPTIONS requests
```go
runtime.WithRespondOnHttpOptions()
```

Maybe a fix for #1680

Adds options to have a simple HTTP HEAD and HTTP OPTIONS response.

#### Other comments
It does not interfere with old behavior where the user could implement their custom HEAD or OPTIONS functionality.

There is a TODO in where you potentially could parse the handlers to get the correct list of http methods supported.


johanbrandhorst:
Hi, thank you for your PR! Generally, for new features like this, I much prefer to see an issue created first, to describe the use case. There we can decide whether it is something we want to implement, which can save everyone time if we decide not to, or find a workaround for it.

In this case, I'm inclined to say that this functionality does not belong in the gateway. For your use case, does it not suffice to simply wrap the gateway with another `http.Handler` that handles these interactions for you? It should be as simple as

```go
func handleHeadOrOptions(next http.Handler) http.Handler {
	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		if r.Method == http.MethodHead && s.respondOnHttpHead {
			w.Header().Set("Content-Type", "application/json")
			w.WriteHeader(http.StatusNoContent)
			return
		}
		if r.Method == http.MethodOptions && s.respondOnHttpOptions {
			// TODO: filter out methods that are not supported by this path
			w.Header().Set("Allow", "GET, POST, PUT, PATCH, DELETE, HEAD, OPTIONS")
			w.WriteHeader(http.StatusOK)
			return
		}
		next.ServeHTTP(w, r)
	})
}

...

wrapped := handleHeadAndOptions(mux)
http.ListenAndServe("", wrapped)
```

Or something like that. What do you think?

PR #3247, opened 2023-03-16
add macros example to readme

### LINKED ISSUE #3198: Use of macros for multipart-formdata bytes fields [enchancement]
Hi! First of all, thanks for creating and supporting grpc-gateway.
We at neiro.ai use it for our ML APIs that process media files such as audio and video. For us, grpc-gateway was an easy way to support both rest and grpc in our API. However, almost every request has a media file in bytes. There must be a more convenient way to handle files then explicitly converting them to base64 in REST API. There is a section in the documentation about how to do binary upload and it looks like "handle raw data by yourself", but we wanted to keep our data structure and use the advantages of your proxy.

Right now docs do not give you much information about how to handle this. Grpc-gateway doesn't support it too.

We solved this issue with ad-hoc macros processing. How it works: you send multipart/form-data with your request and media files. In your request body, you mark media files with macroses. Then, our middleware recursively parses json and looks for a place to expand macros with the file content. Then it goes the standard way through grpc-gateway proxy. Yes, it may be extra latency, but for manual testing, it was a must-have for us.

Example (out speech recognition request):
```jsx
{
    "token": "",
    "audio": {
        "data": "$speech"
    }
}
```

```bash
curl $HOST/v1/asr -X POST 
                -F 'data=@json/asr.json' -F '$speech=@voice.wav'
```

Technically, what we do inside is filter multipart/form-data requests, get their json body in reserved field `data`, and then create standard json request with substitutions

```golang
func EncodeAndSubstitute(jsonData map[string]interface{}, r *http.Request, res *map[string]interface{}) error {
	for name, val := range jsonData {
		rec_map, ismap := val.(map[string]interface{})
		if ismap {
			(*res)[name] = map[string]interface{}{}
			rec_res, _ := (*res)[name].(map[string]interface{})
			err := EncodeAndSubstitute(rec_map, r, &rec_res)
			if err != nil {
				return err
			}
		} else if valStr, ok := val.(string); ok && strings.HasPrefix(valStr, "$") {
			data, err := GetFormFile(r, valStr)
			if err != nil {
				return fmt.Errorf("can not get file %s", val.(string))
			}
			(*res)[name] = base64.StdEncoding.EncodeToString(data)
		} else {
			(*res)[name] = val
		}
	}
	return nil
}
```

It seems like this approach can be used in other projects as well. Maybe we should add another section in https://grpc-ecosystem.github.io/grpc-gateway/docs/mapping/binary_file_uploads/ or another page in the documentation? Maybe it can be implemented as part of grpc-gateway? What do you think?

KiK0S:
This pull request adds an example of the gRPC-gateway service in the documentation (there can't be too much!).  The provided service shows one possible approach to fit media files in the request. It can be useful either for those who want to write their own middleware and want ideas or for those who look for a way to send media files with other data in one request.

Fixes #3198

Yes.

Added an example of repository with:

* Macros for multipart/form-data requests
* Shared wrapper for common logic such as latency measurements

### DIFF
--- a/README.md
+++ b/README.md
@@ -556,7 +556,7 @@ gRPC-Gateway, and a gRPC server, see
 [this example by CoreOS](https://github.com/philips/grpc-gateway-example/blob/master/cmd/serve.go)
 (and its accompanying [blog post](https://web.archive.org/web/20201112010739/https://coreos.com/blog/grpc-protobufs-swagger.html)).
 
-[This example by neiro.ai](https://github.com/mynalabsai/grpc_gateway_media_example) (and its accompanying [blog post](https://medium.com/neiro-ai/grpc-gateway-for-media-api-by-neiro-9033caab12c8)) shows how mediafiles can be integrated in rpc message by middleware for `multipart/form-data` and macros.
+[This example by neiro.ai](https://github.com/mynalabsai/grpc_gateway_media_example) (and its accompanying [blog post](https://medium.com/neiro-ai/grpc-gateway-for-media-api-by-neiro-9033caab12c8)) shows how mediafiles using `multipart/form-data` can be integrated into rpc messages using a middleware.
 
 ## Features
 
--- a/README.md
+++ b/README.md
@@ -556,6 +556,8 @@ gRPC-Gateway, and a gRPC server, see
 [this example by CoreOS](https://github.com/philips/grpc-gateway-example/blob/master/cmd/serve.go)
 (and its accompanying [blog post](https://web.archive.org/web/20201112010739/https://coreos.com/blog/grpc-protobufs-swagger.html)).
 
+[This example by neiro.ai](https://github.com/mynalabsai/grpc_gateway_media_example) (and its accompanying [blog post](https://medium.com/neiro-ai/grpc-gateway-for-media-api-by-neiro-9033caab12c8)) shows how mediafiles can be integrated in rpc message by middleware for `multipart/form-data` and macros.
+
 ## Features
 
 ### Supported


README.md
```diff
@@ -556,6 +556,8 @@ gRPC-Gateway, and a gRPC server, see
 [this example by CoreOS](https://github.com/philips/grpc-gateway-example/blob/master/cmd/serve.go)
 (and its accompanying [blog post](https://web.archive.org/web/20201112010739/https://coreos.com/blog/grpc-protobufs-swagger.html)).
 
+[This example by neiro.ai](https://github.com/mynalabsai/grpc_gateway_media_example) (and its accompanying [blog post](https://medium.com/neiro-ai/grpc-gateway-for-media-api-by-neiro-9033caab12c8)) shows how mediafiles can be integrated in rpc message by middleware for `multipart/form-data` and macros.
```
johanbrandhorst, line 559:
```suggestion
[This example by neiro.ai](https://github.com/mynalabsai/grpc_gateway_media_example) (and its accompanying [blog post](https://medium.com/neiro-ai/grpc-gateway-for-media-api-by-neiro-9033caab12c8)) shows how mediafiles using `multipart/form-data` can be integrated into rpc messages using a middleware.
```

johanbrandhorst (commented):
Thanks for your PR! What do you think about the suggested change?

README.md
KiK0S: yeah, it looks nice

johanbrandhorst (approved):

PR #3252, opened 2023-03-20
New `ignore_comments` option for excluding proto comments from OpenAPI output

### LINKED ISSUE #2684: Ignore buf lint comment in swagger definition title
## 🐛 Bug Report

The definition title in swagger uses comment as title, but even if the comment is buf lint, it still uses the lint comment as title.

## To Reproduce

Given below message with inline buf ignore
```
// buf:lint:ignore FIELD_LOWER_SNAKE_CASE
message GetDataResponse {
}
```

## Expected behavior

The generated swagger title should ignore the buf:lint and use `GetDataResponse`.

## Actual Behavior

The title is `// buf:lint:ignore FIELD_LOWER_SNAKE_CASE`

keyz:
Fixes #2684

Yes

It's pretty common to have proto comments that are irrelevant to OpenAPI output. For instance, if you have a `// buf:lint:ignore` comment, the only way to exclude that today is to manually add an override via `grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field` per https://github.com/grpc-ecosystem/grpc-gateway/issues/2684#issuecomment-1126475877.

In this PR I'm adding a new config option: `ignore_comments`. It's disabled by default and only affects the OpenAPI output; when you enable the option, all proto comments are excluded from the output. With the new option on, annotations can still be explicitly added via `grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field`.

Let me know if you have any feedback!


protoc-gen-openapiv2/main.go
```diff
@@ -30,6 +30,7 @@ var (
 	useFQNForOpenAPIName           = flag.Bool("fqn_for_openapi_name", false, "if set, the object's OpenAPI names will use the fully qualified names from the proto definition (ie my.package.MyMessage.MyInnerMessage). DEPRECATED: prefer `openapi_naming_strategy=fqn`")
 	openAPINamingStrategy          = flag.String("openapi_naming_strategy", "", "use the given OpenAPI naming strategy. Allowed values are `legacy`, `fqn`, `simple`. If unset, either `legacy` or `fqn` are selected, depending on the value of the `fqn_for_openapi_name` flag")
 	useGoTemplate                  = flag.Bool("use_go_templates", false, "if set, you can use Go templates in protofile comments")
+	ignoreProtoComments            = flag.Bool("ignore_proto_comments", false, "if set, all protofile comments are excluded from output")
```
johanbrandhorst, line 33:
```suggestion
	ignoreProtoComments            = flag.Bool("ignore_comments", false, "if set, all comments are excluded from output")
```

johanbrandhorst (changes requested):
I can actually get behind this option. It's consistent and reasonably isolated from other functions. I would request three things:

1. Lets just call it `ignore_comments`.
2. Please add a check to startup that errors if both `ignore_comments` and `use_go_template` is set. These are mutually exclusive options.
3. Lets add something to the docs, specifically we usually have a section for each flag in https://github.com/grpc-ecosystem/grpc-gateway/blob/main/docs/docs/mapping/customizing_openapi_output.md.
Thank you!

keyz:
@johanbrandhorst sounds good -- just updated the PR. Thank you!

johanbrandhorst (approved):
LGTM, thank you!

Issue #3266, opened 2023-03-28
The PathLengthFallback option picks fallback methods non-deterministically


nacx:
## 🐛 Bug Report

When the `disablePathLengthFallback` option is enabled (the default behavior), the handler selection is non-deterministic, leading to unexpected (and potentially dangerous behavior, especially if DELETE methods are present in the handler.

The following code in `mux.go`, iterates over the map of handlers, but map iteration order is non-deterministic. The method assumes a fallback to POST, but if a DELETE method was defined for the same request path, it could be selected by mistake.
https://github.com/grpc-ecosystem/grpc-gateway/blob/836817217b6fe57bd95421b5fe7589c542e549d8/runtime/mux.go#L392-L441

This could be an important issue as a deletion handler could be called by mistake, potentially destroying data.

## To Reproduce

Say the following URIs and methods are valid methods of an API:
```
GET /resources/foo
DELETE /resources/foo
```
And someone by mistake runs a request to: `POST /resources/foo`.
If the fallback mechanism is enabled, it will fallback either to the GET or the DELETE handler (because if the POST request has a body the `isPathLengthFallback` method would return `true`). If the DELETE method is selected, a DELETE operation could be triggered by mistake.

## Expected behavior

The fallback mechanism should only fallback from GET to POST, but not to other HTTP verbs.

## Actual Behavior

The fallback mechanism fallbacks to any other method in a non-deterministic way.

johanbrandhorst:
Hi, thanks for raising this issue. That's definitely a bug (whoops!). Would you be interested in implementing a fix? At the very least we should make this deterministic, but you're right that it should probably also prioritize GETs.

nacx:
Sure! I've raised a small PR changing the logic to only allow fallback for POST -> GET, as looking at the original feature discussions, I think that's the only case this code was targeting.

Issue #447, opened 2017-08-27
"GET" route matches both GET & POST requests?


uynap:
I found a "get" route will match both HTTP GET and HTTP POST requests. For example:

```
rpc aaa(User) returns (User) {
  option (google.api.http) = {
  get: "/api/v1/aaa"
};
```

It matches both `curl -v -X GET -k https://127.0.0.1/api/v1/aaa` and `curl -v -X POST -k https://127.0.0.1/api/v1/aaa`.

I was wondering if it's possible to strictly match all routes including methods?

achew22:
Hrm, that's really unfortunate. Do you think you could put together a PR with a test case in our [integration tests](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/examples/integration_test.go) to demonstrate this error?

UladzimirTrehubenka:
From https://stackoverflow.com/questions/45901389/why-golang-grpc-gateways-get-route-matches-post-request

By spec, it is allowed to encode GET request into a POST method with Content-Type: application/x-www-url-encoded.

See https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/mux.go#L254

So the request routing table in grpc-gateway tries to fallback from POST method to GET when the Content-Type of your request is application/x-www-url-encoded.

From https://groups.google.com/d/msg/grpc-io/Xqx80hG0D44/1gwmwBcnNScJ

```
(Note) URL has a length limitation (go/longer-urls). It's enfoced by some browsers and
proxies. If your GET request exceeds the limitation, browser may reject to send them.
You may change to use a POST request with content type x-www-form-urlencoded instead.
```

Another one example:

```
rpc Read(ReadReq) returns (ReadResp) {
  option (google.api.http) = {
  get: "/api/v1/user/{id}"
};

rpc Create(CreateReq) returns (CreateResp) {
  option (google.api.http) = {
  post: "/api/v1/user"
};
```
curl -X POST http://127.0.0.1:8040/api/v1/user/127 -d '{}'
200 OK ("Content-Type: application/x-www-url-encoded" is in header and POST becomes GET, actually we have no route POST "/api/v1/user/{id}" at all)

curl -X POST http://127.0.0.1:8040/api/v1/user/127
501 NOT_IMPLEMENTED (expected behaviour)

The question is - how to disable this behaviour? This is weird that any "Content-Type" causes expected error 501 NOT_IMPLEMENTED except "application/x-www-url-encoded".

[...]

PR #3272, opened 2023-04-03
Only allow POST -> GET path length fallback

nacx:

Fixes https://github.com/grpc-ecosystem/grpc-gateway/issues/3266

Yes

This PR changes the mux logic that implements the HTTP length fallback to prevent it from falling back to potentially dangerous methods. With these changes, the fallback mechanism is only allowed for `POST -> GET` instead of allowing it for any defined method.

#### Other comments

The discussion about this mechanism can be found at https://github.com/grpc-ecosystem/grpc-gateway/issues/447, and it is specifically designed to target `POST -> GET` fallbacks. However, the current code could fallback to other methods like DELETE, which could be potentially dangerous.

### DIFF
--- a/runtime/mux.go
+++ b/runtime/mux.go
@@ -389,8 +389,12 @@ func (s *ServeMux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	// lookup other methods to handle fallback from GET to POST and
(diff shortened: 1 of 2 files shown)


johanbrandhorst (approved):
LGTM!

johanbrandhorst:
Thanks for your contribution!

PR #3292, opened 2023-04-19
Add the ability to configure HTTP query parameters

### LINKED ISSUE #3290: support custom HTTP Query URI names to make a nicer UX for deeply nested messages
## 🚀 Feature

`protobuf` prefers composition of messages instead of inheritance. Its very common for HTTP GET requests to take common nested filter messages which then leads to a verbose URI.

e.g. if you use a common `Filter` object (for search text, sort, pagination etc) in `SomeRequest ` then your HTTP GET URI would be something like `?filter.query=foo&filter.sort=ASC&filter.page=1`

in protobuf we can already configure json marshalling names via the `json` tag. It would be nice to be able to support a `query` tag for providing a custom HTTP GET query URI parameter name used when exposing the service over HTTP GET.

We can then leave the POST / proto / JSON encoding alone but have nicer GET URIs if required. e.g. `?q=foo&sort=ASC&page=1`

I've an experimental PR on the way soon that implements this to see what folks think...

jstrachan:
in protobuf we can already configure json marshaling names via the `json` tag. This PR adds a similar capability to customize URL query names via a `query` tag to either:

* use a short alias for HTTP GET URL query arguments
* use a blank name to avoid a prefix of nested objects

e.g. if you use a common `Filter` object in `SomeRequest ` then your GET URI may be of the form: `?filter.query=foo&filter.sort=ASC&filter.page=1`

By adding the tag `query: ""` to your `filter` property you can then omit the redundant `filter.` prefix on the query URL values for properties in the filter message - e.g. `?q=foo&sort=ASC&page=1`

Fixes #3290

Yes

This PR adds the detection of `query` tags on Protobuf structs which if populated will use that query name to parse HTTP GET URI names. The usual more verbose URIs are still supported


jstrachan:
closing as I finally figured out the git config magic to get past the cla/google check - sorry for the noise!

PR #3299, opened 2023-04-25
gofmt and fix some comments

cuishuang:
gofmt and fix some comments


johanbrandhorst:
These files are both generated, you'll have to raise issues with the generators.

PR #3345, opened 2023-06-06
Rename LICENSE.txt to LICENSE

pgmitche:
buf cli only accepts license files as `LICENSE` sans extension.

https://github.com/bufbuild/modules/pull/88


johanbrandhorst (approved):
Hopefully this won't break anyone 😂

PR #3364, opened 2023-06-16
Improve AIP support in field_configuration.path_param_name containing

same-id:
When adding "=" to path_param_name:

```
  field_configuration: {
    path_param_name: "item=items/*"
  }
}
```

Like Google AIP APIs like to do we get the following output.

```
"/v1/{item=items/*}": {
  "get": {
    "summary": "GetItem",
    "parameters": [
      {
        "name": "item=items/*",
      }
    ]
  }
}
```

But we actually want something like this, which looks better in docs:

```
"/v1/{item=items/*}": {
  "get": {
    "summary": "GetItem",
    "parameters": [
      {
        "name": "item",
      }
    ]
  }
}
```

Not adding a feature flag since I assume no one uses "=" in path param overrides today since it doesn't look that good.


johanbrandhorst (commented):
Thanks for your PR! I'm a little confused, can't you just set the value to `item` in your field configuration? There's nothing parsing this field (this PR hardly does it either), so why use the full definition?

same-id:
Hi,

Some minor thing first:
I updated the commit message / PR comment from
`item=items/*` to `name=items/*`


To your question,

Before this commit I had just `name` - didn't even have to add an annotation (like you suggest).
It looks like this:

image
image

Notice that `name` follows https://google.aip.dev/122

But it doesn't look that great in the url: `v1/{name}` - since it hides the `vaults/` sub-part.

If you look at how Google shows this:
image

A little bit nicer.

If I use `path_param_name` it changes the param name in both places:

```
  field_configuration: {
    path_param_name: "name=items/*"
  }
}
```

I get

image

Which is nice.
But then I also get this:
image
Which looks meh

This PR simply makes it look better in the openapi.yml and then the subsequent generated docs:
image
image

I think this could be nicer for other peeps developing according to Google's AIP.

johanbrandhorst (approved):
Thanks for the illustrative example, one of my worries here was that introducing any sort of parsing (which this does) might mean we should just properly parse the whole thing, but I'm satisfied that this should be a pretty harmless change. That said, this is technically breaking, since users may be using `=` for other things in this field today, so if we get any reports of breakages we may have to revert this and put it behind a flag.

same-id:
Exactly my thoughts.
I don't think this will break any users' current (fingers crossed - who uses `=` anyway :-) )

Maybe was worth documenting this somehow, or add a test.
Anyway, glad it got it.

PR #3366, opened 2023-06-18
Hide comments on omitted enum default values

same-id:
Omitted enum default value should not show in comments.


johanbrandhorst (approved):
LGTM, but have you got an example of what this looks like?

same-id:
I set "omit default values"

Before the PR:
image

After the PR:
image

johanbrandhorst:
Thanks!

PR #3413, opened 2023-07-21
bazel: bump Bazel repositories versions

### LINKED ISSUE #3407: rules_go 0.41.0 upgrade breaking on missing go_googleapis
## 🐛 Bug Report

I upgraded to rules_go to 0.41.0 which removed `@go_googleapis` as a provided repo. However, `grpc-gateway` still references `@go_googleapis` internally so I had to enable `build_file_generation` to regenerate grpc-gateway's BUILD files as well as add resolve clauses for proto references.

While diagnosing the issue, I noticed that grpc gateway's renovate attempted to upgrade to rules_go 0.41.0 but did so in an inconsistent/invalid manner such that the upgrade doesn't take effect.
https://github.com/grpc-ecosystem/grpc-gateway/pull/3397 has the 0.41.0 version only on the fallback URL and doesn't apply to the file archive either.

## To Reproduce

Upgrade to rules_go 0.41.0
Upgrade to gazelle 0.32.0
Run `bazel build @com_github_grpc_ecosystem_grpc_gateway_v2//internal/descriptor/apiconfig:apiconfig_proto`

## Expected behavior

Bazel build succeeds

## Actual Behavior

The bazel build will fail because `@go_googleapis` is no longer provided and gazelle searches for `google/api` in the local workspace.

```
external/com_github_grpc_ecosystem_grpc_gateway_v2/internal/descriptor/apiconfig/BUILD.bazel:7:14: no such package '@com_github_grpc_ecosystem_grpc_gateway_v2//google/api': BUILD file not found in directory 'google/api' of external repository @com_github_grpc_ecosystem_grpc_gateway_v2. Add a BUILD file to a directory to mark it as a package. and referenced by '@com_github_grpc_ecosystem_grpc_gateway_v2//internal/descriptor/apiconfig:apiconfig_proto'
```

## Your Environment

OSX Ventura 13.4.1

adambabik:

Hopefully fixes #3407

I had a problem to generate Bazel build files from main and noticed #3407. I bumped both `rules_go` and `gazelle` which resulted in getting rid of `@go_googleapis` references as expected.

One more problem I had after bumping `rules_proto`, which was also outdated, was that Bazel used older C++ standard by default. I enforced the C++14 with `BAZEL_CXXOPTS="-std=c++14" bazel run :buildifier`, however, there might be a better way to achieve this.


adambabik:
I moved it to drafts as there are more problems to fix.

BUILD.bazel
```diff
@@ -19,6 +19,18 @@ buildifier(
 # gazelle:go_proto_compilers //:go_apiv2
 # gazelle:go_grpc_compilers //:go_apiv2, //:go_grpc
 # gazelle:go_naming_convention import_alias
+# gazelle:resolve proto proto google/api/annotations.proto @googleapis//google/api:annotations_proto
```
adambabik, line 22: I am not sure if this is the best place for these directives.

CONTRIBUTING.md
```diff
@@ -19,7 +19,7 @@ docker run -itv $(pwd):/grpc-gateway -w /grpc-gateway --entrypoint /bin/bash --r
     ghcr.io/grpc-ecosystem/grpc-gateway/build-env:1.19 -c '\
         bazel run :gazelle -- update-repos -from_file=go.mod -to_macro=repositories.bzl%go_repositories && \
         bazel run :gazelle && \
-        bazel run :buildifier'
+        BAZEL_CXXOPTS="-std=c++14" bazel run :buildifier'
```
adambabik, line 22: Alternatively, we could put `.bazelrc` in the project root. I found that in [`ci.yml`](https://github.com/grpc-ecosystem/grpc-gateway/blob/dd7c267de09fe2bb44ddd5e8214ae9a46527d6ef/.github/workflows/ci.yml#L79) we are setting exactly the same env.

BUILD.bazel
johanbrandhorst: I will defer to @achew22 on this

CONTRIBUTING.md
johanbrandhorst: We have various environment specific options like the CPU and memory limiters that shouldn't go in `.bazelrc`, but this option does seem like a reasonable thing to put in a repo wide `.bazelrc`.

johanbrandhorst (approved):
🙏🏻 thank you for this. I will give Andrew a chance to look at this before merging, but it LGTM.

BUILD.bazel
achew22: I think this is the correct place for them

CONTRIBUTING.md
achew22: I would probably drop it in the `.bazelrc`, 

```
build --cxxopt='-std=c++14'
```

but I have a pathological aversion to environment variables.

WORKSPACE
```diff
@@ -18,6 +18,21 @@ http_archive(
...
+    strip_prefix = "googleapis-64926d52febbf298cb82a8f472ade4a3969ba922",
+    urls = [
+        "https://github.com/googleapis/googleapis/archive/64926d52febbf298cb82a8f472ade4a3969ba922.zip",
+    ],
+)
+
+load("@googleapis//:repository_rules.bzl", "switched_rules_by_language")
+
+switched_rules_by_language(
+    name = "com_google_googleapis_imports",
```
achew22, line 33: Don't you need to set `go = true`?

https://github.com/googleapis/googleapis/blob/master/repository_rules.bzl#L87

achew22 (approved):

WORKSPACE
adambabik: It seems we don't need this, because we use `@googleapis` only in `proto_library`. Go deps are resolved to `@org_golang_google_genproto_googleapis_api`. 

I suppose that I could change:

```
# gazelle:resolve proto go google/api/annotations.proto  @org_golang_google_genproto_googleapis_api//annotations
```

to

```
# gazelle:resolve proto go google/api/annotations.proto  @googleapis//google/api:annotations_go_proto
```

then `go = True` would be necessary. But I am not sure if it gives us any benefits. `@org_golang_google_genproto_googleapis_api//annotations` was recommended in `go_rules` and `gazelle`.

PR #3415, opened 2023-07-23
Handle httpbody not having data to return

achew22:

I was testing out a case in my code where I was trying to make it fail. An abstract caller of this method was passed a `nil` that was typed properly and discovered that instead of handling the zero value properly, it panics. This should resolve that.


johanbrandhorst (approved):
Neat, you broke it!

PR #3500, opened 2023-08-15
Feature/issue 3051 swagger preserve rpc order

### LINKED ISSUE #3051: Add option to preserve RPC paths order in swagger output
## 🚀 Feature

The generated swagger files from the proto files always emits the "paths" list in alphabetical order apparently.
We need these in the same order as in the proto file. 

An option to the openapiv2 protoc plugin such as "paths_in_emit_order=true" with a default value of false would be exceedingly useful.

CemGurhan:

Fixes #3051

Yes

Currently, generated swagger files representing openapi definitions emit paths found in proto files in alphabetical order. This feature provides a new command line option called `preserve_rpc_order`. If set to true, once swagger files are generated, paths will maintain the order found in proto files. This should work for both json and yaml formats. Path ordering is also preserved in the following scenarios (as specified):

1. When using additional bindings, paths should preserve this ordering within an RPC
2. When using multiple services, paths should preserve the ordering between RPCs in the whole protobuf file
3. When merging protobuf files, paths should preserve the ordering specified on the command line.

Alphabetical ordering of paths was due to paths being keys to the `openapiPathsObject` map, stored inside the `openapiSwaggerObject`. Hence, upon encoding to JSON or YAML, paths were automatically ordered alphabetically. In order to resolve this issue, an ordered data structure was used instead for the `openapiPathsObject`.

#### Other comments
This is my first open source contribution, so do let me know if there's anything I can improve on!


examples/internal/helloworld/helloworld.swagger.json
```diff
@@ -15,918 +15,948 @@
...
-      "get": {
-        "operationId": "Greeter_SayHello5",
-        "responses": {
-          "200": {
-            "description": "A successful response.",
-            "schema": {
-              "$ref": "#/definitions/helloworldHelloReply"
+  "paths": [
+    {
+      "Path": "/say/boolval/{boolVal}",
```
johanbrandhorst, line 20: This doesn't look quite right, I think we want the structure to remain the same, right? We still want this to be a key, not a value.

johanbrandhorst (changes requested):
Hi Cem, thanks a lot for taking this on. I think there's a fundamental problem with the change in that it doesn't output the correct swagger structure anymore. Instead of

```json
 "paths": {
    "/v1/example/a_bit_of_everything": {
      "get": {
```

It's now doing

```json
  "paths": [
    {
      "Path": "/v1/example/a_bit_of_everything",
      "PathItemObject": {
        "get": {
```

I think you'll need to go back to the structure and make sure it produces the correct format. Let me know if you need me to dig deeper.

examples/internal/helloworld/helloworld.swagger.json
CemGurhan: I'm so sorry about missing this, this was a silly mistake on my part. I've pushed a new commit to try and address this problem. I've implemented some custom marshallers for the `openapiPathsObject` for both yaml and json to keep using the ordered data structure, but also preserve the desired swagger file outputs.

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -1661,6 +1665,33 @@ func renderServices(services []*descriptor.Service, paths openapiPathsObject, re
...
+		if p.Path == path {
+			(*paths)[i].PathItemObject = pathItemObject
+			return
+		}
+	}
+	*paths = append(*paths, pathData{
+		Path:           path,
+		PathItemObject: pathItemObject,
+	})
+}
```
johanbrandhorst, line 1693: This isn't something we need to change, but it upsets me that `renderServices` has to mutate one of the input paramers 😂.

johanbrandhorst (approved):
This is an excellent submission, sorry for the slow review!

protoc-gen-openapiv2/internal/genopenapi/generator.go
```diff
@@ -114,6 +116,58 @@ func (so openapiSwaggerObject) MarshalYAML() (interface{}, error) {
 	}, nil
```
johanbrandhorst, line 116: I think we can make this change (see GolangCI warning above)

johanbrandhorst:
There are two warnings from golangci:
``` 
Error: func `isValidPathOrder` is unused (unused)
Error: S1001: should use copy(to, from) instead of a loop (gosimple)
```

Otherwise I think this is good to go!

protoc-gen-openapiv2/internal/genopenapi/generator.go
CemGurhan: I've just made the changes listed in the GolangCI warnings, hopefully everything is okay now!

CemGurhan:
Thanks for the review Johan, let me know if there is anything more I can do/I've missed!

protoc-gen-openapiv2/internal/genopenapi/template.go
johanbrandhorst: No, as I said, this is fine. It would be nice to refactor this code in a separate PR though if you are interested. Ideally the rendering should be separate from the mutation of the inputs.

johanbrandhorst (approved):

johanbrandhorst (changes requested):
Thanks a lot for the fixes Cem. I just remembered that we should add a section to the docs about this new functionality so users can find it easily. Would you mind adding something to https://github.com/grpc-ecosystem/grpc-gateway/blob/main/docs/docs/mapping/customizing_openapi_output.md to explain what this does and how to use it?

CemGurhan:
Sure thing, will get on that now!

johanbrandhorst (approved):
Awesome, thank you!

CemGurhan:
No problem, thanks for the reviews! Will hopefully be working on more issues for gRPC gateway soon : )

eccles:
A big Thankyou to @cemGurhan and @johanbrandhorst for doing this. I have just implemented this in our system and it worked cleanly.
Thank you

PR #3546, opened 2023-08-25
Enable a few more golangci-lint linters

pkwarren:
Update grpc-gateway to enable a few more useful golangci-lint linters (and fix the issues reported by them). As part of this, no style linters were added as those can be tailored appropriately by the primary maintainers as needed (except for goimports which is already enabled by CI job).

### DIFF
--- a/protoc-gen-openapiv2/internal/genopenapi/helpers.go
+++ b/protoc-gen-openapiv2/internal/genopenapi/helpers.go
@@ -22,6 +22,7 @@ func getUniqueFields(schemaFieldsRequired []string, fieldsRequired []string) []s
 	for j, schemaFieldRequired := range schemaFieldsRequired {
 		index = nil
 		for i, fieldRequired := range fieldsRequired {
+			i := i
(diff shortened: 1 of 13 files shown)


protoc-gen-openapiv2/internal/genopenapi/helpers.go
```diff
@@ -22,6 +22,7 @@ func getUniqueFields(schemaFieldsRequired []string, fieldsRequired []string) []s
 	for j, schemaFieldRequired := range schemaFieldsRequired {
 		index = nil
 		for i, fieldRequired := range fieldsRequired {
+			i := i
```
pkwarren, line 25: This wasn't technically required in this case (because of the `break` call), however in general it is good practice to avoid these sorts of issues.

pkwarren:
These were the errors reported by the new linters before they were resolved: https://gist.github.com/pkwarren/10cc659ff06ec92644c61ee4cfbee51e

johanbrandhorst (approved):
Thanks a lot for this :)

Issue #3542, opened 2023-08-25
descriptor.File should not contain nil SourceCodeInfo


ebilling:
## 🐛 Bug Report

As of v2.17.0 we have started getting errors when running protoc-gen-openapiv2.

We get the message: “descriptor.File should not contain nil SourceCodeInfo” with no other details about the issue.  

It seems to be coming from here: https://github.com/grpc-ecosystem/grpc-gateway/blob/main/protoc-gen-openapiv2/internal/genopenapi/template.go#L2333

If I revert back to v2.16.2, everything works fine.  I can't tell if this is a problem or not.  

## To Reproduce

1. protoc --plugin protoc-gen-openapiv2=protoc-gen-openapiv2 --openapiv2_out="./openapiV2" -I=./proto:$(GOOGLE_DIR)/api:$(ANNOTATIONS_DIR) *.proto

## Expected behavior

Works like v2.16.2

Alternatively, tell us which file is having the issue.

## Actual Behavior

results in a set of these messages for each proto file:

descriptor.File should not contain nil SourceCodeInfo
descriptor.File should not contain nil SourceCodeInfo
descriptor.File should not contain nil SourceCodeInfo
descriptor.File should not contain nil SourceCodeInfo
descriptor.File should not contain nil SourceCodeInfo
descriptor.File should not contain nil SourceCodeInfo
descriptor.File should not contain nil SourceCodeInfo
descriptor.File should not contain nil SourceCodeInfo
descriptor.File should not contain nil SourceCodeInfo

## Your Environment

Mac OS - Darwin Kernel Version 22.6.0: Wed Jul  5 22:22:52 PDT 2023; root:xnu-8796.141.3~6/RELEASE_ARM64_T8103 arm64
go version go1.20.7 darwin/arm64

johanbrandhorst:
Thanks for the bug report, this is definitely concerning. It doesn't seem to be happening in our own tests, so I'm not sure what could be going on here. This code also hasn't changed in the last 6 years as far as I can tell. What version of protoc are you using? I'll have a look again to see if any changes stand out to me but at the moment I'm at a loss.

ebilling:
protoc --version
libprotoc 23.2

ebilling:
I'm forking the code to see if I can at least print the filename that is causing the problem and get more details about why this is happening.

johanbrandhorst:
[...]

PR #3573, opened 2023-09-06
wip: Fix nil SourceCodeInfo in descriptor.File

aimuz:
In the `protoComments` function, if the `file.SourceCodeInfo` is nil,
print an error message and exit with a non-zero status code.

The change fixes an issue where the `descriptor.File` should not contain
a nil `SourceCodeInfo`, providing better error handling.

Signed-off-by: aimuz <mr.imuz@gmail.com>

Update #1870
Update #3542

### Summary by CodeRabbit

"Bug Fix":
- Updated error handling in `protoc-gen-openapiv2/internal/genopenapi/template.go`. The program will now exit with a non-zero status code when the `file.SourceCodeInfo` is nil, preventing potential silent failures.
- Enhanced logging by replacing `fmt.Fprintln` with `log.Fatalf` for more informative error messages and immediate program termination on fatal conditions.

### DIFF
--- a/protoc-gen-openapiv2/internal/genopenapi/template.go
+++ b/protoc-gen-openapiv2/internal/genopenapi/template.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"log"
 	"math"
 	"net/textproto"
 	"os"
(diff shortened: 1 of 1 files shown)


johanbrandhorst:
As well intentioned as this change is, I don't think we can do this. There may be existing users that use the gateway successfully with stripped protobuf definitions, and we should continue to support them. This would be a breaking change.

PR #3595, opened 2023-09-12
Use go templates on tag descriptions too

same-id:
Using go templates is extremely useful for tag sections such as "Getting Started" etc, which it is a lot easier to just simply import a markdown file.

P.S.
I don't super like the fact that if the file is not found, it fails silently - I imagine cases where documentation will be missing in CI systems (internal repo paths will be exposed), and all for the sake of "debuggability" - might as well fail the process.
But I kept the current implementation for missing imports.

### Summary by CodeRabbit

- New Feature: Enhanced the `renderServiceTags` and `applyTemplate` functions in `template.go`. These changes allow for dynamic modification of the `Description` field of tags based on certain conditions, improving flexibility and customization.
- Test: Added a new test function `TestTagsWithGoTemplate` in `template_test.go`. This function tests the generation of tags using Go templates, ensuring the accuracy and reliability of the new feature.

### DIFF
--- a/protoc-gen-openapiv2/internal/genopenapi/template.go
+++ b/protoc-gen-openapiv2/internal/genopenapi/template.go
@@ -1096,10 +1096,16 @@ func renderServiceTags(services []*descriptor.Service, reg *descriptor.Registry)
 		}
 		if opts != nil {
 			tag.Description = opts.Description
+			if reg.GetUseGoTemplate() {
+				tag.Description = goTemplateComments(tag.Description, svc, reg)
+			}
 			if opts.ExternalDocs != nil {
 				tag.ExternalDocs = &openapiExternalDocumentationObject{
 					Description: opts.ExternalDocs.Description,
 					URL:         opts.ExternalDocs.Url,
+				}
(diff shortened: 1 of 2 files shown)


same-id:
Timeout on golang CI

same-id:
> Timeout on golang CI

Not the first time this happens

same-id:
> Timeout on golang CI

Not the first time this happens.

I think this is because golangci-lint runs only on pull requests (and not on `main` post merge) - so cache is not shared between different pull requests, so usually first push to PR will fail on timeout, second will succeed.

johanbrandhorst:
Yeah it keeps happening, it really shouldn't be a matter of caching, it should take no time to run 🤔

sam-utila:
Not sure about that, golangci-lint is one of the slowest linters I have ever used, it downloads dependencies as well. With cache it is super fast though.

johanbrandhorst (approved):
LGTM!

johanbrandhorst:
Thanks for your contribution!

PR #3646, opened 2023-10-06
Fix 2494

### LINKED ISSUE #2494: An RPC proto oneof field documents the field not used in the path as an optional request parameter

```
rpc GetArticle (GetArticleRequest) returns (GetArticleResponse) {
    option (google.api.http) = {
        get: "/v2/articles/{slug}"
        additional_bindings: {
            get: "/v2/articles/{id.hex}"
        }
     };
}

message GetArticleRequest {
    oneof article_id {
        insider.protobuf.ObjectID id = 1;
        string slug = 2;
    }
}

```

With the annotations above, two "separate" endpoints are defined in the output:

```
/v2/articles/{slug}
/v2/articles/{id.hex}
```

With the non-path parameter of the oneof listed as an optional query parameter, which is incorrect:

```
{
    "name": "id.hex",
    "description": "xyz",
    "in": "query",
    "required": false,
    "type": "string"
 }
```

I expect a single GET endpoint defined with 2 path parameters listed with oneof as required (not both, and nothing listed as a request parameter).

```
"parameters": [
  {
    "name": "slug",
    "description": "Article slug.",
    "in": "path",
    "required": true,
    "type": "oneof"
  },
  {
    "name": "id.hex",
    "description": "Article ID",
    "in": "path",
    "required": true,
    "type": "oneof"
  }
]
```

Your answer here.

I suspect there is simply no code to deal with the protobuf oneof type.

amanraja:
Fixes #2494

### DIFF
--- a/protoc-gen-openapiv2/internal/genopenapi/testdata/generator/path_item_object.swagger.yaml
+++ b/protoc-gen-openapiv2/internal/genopenapi/testdata/generator/path_item_object.swagger.yaml
@@ -22,6 +22,75 @@ paths:
           in: query
           required: false
           type: string
+        - name: uuid
+          in: query
+          required: false
+          type: string
+        - name: id
+          in: query
+          required: false
+          type: string
+      tags:
+        - YourService
+  /api/echo/{id}:
+    get:
+      operationId: YourService_Echo4
+      responses:
+        "200":
+          description: A successful response.
+          schema:
+            $ref: '#/definitions/StringMessage'
+      parameters:
+        - name: id
+          in: path
+          required: true
+          type: string
+        - name: value
+          in: query
+          required: false
+          type: string
+      tags:
+        - YourService
+  /api/echo/{uuid}:
+    get:
+      operationId: YourService_Echo3
+      responses:
+        "200":
+          description: A successful response.
+          schema:
+            $ref: '#/definitions/StringMessage'
+      parameters:
+        - name: uuid
+          in: path
+          required: true
+          type: string
+        - name: value
+          in: query
+          required: false
+          type: string
+      tags:
+        - YourService
+  /api/echo/{value}:
+    get:
+      operationId: YourService_Echo2
+      responses:
+        "200":
+          description: A successful response.
+          schema:
+            $ref: '#/definitions/StringMessage'
+      parameters:
+        - name: value
+          in: path
+          required: true
+          type: string
+        - name: uuid
+          in: query
+          required: false
+          type: string
+        - name: id
+          in: query
+          required: false
+          type: string
       tags:
         - YourService
 definitions:
@@ -29,4 +98,8 @@ definitions:
     type: object
     properties:
       value:
+        type: string
+      uuid:
+        type: string
+      id:
         type: string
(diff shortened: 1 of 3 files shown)


amanraja:
Hey @johanbrandhorst , Would you please look into this PR ?

johanbrandhorst (changes requested):
Hi! Thanks for this PR. The change looks reasonable, but I'm not sure I understand what the testdata changes do. Could you add a test to `template_test.go`? Or perhaps add an example of this circumstance to one of the test protobuf files and regenerate the files? Thanks!

amanraja:
> Hi! Thanks for this PR. The change looks reasonable, but I'm not sure I understand what the testdata changes do. Could you add a test to `template_test.go`? Or perhaps add an example of this circumstance to one of the test protobuf files and regenerate the files? Thanks!

Hey @johanbrandhorst , sure i will add tests for `template_test.go`, i made changes to testdata file to show the issue that we are addressing, hope that would suffice for us ?

protoc-gen-openapiv2/internal/genopenapi/testdata/generator/path_item_object.swagger.yaml
```diff
@@ -22,6 +22,75 @@ paths:
...
+  /api/echo/{uuid}:
+    get:
+      operationId: YourService_Echo3
+      responses:
+        "200":
+          description: A successful response.
+          schema:
+            $ref: '#/definitions/StringMessage'
+      parameters:
+        - name: uuid
```
amanraja, line 63: without this fix `Id` would also come up here, it's not now because `uuid` is of type `one_of`

zfy0701:
any change  this get merge? @johanbrandhorst it's super useful

johanbrandhorst (approved):
Don't know how I forgot about this PR, but it looks reasonable to me.

johanbrandhorst:
I'm going to close and reopen this to re-trigger the CI

johanbrandhorst:
Looks like we still need to rerun the generation script (see CONTRIBUTING.md). If the original author @amanraja cannot pick this back up, feel free to copy the changes to a new PR.

amanraja:
Hey @johanbrandhorst , have ran the generation script now, Please check the PR again , there are few changes changes have come up because of it.

examples/internal/proto/examplepb/echo_service.swagger.json
```diff
@@ -473,46 +473,13 @@
...
-            "required": false,
-            "type": "string",
-            "format": "int64"
-          },
-          {
-            "name": "status.note",
-            "in": "query",
-            "required": false,
-            "type": "string"
-          },
```
johanbrandhorst, line 495: I think these were removed incorrectly, no? There's nothing in this RPC definition that means the status query parameters shouldn't show up, I think? Here's the definition: https://github.com/grpc-ecosystem/grpc-gateway/blob/442bd3ab9497e4f9b3bbcbf4e53854300a68dd5a/examples/internal/proto/examplepb/echo_service.proto#L68. I think it's removing all oneofs instead of just the `lineNum` oneof.

johanbrandhorst (changes requested):
I think there's still a bug here

PR #3361, opened 2023-06-13
remove glog from a bunch of spots

kn100:
#### References to other Issues or PRs

Hopefully fixes https://github.com/grpc-ecosystem/grpc-gateway/issues/3342

#### Have you read the [Contributing Guidelines]

yes

#### Brief description of what is fixed or changed
Remove Glog from protoc-gen-grpc-gateway and also protoc-gen-openapiv2. Glog is a problematic dependency for downstream consumers of this repo.

Issue #3342, opened 2023-06-05
Use of Glog is problematic for users of Klog


kn100:
## 🐛 Bug Report

In repos that use Klog, Glog causes problems. This is because Glog has an uncontrollable init() func that registers flags that may have been registered elsewhere in the codebase. For my use case, it is specifically `protoc-gen-grpc-gateway` and `protoc-gen-openapiv2` that pose the problem. I see in other areas of grpc-gateway `grpclog` has been swapped in for Glog ([for example](https://github.com/grpc-ecosystem/grpc-gateway/pull/3205)). It seems Glog is not actively maintained anymore.

Would the repository owners be amenable to a PR that swaps out Glog for grpclog for these two areas? If so, I'm happy to make one.

## To Reproduce

1. Have a project that depends on grpc-gateway `protoc-gen-grpc-gateway` and `protoc-gen-openapiv2`
1. Add klog as a dependency of that project, and initialise the flags.
1. Observe errors relating to flags being redeclared.

## Expected behavior

I would not expect an upstream logging dep to break our logging solution.

## Actual Behavior

```

panic: v1.test flag redefined: log_dir
--
  |  
  | goroutine 1 [running]:
  | flag.(*FlagSet).Var(0xc0000e0120, {0x3037f28, 0xc00057e3f0}, {0x2c43256, 0x7}, {0x2c8ce5e, 0x2f})
  | /usr/local/go/src/flag/flag.go:982 +0x3bb
  | flag.(*FlagSet).StringVar(...)
  | /usr/local/go/src/flag/flag.go:847
  | flag.(*FlagSet).String(0x4089600?, {0x2c43256, 0x7}, {0x0, 0x0}, {0x2c8ce5e, 0x2f})
  | /usr/local/go/src/flag/flag.go:860 +0xbe
  | flag.String(...)
  | /usr/local/go/src/flag/flag.go:867
  | github.com/golang/glog.init()
```

## Your Environment
* k8s.io/klog/v2 v2.90.1
* github.com/grpc-ecosystem/grpc-gateway/v2 v2.15.1

johanbrandhorst:
Yep we should get rid of glog, I'd be happy to see that contribution if you have the time. Thanks for the detailed issue.

On a sidenote, I recommend versioning generators using `go install`, which might avoid this sort of issue (https://jbrandhorst.com/post/plugin-versioning/).

PR #3652, opened 2023-10-10
Remove glog dependency

ash2k:
This is a follow up for https://github.com/grpc-ecosystem/grpc-gateway/pull/3361 which hasn't removed all the `glog` references.

https://github.com/grpc-ecosystem/grpc-gateway/issues/3342

Yes.

Replaces `glog` with `grpclog`.

### Summary by CodeRabbit

- Refactor: Replaced the deprecated logging package `glog` with `grpclog` across the entire codebase. This change enhances the reliability of the logging system and aligns with the latest standards.
- Chore: Removed the `logtostderr=true` option from various `protoc` commands, simplifying the command-line interface and reducing potential confusion for users.
- Documentation: Updated the documentation to reflect the changes in logging and command-line options, making it easier for users to understand and use the system.


johanbrandhorst (approved):
Nice!

johanbrandhorst:
Not sure about that Bazel build error, the flag should still exist? Or does removing glog remove the flag? Maybe we need to add the flag back?

ash2k:
@johanbrandhorst The flag was part of glog, it was added by the library. Removing glog removed the flag too. I personally don't think the flag is needed, but this might be a breaking change for users when they upgrade the version. It's trivial to fix though - remove the flag from where it's set.

johanbrandhorst:
I see, yes, it will be a breaking change, and I think we're best of just including a new flag with the same purpose in our application so as not to break users. Would you mind adding a new flag for this? Thanks.

ash2k:
I don't think we can include such a flag as it will conflict with glog's flag if user's application is using glog.

johanbrandhorst:
This is `protoc-gen-grpc-gateway`, we shouldn't have anyone using it as a library, so it should be fine to add, I should think?

ash2k:
@johanbrandhorst What about other flags that users might have been setting? See https://pkg.go.dev/github.com/golang/glog#pkg-overview

Also, strictly speaking turning flags into no-ops is also a breaking change. Silently not doing what they used to do is even worse than failing explicitly in my opinion.

Honestly, I don't think many (if anyone) would notice the removal. Again, fixing the problem is trivial for the user but this "fix" to add the flags is clearly a hack. I'd personally not do it.

johanbrandhorst:
Believe me, I'd rather not do it too. I do think it would have a pretty big impact on our users though. During much of the gateways existence the canonical example of the use of the generator has included the use of the `logtostderr` flag. I believe many of our users out there set this flag without really caring about what it does just because the README. As a result, I think we need to add this particular flag back, and honestly it could just be a noop. If other flags are used by users we can add them back as we get bug reports about it, but I think this will affect a large number of users. This codesearch shows that it's quite widely used (and in popular repositories): https://github.com/search?type=code&auto_enroll=true&q=%22logtostderr%3Dtrue%2C%22. I think we'll actually need to add it back to both `protoc-gen-grpc-gateway` and `protoc-gen-openapiv2`. Again, I'd prefer this wasn't the case, but I want to be conscious of our users. What do you think?

ash2k:
I'll add the flag. I don't use either of the projects you mentioned directly, but I have them as transitive deps (for some reason).

If I used the project, I'd rather cleaned up the invocation than accumulated useless cruft but  🤷

README.md
```diff
@@ -532,19 +527,19 @@ through `protoc` using one of 2 patterns:
...
---grpc-gateway_opt logtostderr=true,repeated_path_param_separator=ssv
+--grpc-gateway_opt param1=value1,repeated_path_param_separator=ssv
 # or separately
---grpc-gateway_opt logtostderr=true --grpc-gateway_opt repeated_path_param_separator=ssv
---openapiv2_opt logtostderr=true,repeated_path_param_separator=ssv
+--grpc-gateway_opt param1=value1 --grpc-gateway_opt repeated_path_param_separator=ssv
+--openapiv2_opt param1=value1,repeated_path_param_separator=ssv
 # or separately
---openapiv2_opt logtostderr=true --openapiv2_opt repeated_path_param_separator=ssv
+--openapiv2_opt param1=value1 --openapiv2_opt repeated_path_param_separator=ssv
```
johanbrandhorst, line 542: Thank you for removing the old invocations, but I think we can just remove these example invocations too (the `param1=value1` stuff).

protoc-gen-grpc-gateway/main.go
```diff
@@ -35,6 +35,8 @@ var (
 	versionFlag                = flag.Bool("version", false, "print the current version")
 	warnOnUnboundMethods       = flag.Bool("warn_on_unbound_methods", false, "emit a warning message if an RPC method has no HttpRule annotation")
 	generateUnboundMethods     = flag.Bool("generate_unbound_methods", false, "generate proxy methods even for RPC methods that have no HttpRule annotation")
+
+	_ = flag.Bool("logtostderr", false, "Legacy glog compatibility. This flag is a no-op, you may removed it")
```
johanbrandhorst, line 39:
```suggestion
	_ = flag.Bool("logtostderr", false, "Legacy glog compatibility. This flag is a no-op, you can safely remove it")
```

protoc-gen-openapiv2/main.go
```diff
@@ -47,6 +47,8 @@ var (
 	useAllOfForRefs                = flag.Bool("use_allof_for_refs", false, "if set, will use allOf as container for $ref to preserve same-level properties.")
 	allowPatchFeature              = flag.Bool("allow_patch_feature", true, "whether to hide update_mask fields in PATCH requests from the generated swagger file.")
 	preserveRPCOrder               = flag.Bool("preserve_rpc_order", false, "if true, will ensure the order of paths emitted in openapi swagger files mirror the order of RPC methods found in proto files. If false, emitted paths will be ordered alphabetically.")
+
+	_ = flag.Bool("logtostderr", false, "Legacy glog compatibility. This flag is a no-op, you may removed it")
```
johanbrandhorst, line 51:
```suggestion
	_ = flag.Bool("logtostderr", false, "Legacy glog compatibility. This flag is a no-op, you can safely remove it")
```

johanbrandhorst (changes requested):
Thanks a lot for adding back the flag, I appreciate it 👍🏻. Some minor questions.

README.md
ash2k: Done

johanbrandhorst (approved):
LGTM

johanbrandhorst:
Thanks for your contribution!

ash2k:
Happy this got merged quickly! Thank you!

PR #3711, opened 2023-11-02
feat: support url form encoded marshaler and response in JSON

### LINKED ISSUE #7: Encoding with application/x-www-form-urlencoded
Today request/response body is always encoded in JSON.
Should accept application/x-www-form-urlencoded too.

1. Define a new field in the custom option to define a list of acceptable encoding of request body
2. Similar field to define the list of possible response body
    * maybe should prefer the same encoding as the request
3. Make the parameter deserialization/serialization pluggable

jhowliu:
Fixed #7. Support request in x-www-form-urlencoded and response in JSON format

## Summary by CodeRabbit

New Features:
- Added a new `runtime` file with a `UrlEncodeMarshal` struct, enhancing the system's ability to handle URL encoding.
- The `UrlEncodeMarshal` struct now implements the `Marshaler` interface, improving the system's data marshaling capabilities.
- Introduced methods for `ContentType`, `Marshal`, and `NewDecoder` in the `UrlEncodeMarshal` struct, providing more control over content type determination, JSON marshaling, and request body parsing.
- The `NewDecoder` method now populates the query parameters of the proto message, improving the system's ability to parse form data from request bodies.

### DIFF
--- a/runtime/marshal_urlencode.go
+++ b/runtime/marshal_urlencode.go
@@ -1,0 +1,56 @@
+package runtime
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/url"
+
+	"github.com/grpc-ecosystem/grpc-gateway/v2/utilities"
+	"google.golang.org/protobuf/proto"
+)
+
+type UrlEncodeMarshal struct {
+	Marshaler
+}
+
+// ContentType means the content type of the response
+func (u UrlEncodeMarshal) ContentType(_ interface{}) string {
+	return "application/json"
+}
+
+func (u UrlEncodeMarshal) Marshal(v interface{}) ([]byte, error) {
+	// can marshal the response in proto message format
+	j := JSONPb{}
+	return j.Marshal(v)
+}
+
+// NewDecoder indicates how to decode the request
+func (u UrlEncodeMarshal) NewDecoder(r io.Reader) Decoder {
+	return DecoderFunc(func(p interface{}) error {
+		msg, ok := p.(proto.Message)
+		if !ok {
+			return fmt.Errorf("not proto message")
+		}
+
+		formData, err := ioutil.ReadAll(r)
+		if err != nil {
+			return err
+		}
+
+		values, err := url.ParseQuery(string(formData))
+		if err != nil {
+			return err
+		}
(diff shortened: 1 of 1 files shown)


runtime/marshal_urlencode.go
```diff
@@ -0,0 +1,56 @@
...
+
+// NewDecoder indicates how to decode the request
+func (u UrlEncodeMarshal) NewDecoder(r io.Reader) Decoder {
+	return DecoderFunc(func(p interface{}) error {
+		msg, ok := p.(proto.Message)
+		if !ok {
+			return fmt.Errorf("not proto message")
+		}
+
+		formData, err := ioutil.ReadAll(r)
```
johanbrandhorst, line 36: Use `io.ReadAll`

runtime/marshal_urlencode.go
```diff
@@ -0,0 +1,56 @@
...
+
+		filter := &utilities.DoubleArray{}
+
+		err = PopulateQueryParameters(msg, values, filter)
+
+		if err != nil {
+			return err
+		}
+
+		return nil
```
johanbrandhorst, line 54: Since we're not mutating the error before returning, we can just
```suggestion
		return PopulateQueryParameters(msg, values, filter)
```

johanbrandhorst (changes requested):
Hey, thanks for this contribution! We'll definitely want to add some tests, perhaps you could look at adding an integration test in https://github.com/grpc-ecosystem/grpc-gateway/blob/main/examples/internal/integration/integration_test.go? There are some example tests in that you could take inspiration from. Presumably, you should be able to add a test against the ABitOfEverything endpoint with the query parameters in the body of the request. Let me know if you need more help!

johanbrandhorst:
@mickael-kerjean if you want to contribute tests to this PR you may need to get access to @jhowliu's fork.

paullen:
Hi! Please consider merging this and adding its usage to the documentation

johanbrandhorst:
Please reopen a PR with the same contents, this PR looks abandoned.

mickael-kerjean:
I am no longer interested in pursuing this as I was let go from the employer who needed it

PR #2124, opened 2021-05-04
Send trailers only when `TE: trailers` is set

zoido:
#### References to other Issues or PRs
Fixes #1709

#### Have you read the [Contributing Guidelines](https://github.com/grpc-ecosystem/grpc-gateway/blob/master/CONTRIBUTING.md)?
Yes.

#### Brief description of what is fixed or changed

Sends trailers back only when client specifies `TE: trailers`. 

#### Other comments

I borrowed the code from `errors.go` from https://github.com/grpc-ecosystem/grpc-gateway/pull/1499 as well as the approach to the tests.
I was thinking about increasing the reuse but did not figure out anything elegant. Will gladly adjust.

PR #3725, opened 2023-11-07
runtime: Add outgoing trailer matching

adriansmares:

Closes https://github.com/grpc-ecosystem/grpc-gateway/issues/1697

Yes.

- Avoid partially forwarding trailers if the caller did not accept them explicitly.
  - This seems to be a copy paste failure of the original PR (see [this](https://github.com/grpc-ecosystem/grpc-gateway/pull/2124/files#diff-4e2f81dd46cb78503d268ddc2f7b88ac214651ac316dc000ec32f3d180464203R154)).
  - It does not make sense to send the `Trailers` header if the trailers themselves will not be included, especially if the caller did not signal that it accepts trailers.
  - The error handling code did not do this duplication, so more reasons to consider this a bug.
- Add outgoing trailer matching.
  - The previous behavior is the default one.

#### Other comments

1. I implemented this change due to https://github.com/grpc/grpc-go/pull/6662 , as I would like to skip this trailer completely since the error details are present in the body. But maybe `grpc-gateway` should convert back binary header/trailer values to base64 before sending them to the HTTP caller, by default ? (similar to how binary metadata is decoded from the HTTP caller to the gRPC client)

2. There is no trailer support for streaming responses - should this be added ?

3. Should we expose the default matchers ? Currently only the incoming header matcher is a public function, and it is badly named given that there are three matchers now.

## Summary by CodeRabbit

- **New Features**
	- Enhanced error handling with improved access to server instances.
	- Introduced matching for outgoing headers and trailers.
- **Bug Fixes**
	- Adjusted function signatures for better operational efficiency.
- **Tests**
	- Added new tests for outgoing header and trailer matching.
- **Refactor**
	- Removed unnecessary package imports.
	- Updated server initialization to set default matchers if not set.

### DIFF
--- a/runtime/mux.go
+++ b/runtime/mux.go
@@ -126,7 +126,7 @@ func defaultOutgoingTrailerMatcher(key string) (string, bool) {
 // WithIncomingHeaderMatcher returns a ServeMuxOption representing a headerMatcher for incoming request to gateway.
 //
 // This matcher will be called with each header in http.Request. If matcher returns true, that header will be
-// passed to gRPC context. To transform the header before passing to gRPC context, matcher should return modified header.
+// passed to gRPC context. To transform the header before passing to gRPC context, matcher should return the modified header.
 func WithIncomingHeaderMatcher(fn HeaderMatcherFunc) ServeMuxOption {
 	for _, header := range fn.matchedMalformedHeaders() {
 		grpclog.Warningf("The configured forwarding filter would allow %q to be sent to the gRPC server, which will likely cause errors. See https://github.com/grpc/grpc-go/pull/4803#issuecomment-986093310 for more information.", header)
@@ -156,7 +156,7 @@ func (fn HeaderMatcherFunc) matchedMalformedHeaders() []string {
 //
 // This matcher will be called with each header in response header metadata. If matcher returns true, that header will be
 // passed to http response returned from gateway. To transform the header before passing to response,
-// matcher should return modified header.
+// matcher should return the modified header.
 func WithOutgoingHeaderMatcher(fn HeaderMatcherFunc) ServeMuxOption {
 	return func(mux *ServeMux) {
 		mux.outgoingHeaderMatcher = fn
@@ -167,7 +167,7 @@ func WithOutgoingHeaderMatcher(fn HeaderMatcherFunc) ServeMuxOption {
 //
 // This matcher will be called with each header in response trailer metadata. If matcher returns true, that header will be
 // passed to http response returned from gateway. To transform the header before passing to response,
-// matcher should return modified header.
+// matcher should return the modified header.
 func WithOutgoingTrailerMatcher(fn HeaderMatcherFunc) ServeMuxOption {
 	return func(mux *ServeMux) {
 		mux.outgoingTrailerMatcher = fn
--- a/runtime/mux.go
+++ b/runtime/mux.go
@@ -57,6 +57,7 @@ type ServeMux struct {
 	marshalers                marshalerRegistry
 	incomingHeaderMatcher     HeaderMatcherFunc
 	outgoingHeaderMatcher     HeaderMatcherFunc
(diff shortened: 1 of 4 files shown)


runtime/mux.go
```diff
@@ -154,6 +163,17 @@ func WithOutgoingHeaderMatcher(fn HeaderMatcherFunc) ServeMuxOption {
 	}
 }
 
+// WithOutgoingTrailerMatcher returns a ServeMuxOption representing a headerMatcher for outgoing response from gateway.
+//
+// This matcher will be called with each header in response trailer metadata. If matcher returns true, that header will be
+// passed to http response returned from gateway. To transform the header before passing to response,
+// matcher should return modified header.
```
johanbrandhorst, line 170: I know this is the same language as for `WithOutgoingHeaderMatcher`, but do you mind correcting it too?
```suggestion
// matcher should return the modified header.
```

runtime/mux.go
```diff
@@ -270,13 +290,13 @@ func NewServeMux(opts ...ServeMuxOption) *ServeMux {
...
-	if serveMux.outgoingHeaderMatcher == nil {
-		serveMux.outgoingHeaderMatcher = func(key string) (string, bool) {
-			return fmt.Sprintf("%s%s", MetadataHeaderPrefix, key), true
+	for matcher, fallback := range map[*HeaderMatcherFunc]HeaderMatcherFunc{
+		&serveMux.incomingHeaderMatcher:  DefaultHeaderMatcher,
+		&serveMux.outgoingHeaderMatcher:  defaultOutgoingHeaderMatcher,
+		&serveMux.outgoingTrailerMatcher: defaultOutgoingTrailerMatcher,
+	} {
+		if *matcher == nil {
+			*matcher = fallback
```
johanbrandhorst, line 299: This is clever, but I think I'd prefer that it was just 3 if statements along the line of the code you replaced 🙏🏻.

johanbrandhorst (approved):
This looks great, I just want to double check with you, this doesn't change the behavior unless someone configures a trailer matcher, right?

johanbrandhorst (changes requested):
Oops, meant to request changes

adriansmares:
> This looks great, I just want to double check with you, this doesn't change the behavior unless someone configures a trailer matcher, right?

Indeed:

adriansmares:
> This looks great, I just want to double check with you, this doesn't change the behavior unless someone configures a trailer matcher, right?

Indeed, before:
https://github.com/grpc-ecosystem/grpc-gateway/blob/1be8098bafc2081201753faf425e206c8e856cbd/runtime/handler.go#L112-L117
After:
https://github.com/grpc-ecosystem/grpc-gateway/blob/b1e60cdb4627caa327d8dbb21217a604df959196/runtime/mux.go#L122-L124

---

Could you please comment on the 'other comments' from the top level comment ?

johanbrandhorst:
> I implemented this change due to https://github.com/grpc/grpc-go/pull/6662 , as I would like to skip this trailer completely since the error details are present in the body. But maybe grpc-gateway should convert back binary header/trailer values to base64 before sending them to the HTTP caller, by default ? (similar to how binary metadata is decoded from the HTTP caller to the gRPC client)

Hm, I'm just worried about changing behaviour that we already have. Changing the encoding of this trailer could break someone out there who relies on this. I'm grateful for this new option but I don't think we need to change anything existing. Does that make sense?

> There is no trailer support for streaming responses - should this be added ?

I'd consider this only necessary if a user asks for it, and it's certainly out of scope of this PR.

> Should we expose the default matchers ? Currently only the incoming header matcher is a public function, and it is badly named given that there are three matchers now.

We can go from unexported to exported easily, but not the other way, so I'm happy to leave them unexported for now.

johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution!

PR #3727, opened 2023-11-07
fix: use req.Body instead of IOReaderFactory when possible

### LINKED ISSUE #3714: utilities.IOReaderFactory is inefficient
## 🐛 Bug Report

`utilities.IOReaderFactory` relies on `io.ReadAll` which is [documented to be inefficient](https://github.com/golang/go/issues/50774) for large inputs. 
For our use case, we had a 200MB JSON body that resulted in > 5GB transient memory allocations. 

It looks like in most cases, the `req.Body` could be passed directly into the marshaler. The exception to that is when field_masks are used then the body is parsed twice.  

## To Reproduce

Create a proto definition with repeated fields and send a relatively large input payload to monitor memory utilization. 

## Expected behavior

Memory utilization doesn't grow several times the input size. 

## Actual Behavior

Our 200MB input turned into more than 5GB memory allocations that the GC had to reclaim. This transient explosion in memory usage causes our service to run out of memory.

leungster:

Fixes #3714

yes

Changes request RPCs to use req.Body instead of reading into an in memory byte slice via IOReaderFactory. The IOReaderFactory is still available for the PATCH field_mask feature.

## Summary by CodeRabbit

- **New Features**
	- Enhanced request body processing logic to handle field masks more efficiently.
- **Tests**
	- Updated test cases to reflect changes in request body processing.

### DIFF
--- a/examples/internal/proto/examplepb/a_bit_of_everything.pb.gw.go
+++ b/examples/internal/proto/examplepb/a_bit_of_everything.pb.gw.go
@@ -488,11 +488,7 @@ func request_ABitOfEverythingService_CreateBody_0(ctx context.Context, marshaler
 	var protoReq ABitOfEverything
 	var metadata runtime.ServerMetadata
 
-	newReader, berr := utilities.IOReaderFactory(req.Body)
-	if berr != nil {
-		return nil, metadata, status.Errorf(codes.InvalidArgument, "%v", berr)
-	}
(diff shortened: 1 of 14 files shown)


johanbrandhorst (approved):
Amazing, thank you!

johanbrandhorst:
Thanks for your contribution!

Issue #3700, opened 2023-10-29
Support improved tracing integration by exposing router / path earlier


jalaziz:
## 🚀 Feature

I should note that I am trying to integrate the Datadog APM, but I believe this applies to OpenTracing/OpenTelemetry as well.

In trying to integration the Datadog APM, I realized that the the annotated context containing the path pattern is available only after HTTP handlers are called. At a high level, I understand why this is the case, but it complicates adding context to telemetry.

For example, when integrating the Datadog APM, the most straightforward way is to use the [WrapHandler](https://pkg.go.dev/gopkg.in/DataDog/dd-trace-go.v1/contrib/net/http#WrapHandler) function:

```
mux := runtime.NewServeMux()
handler = httptrace.WrapHandler(mux, "service_name", "")
```

The second parameters here is meant to be the resource name. Datadog also offers a [WithResourceNamer](https://pkg.go.dev/gopkg.in/DataDog/dd-trace-go.v1/contrib/net/http#WithResourceNamer) to dynamically generate the resource name (typically the HTTP method and path pattern). Unfortunately, `WithResourceNamer` works at the handler level and therefore it cannot be used to access the path pattern, just the full URL.

When looking at how Datadog integrates with other HTTP libraries, including the [standard ServerMux](https://github.com/DataDog/dd-trace-go/blob/568dacd4fd2daed508ac5c78fdf60eb1c0a52835/contrib/net/http/http.go#L46-L66), there is usually some capability to lookup the handler and route based on the request.

Unfortunately, `grpc-gateway` doesn't expose the route matching logic in a way that can be accessed outside of `ServeHTTP`, making it very difficult to know the parameterized route at the "wrapping" layer.

Ultimately, it would be nice if the route matching logic could be extracted from `ServeHTTP` in a way that it was both fast and re-usable to be able to lookup the route given a request.

## Workarounds

Yes, I could also use the gRPC client/server integration as the [docs suggest](https://grpc-ecosystem.github.io/grpc-gateway/docs/operations/tracing/). In fact, I also do this. However, telemetry at the HTTP layer is very useful.

As a workaround, I am able to use `runtime.WithMetadata` as [recommended here](https://grpc-ecosystem.github.io/grpc-gateway/docs/operations/annotated_context/) to _inject_ the resource name as a span tag, but it feels a bit hacky and disjointed:

```
runtime.WithMetadata(func(ctx context.Context, req *http.Request) metadata.MD {
[...]

PR #3728, opened 2023-11-07
chore: expose request func signatures

leungster:

yes

Assign request-func-signatures to unexported fields to allow overriding of the logic.

#### Other comments

I have a use case where I want to push additional info into the ctx prior to calling my RPC method. At the moment I can only set grpc Metadata which only allows `string` types. 
The request function is already a separate func in the generated code and I'd like to override it to inject additional info into my ctx prior to calling the predefined request-func.

## Summary by CodeRabbit

- **Refactor**
	- Improved the code structure for better handling of client and server streaming scenarios, and in-process request handling.
- **Tests**
	- Updated tests to align with the refactored code structure.

### DIFF
--- a/examples/internal/helloworld/helloworld.pb.gw.go
+++ b/examples/internal/helloworld/helloworld.pb.gw.go
@@ -35,7 +35,7 @@ var (
 	filter_Greeter_SayHello_0 = &utilities.DoubleArray{Encoding: map[string]int{"name": 0}, Base: []int{1, 2, 0, 0}, Check: []int{0, 1, 2, 2}}
 )
 
-func request_Greeter_SayHello_0(ctx context.Context, marshaler runtime.Marshaler, client GreeterClient, req *http.Request, pathParams map[string]string) (proto.Message, runtime.ServerMetadata, error) {
+var request_Greeter_SayHello_0 = func(ctx context.Context, marshaler runtime.Marshaler, client GreeterClient, req *http.Request, pathParams map[string]string) (proto.Message, runtime.ServerMetadata, error) {
 	var protoReq HelloRequest
 	var metadata runtime.ServerMetadata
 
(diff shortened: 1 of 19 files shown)


johanbrandhorst:
Hi, thanks for the PR. I think I'd like to understand the use case better before approving this. Are you saying that you'd like to add some non-string metadata into the context in the gRPC gateway request path? Is this similar to the request in https://github.com/grpc-ecosystem/grpc-gateway/issues/3700? Can you do what you want on the `http.Handler` level, or the gRPC interceptor level? In the future I'd appreciate if you open an issue to discuss any changes before going to create a PR as well. Thanks!

leungster:
@johanbrandhorst apologies for not creating an issue. I patched this for my own purposes via bazel and wanted to port over the patch to see if it is a reasonable change.

The concrete use case is more convoluted than what the PR says and I think for most use cases, someone can inject things into the request ctx via a handler wrapper.

We're trying to bypass the default decoder logic and manually unmarshal in our service RPC. We're not using the proto request message and instead directly unmarshaling into our internal datastructure to avoid duplicative memory allocations.
Our incoming request JSON is large and while we want to stop supporting this API, there are legacy clients that prevent us removing it entirely. This is a stop-gap until all users are switched over.

#3700 isn't quite the same because it sounds like they want to extract the HTTP pattern outside of the generated handler. Coincidentally we use `runtime.WithMetadata` to do what they're trying to do with opentelemetry.

johanbrandhorst:
If you already have a way to patch this such that we don't need to make any changes to the gateway, I'd prefer not to merge this as is. It seems a very niche use case to me, and in general I'd recommend avoiding making any changes to the contents of generated packages as it may lead to breakages down the road. We certainly don't consider changing unexported variables in generated packages a breaking change. Please close this PR if you're happy to keep it as a private patch.

leungster:
Sure, we'll keep it as a private patch then.

PR #3918, opened 2024-01-22
Parse request body when using X-HTTP-Method-Override header

### LINKED ISSUE #3688: Request body not parsed when using X-HTTP-Method-Override header
## 🐛 Bug Report

There is a bug around this code: https://github.com/grpc-ecosystem/grpc-gateway/blob/1c444cc622f9a1b9d7af04564e28c3a066271207/runtime/mux.go#L324-L330

 `r.Method` is overridden to `GET` **before** `r.ParseForm()` which causes `ParseForm` to skip parsing the request body and goes straight to parsing from the query string parameters. Since the whole point of using `X-HTTP-Method-Override` is to move the query string parameters to the request body, no query string parameters are present in the URL
 
 One possible solution is to move `r.Method = strings.ToUpper(override)` **after** the `r.ParseForm()` block.

## To Reproduce

Send a `POST` request to a `GET` endpoint with `Content-Type: application/x-www-form-urlencoded` and `X-HTTP-Method-Override: GET` headers. Set the request body with what would be query string parameters e.g. `limit=15&page=1`

## Expected behavior

Form parameters are parsed and passed along to the API handler

## Actual Behavior

Form parameters are not parsed and none of the parameters are populated when reaching the API handler

## Your Environment

(no info)

eyasy1217:

Fixes #3688

Yes

Fix to parse request body when using X-HTTP-Method-Override header

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved logic flow for HTTP method override in conjunction with form parsing.

- **Tests**
	- Added tests to ensure proper behavior of form parsing with HTTP method override.

### DIFF
--- a/runtime/mux.go
+++ b/runtime/mux.go
@@ -341,13 +341,13 @@ func (s *ServeMux) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if override := r.Header.Get("X-HTTP-Method-Override"); override != "" && s.isPathLengthFallback(r) {
-		r.Method = strings.ToUpper(override)
 		if err := r.ParseForm(); err != nil {
 			_, outboundMarshaler := MarshalerForRequest(s, r)
 			sterr := status.Error(codes.InvalidArgument, err.Error())
 			s.errorHandler(ctx, s, outboundMarshaler, w, r, sterr)
 			return
 		}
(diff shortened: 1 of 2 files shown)


johanbrandhorst (approved):
LGTM

johanbrandhorst:
Thanks for your contribution!

Issue #3848, opened 2023-12-28
Populating query param semantics changed since v2.15.0


roehrijn:
## 🐛 Bug Report

When dealing with nested messages having same field name like the field name of nested message itself in its parent, the population of HTTP Query parameters seems to be broken since v2.15.0.

**Example:** 

        message Identity {
          string id = 1;
          string foo = 2;
        }
        
        message Entity {
          Identity id = 1;
        }
        
        rpc Echo(Entity) returns (Entity) {
          option (google.api.http) = {
            get: "/{id.id}"
          };
        }

For the sake of simplicity the implementation of this endpoint just returns the message it received. 

## Expected behavior

When I query the endpoint using a HTTP GET like `/someid?id.foo=myqueryvalue` I would expect `id.foo` in the resulting response message to be set to `myqueryvalue`, like it was the case prior to v2.15.0

## Actual Behavior

The field `id.foo` isn't populated and remains empty.

## To Reproduce

I forked the repo and created an integration test showing exactly this behavior: https://github.com/roehrijn/grpc-gateway/commit/012a9d9f5dc210695fe435f83ce4fa4fde9fc84a

## Your Environment

Discovered this issue with v2.15.2 but `main` is affected since v2.15.0. 

## Additional comments 

Most properly the behavior has been introduced with #3072. I tracked it further down the following lines:
https://github.com/grpc-ecosystem/grpc-gateway/pull/3072/files#diff-9648d8792ce733c8bbce0d613a620fffdb73cff98a91763507a1984eb520d5feR80-R82

Looks like the added awareness of JSON names breaks the query param filtering behavior when using nested fields with colliding names. IMHO in the example case only `id.id` should be excluded from being a possible query param field name but actually `id.*` is too because the JSON name of the field in the nested message is not set and thus, just the field name is used. 

When changing definition of the `Identity` message to:

        message Identity {
          string id = 1 [json_name = "somethingElseThanId"];
          string foo = 2;
        }

the particular issue is resolved and the created test turns green. However, this obviously breaks the JSON representation of the messages when using `protojson` or `jsonpb`.

<sub>Jan Roehrich <jan.roehrich@mercedes-benz.com>, Mercedes-Benz Tech Innovation GmbH, [legal info/Impressum](https://github.com/mercedes-benz/foss/blob/master/LEGAL_IMPRINT.md)</sub>

johanbrandhorst:
[...]

PR #3946, opened 2024-01-29
fix path param override by query params

### LINKED ISSUE #3069: path parameters are overwritten by query parameters with the same name

ljmsc:
Fixes #3069 
original fix was reverted because of #3848

Yes

This PR should fix the possibility to override path parameters with query parameters. It was originally fixed with #3072 but a year later reverted because of #3848. In this PR the fix is a different approach, since the JSON names are the problem, I added a normalization function to rename query param names to the corresponding field name. The validation logic will now drop values which would override path parameters.

#### Other comments
I once again want to point out the criticality of this issue. A lot of authorization systems validate access based on the requested resource url. With this **security** issue it is easy to bypass those validations to get unauthorized access to other resources.

## Summary by CodeRabbit

- **New Features**
	- Introduced new optional parameters (`ResourceId`, `NIdNId`, and `NIdVal`) across various methods to enhance query capabilities.
	- Added `ExamplepbNestedMessage` and `ExamplepbUnannotatedNestedMessage` structs for improved data structuring.
	- Expanded test coverage with new test functions focusing on HTTP GET requests and response handling.
	- Updated `EchoService` and `UnannotatedEchoService` with new message types and additional route bindings.

- **Enhancements**
	- Added new fields (`ResourceId` and `NId`) to `ExamplepbSimpleMessage` and `ExamplepbUnannotatedSimpleMessage` structs for better resource identification and nested message support.

- **Refactor**
	- Implemented `normalizeFieldPath` and `findFieldFromPath` functions in `query.go` for optimized field path normalization and retrieval, enhancing parsing and field population logic.


ljmsc:
@coderabbitai review

johanbrandhorst (approved):
LGTM

PR #3966, opened 2024-02-04
api_visibility option now transitively hides request/response messages

### LINKED ISSUE #3962: api_visibility doesn't transitively apply to rpc request/response messages
## 🐛 Bug Report

Hello, it seems like `visibility_restriction_selectors` don't apply transitively to rpc messages when using the service option for `api_visibility`. 

Here's an example grpc spec:

```protobuf
service Example {
    option (google.api.api_visibility).restriction = "INTERNAL";

    rpc Greet(GreetRequest) returns (GreetResponse) {}
}

message GreetRequest {}

message GreetResponse {}
```

The `GreetRequest` and `GreetResponse` messages should not be included in the generated openapi spec if the `visibility_restriction_selectors` excludes the service but currently they will be included.

I believe the bug is occurring here: https://github.com/grpc-ecosystem/grpc-gateway/blob/bc00be5a204f0a2632437f8c4212469151e8c72e/protoc-gen-openapiv2/internal/genopenapi/template.go#L433-L440

Looks like the above code needs to check for service visibility and conditionally skip the service.

Here are some other spots where the service visibility **is** correctly checked for reference:

- https://github.com/grpc-ecosystem/grpc-gateway/blob/bc00be5a204f0a2632437f8c4212469151e8c72e/protoc-gen-openapiv2/internal/genopenapi/template.go#L1151-L1153
- https://github.com/grpc-ecosystem/grpc-gateway/blob/bc00be5a204f0a2632437f8c4212469151e8c72e/protoc-gen-openapiv2/internal/genopenapi/template.go#L1151-L1153

The above two spots mean that the service **is** correctly hidden from the openapi spec's service tags and routes/operations.

I believe messages should be transitively skipped to match the behaviour of the `method_visibility` option which also had this bug a while ago (https://github.com/grpc-ecosystem/grpc-gateway/issues/2691)

Place1:
Fixes #3962 

## Summary by CodeRabbit

- **New Features**
	- Introduced an API visibility option to allow skipping of services that are not marked as visible.
- **Tests**
	- Added tests to verify that services marked as "INTERNAL" are correctly excluded from processing.

### DIFF
--- a/protoc-gen-openapiv2/internal/genopenapi/template.go
+++ b/protoc-gen-openapiv2/internal/genopenapi/template.go
@@ -432,6 +432,10 @@ func getMapParamKey(t descriptorpb.FieldDescriptorProto_Type) (string, error) {
 // findServicesMessagesAndEnumerations discovers all messages and enums defined in the RPC methods of the service.
 func findServicesMessagesAndEnumerations(s []*descriptor.Service, reg *descriptor.Registry, m messageMap, ms messageMap, e enumMap, refs refMap) {
 	for _, svc := range s {
+		if !isVisible(getServiceVisibilityOption(svc), reg) {
+			continue
+		}
+
 		for _, meth := range svc.Methods {
 			// Request may be fully included in query
 			{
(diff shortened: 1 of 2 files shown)


johanbrandhorst (approved):
LGTM!

johanbrandhorst:
Thanks for your contribution!

PR #4027, opened 2024-02-22
Marshal non proto fields with idents

### LINKED ISSUE #4026: JSONPb marshaler doesn't use ident on non proto fields
## 🐛 Bug Report

The JSONPb marshaler does not use ident on non proto fields. [Here](https://github.com/grpc-ecosystem/grpc-gateway/blob/main/runtime/marshal_jsonpb.go#L153-L156) the value of JSONPb.MarshalOptions.Ident is checked once, but not checked below

## To Reproduce

```
package main

import (
	"fmt"

	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
	"google.golang.org/protobuf/encoding/protojson"
)

type S struct {
	A string
	B int64
	C map[string]int
}

func main() {
	m := &runtime.JSONPb{
		MarshalOptions: protojson.MarshalOptions{
			Indent: "    ",
		},
	}

	var err error
	var b []byte

	b, err = m.Marshal(&S{"GG", 3, map[string]int{"FOO": 0, "BAR": -1}})
	if err != nil {
		panic(err.Error)
	}
	fmt.Println(string(b))

	b, err = m.Marshal(map[string]int{"FOO": 1, "BAR": 2})
	if err != nil {
		panic(err.Error)
	}
	fmt.Println(string(b))
}
```

## Expected behavior

```
{
    "A": "GG",
    "B": 3,
    "C": {
        "BAR": -1,
        "FOO": 0
    }
}
{
    "BAR": 2,
    "FOO": 1
}
```

## Actual Behavior

```
{"A":"GG","B":3,"C":{"BAR":-1,"FOO":0}}
{
    "BAR": 2,
    "FOO": 1
}
```

## Your Environment

go version go1.21.6 linux/amd64

gknw:
Fixes #4026

### DIFF
--- a/runtime/marshal_jsonpb.go
+++ b/runtime/marshal_jsonpb.go
@@ -156,7 +156,13 @@ func (j *JSONPb) marshalNonProtoField(v interface{}) ([]byte, error) {
 		return json.Marshal(m)
 	}
 	if enum, ok := rv.Interface().(protoEnum); ok && !j.UseEnumNumbers {
+		if j.Indent != "" {
+			return json.MarshalIndent(enum.String(), "", j.Indent)
+		}
 		return json.Marshal(enum.String())
+	}
+	if j.Indent != "" {
+		return json.MarshalIndent(rv.Interface(), "", j.Indent)
 	}
 	return json.Marshal(rv.Interface())
 }


runtime/marshal_jsonpb.go
```diff
@@ -30,10 +30,6 @@ func (*JSONPb) ContentType(_ interface{}) string {
 
 // Marshal marshals "v" into JSON.
 func (j *JSONPb) Marshal(v interface{}) ([]byte, error) {
-	if _, ok := v.(proto.Message); !ok {
-		return j.marshalNonProtoField(v)
-	}
```
johanbrandhorst, line 35: I'm a bit worried about removing this, why was it necessary?

gknw: This appears to be unused code.  After that type checking calls j.marshalTo, where the same type checking is performed and the same call of j.marshalNonProtoField if the interface type is not proto.Message.

After removing the j.marshalNonProtoField call from j.Marshal, we can use json.Indent to adjust the indentation after the j.marshalNonProtoField call in one place. Otherwise we need to use json.Indent after j.marshalNonProtoFields in j.Marshal too.

johanbrandhorst: Ah, fair, thank you.

johanbrandhorst (approved):

PR #4180, opened 2024-04-07
Omit enum field when value is empty

### LINKED ISSUE #4179: omit-enum-default-value Generates Invalid Spec for Enums with a Single Value
## 🐛 Bug Report

When the `omit-enum-default-value` option is enabled it will set the `enum` field for an enum with just just a single value to `null` which is invalid against swagger specification 2.0.

## To Reproduce

I ran into this issue with a protobuf file that imports `google/protobuf/struct.proto`. The `struct.proto` file defines a [`NullValue`](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/struct.proto#L80) type that has just a single value, `NULL_VALUE` and the generated Swagger definition for the field is invalid. The generated Swagger spec for the following example file will produce the invalid spec:

```protobuf
syntax = "proto3";

package example;

option go_package = "example";

message GetExampleRequest {
  google.protobuf.Struct struct_value = 1;
}

message GetExampleResponse {}

service ExampleService {
  rpc GetExample(GetExampleRequest) returns (GetExampleResponse) {
    option (google.api.http) = {
      get: "/example"
    };
  };
}
```

The generated spec for the `NullValue` type, with the invalid `enum` field, is:

```json
"protobufNullValue": {
  "type": "string",
  "enum": null,
  "description": "`NullValue` is a singleton enumeration to represent the null value for the\n`Value` type union.\n\n The JSON representation for `NullValue` is JSON `null`."
}
```

## Expected behavior

To avoid generating an invalid spec the generator could omit the `enum` field or for types with just a single value. From what I can tell this was the behavior of the generator prior to #3167 but the refactoring in that commit changed the behavior because the `Enum` field was changed to an interface and when the `omit-enum-default-value` option removes the single value the field is a nil slice but [not a nil interface since it is typed](https://go.dev/doc/faq#nil_error) and therefore the `omitempty` option field does not omit it.

Alternatively, the `omit-enum-default-value` option could skip enum types with just a single (this is admittedly an edge case) or there could be a per-type option to configure `omit-enum-default-value` (I ran into this issue because for all other enums types in the proto file I'm using have more than a single value and for those the behavior of `omit-enum-default-value` is exactly what I need).

## Actual Behavior

The Swagger spec generated is invalid because the `enum` field is set to nil.

## Your Environment

I'm using grpc-gateway on Linux and MacOS with version v2.19.1 though as noted above I think this issue was introduced in 2.16.0 with #3167.

jeromefroe:

Fixes #4179.

Yes.

This commit fixes by #4179 by updating the `listEnumNumbers` and `listEnumNames` functions to return a nil interface when the set of enums is empty. The functions currently returned typed slices but when the `Enum` field in the `openapiParameterObject` type is set to the returned value, the values [aren't considered nil even when the slices are empty because the interfaces contain a type value](https://go.dev/doc/faq#nil_error). And the `omitempty` option for the field therefore won't omit the field as a result.


johanbrandhorst (changes requested):
This is a great fix, do you think you could write a test that exhibits the behavior? We have some integration tests in `examples/internal/integration`.

examples/internal/proto/examplepb/enum_with_single_value.swagger.json
```diff
@@ -0,0 +1,97 @@
...
+          }
+        ],
+        "tags": [
+          "EnumWithSingleValueService"
+        ]
+      }
+    }
+  },
+  "definitions": {
+    "examplepbEnumWithSingleValue": {
```
jeromefroe, line 53: With this change, we can see that the `enum` field is now omitted from the Swagger spec for this enum that has just a single value.

johanbrandhorst (approved):
Thanks!

johanbrandhorst:
Looks like there are still some lint errors 😬

johanbrandhorst:
😬

johanbrandhorst:
Thank you!

PR #4239, opened 2024-04-22
#4238 Added the WithWriteContentLength ServerMuxOption

### LINKED ISSUE #4238: Content-Length not set in ForwardResponseMessage

joshgarnett:
This PR Fixes #4238.  The GRPC Gateway is currently relying on the default behavior of w.Write, which will only set Content-Length on small messages by default.  For users who are running GRPC Gateway behind a CDN (i.e. CloudFront), this means they cannot take advantage of automatic compression handling for client requests.

Always setting the Content-Length header, isn't a great option, as middleware/forwardResponseOptions may be mutating the response and could lead to an invalid Content-Length being printed.  While a forwardResponseOption could be used to write Content-Length, this would result in a double marshaling of the protobuf message and could potentially be incorrect in special cases.

This fix should be backwards compatible and avoids the downsides discussed above.

### DIFF
--- a/runtime/handler.go
+++ b/runtime/handler.go
@@ -177,7 +177,7 @@ func ForwardResponseMessage(ctx context.Context, mux *ServeMux, marshaler Marsha
 		return
 	}
 
-	if mux.writeContentLength {
+	if !doForwardTrailers && mux.writeContentLength {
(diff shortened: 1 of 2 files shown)


joshgarnett:
Going to recreate the PR

PR #4259, opened 2024-04-25
Write Content-Length header if doForwardTrailers is not set

### LINKED ISSUE #4238: Content-Length not set in ForwardResponseMessage

joshgarnett:
This PR Fixes #4238

### DIFF
--- a/runtime/handler.go
+++ b/runtime/handler.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"net/http"
 	"net/textproto"
+	"strconv"
 	"strings"
 
 	"google.golang.org/genproto/googleapis/api/httpbody"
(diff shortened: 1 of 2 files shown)


runtime/handler.go
```diff
@@ -176,6 +177,10 @@ func ForwardResponseMessage(ctx context.Context, mux *ServeMux, marshaler Marsha
 		return
 	}
 
+	if !doForwardTrailers {
+		w.Header().Set("Content-Length", strconv.Itoa(len(buf)))
+	}
```
johanbrandhorst, line 182: Mind adding a test for this to our integration tests? https://github.com/grpc-ecosystem/grpc-gateway/blob/main/examples/internal/integration/integration_test.go

johanbrandhorst (changes requested):

runtime/handler.go
joshgarnett: Hi @johanbrandhorst, I've pushed a change that updates the existing tests to check for the header

johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution!

PR #4262, opened 2024-04-26
Added support for writing Etag headers and handling If-None-Match request headers

joshgarnett:
* Entity tags (Etag) are used to identify specific versions of a resource.  CDNs and Clients will use this in future requests to the server, using the If-None-Match header, to specify what data they already have and reduce the amount of data transferred if they already have the latest version.
* The Etag is only written for GET requests with messages greater than 100 bytes and is set to the md5 hash of the message.
* If a client sends the If-None-Match request header, the server will compare that value to the calculated etag.  If they match, it will return http.StatusNotModified (304) and not write the message in the response.
* Added additional tests to ensure the correct headers are being returned and that If-None-Match works.

### DIFF
--- a/runtime/handler.go
+++ b/runtime/handler.go
@@ -2,6 +2,8 @@ package runtime
 
 import (
 	"context"
+	"crypto/md5"
+	"encoding/hex"
 	"errors"
 	"io"
 	"net/http"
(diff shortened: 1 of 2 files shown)


joshgarnett:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag has more details on how this should be implemented

johanbrandhorst:
I appreciate the work put into this, but please open an issue first to discuss how to move forward with a feature request before opening a PR, to save us both time.

joshgarnett:
Good call out @johanbrandhorst, I almost created one first, but ended up being able to throw the code together fairly quickly.   Do you want me to still open one?

johanbrandhorst:
Yes please

joshgarnett:
Fixes #4263

PR #4265, opened 2024-04-27
Fixing TestOutgoingTrailerMatcher, which was non-deterministic

### LINKED ISSUE #4264: TestOutgoingTrailerMatcher is non-deterministic
PR #4259 added comparing the headers in TestOutgoingTrailerMatcher.  While working on another PR, I noticed that this would sometimes fail when the github hooks run the tests.  This is caused by the TrailerMD being a map, which has a non-deterministic read order.  A simple fix is to just sort the headers received before comparing. 

I'll create a PR shortly.

joshgarnett:
Recently the test was updated to also check the headers.  The test will sometimes fail when checking the headers returned, when there were multiple pairs added to the TrailerMD.  This is caused by the TrailerMD being a map, which has a non-deterministic read order.  The simple fix is to just sorts the Trailer headers returned before comparing.

Fixes #4264

### DIFF
--- a/runtime/handler_test.go
+++ b/runtime/handler_test.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"reflect"
+	"sort"
 	"testing"
 
 	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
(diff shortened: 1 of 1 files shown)


johanbrandhorst (approved):
Thanks!

PR #4290, opened 2024-05-06
feat: add mux middlewares and inject http pattern

nikitaksv:

Issue #4319

Yes

Added Middleware for ServerMux
Added the option of passing a HTTP Pattern to the request context

#### Other comments

Our team encountered a problem that middleware can only be applied to the Mux "http.Handler", which is unsafe for our application. Middleware is triggered regardless of whether a route exists or not, which leads to unnecessary triggering of middlewares. Also in http.Handler middleware there is no way except "defer" to get http pattern, so we added an option to pass HttpPattern in the request context before triggering middlewares.


johanbrandhorst:
Hi, thanks for your PR. However, please open an issue to discuss the problem you're having before submitting a PR. I'm not convinced this is functionality we want, so please open a PR to discuss your specific problem first. Thanks!

runtime/mux.go
```diff
@@ -89,6 +94,20 @@ func WithUnescapingMode(mode UnescapingMode) ServeMuxOption {
...
+		serveMux.middlewares = append(serveMux.middlewares, middlewares...)
+	}
+}
+
+// WithInjectHTTPPattern sets the current HTTP Pattern in the request context
+func WithInjectHTTPPattern() ServeMuxOption {
+	return func(serveMux *ServeMux) {
+		serveMux.injectHTTPPattern = true
+	}
+}
```
johanbrandhorst, line 109: I don't think we need an option for this, is there any downside to doing this unconditionally in the handler wrapper?

runtime/mux.go
```diff
@@ -48,11 +48,15 @@ var encodedPathSplitter = regexp.MustCompile("(/|%2F)")
 // A HandlerFunc handles a specific pair of path pattern and HTTP method.
 type HandlerFunc func(w http.ResponseWriter, r *http.Request, pathParams map[string]string)
 
+// A Middleware handler is simply an HandlerFunc that wraps another HandlerFunc to do some pre- and/or post-processing of the request
+type Middleware func(HandlerFunc) HandlerFunc
```
johanbrandhorst, line 52: Lets make sure users know that this is an alternative to using gRPC client interceptors when using the direct-to-implementation registration functions.
```suggestion
// A Middleware handler wraps another HandlerFunc to do some pre- and/or post-processing of the request. This is used as an alternative to gRPC interceptors when using the direct-to-implementation
// registration methods. It is generally recommended to use gRPC client or server interceptors instead
// where possible.
type Middleware func(HandlerFunc) HandlerFunc
```

runtime/mux_test.go
```diff
@@ -867,3 +867,63 @@ func (g *dummyHealthCheckClient) Check(ctx context.Context, r *grpc_health_v1.He
...
+		t.Errorf("The route test with method GET and path /test invalid, got %v", err)
+	}
+
+	r := httptest.NewRequest("GET", "/test", nil)
+	w := httptest.NewRecorder()
+	mux.ServeHTTP(w, r)
+	if w.Code != 200 {
+		t.Errorf("request not processed")
+	}
+}
```
johanbrandhorst, line 929: These tests are great, by the way. Thank you!

runtime/mux.go
```diff
@@ -89,6 +94,20 @@ func WithUnescapingMode(mode UnescapingMode) ServeMuxOption {
 	}
 }
 
+// WithMiddlewares sets server middleware for all handlers
```
johanbrandhorst, line 97:
```suggestion
// WithMiddlewares sets server middleware for all handlers. This is useful as an alternative to gRPC
// interceptors when using the direct-to-implementation registration methods and cannot rely
// on gRPC interceptors. It's recommended to use gRPC interceptors instead if possible.
```

johanbrandhorst (changes requested):
OK, I'm happy to go ahead with this feature. I would request that you update the comment on the generated direct-to-implementation registration functions in template.go: https://github.com/grpc-ecosystem/grpc-gateway/blob/33ca56eeaf85003043d90915d01d79a95dae870c/protoc-gen-grpc-gateway/internal/gengateway/template.go#L704-L706

to nudge people towards the use of this middleware functionality as an alternative to gRPC interceptors.

runtime/mux.go
nikitaksv: Nothing critical will happen if the pattern is added all the time. But it seemed to me that when updating the library, users would be concerned that there would be more contexts, so I thought it would be nice to be able to control this

johanbrandhorst: Nah, remove it.

nikitaksv: @johanbrandhorst done

johanbrandhorst (approved):
LGTM!

johanbrandhorst:
Looks like we need to regenerate the files after the update?

nikitaksv:
> Looks like we need to regenerate the files after the update?

It is not possible to regenerate files. Inside make generate, buf build is used, which has blocked its server for my country :(
Could you restore the files yourself?

johanbrandhorst:

runtime/mux.go
```diff
@@ -48,11 +48,15 @@ var encodedPathSplitter = regexp.MustCompile("(/|%2F)")
 // A HandlerFunc handles a specific pair of path pattern and HTTP method.
 type HandlerFunc func(w http.ResponseWriter, r *http.Request, pathParams map[string]string)
```
seriouspoop, line 49: @nikitaksv what is the need of pathParams here, can't we just implement a http.HandlerFunc? because of this i'm not able to use normal http middlewares

seriouspoop: @johanbrandhorst can you please help me?

johanbrandhorst: You can wrap the `runtime.ServeMux` in normal HTTP middleware, see https://grpc-ecosystem.github.io/grpc-gateway/docs/operations/logging/.

PR #4334, opened 2024-05-15
fix: handle `X-Forwarded-*` headers correctly

### LINKED ISSUE #4320: `X-Forwarded-For` header is not handled correctly
## 🐛 Bug Report

The remote IP is appended to the `X-Forwarded-For` header, but the original value is not overwritten, leading to duplication.

If multiple `X-Forwarded-For` headers are present, the remote IP is appended to the first one, when it should be appended to the last one.

## To Reproduce

1. Send a request to a gRPC-Gateway server with headers:
  ```
  X-Forwarded-For: 1.2.3.4, 5.6.7.8
  X-Forwarded-For: 9.10.11.12
  ```

## Expected behavior

Either
- the remote IP is appended to the value of the last `X-Forwarded-For` header pair using `", "` as a delimiter, and that pair is replaced; or
- an additional `X-Forwarded-For` pair is appended; or
- the existing `X-Forwarded-For` pairs are merged by joining their values with the delimiter `", "` and the remote IP is appended.

so that the resulting metadata contains either

```go
"X-Forwarded-For": []string{"1.2.3.4, 5.6.7.8", "9.10.11.12, ${remoteIP}"},
```

or

```go
"X-Forwarded-For": []string{"1.2.3.4, 5.6.7.8", "9.10.11.12", "${remoteIP}"},
```

or

```go
"X-Forwarded-For": []string{"1.2.3.4, 5.6.7.8, 9.10.11.12, ${remoteIP}"},
```

Per https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For#parsing

>There may be multiple `X-Forwarded-For` headers present in a request. The IP addresses in these headers must be treated as a single list, starting with the first IP address of the first header and continuing to the last IP address of the last header. There are two ways of making this single list:
>
>- join the `X-Forwarded-For` full header values with commas and then split by comma into a list, or
>- split each `X-Forwarded-For` header by comma into lists and then join the lists
>
>It is insufficient to use only one of multiple `X-Forwarded-For` headers.
>
>(Some reverse proxies will automatically join multiple X-Forwarded-For headers into one, but it is safest to not assume that this is the case.)

## Actual Behavior

With v2.19.1, the resulting metadata has `X-Forwarded-For: []string{"1.2.3.4, 5.6.7.8", "1.2.3.4, 5.6.7.8, ${remoteIP}"}`

haines:

Fixes #4320

Yes

This fixes the handling of `X-Forwarded-For` and `X-Forwarded-Host` headers in two ways:

1. They are only added to the metadata once, even if they are allowed by the incoming header matcher.
2. If multiple `X-Forwarded-For` headers are present, they are joined, with the remote IP from the request appended.

### DIFF
--- a/runtime/context.go
+++ b/runtime/context.go
@@ -148,6 +148,12 @@ func annotateContext(ctx context.Context, mux *ServeMux, req *http.Request, rpcM
 	var pairs []string
 	for key, vals := range req.Header {
 		key = textproto.CanonicalMIMEHeaderKey(key)
+		switch key {
+		case xForwardedFor, xForwardedHost:
+			// Handled separately below
+			continue
+		}
+
 		for _, val := range vals {
 			// For backwards-compatibility, pass through 'authorization' header with no prefix.
 			if key == "Authorization" {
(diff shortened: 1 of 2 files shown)


johanbrandhorst (approved):
Great, thank you!

johanbrandhorst:
Thanks for your contribution!

Issue #4358, opened 2024-05-22
proposal: Use version info from runtime/debug:BuildInfo when installed by go install


MakDon:
As #3967 says, if we install gRPC-gateway using `go install`, then `protoc-gen-grpc-gateway --version` would outputs `Version dev, commit unknown, built at unknown` because the build info is not linked during build time.

But i found that golang provides [runtime/debug:BuildInfo](https://pkg.go.dev/runtime/debug#BuildInfo) ,which may helps.

We could use the `BuildInfo` from it if vsc info is not linked by goreleaser:

```golang
	if commit == "unknown" {
		buildInfo, ok := debug.ReadBuildInfo()
		if ok {
			for _, setting := range buildInfo.Settings {
				if setting.Key == "vcs.revision" {
					commit = setting.Value
				}
				if setting.Key == "vcs.time" {
					date = setting.Value
				}
			}
			fmt.Printf("commit %v, built at %v\n", commit, date)
			os.Exit(0)
		}
	}
			
```

And then check if it works:

```bash
➜  grpc-gateway git:(master) ✗ cd ..
➜  dev mkdir tmp
➜  dev 
➜  dev cd tmp 
➜  tmp git clone git@github.com:MakDon/grpc-gateway.git
Cloning into 'grpc-gateway'...
remote: Enumerating objects: 22297, done.
remote: Counting objects: 100% (3444/3444), done.
remote: Compressing objects: 100% (443/443), done.
remote: Total 22297 (delta 3295), reused 3036 (delta 2997), pack-reused 18853
Receiving objects: 100% (22297/22297), 9.03 MiB | 3.20 MiB/s, done.
Resolving deltas: 100% (16799/16799), done.
➜  tmp cd grpc-gateway/protoc-gen-grpc-gateway 
➜  protoc-gen-grpc-gateway git:(master) ls
BUILD.bazel internal    main.go
➜  protoc-gen-grpc-gateway git:(master) go build
➜  protoc-gen-grpc-gateway git:(master) ✗ ./protoc-gen-grpc-gateway --version
commit 09bd9fef8e214eeb8487844d6f25f7f6c1101c22, built at 2024-05-22T17:36:03Z
➜  protoc-gen-grpc-gateway git:(master) ✗
```

MakDon:
Found it not working when installed by `go install`.

```bash
[...]

PR #4359, opened 2024-05-22
feat: Use version info from runtime/debug:BuildInfo when installed by go build

MakDon:
Implement of #4358

### DIFF
--- a/protoc-gen-grpc-gateway/main.go
+++ b/protoc-gen-grpc-gateway/main.go
@@ -13,6 +13,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"runtime/debug"
 	"strings"
 
 	"github.com/grpc-ecosystem/grpc-gateway/v2/internal/codegenerator"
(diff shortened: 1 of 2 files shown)


johanbrandhorst (commented):
This looks great! Just wondering if you could check that the time and commit format is consistent between this output and the released builds. Can you check what format `--version` returns on a released build?

MakDon:
> Can you check what format `--version` returns on a released build?

Yes, they are using the same format `RFC3339`

[runtime/debug:BuildSetting](https://pkg.go.dev/runtime/debug#BuildSetting): vcs.time: the modification time associated with vcs.revision, in RFC3339 format

[goreleaser-common-fields](https://goreleaser.com/customization/templates/#common-fields): current UTC date in RFC 3339 format


And I found that [runtime/debug:BuildSetting](https://pkg.go.dev/runtime/debug#BuildSetting) equals to [goreleaser](https://goreleaser.com/customization/templates/#common-fields)'s `.CommitDate`, but not the currently using `.Date` field. If we should do something with this gap?

johanbrandhorst:
Thanks for checking that, this looks great as is 😄.

PR #4372, opened 2024-05-25
Fixes #3917: Add package openapi naming strategy.

### LINKED ISSUE #3917: genopenapi: Stable openapi names for nested types
### Summary
An option for output similar to the current `openapi_naming_strategy=simple` output but unconditionally qualifying nested type names rather than only doing so dynamically upon collision. This could be a modification of the behavior of the "simple" output (if the current behavior is viewed as a bug), or it could be a new option.

### Current Behavior
Currently, strategies other than `fqn` dynamically determine how to qualify type names based on detecting collisions. This means that the generated names for a given type can change as new types are added to the genopapi invocation scope.

### Current Behavior Example

**Before File**

Simple service getting entities Foo. Foo has a State.

```proto
syntax = "proto3";
package example.package;
option go_package = "github.com/example/packagepb";

service FooService {
    rpc GetFoo (GetFooRequest) returns (Foo) {
        option (google.api.http) = {
            get: "/foo"
        };
    }
}

message GetFooRequest {}

message Foo {
    enum State {
        STATUS_UNKNOWN = 0;
        PENDING = 1;
        FOOED = 2;
    }
    State state = 1;
}
```

**After File**

We add an entity Bar that also has a State in the same proto package.

```proto
syntax = "proto3";
package example.package;
option go_package = "github.com/example/packagepb";

service FooService {
    rpc GetFoo (GetFooRequest) returns (Foo) {
        option (google.api.http) = {
            get: "/foo"
        };
    }
}

message GetFooRequest {}

message Foo {
    enum State {
        STATUS_UNKNOWN = 0;
        PENDING = 1;
        FOOED = 2;
    }
    State state = 1;
}

service BarService {
    rpc GetBar (GetBarRequest) returns (Bar) {
        option (google.api.http) = {
            get: "/bar"
        };
    }
}

message GetBarRequest {}

message Bar {
    enum State {
        STATUS_UNKNOWN = 0;
        PENDING = 1;
        BARRED = 2;
    }
    State state = 1;
}
```

**Emitted Type Names With Each openapi_naming_strategy**

 - `legacy`
   - Before:  `FooState`
   - After: `packageFooState`, `packageBarState`
   - Notes: Adding another type changed the existing type's name, and the generated names in "after" are unnecessarily qualified with `package`
 - `fqn`
   - Before: `example.package.Foo.State`
   - After: `example.package.Foo.State`, `example.package.Bar.State`
   - Notes: The output is predictable and stable but overly verbose for most use cases.
 - `simple`
   - Before: `State`
   - After: `Foo.State`, `Bar.State`
   - Notes: Adding another type changed the existing type's name. In the "Before" name, the generic "State" is not very meaningful - the name is meant to be meaningful in the context of Foo. The After output is desirable for that case (clear, no collisions, not overly verbose).

### Desired Behavior

Always output `Foo.State` for the nested type (unless `Foo.State` collides with a name from another package, in which case qualify it like `package.Foo.State`). This is like in the "after" output from the `simple` strategy, but it would not depend on whether another type `Bar.State` happened to be declared in the package.

In general, it would be nice to have options for more predictable and explicit control of name behavior in OpenAPI generation, rather than trying to "fix" collisions automatically as they are discovered at generation time. That is a larger issue, because there are complexities when colliding types are pulled in from other packages (I use State in this example because Status collides with the common protobuf Status type, which complicates the example). However, within a given package we should be able to mirror how nested-type names are generated in other language plugins (by qualifying the nested types), and as long as we use a separator that cannot appear in the proto type names themselves we will not have collisions within the package (see https://github.com/grpc-ecosystem/grpc-gateway/issues/1066 for an issue where qualification without separators can still result in collisions).

jgiles:
Support passing "package" for openapi_naming_strategy to create in-package unique names by qualifying nested types.

This strategy operates like "simple" except that we start by qualifying nested types, so that nested type names stay stable as new types are added in the same package.

We rely on package and type naming conventions to identify where the package name ends and the nested type name begins. This is not ideal, but avoids a larger refactor to provide package and type names as separate inputs to all the name resolution strategies. Longer term, that refactor may be desirable.


protoc-gen-openapiv2/internal/genopenapi/naming.go
```diff
@@ -52,20 +55,46 @@ func resolveNamesLegacy(messages []string) map[string]string {
...
 	uniqueNames := make(map[string]string)
 
 	hierarchy := func(pkg string) []string {
-		return strings.Split(pkg, ".")
+		if !qualifyNestedMessages {
+			return strings.Split(pkg, ".")
+		}
+		// We rely on the convention that package names are lowercase but message names are
+		// capitalized.
+		pkgEnd := regexp.MustCompile(`\.[A-Z]`).FindStringIndex(pkg)
```
johanbrandhorst, line 88: Could we move this regexp compilation into a global variable?

johanbrandhorst (changes requested):
Looks great, just a minor question. I'm starting to think that we need a whole docs page for the different types of name resolutions that exist but you don't need to add it here.

protoc-gen-openapiv2/internal/genopenapi/naming.go
jgiles: Sure!

I actually had it as a global var initially, as is often done for regexp compilation... but I moved it into the function to have the pattern defined in the same context as it it used for readability. Generally name resolution only happens once per process anyways, so I don't think the performance optimization of doing the regexp compilation one time up-front really gains us anything here.

I've moved it out though into a global var though, let me know what you think.

johanbrandhorst: Thanks, I appreciate the explanation. All the same, I think anyone reading an inline `regexp.MustCompile` will be surprised, so lets leave it as a package level variable.

johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution :)!

Issue #2906, opened 2022-09-26
protoc-gen-openapiv2: support deprecated field option


lachlancooper:
## 🚀 Feature

When generating an openapiv2 definition, any [protobuf `deprecated` field options](https://developers.google.com/protocol-buffers/docs/proto3#options) are currently ignored. It would be nice if the generator automatically set the [deprecated field on the resulting parameter object](https://swagger.io/specification/#parameter-object) accordingly.

Note that the deprecated field on [_operation_ objects](https://swagger.io/specification/#operation-object) is already supported, though this is handled through a custom annotation instead of the official `deprecated` option. This was already touched on in #2856, but in my opinion it makes sense to support the official option ([`deprecated = true` can be set on `rpc`, `message`, `enum`, and `field`](https://github.com/protocolbuffers/protobuf/issues/1734#issuecomment-234681729)) wherever possible instead of requiring the use of custom annotations.

johanbrandhorst:
Hi, thanks for your issue! I'm surprised we didn't already have an issue for this, I agree that it makes sense to automatically apply this. Furthermore, if both an annotation and a protobuf level option is set, the annotation should take precedence (so that it's always possible to change the output of the openapi generator without changing the protobuf definition).

denisdubovitskiy:
@johanbrandhorst i think that this annotation should also add some context value to indicate that the method is deprecated. Having this, we can use an interceptor to collect a metric and notify clients about new implementation. What do you think?

johanbrandhorst:
I don't think that should be necessary - with APIv2 it should be possible to tell already through reflection whether a method is deprecated. I would consider this ask outside the scope of the gateway.

ankur-kalita:
can you please assign this issue to me ?

johanbrandhorst:
Please just submit a PR if you want to work on this

PR #4373, opened 2024-05-26
Add support for method's deprecated option

vajexal:
https://github.com/grpc-ecosystem/grpc-gateway/issues/2906
https://github.com/grpc-ecosystem/grpc-gateway/pull/2872

this pr adds support for `option deprecated = true;` in method description, e. g.

```
service Greeter {
    option deprecated = true;
  }
}
```

instead of

```
service Greeter {
      deprecated: true
    }
  }
}
```

some breaking change: we will mark method deprecated if any option says so, thus if method's deprecated option is true and openapiv2_operation's deprecated is false, openapiv2 method would be deprecated (before this pr method wouldn't be deprecated). This will happen because we can't differ openapiv2_operation's deprecated = false from absence of the field, so we process openapiv2_operation's deprecated field only if it's true

### DIFF
--- a/internal/descriptor/registry.go
+++ b/internal/descriptor/registry.go
@@ -160,6 +160,9 @@ type Registry struct {
 	// preserveRPCOrder, if true, will ensure the order of paths emitted in openapi swagger files mirror
 	// the order of RPC methods found in proto files. If false, emitted paths will be ordered alphabetically.
 	preserveRPCOrder bool
+
+	// enableRpcDeprecation whether to process grpc method'd deprecated option
+	enableRpcDeprecation bool
 }
 
 type repeatedFieldSeparator struct {
@@ -874,4 +877,14 @@ func (r *Registry) SetPreserveRPCOrder(preserve bool) {
 // IsPreserveRPCOrder returns preserveRPCOrder
 func (r *Registry) IsPreserveRPCOrder() bool {
 	return r.preserveRPCOrder
+}
+
+// SetEnableRpcDeprecation sets enableRpcDeprecation
+func (r *Registry) SetEnableRpcDeprecation(enable bool) {
+	r.enableRpcDeprecation = enable
+}
+
+// GetEnableRpcDeprecation returns enableRpcDeprecation
+func (r *Registry) GetEnableRpcDeprecation() bool {
+	return r.enableRpcDeprecation
 }
--- a/internal/descriptor/registry.go
+++ b/internal/descriptor/registry.go
@@ -161,7 +161,7 @@ type Registry struct {
 	// the order of RPC methods found in proto files. If false, emitted paths will be ordered alphabetically.
 	preserveRPCOrder bool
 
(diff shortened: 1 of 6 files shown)


johanbrandhorst (changes requested):
Thank you for this thoughtful PR! I especially appreciate that you outlined the breaking changes involved. I think, because this introduces behavior that can't be turned off, we need to only enable this behavior under a flag. If it was possible to have the OpenAPIv2 annotation override the protobuf annotation I think it'd be fine to have this enabled by default, but as it is I think we need an option. How about `enable_rpc_deprecation`?

internal/descriptor/registry.go
```diff
@@ -160,6 +160,9 @@ type Registry struct {
 	// preserveRPCOrder, if true, will ensure the order of paths emitted in openapi swagger files mirror
 	// the order of RPC methods found in proto files. If false, emitted paths will be ordered alphabetically.
 	preserveRPCOrder bool
+
+	// enableRpcDeprecation whether to process grpc method'd deprecated option
```
johanbrandhorst, line 164:
```suggestion
	// enableRpcDeprecation whether to process grpc method's deprecated option
```

protoc-gen-openapiv2/main.go
```diff
@@ -49,6 +49,7 @@ var (
 	useAllOfForRefs                = flag.Bool("use_allof_for_refs", false, "if set, will use allOf as container for $ref to preserve same-level properties.")
 	allowPatchFeature              = flag.Bool("allow_patch_feature", true, "whether to hide update_mask fields in PATCH requests from the generated swagger file.")
 	preserveRPCOrder               = flag.Bool("preserve_rpc_order", false, "if true, will ensure the order of paths emitted in openapi swagger files mirror the order of RPC methods found in proto files. If false, emitted paths will be ordered alphabetically.")
+	enableRpcDeprecation           = flag.Bool("enable_rpc_deprecation", false, "whether to process grpc method'd deprecated option.")
```
johanbrandhorst, line 52:
```suggestion
	enableRpcDeprecation           = flag.Bool("enable_rpc_deprecation", false, "whether to process grpc method's deprecated option.")
```

johanbrandhorst (changes requested):
Could you please add a bazel flag for the new option too?

vajexal:
i'm not good at bazel, so correct me if i'm wrong somewhere

vajexal:
and should this option be added to `docs/docs/mapping/customizing_openapi_output.md`?

johanbrandhorst:
> and should this option be added to `docs/docs/mapping/customizing_openapi_output.md`?

Sure, that'd be great!

johanbrandhorst:
> i'm not good at bazel, so correct me if i'm wrong somewhere

This looks fine, thank you.

johanbrandhorst (approved):
Looks great!

Issue #4568, opened 2024-07-30
Dealing with 304 response without printing errors


taichimaeda:
## 🐛 Bug Report

(A clear and concise description of what the bug is.)

This is not really a bug to be exact, but I've been struggling with the error logs coming from `grpc-gateway` and was wondering if I could get some help...

I'm using `WithForwardResponseOption(forwardResponseOption)` to inspect the response message and set the 304 Not Modified status when necessary, but doing so results in the following error:

```console
Failed to write response: http: request method or response status code does not allow body
```

which comes from this [Write](https://github.com/grpc-ecosystem/grpc-gateway/blob/8508633ca6fc9448d1db4f6ed49a6b0a3e23601a/runtime/handler.go#L184) call, because you're [not allowed](https://github.com/golang/go/blob/3959d54c0bd5c92fe0a5e33fedb0595723efc23b/src/net/http/transfer.go#L460) to write a response body when it's 304.

This is not the behaviour I'm expecting - a 304 response is not an error, and printing them on every request just makes them noisy.

The way I used to avoid this issue was simply returning an error in the `forwardResponseOption` function, so that the above `Write` gets skipped by an early return, and ignoring this error in a custom error handler using the`WithErrorHandler`.

But unfortuantely, this trick is no longer working for me, since [this PR](https://github.com/grpc-ecosystem/grpc-gateway/pull/4327/files#diff-4e2f81dd46cb78503d268ddc2f7b88ac214651ac316dc000ec32f3d180464203L204) changed `grpclog.Infof("Error handling ForwardResponseOptions: %v", err)` to `grpclog.Errorf("Error handling ForwardResponseOptions: %v", err)`.
(To be clear, this means it still did print *info* logs before this PR, but it was ok for me - the *error* logs are the real problem)

And now I'm out of options...
Is there a good way to get around this, perhaps by modifying [handleForwardResponseOptions](https://github.com/grpc-ecosystem/grpc-gateway/blob/8508633ca6fc9448d1db4f6ed49a6b0a3e23601a/runtime/handler.go#L198) so that it leaves the printing to the [error handler](https://github.com/grpc-ecosystem/grpc-gateway/blob/8508633ca6fc9448d1db4f6ed49a6b0a3e23601a/runtime/mux.go#L68)?

## To Reproduce

(Write your steps here:)

1. Step 1...
1. Step 2...
1. Step 3...

## Expected behavior

(Write what you thought would happen.)

[...]

PR #4594, opened 2024-08-06
stop printing error logs in handleForwardResponseOptions

taichimaeda:

Fixes: https://github.com/grpc-ecosystem/grpc-gateway/issues/4568

✅ Done

This will stop printing error messages when [handleForwardResponseOptions](https://github.com/taichimaeda/grpc-gateway/blob/52698e52d9261b141d8b062b1bd60a84bb47fae2/runtime/handler.go#L199) fails, and instead wraps the error and lets the error handler deal with the printing, which should hopefully fix the issue around 304 status responses.

#### Other comments

My major concern is that [DefaultHTTPErrorHandler](https://github.com/taichimaeda/grpc-gateway/blob/52698e52d9261b141d8b062b1bd60a84bb47fae2/runtime/errors.go#L93) seems to use the `err` only for getting the error status, and not for printing its message (correct me if I'm wrong) - will consider other solutions if this is not okay.

### DIFF
--- a/runtime/handler.go
+++ b/runtime/handler.go
@@ -3,6 +3,7 @@ package runtime
 import (
 	"context"
 	"errors"
+	"fmt"
 	"io"
 	"net/http"
 	"net/textproto"
@@ -201,8 +202,7 @@ func handleForwardResponseOptions(ctx context.Context, w http.ResponseWriter, re
 	}
 	for _, opt := range opts {
 		if err := opt(ctx, w, resp); err != nil {
-			grpclog.Errorf("Error handling ForwardResponseOptions: %v", err)
-			return err
+			return fmt.Errorf("error handling ForwardResponseOptions: %w", err)
 		}
 	}
 	return nil


johanbrandhorst (approved):
LGTM!

johanbrandhorst:
> My major concern is that [DefaultHTTPErrorHandler](https://github.com/taichimaeda/grpc-gateway/blob/52698e52d9261b141d8b062b1bd60a84bb47fae2/runtime/errors.go#L93) seems to use the err only for getting the error status, and not for printing its message (correct me if I'm wrong) - will consider other solutions if this is not okay.

I think this is fine, in effect all we've done is removed a log message and added some context to the error. I wouldn't consider it a breaking change.

taichimaeda:
HI thanks for the merge!

I was just curious, would you think this change will take some time before getting rolled out in the next release?

If that's the case I might need to think of other temporary solutions, since it's getting somewhat urgent (the logs started overwhelming the metrics collector also...)

PR #4602, opened 2024-08-07
Add some documenation for CustomQueryParser with examples

texzone:
Yes, I do not think it applies to me

Just adding some hopefully helpful docs to the README about the Custom Query Parser, when to use it, and how to use it! Hopefully it will be helpful for someone else!

#### Other comments
I did not look with a strongly discerning eye at all the places this documentation this *may* be more appropriately placed. If you have any suggestions about that, please let me know!


johanbrandhorst (changes requested):
Ah, sorry for not being clearer, we actually have a docs site with content in the `docs` folder. You might want to copy most of this into the `docs/docs/customizing_your_gateway.md` instead? Thanks!

texzone:
> Ah, sorry for not being clearer, we actually have a docs site with content in the `docs` folder. You might want to copy most of this into the `docs/docs/customizing_your_gateway.md` instead? Thanks!

Totally agree that they should not be in root readme - just wasn't sure where to put them.

docs/docs/using_custom_query_parser.md
```diff
@@ -0,0 +1,119 @@
...
+	"your_module/path/customparser"
+)
+
+// create a new ServeMux with custom parser and other runtime options as needed
+func createGRPCGatewayMux() *runtime.ServeMux {
+	// whatever custom code you may need before you create the mux...
+	
+	return runtime.NewServeMux(
+		// Custom query parameter parser
+		runtime.WithQueryParameterParser(&customparser.CustomQueryParameterParser{}),
```
johanbrandhorst, line 113: I don't think this option exists, does it? You have to use `SetQueryParameterParser`: https://pkg.go.dev/github.com/grpc-ecosystem/grpc-gateway/v2/runtime#SetQueryParameterParser. Note that this will set the query parameter parser globally. It's honestly a bit gross, but it's the way it is because the parse function is called directly from generated code and not on the serve mux.

johanbrandhorst (changes requested):
Thanks, this is great, just some minor comments.

docs/docs/using_custom_query_parser.md
texzone: oof, good catch, yes

johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution!

PR #4619, opened 2024-08-10
Fix #4614 Unique Items placement

### LINKED ISSUE #4614: Placement for unique items in array type is located in the wrong place
## 🐛 Bug Report

This bug relates to protoc-gen-openapiv2.

It looks like `uniqueItems` option with array type is placed in the wrong place when compiled to json. Current formatting places the field within the `items` field of the property object. But according to the [spec](https://swagger.io/specification/v2/#items-object): it should be in the root property object.

The problem is located here: https://github.com/grpc-ecosystem/grpc-gateway/blob/main/protoc-gen-openapiv2/internal/genopenapi/template.go#L2939 - where the unique field is placed within the items object instead of the root object

## To Reproduce

1. Create in your proto a repeated string field with annotations like:
```
  repeated string url = 1 [
      min_length: 1,
      max_length: 512,
      max_items: 3
      unique_items: true,
      description: "A cool set of URLs",
      example: "[\"https://www.example.com/\", \"https://www.github.com/\"]",
  }];
``` 

1. Run the generator
1. See that the output has:
```
"url": {
  "type": "array",
  "example": [
    "https://www.example.com/",
    "https://www.google.com/"
  ],
  "items": {
    "type": "string",
    "maxLength": 512,
    "minLength": 1
    "uniqueItems": true,
  },
  "description": "a cool set of URLs",
  "maxItems": 3
},
```

## Expected behavior

```
"url": {
  "type": "array",
  "example": [
    "https://www.example.com/",
    "https://www.google.com/"
  ],
  "items": {
    "type": "string",
    "maxLength": 512,
    "minLength": 1
  },
  "uniqueItems": true,
  "description": "a cool set of URLs",
  "maxItems": 3
},
```

Using the rendered swagger output in editor.swagger.io I would expect to see an error when I put in two duplicate entries

## Actual Behavior

Using the rendered swagger output in editor.swagger.io I see no error when I put in two duplicate entries, manually adjusting the output to the "expected output" I do get an error from editor.swagger.io

## Your Environment

Was using grpc gateway 2.19.1

sebaraj:

Fixes #4614

Yes.

Corrects placement of "Unique Items" in output.

#### Other comments

This is my first contribution in this project, so I welcome any feedback/recommendations.


johanbrandhorst (changes requested):
Thanks for your PR! Don't we also need to remove the `UniqueItems` field from the `openapiItemsObject` type?

sebaraj:
I apologize if I am mistaken, but I believe that `openapiItemsObject` is defined as an `openapiSchemaObject` (https://github.com/grpc-ecosystem/grpc-gateway/blob/main/protoc-gen-openapiv2/internal/genopenapi/types.go#L238), and removing `UniqueItems` from the schema object would be destructive. Also, I apologize for not updating the test cases in `template_test.go`.

johanbrandhorst (approved):
Ah, you are right, both the Schema object and Parameter object appear to have a UniqueItems field. Then this looks good to me! Thank you for updating the tests too.

johanbrandhorst:
Congratulations on your first contribution :)

Issue #4483, opened 2024-07-01
[Feature Request] Provide a legit way to override HTTP Response.


nguyentranbao-ct:
## 🚀 Feature Request

**Issue with error logs for ForwardResponseOptions**

PR #4245 changed how errors are logged, which is cool. But it also started logging a bunch of stuff we already aware. 

**My situation:**

I need to override the response sometimes. To do that, I use `ForwardResponseOptions` with a custom function that might return an error. This tells grpc-gateway to chill and not write anything else. Works great, but then it fills the logs with these messages after `v2.20.0`:

```
Error handling ForwardResponseOptions: xxx
```

**Tried a few things:**

* Setting `GRPC_GO_LOG_SEVERITY_LEVEL=none` would stop all logs, but I still need to see other errors.
* Overriding `grpc.Logger` is also an option, but that's kinda overkill just to omit the simple log.

**Proposal:**

Add a new variable like `ErrResponseHandled`. This would tell `grpc-gateway` to:

* Stop logging "Error handling ForwardResponseOptions" for these cases.
* Not rewrite the response body (which I already took care of).
* Less code clutter by avoiding unnecessary logging.

**Similar issues but have been closed:**

* #517
* #1610

**Code examples (in case they help):**

Here's what the change in `grpc-gateway/runtime/handler.go` would look like:

```go
// runtime/handler.go
package runtime

var ErrResponseHandled = errors.New("response handled")

func handleForwardResponseOptions(ctx context.Context, w http.ResponseWriter, resp proto.Message, opts []func(context.Context, http.ResponseWriter, proto.Message) error) error {
  if len(opts) == 0 {
    return nil
  }
  for _, opt := range opts {
    if err := opt(ctx, w, resp); err != nil {
      if !errors.Is(err, ErrResponseHandled) {
        grpclog.Errorf("Error handling ForwardResponseOptions: %v", err)
      }
      return err
    }
  }
  return nil
}
```

And here's how I use it in my code:

```go
// in your files
func forwardResponse(ctx context.Context, w http.ResponseWriter, m protoreflect.ProtoMessage) error {
  switch v := m.(type) {
  case A, B, C:
    w.WriteHeader(http.StatusNoContent)
	// response has been written, don't write anything else
    return runtime.ErrResponseHandled
  case X, Y, Z:
    resp := map[string]any{
      "Success": true,
      "Data":    m,
    }
    w.WriteHeader(http.StatusOK)
[...]

Issue #1610, opened 2020-08-24
How to elegant rewrite/custom resp body from gRPC resp ? 


zhangjie2012:
My business scene want all http resp format like: 

```json
{ "code": "0/1", "data": {}, "error": ""}
```

**For error response**, I use `runtime.WithProtoErrorHandler` get grpc error, and custom it, work well.

but, **For success response**, I did not find suitable ones, I tried `runtime.WithForwardResponseOption`, it work like a 
hook. When I call `w.Write`, it will output data _twice copy_. like:

```json
{"code": 1, "data": {"key": "hello"}, "error": ""}{"key": "hello"}
``` 

So, I check out source, found https://github.com/grpc-ecosystem/grpc-gateway/blob/master/runtime/handler.go#L128 `ForwardResponseMessage`.  It realy like a hook. 

A hack skill to fixed my problem, all resp proxy by `runtime.WithProtoErrorHandler`:

``` go
type StandardResp struct {
	Code  int         `json:"code"`
	Data  interface{} `json:"data"`
	Error string      `json:"error"`
}

const (
	proxyFlag = "__succ__"
)

func HttpSuccHandler(ctx context.Context, w http.ResponseWriter, p proto.Message) error {
	resp := StandardResp{
		Code:  0,
		Data:  p,
		Error: "",
	}
	bs, _ := json.Marshal(&resp)
	return errors.New(proxyFlag + string(bs))
}

func HttpErrorHandler(ctx context.Context, mux *runtime.ServeMux, m runtime.Marshaler, w http.ResponseWriter, r *http.Request, err error) {
	w.Header().Set("Content-Type", "application/json")

	// success proxy
	raw := err.Error()
	if strings.HasPrefix(raw, proxyFlag) {
		raw = raw[len(proxyFlag):]
		w.Write([]byte(raw))
		return
	}

	// error handler
	s, ok := status.FromError(err)
	if !ok {
		s = status.New(codes.Unknown, err.Error())
	}
	resp := StandardResp{
		Code:  1,
		Data:  nil,
		Error: s.Message(),
	}
	bs, _ := json.Marshal(&resp)
	w.Write(bs)
}
```

It worked for me, but I don't think it's a elegant way. 

So, my question is can I custom resp body by more direct or simple way, similar programmer can define a `ForwardResponseMessage` method, framework provide a `DefaultForwardResponseMessage` method.

go version go1.14.7 darwin/amd64,  grpc-gateway v1.14.7

[...]

PR #4622, opened 2024-08-12
feat: Add WithForwardResponseRewriter to allow easier/more useful response control

nkcmr:
## Context/Background
I am working with grpc-gateway to mimick an older REST API while
implementing the core as a gRPC server. This REST API, by convention,
emits a response envelope.

After extensively researching grpc-gateway's code and options, I think
that there is not a good enough way to address my use-case; for a few
more nuanced reasons.

The sum of my requirements are as follows:
- 1) Allow a particular field of a response message to be used as the main
response. ✅ This is handled with `response_body` annotation.
- 2) Be able to run interface checks on the response to extract useful
information, like `next_page_token` [0] and surface it in the final
response envelope. (`npt, ok := resp.(interface { GetNextPageToken() string })`).
- 3) Take the true result and place it into a response envelope along with
other parts of the response by convention and let that be encoded and
sent as the response instead.

 ### Implementing a response envelope with `Marshaler`
My first attempt at getting my gRPC server's responses in an envelope
led me to implement my own Marshaler, I have seen this approach
discussed in https://github.com/grpc-ecosystem/grpc-gateway/issues/4483.

This does satisfy requirements 1 and 3 just fine, since the HTTP
annotations helpfully allow the code to only receive the true result,
and the Marshal interface has enough capabilities to take that and wrap
it in a response envelope.

However, requirements 1 and 2 are not _both_ satisfiable with the
current grpc-gateway code because of how the `XXX_ResponseBody()` is
called _before_ passing to the `Marshal(v)` function. This strips out
the other fields that I would normally be able to detect and place in
the response envelope.

I even tried creating my _own_ protobuf message extension that would let
me define another way of defining the "true result" field. But the
options for implementing that are either a _ton_ of protoreflect at
runtime to detect and extract that, or I am writing another protobuf
generator plugin (which I have done before [1]), but both of those
options seem quite complex.

 ### Other non-starter options
Just to get ahead of the discussion, `WithForwardResponseOption` clearly
was not meant for this use-case. At best, it seems to only be a way to
take information that might be in the response and add it as a header.

[0]: https://google.aip.dev/158#:~:text=Response%20messages%20for%20collections%20should%20define%20a%20string%20next_page_token%20field
[1]: https://github.com/nkcmr/protoc-gen-twirp_js

 ### In practice
This change fulfills my requirements by allowing logic to be inserted
right before the Marshal is called:

```go
gatewayMux := runtime.NewServeMux(
  runtime.WithForwardResponseRewriter(func(ctx context.Context, response proto.Message) (interface{}, error) {
    if s, ok := response.(*statuspb.Status); ok {
      return rewriteStatusToErrorEnvelope(ctx, s)
    }
    return rewriteResultToEnvelope(ctx, response)
  }),
)
```

 ## In this PR
This PR introduces a new `ServeMuxOption` called
`WithForwardResponseRewriter` that allows for a user-provided function
to be supplied that can take a response `proto.Message` and return `any`
during unary response forwarding, stream response forwarding, and error
response forwarding.

The implementation ensures that checks for `XXX_ResponseBody()` are
always done, even on the results returned from user-provided forward
response rewriters.

Also added `GetResponseBody()` to allow any users that use the
`WithForwardResponseRewriter` option to still rely on the `runtime` to
know how to extract the configured `response_body`.

The code generation was also updated to make the `XXX_ResponseBody()`
response wrappers embed the concrete type instead of just
`proto.Message`. This allows any code in response rewriter functions to
be able to have access to the original type, so that interface checks
against it should pass as if it was the original message.

 ## Testing
Added some basic unit tests to ensure Unary/Stream and error handlers
invoke `ForwardResponseRewriter` correctly.

  ## Looking for feedback on
* The function signature of `ForwardResponseRewriter` is very light at
the moment. Input about what else might be needed here to serve
potential _future_ uses of this are appreciated.


johanbrandhorst (changes requested):
Just having a brief look at this before diving deeper, but you'll need to regenerate the files in the repo using the new template.go logic. See `CONTRIBUTING.md` in the root of the repo for the steps.

johanbrandhorst:
Thanks for your thoughtful and detailed description. I'm hoping to exhaust all options before going for new functionality, as I'm sure you understand. I had some crazy ideas to throw out there:

1. With your custom marshaler, can't you just not use the `response_body` option? You could make the marshaler switch on the type and pick the field you want to be in the envelope. Even the custom extension doesn't sound all that bad (protoreflect is good, actually).
1. It seems like you want to add a static type wrapper around all messages? Have you considered using a gRPC client-side interceptor on the gateway side? I don't even know if it's possible to change the type returned from an interceptor, but thought it'd be worth suggesting.

nkcmr:
Hi @johanbrandhorst!

Thanks for the feedback. I have run the code-generators and pushed those changes up (sorry for missing that step and just breezing by the contributing docs).

---

Let me address your other feedback:

> Have you considered using a gRPC client-side interceptor on the gateway side?

So, I might be misunderstanding what exactly you're recommending here, but I think you're recommending [`grpc.UnaryClientInterceptor`](https://pkg.go.dev/google.golang.org/grpc@v1.65.0#UnaryClientInterceptor) be configred on the Gateway → gRPC-Server end. However, I am confused by how exactly I would be able to go from Response A (the original response of the method) to Response B (My response envelope) while using this:
```go
grpc.NewClient(
	"localhost:8334",
	grpc.WithUnaryInterceptor(func(ctx context.Context, method string, req, reply any, cc *grpc.ClientConn, invoker grpc.UnaryInvoker, opts ...grpc.CallOption) error {
		err := invoker(ctx, method, req, reply, cc, opts...)
		if err != nil {
			// uh..., how do I return my error response envelope?
		}
		// uh..., how do return my normal response envelope?
	}),
)
```
The interceptor does not allow for a different response to be returned. Again, I could be misunderstanding or missing something obvious, apologies if so.

Moreover though, my current approach to using `grpc-gateway` makes use of the registration functions that let you pass in the server implementation directly without using a gRPC connection; thus there is no way to add an interceptor there.

---

> With your custom marshaler, can't you just not use the response_body option? You could make the marshaler switch on the type and pick the field you want to be in the envelope.

This would technically work, but I think it runs contrary to my goals with this change. I'll elaborate: I am actually trying to find a way to make this pattern repeatable across a number of teams inside my company. A lot of teams implement microservices that are upstream of a Load Balancer, and all of those services implement the response envelope convention.

My "vision" is to broaden adoption of gRPC but also remain accessible in the REST/JSON world. `grpc-gateway` is quite perfect to automate this backwards compatibility. The only missing piece is some company-internal library that can configure the `*runtime.ServeMux` just right to make everything else work "out-of-the-box." And thus I am here with this proposed change. It would fit perfectly into an off-the-shelf library that would allow teams to quickly/effortlessly adopt gRPC.

**Exec Summary:** That approach _does_ work, but it does not "scale" well when the aim is to get large groups of separate people to adopt it; to have a manual tedium of needing to hand-implement this sort of neutralizes a lot of the benefits.

I fully understand that my "starry-eyed dreams" of some internal company experiment are _not_ the concern of this project 😅. That being said, I think the proposal made here is a net-benefit for the usefulness/value of this project. Having skimmed the past issues, I do get a sense that it is a missing piece (just my perspective).

Lastly, `protoreflect` isn't _bad_, but the scale of some internal services would disallow so much unnecessary runtime CPU cycles to be spent doing that. I would more than likely end up opting for a new `protoc` plugin to stave off those concerns; but I still believe that would be much more complexity than desired.

johanbrandhorst:
Thanks for elaborating. I expected that this would come down to wanting to make this scalable, which I agree that manual marshaler logic probably isn't.

I want to try to narrow down exactly what the proposal here is though, because it's also not a "give me a way to do anything with the response", which is essentially what the custom marshaler does. This seems to me to be almost an interceptor/middleware/decorator in the middle of the gateway logic, and I don't know how applicable it is to the general user. It seems like it solve this very particular problem of introducing a message envelope, dynamically (rather than statically in your protos), but I'm not yet convinced that it's generally useful. Do you think you could find issues in our backlog that could be resolve by this functionality?

I want this feature to be both generally useful and orthogonal to other functionality of the gateway. I'm worried that this option will be confusing when we have `WithForwardResponseOption`. How would you see them interacting? When would a user want to use both?

nkcmr:
In my mental model of `grpc-gateway`'s configuration surface, there are a few elements that allow control over what the response looks like:

* `WithForwardResponseOption`: This lets the user control headers and status if need be (maybe add `X-Request-ID` or switch the response code to be `201 Created`)
* `WithMarshalerOption`: This lets the user control the byte-for-byte representation of the response; its encoding.
  * Note: Given what I was trying to do and what previous issue discussions were recommending, the requirement of configuring my custom marshaler with a MIME type was a source of confusion/friction. I was not in a mindset where I was needing to think about the actual _encoding_, just the shape/type.
* `WithStreamErrorHandler` / `WithProtoErrorHandler`: This lets the user control nearly every aspect of an HTTP response when an error occurs. 
  * Note: This contrasts with the aforementioned things, in that it _does_ give the user full control over the response. Other options are simply small escape-hatches tucked into immutable implementations.

My proposal of `WithForwardResponseRewriter` fills the need/gap in between `WithForwardResponseOption` and `WithMarshalerOption`.

`ForwardResponseOption` allows control over the headers and status, but not the response _body_. `Marshaler` is intended to let the gateway understand different `Accept` headers and adapt the response's _encoding_ to meet that requirement. The missing piece is allowing for the response's shape/type to be configured at a server level.

So, with that, I do believe it is "orthogonal" to other functionality and fills a missing piece.

---

The other extreme alternative to all this (which I actually do not think is a good idea) is to allow _full_ control over the response. A bad direction (IMO) would be to allow configuration over what gets assigned/used in the generated code here:
```go
// myservice.pb.gw.go
var (
	// Let this be configured to be something else?
	forward_MyService_ListStuff_0 = runtime.ForwardResponseMessage
)
```

I don't think this is a good approach though. Letting `grpc-gateway` encapsulate most of the implementation details of handling trailing headers and picking the right marshaler is actually good. Then simply have escape-hatches to control high-level things like Supplemental Headers, Status Code, Response Shape/Type, and Encoding/Decoding.

---

Some past issues that I think this helps:
#1610: This person was trying to do pretty much the exact same thing as me. This proposal would have suited their ask perfectly.
#4483: I do think this person would have been helped by this as well, their use case did happen to fit into the Marshaler though, still awkwardly though IMO because of the difference between "encoding concerns" and "type/shape concerns."
#517: I think this person could use this proposal to take a special error `*status.Status` to make the response empty (switch it to `*httpbody.HttpBody`) and use `ForwardResponseOption` to set the right status code.

---
Sorry for the meandering response, but to summarize: There is (IMO) a missing piece to response control. I think this is the right direction to go in (Just one more escape-hatch, not necessarily allowing _full_ response control), and this approach does not overlap with anything currently available. This proposal also I believe just lets you control the shape/type of the response, not "give me a way to do anything with the response." 🙂 

**Side-note:** I see "change requested" still, did I address that? Also the `CI / lint (pull_request)` job is failing on things I did not seem to touch, is there a way to fix that?

johanbrandhorst:
The lint failure is not your fault. I will take time to review this when I have a chance. Thanks for your reply.

runtime/handler.go
```diff
@@ -137,6 +138,17 @@ type responseBody interface {
...
+// GetResponseBody will return the field that was configured in the "response_body" annotation for this specific
+// response type. The main use of this is if the [WithForwardResponseRewriter] ServeMuxOption is used, those custom
+// response rewriters can still rely on grpc-gateway to understand how to retrieve the configured response body by
+// calling this.
+func GetResponseBody(msg proto.Message) interface{} {
+	if rb, ok := msg.(responseBody); ok {
+		return rb.XXX_ResponseBody()
+	}
+	return msg
+}
```
johanbrandhorst, line 150: I'm not convinced this needs to be part of the PR. It's a convenience function that could be added at a later stage if necessary, IMO.

runtime/mux.go
```diff
@@ -75,6 +76,30 @@ type ServeMux struct {
...
+func WithForwardResponseRewriter(fwdResponseRewriter ForwardResponseRewriter) ServeMuxOption {
+	return func(sm *ServeMux) {
+		sm.forwardResponseRewriter = func(ctx context.Context, response proto.Message) (interface{}, error) {
+			result, err := fwdResponseRewriter(ctx, response)
+			if err != nil {
+				return nil, err
+			}
+			if rb, ok := result.(responseBody); ok {
+				return rb.XXX_ResponseBody(), nil
+			}
```
johanbrandhorst, line 97: It's a bit confusing to me to have business logic like this in one of these options. Is there any reason we can't return this to the respective handler functions instead?

runtime/mux.go
```diff
@@ -75,6 +76,30 @@ type ServeMux struct {
...
 
+// ForwardResponseRewriter is the signature of a function that is capable of rewriting messages
+// before the are forwarded in unary response, stream response, and error responses.
+type ForwardResponseRewriter func(ctx context.Context, response proto.Message) (interface{}, error)
+
+// WithForwardResponseRewriter returns a ServeMuxOption that allows for implementers to insert logic
+// that can rewrite the final response before it is forwarded.
+//
+// The response rewriter function is called during unary message forwarding, stream message
+// forwarding and when errors are being forwarded.
```
johanbrandhorst, line 87: Could you add a section to say that using this option to modify the response will likely make it incompatible with the schema generated by `protoc-gen-openapiv2` please?

johanbrandhorst (changes requested):
Alright, I buy it. BUT, I'm going to make you work for it. This is probably one of the biggest feature additions in a year or more. Could you please add a section on this new functionality to our docs page on customizing the gateway: https://github.com/grpc-ecosystem/grpc-gateway/blob/main/docs/docs/mapping/customizing_your_gateway.md? Thanks!

runtime/handler.go
nkcmr: That is fine. I can remove it. I think my thought process here was "Is `XXX_ResponseBody()` internal to `grpc-gateway`?" — If it needed to be changed for some reason, would that count as a SemVer major change?

runtime/mux.go
nkcmr: Sure. That sounds better.

runtime/mux.go
nkcmr: Done!

runtime/mux.go
```diff
@@ -75,6 +76,24 @@ type ServeMux struct {
 // ServeMuxOption is an option that can be given to a ServeMux on construction.
 type ServeMuxOption func(*ServeMux)
 
+// ForwardResponseRewriter is the signature of a function that is capable of rewriting messages
+// before the are forwarded in unary response, stream response, and error responses.
```
johanbrandhorst, line 80:
```suggestion
// before they are forwarded in a unary, stream, or error response.
```

runtime/mux.go
```diff
@@ -75,6 +76,24 @@ type ServeMux struct {
...
+type ForwardResponseRewriter func(ctx context.Context, response proto.Message) (interface{}, error)
```
johanbrandhorst, line 81:
```suggestion
type ForwardResponseRewriter func(ctx context.Context, response proto.Message) (any, error)
```

johanbrandhorst (changes requested):
This looks great, just a few minor changes

johanbrandhorst (approved):

nkcmr:
Thank you so much, @johanbrandhorst! Especially for putting up with my short novel comments. 😆

johanbrandhorst:

johanbrandhorst:
I just cut v2.22.0 with this new feature. Thanks again :).

PR #4813, opened 2024-10-09
feat: add expand_slashed_path_patterns flag

### LINKED ISSUE #4784: Openapi generator - explode the AIP style path parameters into URL with multiple path parameters
## 🚀 Feature

We are building our API according to [Google AIP](https://google.aip.dev/general), so our URL's contains typically just one path parameter, either `{name}` or `{parent}` which contains multiple URI segments, some of them dynamic, described via `pattern`, e.g.

```protobuf
service DocumentService {
  rpc ExportResults(ExportResultsRequest) returns (ExportResultsResponse) {
    option (google.api.http) = {
      post: "/v1/{name=projects/*/documents/*}:exportResults"
      body: "*"
    };
}
```

where the URI have just `name` parameter, but this one _hides_ unnamed `projectId` and `documentId` parameters inside. This generates.

```JSON
{
 "/v1/{name}:exportResults": {
      "post": {
        "operationId": "DocumentService_ExportResults",
        "responses": {
          "200": {
            "schema": {
              "$ref": "#/definitions/v1ExportResultsResponse"
            }
          },
          "default": {
            "schema": {
              "$ref": "#/definitions/rpcStatus"
            }
          }
        },
        "parameters": [
          {
            "name": "name",
            "in": "path",
            "required": true,
            "type": "string",
            "pattern": "projects/[^/]+/documents/[^/]+"
          },
          {
            "name": "body",
            "in": "body",
            "required": true,
            "schema": {
              "$ref": "#/definitions/DocumentServiceExportResultsBody"
            }
          }
        ]
      }
    }
}
```

You see, the URL of the resource `"/v1/{name}:exportResults"` is blurry for a documentation purpose. The fact, that the parameter `name` contains more dynamic URI segments is described in the parameters `pattern`, but again, not super readable.

This also causes a problem with OAS tooling, as path parameters in OpenAPI generally don't allow (unescaped) forward slashes, so many tools in OpenAPI space, that conform to OpenAPI spec, don't support parameters like that.

The problem with this is mentioned in the [Documentation > Customizing openapi output > path parameters](https://grpc-ecosystem.github.io/grpc-gateway/docs/mapping/customizing_openapi_output/#path-parameters), where it states

> Note that path parameters in OpenAPI does not support values with /, as discussed in [Support for path parameters which can contain slashes #892](https://github.com/OAI/OpenAPI-Specification/issues/892), so tools as Swagger UI will URL encode any / provided as parameter value. A possible workaround for this is to write a custom post processor for your OAS file to replace any path parameter with / into multiple parameters.

I was thinking briefly about building the post-processor, but it seems to me like a good candidate for a built-in feature, since all people following Google AIP style have the same problem and will most probably end up with the same post-processor design, so why don't provide this feature in the generator? Or do you know any open-source post-processor available?

czabaj:

Closes #4784 - a feature request

Yes 👍 

This adds an experimental flag into the OpenAPI generator, that expands path patterns containing URI sub-paths into the URI with new path parameters. This improves the readability and OpenAPI compatibility of protobuf APIs based on Google AIP. For more context read the [linked issue](https://github.com/grpc-ecosystem/grpc-gateway/issues/4784).

### DIFF
--- a/protoc-gen-openapiv2/internal/genopenapi/template.go
+++ b/protoc-gen-openapiv2/internal/genopenapi/template.go
@@ -1157,7 +1157,7 @@ func renderServiceTags(services []*descriptor.Service, reg *descriptor.Registry)
 // Returns:
 //
 //	The modified pathParts and pathParams slice.
-func expandPathPatterns(pathParts []string, pathParams []descriptor.Parameter) ([]string, []descriptor.Parameter) {
+func expandPathPatterns(pathParts []string, pathParams []descriptor.Parameter, reg *descriptor.Registry) ([]string, []descriptor.Parameter) {
 	expandedPathParts := []string{}
 	modifiedPathParams := pathParams
 	for _, pathPart := range pathParts {
@@ -1192,6 +1192,9 @@ func expandPathPatterns(pathParts []string, pathParams []descriptor.Parameter) (
 			}
 			lastPart := expandedPathParts[len(expandedPathParts)-1]
 			paramName := strings.TrimSuffix(lastPart, "s")
+			if reg.GetUseJSONNamesForFields() {
+				paramName = casing.JSONCamelCase(paramName)
+			}
 			expandedPathParts = append(expandedPathParts, "{"+paramName+"}")
 			newParam := descriptor.Parameter{
 				Target: &descriptor.Field{
@@ -1248,7 +1251,7 @@ func renderServices(services []*descriptor.Service, paths *openapiPathsObject, r
 				parts := templateToParts(b.PathTmpl.Template, reg, meth.RequestType.Fields, msgs)
 				pathParams := b.PathParams
 				if reg.GetExpandSlashedPathPatterns() {
-					parts, pathParams = expandPathPatterns(parts, pathParams)
+					parts, pathParams = expandPathPatterns(parts, pathParams, reg)
 				}
 				// extract any constraints specified in the path placeholders into ECMA regular expressions
 				pathParamRegexpMap := partsToRegexpMap(parts)
--- a/protoc-gen-openapiv2/internal/genopenapi/template.go
+++ b/protoc-gen-openapiv2/internal/genopenapi/template.go
@@ -978,8 +978,18 @@ func resolveFullyQualifiedNameToOpenAPINames(messages []string, namingStrategy s
 
 var canRegexp = regexp.MustCompile("{([a-zA-Z][a-zA-Z0-9_.]*)([^}]*)}")
 
-// templateToParts will split a URL template as defined by https://github.com/googleapis/googleapis/blob/master/google/api/http.proto
-// into a string slice with each part as an element of the slice for use by `partsToOpenAPIPath` and `partsToRegexpMap`.
+// templateToParts splits a URL template into path segments for use by `partsToOpenAPIPath` and `partsToRegexpMap`.
+//
+// Parameters:
+//   - path:	The URL template as defined by https://github.com/googleapis/googleapis/blob/master/google/api/http.proto
+//   - reg:	The descriptor registry used to read compiler flags
+//   - fields:	The fields of the request message, only used when `useJSONNamesForFields` is true
+//   - msgs:	The Messages of the service binding, only used when `useJSONNamesForFields` is true
+//   - pathParams:	The path parameters of the service binding, only used when `expandSlashedPathPatterns` is true
+//
+// Returns:
+//
+//	The path segments of the URL template. When `expandSlashedPathPatterns` is true, also mutates the pathParams slice.
 func templateToParts(path string, reg *descriptor.Registry, fields []*descriptor.Field, msgs []*descriptor.Message, pathParams *[]descriptor.Parameter) []string {
 	// It seems like the right thing to do here is to just use
 	// strings.Split(path, "/") but that breaks badly when you hit a url like
--- a/protoc-gen-openapiv2/internal/genopenapi/template.go
+++ b/protoc-gen-openapiv2/internal/genopenapi/template.go
@@ -1156,19 +1156,19 @@ func renderServiceTags(services []*descriptor.Service, reg *descriptor.Registry)
 	return tags
 }
 
-// expandPathPatterns search the URI parts for path parameters with pattern and when the pattern contains a sub-path,
+// expandPathPatterns searches the URI parts for path parameters with pattern and when the pattern contains a sub-path,
 // it expands the pattern into the URI parts and adds the new path parameters to the pathParams slice.
 //
 // Parameters:
 //   - pathParts:	the URI parts parsed from the path template with `templateToParts` function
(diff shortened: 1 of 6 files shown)


protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -1004,14 +1015,72 @@ pathLoop:
...
+			})
+			if pathParamIndex == -1 {
+				panic(fmt.Sprintf("Path parameter %q not found in path parameters", paramName))
+			}
+			pathParam := (*pathParams)[pathParamIndex]
+			patternParts := strings.Split(pattern, "/")
+			for i, part := range patternParts {
+				modifiedPart := part
+				if part == "*" {
+					paramName := strings.TrimSuffix(parts[len(parts)-1], "s")
```
czabaj, line 1051: The wildcards from the path pattern create new path parameters, I decided to infer a name for these parameters from the previous path segment, as this works well for _my_ current use case, but I understand that this problematic for many reasons so I would like to discuss how to name the new path params?

johanbrandhorst: Yeah, obviously that heuristic isn't always going to work. Not that I have a better suggestion to be honest. I think most users of this pattern will be using the AIP guidelines, so this should work for them. Lets just make sure we document this clearly.

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -1004,14 +1015,72 @@ pathLoop:
...
+						*pathParams = append((*pathParams)[:pathParamIndex], (*pathParams)[pathParamIndex+1:]...)
+						pathParamIndex = -1
+					}
+				}
+				isLastPart := i == len(patternParts)-1
+				if isLastPart {
+					buffer = modifiedPart
+				} else {
+					parts = append(parts, modifiedPart)
+				}
```
johanbrandhorst, line 1083: This is a lot of logic to add to an already complicated function. Do you think you could break this out into a separate function with separate tests, so we can cover the different scenarios more easily?

johanbrandhorst (changes requested):
This is a great start. Could you please make sure to also:
1. Add this new flag to the bazel rules for the openapiv2 generator. Just follow the pattern for the other flags.
2. Add a small section to our docs about this new flag with some example use cases? The file you want is https://github.com/grpc-ecosystem/grpc-gateway/blob/main/docs/docs/mapping/customizing_openapi_output.md.

Thanks!

czabaj:
I created a single clean function that patches the path parts & params only when the flag is turned on. It consumes the parts from the `templateToParts` function which I reverted to the original shape. During running the test, I encountered a bug in `templateToParts`, or at least, I think it is a bug.

If you pass into the `templateToParts` URI template with a path param with pattern, and that pattern uses a snake-case URI, like

```
/test/{name=test_cases/*}/
```
 
and the `useJSONNamesForFields` is turned on, then this is split into parts

```
test
name=testCases/*
```

so the _path pattern is camelcased_. I believe this is wrong, since the only path parameter names should be camelcased, not other parts of the URI - am I right?

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -1145,6 +1156,78 @@ func renderServiceTags(services []*descriptor.Service, reg *descriptor.Registry)
...
+//
+// Parameters:
+//   - pathParts:	the URI parts parsed from the path template with `templateToParts` function
+//   - pathParams: the path parameters of the service binding, this slice will be mutated when the path pattern contains
+//     a sub-path with wildcard.
+//
+// Returns:
+//
+//	The modified pathParts. Also mutates the pathParams slice.
+func expandPathPatterns(pathParts []string, pathParams *[]descriptor.Parameter) []string {
```
johanbrandhorst, line 1170: Since this mutates an input, what do you think about returning a parameter of the same type instead? It's a bit gross IMO to mutate the input slice and returning it will make it more explicit.
```suggestion
//	The modified pathParts and pathParams slice.
func expandPathPatterns(pathParts []string, pathParams []descriptor.Parameter) ([]string, []descriptor.Parameter) {
```

johanbrandhorst, line 1170: I just realized that we have a package for parsing these paths already: https://github.com/grpc-ecosystem/grpc-gateway/blob/main/internal/httprule. What do you think about reusing this?

johanbrandhorst (changes requested):
> I believe this is wrong, since only the path parameter names should be camelcased, not other parts of the URI - am I right?

Yeah this looks wrong, thanks for finding it. Can you fix it?

protoc-gen-openapiv2/internal/genopenapi/template.go
czabaj: I used that package but unfortunately, it don't fit as a replace for the `templateToParts` function, I encountered two problems

- the `templateToParts` splits the template by slashes, so template with a leading slash creates an empty string at the start of the output slice, trailing slash in the template creates an empty string at the end of the slice - I can go around this, but `httprule.Parse` do not even accept templates without trailing slash, which broke some tests - I guess that all templates should be generally with trailing slash, but I'm not certain.
- the `httprule.Parse` ignores and omits the path patterns, these are completely lost in the `Compile` output object, this was a blocker,
- the `Compile.Pool` contains template segments, but the order is not guaranteed, I encounter a test with a path param where the order of the segments was different than in the original template, this is hard to resolve,

all in all, `httprule.Parse` does a lot of stuff but most of it is not useful for the `templateToParts` and the important stuff is missing and hard to add. At least, I managed to simplify the `templateToParts` and fix the camel-casing, which wrongly affected the parameter pattern.

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -988,30 +998,24 @@ func templateToParts(path string, reg *descriptor.Registry, fields []*descriptor
 	var parts []string
 	depth := 0
 	buffer := ""
-	jsonBuffer := ""
```
czabaj, line 991: This second buffer was just a complexity gainer, I added a single branch to the `}` case, where it extracts the parameter name and camel-case it, when the camel-casing is enabled.

protoc-gen-openapiv2/internal/genopenapi/template.go
johanbrandhorst: Thanks for the explanation, that makes sense!

johanbrandhorst (changes requested):
This is really cool, as a final request, do you think you could add a little blurb about this setting to our docs? This page: https://github.com/grpc-ecosystem/grpc-gateway/blob/main/docs/docs/mapping/customizing_openapi_output.md

johanbrandhorst:
Looks like there was a test failure, I think it's from the camelcase fix you made.

johanbrandhorst:

PR #4819, opened 2024-10-10
feat: expose invalid argument error to clients in bidirectional streaming (#4795)

### LINKED ISSUE #4795: Best practices for handling decode failures in a websocket connection
## 📚 Documentation

(A clear and concise description of what the issue is.)

I wasn't able to find any documentation around handling decoding failures. I'm new to GRPC ecosystem and we have a GRPC server that needs to accept requests from clients that don't use GRPC. often times, our clients do not follow the best practices of GRPC and send invalid requests. such as

```
2024/10/02 11:02:05 ERROR: Failed to decode request: proto: (line 1:14): invalid google.protobuf.Timestamp value "2024-10-02T11.02.05.5880000Z"
2024/10/04 13:43:29 ERROR: Failed to decode request: proto: (line 1:190): invalid google.protobuf.Duration value "1.541S"
```

This a developer error but the problem I'm running into is that there is no easy way for us to communicate with our clients of these issues. As a results, these requests just get dropped in our websocket streaming connection and never reach our GRPC servers. So everytime, this happens, we can only retroactively inform our users of these issues.

Any advice on this issue would be greatly appreciated!

ianbbqzy:

Fixes #4795

yes

For bidirectional streaming, allows request decoder to alert the clients of invalid requests.

### DIFF
--- a/examples/internal/proto/examplepb/stream.pb.gw.go
+++ b/examples/internal/proto/examplepb/stream.pb.gw.go
@@ -130,13 +130,13 @@ func request_StreamService_BulkEcho_0(ctx context.Context, marshaler runtime.Mar
 		return nil
 	}
 	go func() {
+		defer close(errChan)
 		for {
 			if err := handleSend(); err != nil {
 				errChan <- err
 				break
 			}
 		}
-		close(errChan)
 	}()
 	header, err := stream.Header()
 	if err != nil {
@@ -314,9 +314,10 @@ func RegisterStreamServiceHandlerClient(ctx context.Context, mux *runtime.ServeM
 			return
 		}
 		go func() {
-			err := <-reqErrChan
-			if err != io.EOF {
-				runtime.HTTPError(annotatedContext, mux, outboundMarshaler, w, req, err)
+			for err := range reqErrChan {
+				if err != nil && err != io.EOF {
+					runtime.HTTPError(annotatedContext, mux, outboundMarshaler, w, req, err)
+				}
 			}
 			if err := resp.CloseSend(); err != nil {
 				grpclog.Errorf("Failed to terminate client stream: %v", err)
--- a/examples/internal/proto/examplepb/stream.pb.gw.go
+++ b/examples/internal/proto/examplepb/stream.pb.gw.go
@@ -136,6 +136,7 @@ func request_StreamService_BulkEcho_0(ctx context.Context, marshaler runtime.Mar
 				break
 			}
 		}
+		close(errChan)
 	}()
 	header, err := stream.Header()
 	if err != nil {
@@ -313,14 +314,12 @@ func RegisterStreamServiceHandlerClient(ctx context.Context, mux *runtime.ServeM
 			return
 		}
 		go func() {
-			for err := range reqErrChan {
-				if err != nil {
-					runtime.HTTPError(annotatedContext, mux, outboundMarshaler, w, req, err)
-					if err := resp.CloseSend(); err != nil {
-						grpclog.Errorf("Failed to terminate client stream: %v", err)
-					}
-					return
-				}
+			err := <-reqErrChan
+			if err != io.EOF {
+				runtime.HTTPError(annotatedContext, mux, outboundMarshaler, w, req, err)
+			}
+			if err := resp.CloseSend(); err != nil {
+				grpclog.Errorf("Failed to terminate client stream: %v", err)
 			}
 		}()
 
--- a/examples/internal/proto/examplepb/stream.pb.gw.go
+++ b/examples/internal/proto/examplepb/stream.pb.gw.go
@@ -110,6 +110,7 @@ func request_StreamService_BulkEcho_0(ctx context.Context, marshaler runtime.Mar
 	stream, err := client.BulkEcho(ctx)
 	if err != nil {
 		grpclog.Errorf("Failed to start streaming: %v", err)
+		close(errChan)
 		return nil, metadata, errChan, err
 	}
 	dec := marshaler.NewDecoder(req.Body)
@@ -136,6 +137,9 @@ func request_StreamService_BulkEcho_0(ctx context.Context, marshaler runtime.Mar
 				errChan <- err
 				break
 			}
+		}
+		if err := stream.CloseSend(); err != nil {
+			grpclog.Errorf("Failed to terminate client stream: %v", err)
 		}
 	}()
 	header, err := stream.Header()
@@ -318,9 +322,6 @@ func RegisterStreamServiceHandlerClient(ctx context.Context, mux *runtime.ServeM
 				if err != nil && err != io.EOF {
 					runtime.HTTPError(annotatedContext, mux, outboundMarshaler, w, req, err)
 				}
-			}
-			if err := resp.CloseSend(); err != nil {
-				grpclog.Errorf("Failed to terminate client stream: %v", err)
 			}
 		}()
 
--- a/examples/internal/proto/examplepb/stream.pb.gw.go
+++ b/examples/internal/proto/examplepb/stream.pb.gw.go
@@ -104,12 +104,13 @@ func request_StreamService_List_0(ctx context.Context, marshaler runtime.Marshal
 
 }
 
-func request_StreamService_BulkEcho_0(ctx context.Context, marshaler runtime.Marshaler, client StreamServiceClient, req *http.Request, pathParams map[string]string) (StreamService_BulkEchoClient, runtime.ServerMetadata, error) {
+func request_StreamService_BulkEcho_0(ctx context.Context, marshaler runtime.Marshaler, client StreamServiceClient, req *http.Request, pathParams map[string]string) (StreamService_BulkEchoClient, runtime.ServerMetadata, chan error, error) {
 	var metadata runtime.ServerMetadata
+	errChan := make(chan error, 1)
 	stream, err := client.BulkEcho(ctx)
 	if err != nil {
 		grpclog.Errorf("Failed to start streaming: %v", err)
-		return nil, metadata, err
+		return nil, metadata, errChan, err
 	}
 	dec := marshaler.NewDecoder(req.Body)
 	handleSend := func() error {
@@ -120,7 +121,7 @@ func request_StreamService_BulkEcho_0(ctx context.Context, marshaler runtime.Mar
 		}
 		if err != nil {
 			grpclog.Errorf("Failed to decode request: %v", err)
-			return err
+			return status.Errorf(codes.InvalidArgument, "Failed to decode request: %v", err)
 		}
 		if err := stream.Send(&protoReq); err != nil {
 			grpclog.Errorf("Failed to send request: %v", err)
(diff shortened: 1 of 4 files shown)


examples/internal/proto/examplepb/stream.pb.gw.go
```diff
@@ -306,12 +305,25 @@ func RegisterStreamServiceHandlerClient(ctx context.Context, mux *runtime.ServeM
...
+			for {
+				err := <-reqErrChan
+				if err != nil {
+					runtime.HTTPError(annotatedContext, mux, outboundMarshaler, w, req, err)
+					if err := resp.CloseSend(); err != nil {
+						grpclog.Errorf("Failed to terminate client stream: %v", err)
+					}
+					return
+				}
+			}
```
johanbrandhorst, line 325: This loop should be over the channel itself, so it terminates when the channel is closed
```suggestion
			for err := range reqErrChan {
				if err != nil {
					runtime.HTTPError(annotatedContext, mux, outboundMarshaler, w, req, err)
					if err := resp.CloseSend(); err != nil {
						grpclog.Errorf("Failed to terminate client stream: %v", err)
					}
					return
				}
			}
```

examples/internal/proto/examplepb/stream.pb.gw.go
```diff
@@ -104,12 +104,13 @@ func request_StreamService_List_0(ctx context.Context, marshaler runtime.Marshal
 
 }
 
-func request_StreamService_BulkEcho_0(ctx context.Context, marshaler runtime.Marshaler, client StreamServiceClient, req *http.Request, pathParams map[string]string) (StreamService_BulkEchoClient, runtime.ServerMetadata, error) {
+func request_StreamService_BulkEcho_0(ctx context.Context, marshaler runtime.Marshaler, client StreamServiceClient, req *http.Request, pathParams map[string]string) (StreamService_BulkEchoClient, runtime.ServerMetadata, chan error, error) {
 	var metadata runtime.ServerMetadata
+	errChan := make(chan error, 1)
```
johanbrandhorst, line 109: Who is responsible for closing this channel?

johanbrandhorst (changes requested):
Looks like there's a Bazel failure and a test failure

ianbbqzy:
Hi @johanbrandhorst , I'm having some trouble running the integration tests following the instructions. I'm not really familiar with gulp and having issue trying to run the gulp CLI commands

```
ianlee@Ians-MacBook-Pro-2 browser % gulp serve
AssertionError [ERR_ASSERTION]: Task function must be specified
    at Gulp.set [as _setTask] (/Users/ianlee/grpc-gateway/examples/internal/browser/node_modules/undertaker/lib/set-task.js:10:3)
    at Gulp.task (/Users/ianlee/grpc-gateway/examples/internal/browser/node_modules/undertaker/lib/task.js:13:8)
    at Object.<anonymous> (/Users/ianlee/grpc-gateway/examples/internal/browser/gulpfile.js:26:6)
    at Module._compile (node:internal/modules/cjs/loader:1376:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1435:10)
    at Module.load (node:internal/modules/cjs/loader:1207:32)
    at Module._load (node:internal/modules/cjs/loader:1023:12)
    at Module.require (node:internal/modules/cjs/loader:1235:19)
    at require (node:internal/modules/helpers:176:18)
    at requireOrImport (/usr/local/lib/node_modules/gulp-cli/lib/shared/require-or-import.js:20:11)
    at execute (/usr/local/lib/node_modules/gulp-cli/lib/versioned/^5.0.0/index.js:35:3)
    at onExecute (/usr/local/lib/node_modules/gulp-cli/index.js:230:24)
    at Liftoff.<anonymous> (/usr/local/lib/node_modules/gulp-cli/index.js:153:5)
    at Liftoff.execute (/usr/local/lib/node_modules/gulp-cli/node_modules/liftoff/index.js:320:14)
    at module.exports (/usr/local/lib/node_modules/gulp-cli/node_modules/flagged-respawn/index.js:52:3)
    at Liftoff.<anonymous> (/usr/local/lib/node_modules/gulp-cli/node_modules/liftoff/index.js:310:7) {
  generatedMessage: false,
  code: 'ERR_ASSERTION',
  actual: false,
  expected: true,
  operator: '=='
}
```

would you be able to help?

johanbrandhorst:
I'm... also not much of an expert unfortunately. I usually just let CI run the browser test suite 😬. It looks like the node tests passed in CI on the latest push.

examples/internal/proto/examplepb/stream.pb.gw.go
ianbbqzy: good point. The sender should be responsible after encountering an error imo.

examples/internal/proto/examplepb/stream.pb.gw.go
ianbbqzy: Makes sense. But since it seems that there will ever only be one error received before the channel closing, maybe I should just remove the for loop to make that clear?

johanbrandhorst: I like that the loop works both if the channel closes without an error and if there is one error. If there is more than one error then this will likely fail to do anything, so I still think the for loop is best. Maybe I'm missing something?

examples/internal/proto/examplepb/stream.pb.gw.go
johanbrandhorst: Yes, agreed.

examples/internal/proto/examplepb/stream.pb.gw.go
```diff
@@ -131,20 +132,19 @@ func request_StreamService_BulkEcho_0(ctx context.Context, marshaler runtime.Mar
...
 		for {
 			if err := handleSend(); err != nil {
+				errChan <- err
 				break
 			}
 		}
-		if err := stream.CloseSend(); err != nil {
-			grpclog.Errorf("Failed to terminate client stream: %v", err)
-		}
+		close(errChan)
```
johanbrandhorst, line 139: Could we do this in a defer at the top of the goroutine instead to make it even clearer?

examples/internal/proto/examplepb/stream.pb.gw.go
```diff
@@ -306,12 +306,22 @@ func RegisterStreamServiceHandlerClient(ctx context.Context, mux *runtime.ServeM
...
+
+		resp, md, reqErrChan, err := request_StreamService_BulkEcho_0(annotatedContext, inboundMarshaler, client, req, pathParams)
 		annotatedContext = runtime.NewServerMetadataContext(annotatedContext, md)
 		if err != nil {
 			runtime.HTTPError(annotatedContext, mux, outboundMarshaler, w, req, err)
 			return
 		}
+		go func() {
+			err := <-reqErrChan
+			if err != io.EOF {
```
johanbrandhorst, line 318: This will print an error if `err == nil`. I think the for loop is better in that case.

johanbrandhorst (changes requested):
Thanks for the updates, some more questions

johanbrandhorst:
It might be easier to run if you use our CI environment docker image as described in CONTRIBUTING.md, but I'm also happy to just trigger the CI run for you.

ianbbqzy:
I managed to get gulp to work by forcing it to use version 3.9.1 and also forcing the use of node v10. but don't think it helps with running the integration tests.

if i just use the docker command like
```
docker run -v $(pwd):/grpc-gateway -w /grpc-gateway --rm ghcr.io/grpc-ecosystem/grpc-gateway/build-env:1.23 \
    /bin/bash -c 'make install && \
        make clean && \
        make generate && make test'
```

I get similar errors like `2024/10/17 22:11:19 ERROR: http error 502 request body` as in the github CI environment
along with errors like
```
--- FAIL: TestNonStandardNames (20.04s)
    --- FAIL: TestNonStandardNames/Test_standard_update_method (0.02s)
        integration_test.go:2405: {"code":14,"message":"name resolver error: produced zero addresses","details":[]}
        integration_test.go:2408: patchResp.StatusCode= 503; want 200 resp: {"code":14,"message":"name resolver error: produced zero addresses","details":[]}
        integration_test.go:2414: marshaler.Unmarshal failed: proto: (line 1:2): unknown field "code"
```

note that I'm running this on the main branch so it's not related to my changes in this PR, which would've run into the same issues

johanbrandhorst:
Please ignore the gulp tests. It looks like there might be a problem with them, but I don't think this PR will change anything. As you can see, the `go test ./...` CI job is failing right now, which is more concerning. Looks like `TestABE` failed to run:

```

**Comment by ianbbqzy** [comment]
Sorry if I'm misunderstanding the testing structure, but from my perspective, `TestABE ` is part of the integration tests, which I tried to run in the docker command with `Make test` which in turns calls `go test -race ./examples/internal/integration -args -network=unix -endpoint=test.sock`. 

So I would like to debug the test that's failing, but it would be hard without being able to actually run the integration tests properly from my machine.

**Comment by johanbrandhorst** [comment]
All that sounds right, but that doesn't actually require running `gulp`, does it? On `main`, I can run this test locally with:

```
$ go test ./examples/internal/integration/ -run '^TestABE$'
```

Does that work for you?

File: examples/internal/proto/examplepb/stream.pb.gw.go [hunk 5]
```diff
@@ -306,12 +310,20 @@ func RegisterStreamServiceHandlerClient(ctx context.Context, mux *runtime.ServeM
...
+		resp, md, reqErrChan, err := request_StreamService_BulkEcho_0(annotatedContext, inboundMarshaler, client, req, pathParams)
 		annotatedContext = runtime.NewServerMetadataContext(annotatedContext, md)
 		if err != nil {
 			runtime.HTTPError(annotatedContext, mux, outboundMarshaler, w, req, err)
 			return
 		}
+		go func() {
+			for err := range reqErrChan {
+				if err != nil && err != io.EOF {
+					runtime.HTTPError(annotatedContext, mux, outboundMarshaler, w, req, err)
```
johanbrandhorst, line 323: Is it safe to call this function with the shared `w`? Is there some way we can avoid doing this concurrently (a defer? or something).

ianbbqzy: so testing this implementation with my server works fine but I agree that might be more optimal solution.

I'm not sure about the concern around sharing `w`, the timing of this invalid argument error and the timing of a stream of valid responses could happen concurrently.  I don't see any better indicators for when it's appropriate to send the error the client.

But one thing I noticed is that `runtime.HTTPError` modifies the header, which means the header could be modified twice.  I'm thinking about writing the error directly using `w.Write()`. My thinking is that if the error occurs in the middle of the stream, it makes sense that the original response header stays at 200.

I have been trying to modify the [integration test](https://github.com/grpc-ecosystem/grpc-gateway/blob/b18dc950f6c8952da46f8a88689f0e8e190a2e05/examples/internal/integration/integration_test.go#L1341) to test it out but I found that the current test set up waits for the entire request body to be written before actually making the http request. I would expect the response collection to start without waiting for the first goroutine of requests writing to complete. Unfortunately I haven't figured out why that isn't the case yet so it's hard to test the scenario of error occurring in the middle of a stream.

ianbbqzy:
BTW, the vscode dev containers worked like a charm. Shouldn't have fixated on the docker commands and I assumed it would be the same issues with the dev containers.

examples/internal/proto/examplepb/stream.pb.gw.go
johanbrandhorst: It's fine if we can't easily test it. I don't think setting the header will matter once we've started writing to the body, since we can't write the header again. Given that, this is probably fine :). Thanks for testing it manually.

johanbrandhorst (approved):

examples/internal/proto/examplepb/stream.pb.gw.go
ianbbqzy: Hey Johan, thanks for approving this, but I still find that using `ErrorHandler` might not be ideal. PTAL at https://github.com/grpc-ecosystem/grpc-gateway/pull/4864. Let's move the discussion over as well

Issue #4865, opened 2024-10-23
Path params generated with `expand_slashed_path_patterns=true` have type `number`


czabaj:
## 🐛 Bug Report

This is a follow-up to the issue https://github.com/grpc-ecosystem/grpc-gateway/issues/4784 (closed PR https://github.com/grpc-ecosystem/grpc-gateway/pull/4813) - the path parameters generated with the `expand_slashed_path_patterns=true` compiler option have an OpenAPI type `type: number` with `format: double`.

## To Reproduce

Have a protobuf definition with Google AIP style path parameters, e.g. 

```proto
rpc GetBook(GetBookRequest) returns (Book) {
  option (google.api.http) = {
    get: "/v1/{name=publishers/*/books/*}"
  };
}
```

and enable the new  `expand_slashed_path_patterns=true` compiler option (not yet released).

When you generate the OpenAPI file from this, it will generate

```JSON
{
 "/v1/publishers/{publisher}/books/{book}": {
      "get": {
        "parameters": [
          {
            "name": "publisher",
            "in": "path",
            "required": true,
            "type": "number",
            "format": "double"
          },
 {
            "name": "book",
            "in": "path",
            "required": true,
            "type": "number",
            "format": "double"
          },
        ]
      }
    }
}
```

The path parameters `publisher` and `book` are inferred from the proto pattern, but their type is `number` - it should be a `string`.

## Expected behavior

The path parameters generated when expanding the path pattern should have a type of `string`.

## Actual Behavior

The path parameters generated when expanding the path pattern have a type `number` with format `double`.

austingebauer:
Thanks. Also looking for this to land in a release soon!

johanbrandhorst:
https://github.com/grpc-ecosystem/grpc-gateway/releases/tag/v2.23.0 has been released!

PR #4866, opened 2024-10-23
fix: path parametrs of type number

czabaj:
path parameters generated with `expand_slashed_path_patterns` are of type `number`, should be a `string`

Closes https://github.com/grpc-ecosystem/grpc-gateway/issues/4865
Relates https://github.com/grpc-ecosystem/grpc-gateway/issues/4784

Yes

All path parameters generated with `expand_slashed_path_patterns` have a type `string`.

#### Other comments

I fixed that by explicitly setting the type of the newly generated parameter to `string`. I tested that manually with my real API. I tried to write a test for this, but it turned out to be quite complicated


johanbrandhorst (commented):
Doh, that makes sense! Is there a test we can use to assert that this is always a string?

johanbrandhorst (approved):

Issue #4901, opened 2024-11-04
Skip Writing to Response When the Response is Empty


reversearrow:
## 🚀 Feature

The `ForwardResponseMessage` function currently calls the [HTTP response write function](https://github.com/grpc-ecosystem/grpc-gateway/blob/main/runtime/handler.go#L197) even when the response is empty. This occurs when sending a response with a 204 No Content status code.

As a result, the program repeatedly logs the following error:
`2024/11/04 10:49:00 ERROR: Failed to write response: http: request method or response status code does not allow body` 

This proposed change introduces a check to determine if the response is of type grpc empty (emptypb.Empty). If the response is empty, it will skip writing to the response body, preventing the error from occurring.
```
if _, ok := resp.(*emptypb.Empty); !ok {
	if _, err = w.Write(buf); err != nil {
		grpclog.Errorf("Failed to write response: %v", err)
}
```

Issue #240, opened 2016-10-20
allow responding with http status 204


ericlagergren:
google.protobuf.Empty indicates an empty JSON response. I'd _prefer_ to send a 204 for services that respond with google.protobuf.Empty.

Right now all I could come up with is adding a `WithForwardResponseOption` and checks the message's type and calls `WriteHeader(http.StatusNoContent)` if the message is `*empty.Empty`.

This causes the call to `w.Write` (https://github.com/grpc-ecosystem/grpc-gateway/blob/7ba755f85a3dc73224047317dab675b32d3a91e0/runtime/handler.go#L115) to print an error.

Technically it all works, but it feels hacky.

tmc:
It does seem like this should be achievable in a fairly clean fashion.

@EricLagergren what's the error that subsequent Write generates?

EricLagergren:
No error is returned, instead one is printed out. (See: https://golang.org/src/net/http/server.go#L879)

afking:
@tmc is there currently no way to propagate HTTP 2** status codes to the rest proxy?

PR #4902, opened 2024-11-04
feat: skip calling http Write method when response is of empty type

reversearrow:
https://github.com/grpc-ecosystem/grpc-gateway/issues/4901
https://github.com/grpc-ecosystem/grpc-gateway/issues/240

Yes

Skip calling the `Write()` method on the HTTP response writer when the response is empty. This is to avoid code from keep printing the error message below. 

```
2024/11/04 10:49:00 ERROR: Failed to write response: http: request method or response status code does not allow body
```

### DIFF
--- a/runtime/handler.go
+++ b/runtime/handler.go
@@ -15,6 +15,7 @@ import (
 	"google.golang.org/grpc/grpclog"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/emptypb"
 )
 
 // ForwardResponseStream forwards the stream from gRPC server to REST client.
@@ -194,8 +195,10 @@ func ForwardResponseMessage(ctx context.Context, mux *ServeMux, marshaler Marsha
 		w.Header().Set("Content-Length", strconv.Itoa(len(buf)))
 	}
 
-	if _, err = w.Write(buf); err != nil {
-		grpclog.Errorf("Failed to write response: %v", err)
+	if _, ok := resp.(*emptypb.Empty); !ok {
+		if _, err = w.Write(buf); err != nil {
+			grpclog.Errorf("Failed to write response: %v", err)
+		}
 	}
 
 	if doForwardTrailers {
--- a/runtime/handler.go
+++ b/runtime/handler.go
@@ -15,7 +15,6 @@ import (
 	"google.golang.org/grpc/grpclog"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/proto"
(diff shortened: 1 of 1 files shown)


reversearrow:
@johanbrandhorst can you please review this PR? Thank you

runtime/handler.go
```diff
@@ -194,8 +195,10 @@ func ForwardResponseMessage(ctx context.Context, mux *ServeMux, marshaler Marsha
 		w.Header().Set("Content-Length", strconv.Itoa(len(buf)))
 	}
 
-	if _, err = w.Write(buf); err != nil {
-		grpclog.Errorf("Failed to write response: %v", err)
+	if _, ok := resp.(*emptypb.Empty); !ok {
+		if _, err = w.Write(buf); err != nil {
+			grpclog.Errorf("Failed to write response: %v", err)
+		}
```
johanbrandhorst, line 201: I understand the problem, but I don't think this is the right conditional. The error is because we're trying to write a body when the request method doesn't allow it. Shouldn't we just check if `req.Method == http.MethodGet` or something like that?

johanbrandhorst (changes requested):
Could you please add a test for this behavior? It might be possible to capture the log output in an integration test (https://github.com/grpc-ecosystem/grpc-gateway/blob/main/examples/internal/integration/integration_test.go).

runtime/handler.go
reversearrow: Thanks @johanbrandhorst for the review. 

The error occurs because status codes 204 and 304 do not allow a response body. 

Ideally, we should check if the response status is one of these codes before attempting to write a body. However, there isn’t a straightforward way to retrieve the status code from http.ResponseWriter after it has been written.

Also, req.Method == http.MethodGet alone is not sufficient, as a GET request can still return a non-empty response body. We could add a check for DELETE method calls, but again, there’s no guarantee that they will always have an empty response.

johanbrandhorst: OK, alternative idea. How about we intercept the error and just don't log it?
```suggestion
	if _, err = w.Write(buf); err != nil && !errors.Is(err, http.ErrBodyNotAllowed) {
		grpclog.Errorf("Failed to write response: %v", err)
	}
```

johanbrandhorst (changes requested):

runtime/handler.go
reversearrow: Works for me, thank you. Addressed it here [ff571ca](https://github.com/grpc-ecosystem/grpc-gateway/pull/4902/commits/ff571ca2c37de587920d298138199eac6ddfed28)

johanbrandhorst (approved):

Issue #581, opened 2018-03-23
Streamed response is not valid json (or: is this the expected format?)


colini:
I've written a simple a simple proto that includes this rpc with a stream response:
```protobuf
rpc GetSomeData(GetSomeDataRequest) returns (stream SomeData) {
  option (google.api.http) = {
    get: "/v1beta1/getData:raw"
  };
};
```

My server replies with 3 SomeData records.  The result from curl via my generated grpc-gateway is:
```
$ curl -i 'http://localhost:8080/v1beta1/getData:raw'
HTTP/1.1 200 OK
Content-Type: application/json
Grpc-Metadata-Content-Type: application/grpc
Grpc-Metadata-Grpc-Accept-Encoding: gzip
Date: Fri, 23 Mar 2018 12:30:50 GMT
Transfer-Encoding: chunked

{"result":{"fund":"FUND_0","value":123.99}}
{"result":{"fund":"FUND_1","value":123.99}}
{"result":{"fund":"FUND_2","value":123.99}}
```
This appears to be newline-delimited json, which is fine, but it's not valid json.

AFAIK I'm using the latest grpc-gateway code, which I installed per the instructions:
```
go get -u github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway
go get -u github.com/grpc-ecosystem/grpc-gateway/protoc-gen-swagger
go get -u github.com/golang/protobuf/protoc-gen-go
```
With go 1.10
```
$ go version
go version go1.10 darwin/amd64
```

Is this the expected stream response format?  I'd suggest enclosing the results in a list, which would add only 1 byte per line (the comma), and would then be valid json:
```json
[
{"result":{"fund":"FUND_0","value":123.99}},
{"result":{"fund":"FUND_1","value":123.99}},
{"result":{"fund":"FUND_2","value":123.99}}
]
```

(Sorry if this is a repeat of #481, but it wasn't clear how that issue was resolved.)

yugui:
Streaming requests and responses are mapped to JSON streaming by design.

https://en.m.wikipedia.org/wiki/JSON_streaming

colini:
@yugui Thank you for the clarification.

veqryn:
@yugui shouldn't the content type of the response be `application/x-ndjson` instead of `application/json`?

veqryn:
Sure:
http://ndjson.org/

and
https://github.com/ndjson/ndjson-spec

[...]

PR #4926, opened 2024-11-08
Optional specification of stream content type

huin:
We have grpc-gateway based services that include both unary and streamed responses, using `JSONPb` as the marshaler. There is a niggle where the response type for the streamed response is `application/json`, but we'd like it to be `application/x-ndjson` in the streamed response case.

This looks like it would have provided a solution for at least one comment on #581 (https://github.com/grpc-ecosystem/grpc-gateway/issues/581#issuecomment-423053891 in particular).

Yes. This does not appear to need file regeneration.

Support specifying streamed response content types, without changing the existing behaviour of grpc-gateway unless consumers opt in by providing their own `Marshaler` to set the stream content type.

### DIFF
--- a/runtime/handler_test.go
+++ b/runtime/handler_test.go
@@ -178,36 +178,68 @@ func (c *CustomMarshaler) NewDecoder(r io.Reader) runtime.Decoder     { return c
 func (c *CustomMarshaler) NewEncoder(w io.Writer) runtime.Encoder     { return c.m.NewEncoder(w) }
 func (c *CustomMarshaler) ContentType(v interface{}) string           { return "Custom-Content-Type" }
 
+// marshalerStreamContentType implements Marshaler, but with the addition of a custom StreamContentType.
+type marshalerStreamContentType struct {
+	runtime.Marshaler
+	CustomStreamContentType string
+}
+
+func (m marshalerStreamContentType) StreamContentType(interface{}) string {
+	return m.CustomStreamContentType
+}
+
 func TestForwardResponseStreamCustomMarshaler(t *testing.T) {
 	type msg struct {
 		pb  proto.Message
 		err error
 	}
+	marshaler := &CustomMarshaler{&runtime.JSONPb{}}
+
 	tests := []struct {
-		name       string
-		msgs       []msg
-		statusCode int
+		name            string
+		marshaler       runtime.Marshaler
+		msgs            []msg
+		statusCode      int
+		wantContentType string
 	}{{
-		name: "encoding",
+		name:      "encoding",
+		marshaler: marshaler,
 		msgs: []msg{
 			{&pb.SimpleMessage{Id: "One"}, nil},
 			{&pb.SimpleMessage{Id: "Two"}, nil},
 		},
-		statusCode: http.StatusOK,
+		statusCode:      http.StatusOK,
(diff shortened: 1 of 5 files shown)


johanbrandhorst:
Hi, thanks for your PR. I don't think we should need any new functionality for this. Marshalers in the grpc-gateway are registered by incoming `Accept:` header contents. Could you do something like
```go
type myJSONPbWrapper struct {
    runtime.JSONPb
}

func (myJSONPbWrapper) ContentType() string {
    return "application/x-ndjson"
}
```

and

```go
mux := runtime.NewServeMux(
    runtime.WithMarshalerOption("application/json", &runtime.JSONPb{}),
    runtime.WithMarshalerOption("application/x-ndjson", myJSONPbWrapper{}),
)
```
?

huin:
I don't think that quite captures the solution I was seeking.

Using the `Accept` header requires that the client provides the content type that it will receive in the response `Content-Type` header, and the server will blindly repeat back whichever of the two options it is given in code, regardless of the actual content type that it sends back.

I was aiming for something where the server would authoritatively state the content type on endpoints that have a streaming result.

For more context: in our situation, the client is not under our control, as we provide the serving grpc-gateway service, but not the client that is consuming a grpc-gateway service.

runtime/handler_test.go
```diff
@@ -178,36 +178,68 @@ func (c *CustomMarshaler) NewDecoder(r io.Reader) runtime.Decoder     { return c
 func (c *CustomMarshaler) NewEncoder(w io.Writer) runtime.Encoder     { return c.m.NewEncoder(w) }
 func (c *CustomMarshaler) ContentType(v interface{}) string           { return "Custom-Content-Type" }
 
+// MarshalerStreamContentType implements Marshaler, but with the addition of a custom StreamContentType.
+type MarshalerStreamContentType struct {
```
johanbrandhorst, line 182: Can we make this unexported?

johanbrandhorst (commented):
OK, fair enough, I do think there's a pretty good use case for this. It sorta sucks that we can introduce this retroactively though since it might break users. What do you think about adding an example of an ndjson stream to our docs page? It could wrap the `JSONPb` marshaler to implement the streaming content type and to set the delimiter. Maybe a new doc in https://github.com/grpc-ecosystem/grpc-gateway/tree/main/docs/docs/mapping?

runtime/handler_test.go
huin: Done. Was following the pattern of `CommonMarshaler` above, but I'm not attached to it.

huin:
> OK, fair enough, I do think there's a pretty good use case for this. It sorta sucks that we can introduce this retroactively though since it might break users. What do you think about adding an example of an ndjson stream to our docs page? It could wrap the `JSONPb` marshaler to implement the streaming content type and to set the delimiter. Maybe a new doc in https://github.com/grpc-ecosystem/grpc-gateway/tree/main/docs/docs/mapping?

Willdo, I'll take a crack at that now.

huin:
I've added another commit to add another page of documentation, and refer to it from an existing page.

johanbrandhorst (approved):
Excellent, I love the new docs page!

huin:
No problem, thanks for reviewing!

PR #2855, opened 2022-08-18
Issue 2671

MakDon:
Close #2671

PR #4931, opened 2024-11-10
Extend google.protobuf.EnumOptions for Schema

### LINKED ISSUE #2671: Extend google.protobuf.EnumOptions
## 🚀 Feature

Extend google.protobuf.EnumOptions so we can set examples on enums.

```protobuf
enum State {
  option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_enum) = {
    example: "ACTIVE"
  };
  NONE = 0;
  ACTIVE = 1;
  INACTIVE = 2;
}
```

Workaround: set the example on every message using the enum.

hown3d:

Fixes #2671

Allows to set options on proto enums with a new message `EnumSchema`.
`EnumSchema` is a subset of the `Schema` message that contains only fields that are applicable on enums.

#### Other comments
PR is based mostly on https://github.com/grpc-ecosystem/grpc-gateway/pull/2855 with a few adjustments


protoc-gen-openapiv2/options/annotations.proto
```diff
@@ -28,6 +28,13 @@ extend google.protobuf.MessageOptions {
   // different descriptor messages.
   Schema openapiv2_schema = 1042;
 }
+extend google.protobuf.EnumOptions {
+  // ID assigned by protobuf-global-extension-registry@google.com for gRPC-Gateway project.
+  //
+  // All IDs are the same, as assigned. It is okay that they are the same, as they extend
+  // different descriptor messages.
+  Schema openapiv2_enum = 1042;
```
hown3d, line 36: I would like to discuss if we really need to embed the whole Schema message here since most of the fields are not applicable to Enums. When applying the parsed schema I currently only respect a subset of fields.

johanbrandhorst: I'd definitely prefer it if we used a smaller more accurate message definition, since I don't see enums ever supporting the same features as the other options 👍🏻

hown3d: Where would the new message reside? 
https://github.com/grpc-ecosystem/grpc-gateway/blob/442bd3ab9497e4f9b3bbcbf4e53854300a68dd5a/protoc-gen-openapiv2/options/openapiv2.proto currently only contains messages that represent the OpenAPI specification. Having a message that does not represent a OpenAPI spec is somewhat misleading. 
https://github.com/grpc-ecosystem/grpc-gateway/blob/442bd3ab9497e4f9b3bbcbf4e53854300a68dd5a/protoc-gen-openapiv2/options/annotations.proto holds the extensions that register to the descriptors, so I guess that file is not an option either.

What's your opinion here?

johanbrandhorst: My initial thought was `openapiv2.proto`. How do you mean that it doesn't represent the OpenAPI specification? We're talking about defining a subset of `message Schema` as it would apply to enums, right? I was under the impression that using `Schema` wouldn't make sense since many options in there will never be possible for enums. In that sense, I guess we're defining our own subset of the OpenAPI spec?

hown3d: Yes, it would be a subset of the `Schema` message.
What I wanted to say it that currently each message in openapi.proto follows closely to the spec of OpenAPI 2.
For example the `Schema` message maps to the [Message object](https://github.com/OAI/OpenAPI-Specification/blob/3.0.0/versions/2.0.md#schema-object) of the OpenAPI spec.
The new `EnumSchema` would not map to an object of the spec, but I guess that's negligible

hown3d: I've created the new `EnumSchema` message :)

johanbrandhorst (approved):
This looks great, thank you!

johanbrandhorst:
🎉 thank you for your contribution!

PR #4965, opened 2024-11-19
chore: reduce number of empty lines in generated code

mmorel-35:
#### Description

This PR group vars and reduces the number of empty lines in generated code


johanbrandhorst (changes requested):
Thanks for this! Looks like we might need to update some tests too. This is a pretty big change but hopefully the generated file corpus we have should catch any errors.

johanbrandhorst:
Thanks for your contribution!

PR #4971, opened 2024-11-20
chore: use errors.Is to compare errors in generated code

mmorel-35:
#### Description
This uses errors.Is to compare errors in generated code.

This the recommended way to compare errors (with erros.As but do not apply here). There is a linter ([errorlint](https://github.com/polyfloyd/go-errorlint)) that ensure this.

### DIFF
--- a/protoc-gen-grpc-gateway/internal/gengateway/template.go
+++ b/protoc-gen-grpc-gateway/internal/gengateway/template.go
@@ -253,6 +253,7 @@ var (
 	_ codes.Code
 	_ io.Reader
 	_ status.Status
+	_ = errors.New
 	_ = runtime.String
 	_ = utilities.NewDoubleArray
 	_ = metadata.Join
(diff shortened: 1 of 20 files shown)


protoc-gen-grpc-gateway/internal/gengateway/template.go
```diff
@@ -270,35 +270,35 @@ func request_{{ .Method.Service.GetName }}_{{ .Method.GetName }}_{{ .Index }}(ct
 {{ template "request-func-signature" . }} {
 	var metadata runtime.ServerMetadata
 	stream, err := client.{{ .Method.GetName }}(ctx)
-	if err != nil {
+	if !errors.Is(err, nil) {
```
johanbrandhorst, line 273: I don't think we need this change, please undo the instances of this rewrite.

johanbrandhorst (changes requested):
I'm OK with using `errors.Is` for `io.EOF`, but please leave `err != nil` the same as it was.

johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution!

PR #4973, opened 2024-11-20
Errors on valid timestamps

### LINKED ISSUE #4967: Timestamps in year 0000 are accepted as query parameters
## 🐛 Bug Report

Affects APIs with a `google.protobuf.Timestamp` field on a GET request.

When `grpc-gateway` parses the timestamp query parameter it does return an error for dates in the year 0. This is inconsistent with the behavior described in the [well known type test](https://github.com/protocolbuffers/protobuf-go/blob/30f628eeb303f2c29be7a381bf78aa3e3aabd317/types/known/timestamppb/timestamp_test.go#L26) and with the behavior implemented in protojson.

Parsing of the timestamp field is done [here](https://github.com/grpc-ecosystem/grpc-gateway/blob/main/runtime/query.go#L289)

I am happy to submit a fix and include tests for min/max timestamp if this bug is considered an issue.

Seen in `v2.19.1`

## To Reproduce

1. Define an API that uses `google.protobug.Timestamp` as a field on a GET request
1. Make the request with `0000-01-01T00%3A00%3A00.00Z`
1. See that this does not cause an error

## Expected behavior

Should return a 400 Bad Request error

## Actual Behavior

Request is accepted

## Your Environment

N/A

jakec-github:

Fixes #4967

Fixes acceptance of invalid timestamps as query parameters

### DIFF
--- a/runtime/query.go
+++ b/runtime/query.go
@@ -291,7 +291,11 @@ func parseMessage(msgDescriptor protoreflect.MessageDescriptor, value string) (p
 		if err != nil {
 			return protoreflect.Value{}, err
 		}
-		msg = timestamppb.New(t)
+		timestamp := timestamppb.New(t)
+		if ok := timestamp.IsValid(); !ok {
+			return protoreflect.Value{}, fmt.Errorf("%s is not a valid timestamp", value)
+		}
+		msg = timestamp
 	case "google.protobuf.Duration":
 		d, err := time.ParseDuration(value)
 		if err != nil {
--- a/runtime/query.go
+++ b/runtime/query.go
@@ -293,7 +293,7 @@ func parseMessage(msgDescriptor protoreflect.MessageDescriptor, value string) (p
 		}
 		timestamp := timestamppb.New(t)
 		if ok := timestamp.IsValid(); !ok {
-			return protoreflect.Value{}, fmt.Errorf("%s is not a valid timestamp", value)
+			return protoreflect.Value{}, fmt.Errorf("%s before 0001-01-01", value)
 		}
 		msg = timestamp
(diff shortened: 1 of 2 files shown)


runtime/query.go
```diff
@@ -291,7 +291,11 @@ func parseMessage(msgDescriptor protoreflect.MessageDescriptor, value string) (p
 		if err != nil {
 			return protoreflect.Value{}, err
 		}
-		msg = timestamppb.New(t)
+		timestamp := timestamppb.New(t)
+		if ok := timestamp.IsValid(); !ok {
```
johanbrandhorst, line 295: Nit: now that I look at [the docs for `timestamppb`](https://pkg.go.dev/google.golang.org/protobuf/types/known/timestamppb#Timestamp.CheckValid), perhaps we should use `timestamp.CheckValid()` instead so that we can tell the user why it was an invalid timestamp?

johanbrandhorst (changes requested):
Thanks for the fix, but lets use the error with extra context

runtime/query.go
jakec-github: Agreed. Will do

jakec-github: The error returned from `timestamp.CheckValid` wasn't very friendly and was difficult to test. Looks like this was by [design](https://github.com/protocolbuffers/protobuf-go/blob/30f628eeb303f2c29be7a381bf78aa3e3aabd317/internal/errors/errors.go#L27)

But the only possible error here is timestamp before 0001-01-01 so just returned that message

johanbrandhorst: Alright fair enough

johanbrandhorst (approved):

PR #4976, opened 2024-11-21
chore: use golangci-lint in CI

mmorel-35:
#### Description

This uses golangci/golangci-lint-action instead of dominikh/staticcheck-action and define golangci-lint configuration.

golangci-lint is here configured to apply staticcheck rules and also gofmt.

gosec and govet are disabled to make the linting pass but deserve a closer look

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -125,9 +125,10 @@ jobs:
         with:
           go-version: 1.23
           check-latest: true
-      - uses: dominikh/staticcheck-action@fe1dd0c3658873b46f8c9bb3291096a617310ca6 # v1.3.1
+      - uses: golangci/golangci-lint-action@971e284b6050e8a5849b72094c50ab08da042db8 # v6
         with:
-          install-go: false
+          args: --verbose
+          version: v1.62.0
   fuzz:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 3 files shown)


johanbrandhorst:
This I'm not so sure about. I actually consider golangci-lint to be pretty bad software. I intentionally just use staticcheck because it gives me warnings that I care about. golangci-lint contains all kinds of unnecessary and questionable linters that I don't really care about. Did you find any lint failures we should fix?

johanbrandhorst:
I mean that it is hard to use well. The default linters are too broad. It doesn't run well with all linters (including staticcheck). I value having a low number of false positives, something that is true of all staticcheck lint checks but obviously not of all linters in golangci-lint. I have applied all the linters I desire, and it is the staticcheck linter. I do not think golangci-lint adds value.

Generated code is often exempt from lint checks because it's not written by humans. You can manually check the lint code if you want locally, and fix any issues that arises, but I don't think it's important for us to lint it generally.

johanbrandhorst:
I understand that you can configure golangci-lint, but I don't think the value in having it outweighs the cost (CI runtime, configuration complexity, breaking linters). If there is a specific linter you think we'd benefit from, we can add that separately. I'm going to close this PR as I do not intend to replace staticcheck with golangci-lint. Thank you for your consideration.

PR #4978, opened 2024-11-21
Fix marshaler interface function name in examples in docs

o-nix:
It is actually:
```go
// Delimited defines the streaming delimiter.
type Delimited interface {
	// Delimiter returns the record separator for the stream.
	Delimiter() []byte
}
```


johanbrandhorst (approved):
Oops, thanks!

PR #5041, opened 2024-12-09
Correctly expand path variables for Update Methods. 

nullaus:
Refer to [AIP-134](https://google.aip.dev/134) for more information.

Update Methods require the resource type name as the path parameter instead of a simple `name` (e.g. `{vendor_contact.name=vendor_contacts/*}` instead of `{name=vendor_contacts/*}`)

This would case the following error to occur when generating an OpenAPIv2 specification:

```
panic: Path parameter "vendorContact.name" not found in path parameters

goroutine 1 [running]:
github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2/internal/genopenapi.expandPathPatterns({0x14000798600?, 0x33?, 0x14000001b00?}, {0x14000b6a660?, 0x4?, 0x4?}, 0x14000001b00)
```

This was due to `templateToParts` applying the `GetUseJSONNamesForFields` flag and `expandPathPatterns` _not_ factoring in JSON names when finding the path parameter by name in the slice of path parameters. It must look for the path parameter by JSON name. 

I modified the find function to compare using JSON names when appropriate by looking at the UseJSONNamesForFields flag.

Also fixed are minor linter complaints when using `go 1.23`. I can omit these if preferred.

### DIFF
--- a/protoc-gen-openapiv2/internal/genopenapi/template_test.go
+++ b/protoc-gen-openapiv2/internal/genopenapi/template_test.go
@@ -922,8 +922,8 @@ func TestMessageToQueryParametersWithJsonName(t *testing.T) {
 		Params   []openapiParameterObject
 	}
 
-	var requiredField = []annotations.FieldBehavior{annotations.FieldBehavior_REQUIRED}
-	var requiredFieldOptions = new(descriptorpb.FieldOptions)
+	requiredField := []annotations.FieldBehavior{annotations.FieldBehavior_REQUIRED}
+	requiredFieldOptions := new(descriptorpb.FieldOptions)
 	proto.SetExtension(requiredFieldOptions, annotations.E_FieldBehavior, requiredField)
 
 	messageSchema := &openapi_options.Schema{
@@ -2021,7 +2021,8 @@ func TestApplyTemplateExtensions(t *testing.T) {
 		},
 	}
 	verifyTemplateExtensions := func(t *testing.T, reg *descriptor.Registry, file *descriptor.File,
-		opts *openapiconfig.OpenAPIOptions) {
+		opts *openapiconfig.OpenAPIOptions,
+	) {
 		if err := AddErrorDefs(reg); err != nil {
 			t.Errorf("AddErrorDefs(%#v) failed with %v; want success", reg, err)
 			return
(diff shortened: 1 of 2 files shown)


johanbrandhorst (approved):
LGTM

nullaus:
Thank you for the ⚡ fast review @johanbrandhorst!

johanbrandhorst (changes requested):
I was actually surprised this passed without any generation changes. Do you think you could add a test to template_test that exercises this? Alternatively, add an annotation somewhere in one of our protobuf files that would exercise this. I think we should have one with `user_json_names_for_fields` set as an option. Thanks!

protoc-gen-openapiv2/internal/genopenapi/template_test.go
```diff
@@ -4162,6 +4160,8 @@ func TestTemplateToOpenAPIPathExpandSlashed(t *testing.T) {
 		{"/test/{name=*}/", "/test/{name}/", getParameters([]string{"name"}), []string{"name"}, true},
 		{"/test/{name=test_cases/*}/", "/test/test_cases/{testCase}/", getParameters([]string{"name"}), []string{"testCase"}, true},
 		{"/test/{name=test_cases/*}/", "/test/test_cases/{test_case}/", getParameters([]string{"name"}), []string{"test_case"}, false},
+		{"/test/{test_type.name=test_cases/*}/", "/test/test_cases/{testCase}/", getParameters([]string{"test_type.name"}), []string{"testCase"}, true},
+		{"/test/{test_type.name=test_cases/*}/", "/test/test_cases/{test_case}/", getParameters([]string{"test_type.name"}), []string{"test_case"}, false},
```
nullaus, line 4164: @johanbrandhorst Here are the test cases that exercise this.

johanbrandhorst: Welp I missed it in the cosmetic changes, thanks for pointing it out

johanbrandhorst (approved):

protoc-gen-openapiv2/internal/genopenapi/template_test.go
nullaus: No worries! Thanks again for your responsiveness. :)

PR #5072, opened 2024-12-17
fix: exclude synthetic oneofs during populateFieldValueFromPath

aerialls:

Add compatibility with protobuf-go `>1.36.0` and changes regarding oneofs for synthetic fields introduced with https://go-review.googlesource.com/c/protobuf/+/632735

Yes synthetic fields of the initial check.

Exclude


johanbrandhorst (changes requested):
Hey, thanks for the PR, is there a test we can add here?

aerialls:
I've added a dedicated test in `query_test.go` with a new optional field. The error is more straight forward this way. I've also updated to protobuf `1.36.0` in the same PR to make sure this is working as expected. 

I've validated that without the fix, the test is failing with the following error.
```
panic: invalid oneof descriptor grpc.gateway.runtime.internal.examplepb.Proto3Message._optional_value for message grpc.gateway.runtime.internal.examplepb.Proto3Message
```

johanbrandhorst:
Thanks a lot, sorry for the noise, could you rebase on main? We had some other dependency changes go in

johanbrandhorst:
I force pushed an update, but looks like the bazel tests are still failing. It doesn't look like it's related to your change so I'll try to investigate.

johanbrandhorst (approved):
Nice!

PR #5129, opened 2025-01-10
protoc-gen-openapiv2: Do not add invisible enum values as default

navruzm:

Yes 

Right now, if we hide the default enum value with `value_visibility`, it does not appear in the enum list but still shows up in the `default` field. This PR adds a check that skips the default enum value if it is hidden. That way, we will not see a hidden enum as the default. With this PR, if `unknown_type` is hidden, it will no longer appear as the default.

**Example proto:**
```
enum Type {
    unknown_type = 0 [(google.api.value_visibility).restriction = "INTERNAL"];
    created_at = 1;
    updated_at = 2;
}
```

**Example output (before fix):**
```
{
  "name": "type",
  "description": "",
  "in": "query",
  "required": false,
  "type": "string",
  "enum": [
    "created_at",
    "updated_at"
  ],
  "default": "unknown_type"
}
```


johanbrandhorst (changes requested):
I'm a little conflicted about this. On the one hand, it makes sense that if you want to hide the value it shouldn't show up as a default. On the other hand, all protobuf enum values have defaults that will be set if no value is explicitly supplied. What do you think about returning the integer value while hiding the name?

Also could you try adding add an enum with this property to the https://github.com/grpc-ecosystem/grpc-gateway/blob/main/examples/internal/proto/examplepb/visibility_rule_echo_service.proto test file? Thanks!

navruzm:
> What do you think about returning the integer value while hiding the name?

Done.

> Also could you try adding add an enum with this property to the https://github.com/grpc-ecosystem/grpc-gateway/blob/main/examples/internal/proto/examplepb/visibility_rule_echo_service.proto test file?

Sorry, I didn't get this. Can you explain this a bit more? There is already an enum in the file, `VisibilityEnum`, and no Swagger file is generated for the file.

johanbrandhorst (changes requested):
> Sorry, I didn't get this. Can you explain this a bit more? There is already an enum in the file, VisibilityEnum, and no Swagger file is generated for the file.

Sorry, I was unclear. What I meant was, lets add a new enum to the proto file https://github.com/grpc-ecosystem/grpc-gateway/blob/main/examples/internal/proto/examplepb/visibility_rule_echo_service.proto. Lets have the default value be hidden using the visibility specification. This _should_ update the *.swagger.json files generated by various visibility test configs we have living alongside this file (see https://github.com/grpc-ecosystem/grpc-gateway/tree/main/examples/internal/proto/examplepb). It would be great to see that the enum is there but the default value is omitted, in the case where the enum is used as a string. Does that make sense?

navruzm:
Done!

johanbrandhorst (approved):

PR #5151, opened 2025-01-17
Only write Content-Length if the runtime.WithWriteContentLength() option is specified

### LINKED ISSUE #4238: Content-Length not set in ForwardResponseMessage

joshgarnett:
This fixes #4238


johanbrandhorst (approved):

johanbrandhorst:
Thank you for this!

PR #3323, opened 2023-05-25
feat(#3238):Support map in query parameters

li31727:
#### References to other Issues or PRs

Fixes #3238 

#### Have you read the [Contributing Guidelines](https://github.com/grpc-ecosystem/grpc-gateway/blob/main/CONTRIBUTING.md)?
Yes

#### Brief description of what is fixed or changed
Generate the swagger api for Get requests that pass map type parameters

PR #5158, opened 2025-01-21
Prevent overwriting description of map fields

fredde-fisk:
In https://github.com/grpc-ecosystem/grpc-gateway/pull/3323, `map` fields where included in the OpenAPI specification, which is good. However, the documentation for the field might be confusing since it contains the following hard coded example that might be irrelevant and misleading for the one reading the documentation:

> This is a request variable of the map type. The query format is "map_name[key]=value", e.g. If the map name is Age, the key type is string, and the value type is integer, the query parameter is expressed as Age["bob"]=18

It is also not possible to override this hard coded description using `grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field`, like this:
```
map<string, string> mapped_string_value = 1 [(grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
  description: "This will be ignored"
}];
```

This PR removes the hard coded description and instead let's the creator of the protobuf file decide the appropriate descriptions for the field.


johanbrandhorst (approved):
Yeah this makes sense to me. I think the old description was useful, but I agree it's probably confusing to the consumer of the API docs.

PR #5211, opened 2025-02-06
5113. Validation of oneof nested objects

### LINKED ISSUE #5113: Expecting a oneof validation error but only getting it intermittently
## 🐛 Bug Report

When receiving values from query string params, fields that belong to a oneof are not validated consistently.

## To Reproduce

Let’s say you have an Outer message with fields foo and bar which form a oneof and foo is of type int32 and bar is a nested message type called Inner.

```
message Outer {
  oneof value {
    int32 foo = 1;
    Inner bar = 2;
  }
  message Inner {
    int32 width = 1;
    int32 height = 2;
  }
}
```

If QueryParameterParser tries to build a message from a GET request by parsing query parameters in the URL then sometimes it fails to assert the oneof condition: either foo can have a value, or bar can have a value, but not both.

Example requests:
```
http://example.com/endpoint?foo=123  // okay
http://example.com/endpoint?bar.width=100&bar.height=100  // okay
http://example.com/endpoint?bar.width=100&bar.height=100&foo=123   // not okay. should always get the validation error "field already set for oneof"
```

For the last example, sometimes you get a validation error, sometimes you don’t.

## Expected behavior

Should always get a "field already set for oneof" validation error when foo and bar are both present in the URL query string.

## Actual Behavior

Sometimes you get a "field already set for oneof" validation error, sometimes you don’t.

## Your Environment

Version v2.20.0 on Ubuntu 22.04

optician:

Fixes #5113

Yes.

The fix follows the suggestion from the issue. The oneOf check is moved before `msgValue.Mutable()` resets other option of this oneOf. 
Also, previous behavior didn't validate nested messages correctly. After the cycle "fieldDescriptor" points to a leaf of a path tree and it's not a member of a oneOf field. For example, the piece of a proto
```proto
 ...
  oneof nested_oneof_value {
    int32 nested_oneof_int32_value = 50;
    Proto3Message nested_oneof_value_one = 46;
  }
 ...
```
and the path `["nested_oneof_value_one", "int32Value"]`.  `WhichOneof(nested_oneof_value_one)` returns nested_oneof_value, but `WhichOneof(int32Value) returns nil. 

#### Other comments
I struggled with codegen refreshing and spent quite a lot of time on a simple task. The contribution instruction does not regenerate *.pb.go files when protos are changed (and runs very slowly because it recompiles protoc). Other documentation also lacks instructions. I hope I found the correct way and it will save time for others.


CONTRIBUTING.md
```diff
@@ -46,6 +46,11 @@ nested docker container is not possible.
 
 If this has resulted in some file changes in the repo, please ensure you check those in with your merge request.
 
+### Regenerate a code for a single proto file
+1. Install [buf](https://buf.build/docs/installation/).
+2. You may need to install plugins explicitly if they are not in the PATH (brew package does not add them to a visible path for "buf").
+3. `buf generate --path=location/of/proto`
```
johanbrandhorst, line 52: I don't think this section is necessary. I see you reported that using bazel was rebuilding `protoc`, I have also had that issue, but you should have been able to just run `make install && make clean && make generate`. Did you try that? I think if we want to add a section here, lets just make it 
```suggestion
### Regenerate with own environemnt
1. Run `make install` to install the tools at the correct versions
2. Run `make clean` to delete existing generated files
3. Run `make generate` to regenerate all files. Note that this will regenerate the swagger files, which requires `swagger-codegen` in your $PATH at the appropriate version. How to install this is left as an exercise to the reader.

If you experience any errors during any of these steps, try one of the options above, which help set up your environment.
```

Just regenerating a single protobuf file might not necessarily generate all the correct dependent files.

johanbrandhorst (changes requested):
Minus the contribution guide changes, this looks great, thank you so much!

CONTRIBUTING.md
optician: It seems I had a weird problem with virtualization. The instruction
```bash
docker run -v $(pwd):/grpc-gateway -w /grpc-gateway --rm ghcr.io/grpc-ecosystem/grpc-gateway/build-env:1.23 \
    /bin/bash -c 'make install && \
        make clean && \
        make generate'
    ghcr.io/grpc-ecosystem/grpc-gateway/build-env:1.23 -c '\
        bazel run :gazelle -- update-repos -from_file=go.mod -to_macro=repositories.bzl%go_repositories && \
        bazel run :gazelle && \
        bazel run :buildifier'
```
completed successfully but 0 files were updated. It works for me now. Also, it looks like the first command is enough. I removed the documentation change.

johanbrandhorst (approved):

johanbrandhorst:
Seems like some swagger files still need to be regenerated, could you try running it again? See the failures in the `generate` CI job.

optician:
As I understand the file shouldn't change. It seems ok now. I can't explain what happened in the previous iteration 🤷

johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution!

Issue #5244, opened 2025-02-16
Slow test startup


paskozdilar:
## 🐛 Bug Report

When running tests on `main`, it takes 10 seconds for tests to initialize before starting.

## To Reproduce

Run: `go test -v ./examples/internal/integration/`

## Expected behavior

Tests run immediately, or with a sensible init delay.

## Actual Behavior

Tests wait for 10 seconds before starting:

```
$ go test -v ./examples/internal/integration/
2025/02/16 19:15:41 ERROR: http error 502 request body 
2025/02/16 19:15:41 ERROR: http error 502 request body 
2025/02/16 19:15:41 ERROR: http error 502 request body 
2025/02/16 19:15:41 ERROR: http error 502 request body 
2025/02/16 19:15:41 ERROR: http error 502 request body 
2025/02/16 19:15:41 ERROR: http error 502 request body 
2025/02/16 19:15:41 ERROR: http error 502 request body 
[...]
```

## Your Environment

OS: Ubuntu 24.04 LTS on WSL2
Go: go version go1.23.4 linux/amd64
Commit: 250bec2c4f2cbd9a951e5921d1360dbd6468392e

PR #5245, opened 2025-02-16
Fix slow test startup

paskozdilar:

Fixes https://github.com/grpc-ecosystem/grpc-gateway/issues/5244

Yes.

The `waitForGateway` method waited 10 seconds in a loop for `healthzServer` to return StatusOK.
The `healthzServer` function returns StatusOK only if `conn.GetState()` is `connectivity.Ready` - but that never happened because connection stays idle before the first method call.
To resolve this, if we detect that the connection is idle, we call any method (in this case, we chose `"/grpc.health.v1.Health/Check"` as it seemed like an appropriate name) and ignore its result, just to get the connection started.

This fixes the test startup delay.

### DIFF
--- a/examples/internal/gateway/BUILD.bazel
+++ b/examples/internal/gateway/BUILD.bazel
@@ -19,6 +19,7 @@ go_library(
         "@org_golang_google_grpc//connectivity",
         "@org_golang_google_grpc//credentials/insecure",
         "@org_golang_google_grpc//grpclog",
+        "@org_golang_google_protobuf//types/known/emptypb",
     ],
 )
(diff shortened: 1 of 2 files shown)


johanbrandhorst (changes requested):
Great detective work here! Looks like there's a minor bazel diff, after that is fixed we can merge this :).

paskozdilar:
Done.

johanbrandhorst:
Thanks for your contribution!

paskozdilar:
Thank you for your patience and support :)

PR #5265, opened 2025-02-22
Update go version

johanbrandhorst:


RTann:
How come this was ported to v2.26.2? This is a rather breaking change because people who use go1.23 would not be able to upgrade from v2.26.1 to v2.26.2. I'm ok with not being allowed to use v2.27.0 (for example) until upgrading to go1.24, but this is a patch version upgrade, so I'd expect to be able to use it with the same go version (unless the patch requires go1.24, but it does not seem like it's required?)

johanbrandhorst:
This change has been reverted in https://github.com/grpc-ecosystem/grpc-gateway/releases/tag/v2.26.3. Sorry for the inconvience.

RTann:
oh wow that was fast. Thanks!

PR #5278, opened 2025-02-26
feat(protoc-gen-openapiv2/template): updateSwaggerObjectFromFieldBeha…

Ccheers:
…vior add default required rule


johanbrandhorst:
Hi, thanks for your PR. Is there an issue this relates to? What is this fixing? Looks to be something related to field behavior?

Ccheers:
> Hi, thanks for your PR. Is there an issue this relates to? What is this fixing? Looks to be something related to field behavior?

In proto3, all fields are Required by default, and the generated pb files are also Required, so the field behavior should be Required by default.

johanbrandhorst:
Hm, I suppose that might make sense. Looks like there are quite a few test errors, could you try and fix those please? Also you'll need to regenerate the files, see CONTRIBUTING.md.

Ccheers:
> Hm, I suppose that might make sense. Looks like there are quite a few test errors, could you try and fix those please? Also you'll need to regenerate the files, see CONTRIBUTING.md.嗯，我觉得有道理。看起来有不少测试错误，你能试着修复一下吗？此外，你还需要重新生成文件，请参阅 CONTRIBUTING.md。

ok

Makefile
```diff
@@ -83,7 +83,7 @@ $(GENERATE_UNBOUND_METHODS_EXAMPLE_SRCS): $(GENERATE_UNBOUND_METHODS_EXAMPLE_SPE
 
 install:
 	go install github.com/bufbuild/buf/cmd/buf@v1.45.0
-	go install \
+	go install -buildvcs=false \
```
johanbrandhorst, line 86: Why is this necessary?

johanbrandhorst (commented):
I'm sort of worried that this is going to be a breaking change for users. It may be a semantically correct translation from the proto3 definitions, but this is going to change a LOT of people's swagger definitions. What do you think about adding an option for this? `use_proto3_field_semantics` perhaps?

Makefile
Ccheers: it's unnecessary i'll remove it

Ccheers:
> I'm sort of worried that this is going to be a breaking change for users. It may be a semantically correct translation from the proto3 definitions, but this is going to change a LOT of people's swagger definitions. What do you think about adding an option for this? `use_proto3_field_semantics` perhaps?

Good Idea 👍

johanbrandhorst (changes requested):
This looks good, could you also add the new option to the bazel options? They are in https://github.com/grpc-ecosystem/grpc-gateway/blob/main/protoc-gen-openapiv2/defs.bzl

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -849,6 +849,8 @@ func schemaOfField(f *descriptor.Field, reg *descriptor.Registry, refs refMap) o
 		}
 	}
 
+	ret.Required = slices.Compact(ret.Required)
```
johanbrandhorst, line 852: Will the fields in `ret.Required` be sorted already? `slices.Compact` removes _consecutive runs of the same element_ so for deduplication we will need to either guarantee that they are sorted already, or sort first.

Ccheers: If that's the case, would the previous version of uniqArray be a better implementation?

johanbrandhorst: No, as I said, just sort it first
```suggestion
	slices.Sort(ret.Required)
	ret.Required = slices.Compact(ret.Required)
```

johanbrandhorst (approved):

johanbrandhorst:
Thanks for your contribution!

PR #5303, opened 2025-03-03
Revert #4819

johanbrandhorst:
This reverts commit 830ba2782f9f2715fce43027ce10100a8e0096df. Unfortunately this suffered from a bug that could cause a panic. Revert until we can reintroduce it without the bug.

See https://github.com/etcd-io/etcd/issues/19509 for more info.


johanbrandhorst:
I think the problem is that there is no synchronization between returning from the handler and using the `http.ResponseWriter` in the goroutine that's created to handle the errors. We can reintroduce this if we add a channel or something to synchronize exiting from the goroutine with exiting from the handler.

CC @ianbbqzy

ahrtr:
Thanks for the quick fix and release of v2.26.2!!! @johanbrandhorst

PR #5308, opened 2025-03-04
chore(deps): compatibility with at least two major versions of Go

Tiscs:

Revert #5265, maintain compatibility with at least two major versions of Go.

Yes

I think a module that is so widely used should be compatible with at least two major versions of Go, not just the last one.

thank you.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/grpc-ecosystem/grpc-gateway/v2
 
-go 1.24
+go 1.23.0
 
 require (
 	github.com/antihax/optional v1.0.0
--- a/.github/Dockerfile
+++ b/.github/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.24.0
+FROM golang:1.23.6
 
 ENV NVM_DIR="/usr/local/share/nvm"
 ENV NVM_SYMLINK_CURRENT=true \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
           ./node_modules/.bin/gulp
   generate:
     container:
-      image: ghcr.io/grpc-ecosystem/grpc-gateway/build-env:1.24
+      image: ghcr.io/grpc-ecosystem/grpc-gateway/build-env:1.23
       options: "--user root"
       credentials:
         username: ${{ github.actor }}
@@ -59,7 +59,7 @@ jobs:
       - run: git diff --exit-code
   bazel:
     container:
-      image: ghcr.io/grpc-ecosystem/grpc-gateway/build-env:1.24
+      image: ghcr.io/grpc-ecosystem/grpc-gateway/build-env:1.23
       options: "--user root"
       credentials:
         username: ${{ github.actor }}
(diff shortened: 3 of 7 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/grpc-ecosystem/grpc-gateway/v2
 
-go 1.24
```
johanbrandhorst, line 3: I think this change should be the only one necessary to allow the last two major versions, right? It's fine for us to use 1.24 in most of our own testing and so on. Could you revert the other changes please?

johanbrandhorst (changes requested):

go.mod
Tiscs: Whether tests should be run on the minimum supported version of Go?

ivanvc:
For context, Go has support for the latest two minors: https://go.dev/doc/devel/release#policy. So, for widely used libraries I agree that it should stay in the oldest supported version (i.e., 1.23 in this case)

Tiscs:
Task [CI / gorelease (pull_request)](https://github.com/grpc-ecosystem/grpc-gateway/actions/runs/13646930161/job/38148583478?pr=5308) fails because the `github.com/grpc-ecosystem/grpc-gateway` module already requires Go 1.24.

go.mod
```diff
@@ -1,6 +1,6 @@
...
+go 1.23.0
```
ahrtr, line 3: If the suffix ".0" wasn't automatically added by `go mod tidy`, then we should prefer to `1.23`. You can also add `toolchain go1.23.6` if you want.
```suggestion
go 1.23
```

Tiscs: This because the go version of package `google.golang.org/genproto/googleapis/rpc` is declared as `1.23.0`, so `1.23.0` should be the minimum version available here.

ahrtr: OK, then it's automatically added by `go mod tidy`. Technically, I don't think there is any difference between `go 1.23` and `go 1.23.0`; but Semantically, `go 1.23.x` means it requires the minimum go patch version for the downstream modules.

I don't know what motivate some people to add the suffix ".0". On the other hand, the golang team should treat the `go 1.x.0` equivalent to `go 1.x`, and do not require the downstream module to add the suffix no matter the upstream has it or not.

Tiscs: [Starting in Go 1.21](https://go.dev/doc/go1.21#introduction), the go directive in the go.mod file supports specifying minor versions, enhancing versioning precision while maintaining backward compatibility.
Setting it to `1.23.0` means that downstream modules must use a Go higher than `1.23.0`(e.g. `1.23.6`, `1.24.0`), not just `1.23.0`.

ahrtr:
The problem is that grpc-ecosystem doesn't have long-term supported versions or stable versions (stable versions should go with the older supported go minor version, i.e. 1.23 for now); its version number has always been increasing linearly.

Tiscs:
> The problem is that grpc-ecosystem doesn't have long-term supported versions or stable versions (stable versions should go with the older supported go minor version, i.e. 1.23 for now); its version number has always been increasing linearly.

Yes, but I think the Go version should not be increased unnecessarily. The new features in Go 1.24 are not used in the project, and the choice should be given to downstream modules as much as possible.

HaraldNordgren:
Please revert to not requiring 1.24 🤗

I would advise against bumping the major Go version requirement in a patch version like this.

go.mod
johanbrandhorst: We run builds for `stable` and `oldstable`, I think tests can keep running for Go 1.24. Please revert the other changes and I can merge this.

johanbrandhorst:
Thanks everyone for your feedback, I didn't intend for this to break Go 1.23 users, so I think this PR should be merged, but I only want to change the go.mod declaration, not the other files.

johanbrandhorst:
Closing this in favor of #5313 to get this out quicker. Thanks for the PR @Tiscs

PR #5240, opened 2025-02-14
Drain req.body on server stream methods with no body defined

paskozdilar:
This will prevent connection leaking when client sends a body on a method with no body defined. It will cause hanging when client sends a body stream, but such calls will hang anyway, even without the fix.

Fixes #5236.

#### References to other Issues or PRs

https://github.com/grpc-ecosystem/grpc-gateway/issues/5236

#### Have you read the [Contributing Guidelines](https://github.com/grpc-ecosystem/grpc-gateway/blob/main/CONTRIBUTING.md)?

Yes.

#### Brief description of what is fixed or changed

When an HTTP client send body on a method which doesn't have (google.api.http) { body } annotation defined, grpc-gateway never closes context, causing connection leaks.
This PR drains req.Body on every method which doesn't have body annotation defined, so methods will work even if client sends body.

Issue #5236, opened 2025-02-13
ServerStream client context never closes when using google.protobuf.Empty without body annotation


paskozdilar:
## 🐛 Bug Report

In a nutshell, when using `google.protobuf.Empty` in a server-streaming gRPC method without a `body` annotation in `google.api.http` options, client context never closes, even when client disconnects completely. This caused connection leaks in my service which took me a while to figure out.

Example RPC that works fine:

```proto3
rpc ServerStreamOK (google.protobuf.Empty) returns (stream ExampleResponse) {
  option (google.api.http) = {
    post: "/example/v1/ServerStreamOK";
    body: "*";
  };
}
```

Example RPC that has this issue:

```proto3
rpc ServerStreamBroken (google.protobuf.Empty) returns (stream ExampleResponse) {
  option (google.api.http) = {
    post: "/example/v1/ServerStreamBroken";
  };
}
```

## To Reproduce

Please follow the README.md of the repository I've created for this exact issue:

https://github.com/paskozdilar/bug-report-grpc-gateway

## Expected behavior

Client context is closed when client disconnects.

## Actual Behavior

Client context is never closed.

## Your Environment

- OS: Ubuntu 24.04 LTS on WSL2
- Go: go version go1.23.4 linux/amd64
- Package versions: Please see [go.mod file](https://github.com/paskozdilar/bug-report-grpc-gateway/blob/master/go.mod) in the repository I've created for this issue.

johanbrandhorst:
Hi, thanks for the detailed bug report! That's a really unfortunate bug 😁. Would you be willing to submit a fix?

paskozdilar:
Of course, as soon as I figure it out. I'm not too familiar with the grpc-gateway codebase, so it might take a bit of time.

If anyone has any hunch where the problem might be, any hints are welcome.

paskozdilar:
It seems that, during ServerStreamBroken call, the gRPC [HandleStreams call doesn't receive *http2.RSTStreamFrame at all](https://github.com/grpc/grpc-go/blob/v1.70.0/internal/transport/http2_server.go#L679).
No idea why that may be. I'm looking for relevant grpc-gateway code that may have something to do with the HandleStreams method...

paskozdilar:
Indeed, logging frames received in HandleStreams, I get the following output:

ServerStreamOK:
```
2025/02/14 10:33:09 Received frame: *http2.SettingsFrame
[...]

PR #5331, opened 2025-03-06
Discard body of body-less methods in background

### LINKED ISSUE #5326: v2.26.2 breaks grpc-websocket-proxy
I realize https://github.com/tmc/grpc-websocket-proxy is not officially supported by this project, but it's mentioned a lot in issues here. Just flagging that the change in #5240 breaks the websocket-proxy, causing requests to hang indefinitely.

paskozdilar:
When `io.Copy` is called synchronously on HTTP request body, it stalled the clients who may keep the request stream open.
This broke WebSocket gateway implementations which rely on Close detection to know when to terminate.

This PR runs the `io.Copy` method in the background.

Fixes #5326.

Yes.

Added `go` keyword in front of `io.Copy` call in the template when `body` annotation is not set.


paskozdilar:
Forgot to regenerate files. Doing it now.

protoc-gen-grpc-gateway/internal/gengateway/template.go
```diff
@@ -382,7 +382,7 @@ var filter_{{ .Method.Service.GetName }}_{{ .Method.GetName }}_{{ .Index }} = {{
 	}
 	{{- end }}
 {{- else }}
-    io.Copy(io.Discard, req.Body)
+    go io.Copy(io.Discard, req.Body)
```
johanbrandhorst, line 385: I'm a bit nervous about doing this in a goroutine, since it may outlive the handler (especially after https://github.com/grpc-ecosystem/grpc-gateway/pull/5303). The documentation for [http.Handler](https://pkg.go.dev/net/http#Handler) states:

> Depending on the HTTP client software, HTTP protocol version, and any intermediaries between the client and the Go server, it may not be possible to read from the [Request.Body] after writing to the [ResponseWriter](https://pkg.go.dev/net/http#ResponseWriter). Cautious handlers should read the [Request.Body] first, and then reply.

This might cause us to write before reading (since we don't know when the goroutine will be scheduled). I understand that doing the reading synchronously was causing the websocket client to block, but I wonder if we could try doing this in such a way as to guarantee that we don't read after the handler has returned, e.g. something like

```
wg := sync.WaitGroup{}
wg.Add(1)
go func() {
    defer wg.Done()
    _, _ = io.Copy(io.Discard, req.Body)
}()

...

// At the end of this handler
wg.Wait()
```

What do you think?

paskozdilar: As far as I understood, the docs warn that reading Body may fail after writing to ResponseHandler. And please correct me if my understanding is wrong.

With that in mind:
1. we don't check for errors in `io.Copy`, so we don't really care if it fails or not
2. this code only executes when we have already read all request data we need (nothing when `body` is not defined, JSON request when `body` is defined), so we don't really care about the data
3. gateway clients, such as tmc/grpc-websocket-proxy, may not close the Body stream at all so we *cannot* read the whole Body before writing the response (hence the `go io.Copy` in the first place)
4. since we're doing the io.Copy in the goroutine, the scheduling is non-deterministic - it may as well happen that the handler finishes writing to ResponseBody happen before `io.Copy` executes

However, I do dislike firing off uncontrollable goroutines in the background. Seems like an invitation for goroutine leak. This code may help diagnoze any issues if they happen.

paskozdilar: It may also be prudent to defer req.Body.Close() to prevent leaking goroutines, in case req.Body is never closed on the client side?

paskozdilar:
Missing import for sync.WaitGroup.

I don't feel like tinkering with the import machinery - and benchmarks don't show any significant slowdown when using channels instead.

I'll update the implementation to use plain channels.

paskozdilar:
I'll try to write an integration test for this too.

paskozdilar:
Apparently, transactional request-response flow is inherent to Go HTTP, including grpc-gateway runtime.
As such, it is impossible to "stream" data via HTTP client as we would with a WebSocket wrapper, so it's impossible to write an integration test for this without implementing a websocket gateway inside the server.

@johanbrandhorst
Would it be acceptable to add a minimal WebSocket wrapper inside the integration test server?

johanbrandhorst:
Sure, it'd be great to know if we break the websocket proxy again in the future, why not!

paskozdilar:
I've opened a repository for the websocket proxy itself:
https://github.com/paskozdilar/grpc-gateway-websocket

It's fairly covered with tests, which fail on `io.Copy` and pass on `go io.Copy`.
I'll use this code as reference.

paskozdilar:
Please see: https://github.com/grpc-ecosystem/grpc-gateway/pull/5338

paskozdilar:
Integration tests are failing even with the new changes :) good thing we decided to include them.

Unfortunately, seems like the very act of waiting for `io.Copy()` to finish makes the request hang on the client side - even when gRPC method returns, the connection never gets disconnected until client closes the connection.

This is certainly not desirable behavior - we want clients to be disconnected when server returns. That only leaves us with `go io.Copy` as an option.

I will revert the commits replacing the `go io.Copy` with blocking version - tests should then pass.

johanbrandhorst:
Is it only affecting server side streaming methods, or bi-directional methods too?

paskozdilar:
It happens on Unary methods and Server Stream methods (i.e. non-client-stream methods).

Bidirectional and Client Stream methods don't use this template branch, so the problem doesn't apply - and neither does the root issue in which gateway doesn't read client body...

---

In other news, I'm having trouble with properly testing WebSocket context cancellation:
1. in case of Unary and ServerStream methods, closing the WS connection should cancel the context of the method
2. in case of ClientStream and Bidirectional methods, closing the WS connection should send io.EOF to the gRPC stream

Since WebSocket proxy layer cannot possibly know which method is going to be executed, we cannot decide what to do on WS close message:
1. if we close the request pipe, Unary and ServerStream methods will treat that as EOF of request and still execute
2. if we cancel the request context, ClientStream and Bidirectional methods will fail immediately and not return the response.

In an ideal world, grpc-gateway would generate the WebSocket proxy itself, and it would generate two different implementations depending on the method type. Maybe that would be a cool feature in the future?

---

So yeah, this is going to be a hairy one. I'll try to figure something out.

johanbrandhorst:
I think @tmc has wanted to integrate the websocket proxy into the gateway for a long time but we just haven't decided the best way forward. I think it would be a bigger task than this issue. If we have to choose between supporting the websocket proxy and removing a bug in the gateway (#5236), then we're going to fix the bug. It might be good enough that we support the websocket proxy only for bi-directional streaming methods?

paskozdilar:
Eh, you're right. I'm way too deep in the rabbit hole.
I'll take a simpler approach - I'll just use tmc/grpc-gateway-websocket in the first place, and only add tests for context hanging from the bug report.

So I can use both 1) coder/websocket library, as suggested, and 2) your suggestion to wait for io.Copy at end of the request.
The fact that it doesn't quite work with *my* implemetation of websocket proxy is fault of *my* implementation.

For any further bugs and features, I'll open new issues and possibly further PRs.
I'd love to give full websocket integration a try :)

paskozdilar:
Everything passing, tests are failing on `integration-test` branch.

I think this is ready to merge.

paskozdilar:
Will test how this works with https://github.com/jnis23/grpc-gateway-proxy-example provided by the bug report.

paskozdilar:
Integration tests are failing on `integration-tests` branch, bazel workflows are passing on `main`, and my local tests show that tmc/grpc-gateway-websocket proxy works with these changes.
As far as I'm concerned, I think this is ready for merge.

Let me know if there's anything else I need to do.

generate.sh
```diff
@@ -0,0 +1,25 @@
...
+        make clean && \
+        make generate'
+docker run -itv "$(pwd)":/grpc-gateway -v "$(pwd)"/certs:/etc/ssl/certs -w /grpc-gateway --entrypoint /bin/bash --rm \
+    ghcr.io/grpc-ecosystem/grpc-gateway/build-env:1.24 -c '\
+        bazel run :gazelle -- update-repos -from_file=go.mod -to_macro=repositories.bzl%go_repositories && \
+        bazel run :gazelle && \
+        bazel run :buildifier &&
+        bazel test //...'
+
+beep
```
johanbrandhorst, line 25: LOL. I love this, but did you mean to commit this? I think I would prefer not to have this since we use the Makefile for generation

protoc-gen-grpc-gateway/internal/gengateway/template.go
```diff
@@ -381,9 +381,8 @@ var filter_{{ .Method.Service.GetName }}_{{ .Method.GetName }}_{{ .Index }} = {{
 			}
 	}
 	{{- end }}
-{{- else }}
-    io.Copy(io.Discard, req.Body)
 {{- end }}
+    go io.Copy(io.Discard, req.Body)
```
johanbrandhorst, line 385: Did you mean to do this unconditionally? I thought we were only doing this on body-less endpoints.

johanbrandhorst (changes requested):

generate.sh
paskozdilar: > did you mean to commit this?

No, haha. That was a helper script I used for all the Bazel stuff, which sends a beep to my headphones when it's done.
I committed it by mistake, and I'll remove it.

protoc-gen-grpc-gateway/internal/gengateway/template.go
paskozdilar: Yes.

One of the issues that I've noticed in the https://github.com/jnis23/grpc-gateway-proxy-example is that even the "working" example (the one with "body" defined) seems to have issues with closing connections.
After adding the `go io.Copy` to those too, the issue disappeared.

In both Unary and ServerStream methods, client sends a _finite amount of data_ before server stops reading the `req.Body` (zero if `body` is not defined in proto, request-message-length if body is defined in proto).
In both cases, not reading the `req.Body` further causes the connections to hang.
Since server does not care for any input after those, we can safely do `go io.Copy` on the request body.

Here's a screen recording of the issue:

https://github.com/user-attachments/assets/38d6238d-3df4-4c4d-abf9-ec17e6922548

paskozdilar:
I've modified the PR name to better align with changes.

protoc-gen-grpc-gateway/internal/gengateway/template.go
johanbrandhorst: I just had a crazy thought - what about if we defer `req.Body.Close`? How does that change the behavior?

paskozdilar: Unfortunately, it seems that `defer req.Body.Close()` doesn't change the behavior at all.
I'm not sure why - unlike Unary, the ServerStream methods return the stream immediately, and Body should be closed as soon as they start. But the same behavior happens - if client sends Body, the context is never closed...

The `go io.Copy` is the only method I've found works in my experimentation.
I'll try to investigate the root cause some time in the future.

johanbrandhorst (commented):
I'm sorry but I really think we need to get to the bottom of the behavior here before we can merge it. This proposed solution will affect all users of the gateway (the goroutine), and I want to be sure this is really necessary. The original fix (#5240) fixed a minor connection leak. The impact of that change was very small because it only affected RPCs with no body. Somehow it broke the websocket proxy. That is also probably a pretty small impact. Now fixing both of the issues requires creating a goroutine, which may have a big impact on _all_ users of the gateway. I just want to be sure this is the right solution, and honestly I'm still tempted to say we should just break the websocket proxy rather than impose this on all users.

paskozdilar:
Fair enough.

I'll try to investigate the root cause of this issue.

paskozdilar:
I've deduced that the issue appears in the [google.golang.org/grpc/internal/transport/clientStream.waitOnHeader()](https://github.com/grpc/grpc-go/blob/master/internal/transport/client_stream.go#L91) method - when request is not closed (or completely read, in case of no-body methods), the `s.headerChan` is never closed for some reason.
Adding `go io.Copy` has a side-effect of `s.headerChan` being closed, which unblocks the whole thing.

The [newClientWithParams](https://github.com/grpc/grpc-go/blob/master/stream.go#L266) function also mentions a comment that may be related:
```go
	// Possible context leak:
	// The cancel function for the child context we create will only be called
	// when RecvMsg returns a non-nil error, if the ClientConn is closed, or if
	// an error is generated by SendMsg.
	// https://github.com/grpc/grpc-go/issues/1818.
```
Indeed, during tests, the context doesn't seem to be closed when connection closes, but instead blocks forever.

This is what I have for now. I'll try to figure out why this happens in the first place.

paskozdilar:
You're right - it seems that this issue can be fixed in the websocket-proxy itself, instead of in the grpc-gateway.

The root cause of this is the conn-read -> req-write loop in the websocket proxy:
https://github.com/tmc/grpc-websocket-proxy/blob/master/wsproxy/websocket_proxy.go#L255

If nobody is reading the `req.Body`, the `Write` call gets stuck, and the loop never goes on to read the body again, so it never detects the connection close.
One possible solution is to add a "buffer" goroutine in the websocket-proxy itself. That way this PR is not necessary at all.

---

Arguably, it *should* be the server's responsibility to read requests, even when they are not needed e.g. in order to not block the network buffer.
Server should not assume how much data the client will send - and the cancellation of context should *definitely* not depend on client not sending more data than expected.

Of course, in this case, the fault is on the websocket-proxy library, since it mistakenly assumes that request write will always pass.
But the fact that this worked at all is because most Go HTTP servers require clients to send the whole body before even starting to process the data...

---

This has been a wild ride. There's a lot to consider here.
I've created the PR for websocket-proxy here:
https://github.com/tmc/grpc-websocket-proxy/pull/41

paskozdilar:
Wait, that won't work either. The blocking `io.Copy` prevents WebSocket from continuing in this case because `io.Copy` expects Body to be closed, which isn't the case with websocket-proxy - and cannot be, in general case, because proxy doesn't know which method expects request and which expects a stream. Only grpc-gateway knows this.

I'd like to point out that grpc-gateway already use a goroutine in the implementation of client-streaming methods (ClientStream and Bidirectional):

```go
        go func() {
		for {
			if err := handleSend(); err != nil {
				break
			}
		}
		if err := stream.CloseSend(); err != nil {
			grpclog.Errorf("Failed to terminate client stream: %v", err)
		}
	}()
```

For what it's worth, my benchmark with and without `go io.Copy` on Unary method that doesn't do anything (just returns a value) seem pretty close:

```
$ sudo nice -20 go test -bench=. -benchtime=10s -benchmem .
goos: linux
goarch: amd64
pkg: github.com/paskozdilar/bug-report-grpc-gateway
cpu: 13th Gen Intel(R) Core(TM) i7-13800H
BenchmarkUnary/UnaryOK-20                  84042            146442 ns/op           24236 B/op        360 allocs/op
BenchmarkUnary/UnaryGoIOCopy-20            83079            145399 ns/op           24227 B/op        360 allocs/op
PASS
ok      github.com/paskozdilar/bug-report-grpc-gateway  27.313s
```

johanbrandhorst:
So websocket clients with a body defined still work? I'm happy to put constraints on what types of websocket clients we can support, but I'm not really happy breaking any part of the gateway to support the websocket wrapper.

paskozdilar:
I understand. Messing with a widely used library is tricky, and one must be careful.

However, I'd like to present another more arguments for including `go io.Copy` in Unary and ServerStream methods anyway:
1. they don't break behavior, since the handler never reads any data beyond first message (or at all, if body is not defined)
2. if client sends more data than the default `marshaler.Decoder` ingests in the first scan (which is [512 bytes](https://cs.opensource.google/go/go/+/refs/tags/go1.24.1:src/encoding/json/stream.go;l=157)), the issue also happens with methods with body defined.

For reproduction of second bug, please see: https://github.com/paskozdilar/bug-report-grpc-gateway/tree/bug-when-body-defined

So this bug exists even with HTTP clients.
We *could* add a blocking `io.Copy` to body-defined Unary and ServerStream methods anyway, which would fix this issue for all HTTP clients regardless of excess body size, but it would break all WebSocket Unary/ServerStream methods.
In contrast, `go io.Copy` would fix both HTTP and WebSocket clients.

paskozdilar:
> Alternatively, similar mechanism could be used by specifying MethodType query parameter in WebSocket URL, similar to how HTTP method is defined.

Yes, I think this could work. That way we could add blocking `io.Copy` to all Unary and ServerStreaming methods, and everything would work for both HTTP and WebSocket clients.

I have a proof-of-concept example here: https://github.com/paskozdilar/grpc-gateway-proxy-example
This example uses my fork of tmc/grpc-websocket-proxy: https://github.com/paskozdilar/grpc-websocket-proxy

johanbrandhorst:
Thank you for spending so much time investigating this and laying out the options. I absolutely think a blocking `io.Copy` would acceptable. We could perhaps have a

```go
defer func() {
    io.Copy(io.Discard, req.Body)
    req.Body.Close()
}()
```

What do you think?

paskozdilar:
No, that would not work.

1. `req.Body.Close()` would only run after `io.Copy` finishes, and `io.Copy` only finishes when `req.Body` is closed
2. The call causing problems in ServerStream is `stream.Header()`, so the body must be drained until closed *before* `stream.Header()` is closed (more detailed explanation here: https://github.com/grpc-ecosystem/grpc-gateway/pull/5331#issuecomment-2753950071).

Considering many HTTP servers would read the body completely before even starting to process the response, I believe the correct behavior would be adding the following lines before stream.Header() call:

```go
// NBDUSM
[...]
_, err = req.Body.Read(make([]byte, 1))
if err != io.EOF {
    return nil, metadata, status.Errorf(codes.InvalidArgument, "method not expecting body")
}
[...]

// BDUSM
[...]
d := marshaler.NewDecoder(req.Body)
if err := d.Decode(&protoReq); err != nil && !errors.Is(err, io.EOF) {
}
if err := d.Decode(&struct{}{}); err != io.EOF {
    return nil, metadata, status.Errorf(codes.InvalidArgument, "body must be empty")
}
[...]
```

Interestingly, this issue of not finishing `json.Decoder` stream is documented in several blogposts and even an official Go issue:
- https://ahmet.im/blog/golang-json-decoder-pitfalls/
- https://mottaquikarim.github.io/dev/posts/you-might-not-be-using-json.decoder-correctly-in-golang/
- https://github.com/golang/go/issues/36225

---

Do you agree with this approach?

paskozdilar:
I have written a proof-of-concept fix with the above code on a brach `bug-all-together`:
https://github.com/paskozdilar/bug-report-grpc-gateway/tree/bug-all-together

Running the code prints pretty expected scenario - invalid-body gets rejected, valid body handles contexts nicely:

```
> Running invalid requests:
requesting: ServerStreamNoBody
requesting: UnaryNoBody
requesting: UnaryBody
requesting: ServerStreamBody
request failed: ServerStreamNoBody: 400 Bad Request
request failed: UnaryBody: 400 Bad Request
request failed: UnaryNoBody: 400 Bad Request
request failed: ServerStreamBody: 400 Bad Request
> Running valid requests:
requesting: ServerStreamNoBody
requesting: UnaryBody
requesting: UnaryNoBody
requesting: ServerStreamBody
ServerStreamNoBody open
UnaryBody open
UnaryNoBody open
ServerStreamBody open
request failed: ServerStreamNoBody: Post "http://localhost:8081/example/v1/ServerStreamNoBody": context canceled
request failed: UnaryNoBody: Post "http://localhost:8081/example/v1/UnaryNoBody": context canceled
request failed: UnaryBody: Post "http://localhost:8081/example/v1/UnaryBody": context canceled
request failed: ServerStreamBody: Post "http://localhost:8081/example/v1/ServerStreamBody": context canceled
UnaryNoBody close
UnaryBody close
ServerStreamNoBody close
ServerStreamBody close
```

---

As for WebSocket issue, would you be open to adding method-type metadata to response headers? That way, tmc/grpc-websocket-proxy could detect the method type automatically and adjust its behavior accordingly. That way, everything would work correctly!

paskozdilar:
This issue also seems to appear in some of the tests, e.g. many of the `testEcho*` and `testABE*` methods send body to method that doesn't expect it:
- [testEcho](https://github.com/grpc-ecosystem/grpc-gateway/blob/main/examples/internal/integration/integration_test.go#L226)
- [testEchoWithNonASCIIHeaderValues](https://github.com/grpc-ecosystem/grpc-gateway/blob/main/examples/internal/integration/integration_test.go#L2544)
- [testEchoWithInvalidHeaderKey](https://github.com/grpc-ecosystem/grpc-gateway/blob/main/examples/internal/integration/integration_test.go#L2582)
- [testABECreate](https://github.com/grpc-ecosystem/grpc-gateway/blob/main/examples/internal/integration/integration_test.go#L557)

This is a sign that many clients may expect methods to work even with excess body.

So the way I see it, there are two choices:
- use `go io.Copy(io.Discard, req.Body)` to drain body in the background, keeping the current methods working
- use strict request checking as described in [comment-2765779112](https://github.com/grpc-ecosystem/grpc-gateway/pull/5331#issuecomment-2765779112) and risk breaking some (non-conforming) clients

I am currently in favor of the second option - I think _strictness-that-breaks-some-clients_ is better than _relaxedness-that-allows-erroneous-behavior-to-go-unnoticed_.

What do you think?

paskozdilar:
Strange, tests are failing on http.MethodOption calls, which I didn't touch at all... I'll investigate this.

johanbrandhorst:
Thanks again for your work on this. Out of the options, I'm afraid that even if the third option is the more correct, we _cannot_ break users, even if they are doing the wrong thing today. Thus I think option 2 is preferred to me.

I'm confused about the need for the method type in the response headers. What do yo mean by method type? Could it be an option, or even better, a third party middleware?

paskozdilar:
> I'm confused about the need for the method type in the response headers

WebSocket proxy keeps the request writer open until WS conn is closed, while this PR will expect Unary / ServerStream request writers to be closed immediately after sending data.
However, on ClientStream / Bidirectional methods, they need to be kept open.
WebSocket proxy cannot, by itself, know when it is calling a Unary / ServerStream method, and when it is it is calling ClientStream / Bidirectional methods.

WebSocket proxy could add a query parameter, but I figured if it can *automatically* get that information from the response header itself, that would be more convenient for the users. You can see preliminary implementation here:
https://github.com/paskozdilar/grpc-websocket-proxy/blob/master/wsproxy/websocket_proxy.go#L201

paskozdilar:
Alright, the tests are passing now.

---

As for the server metadata, since we're pretty liberal about the request body, we only need *method type* metadata for WebSocket proxy to work fine without changes.

I see there's already some metadata header prefixes used in https://github.com/grpc-ecosystem/grpc-gateway/blob/main/runtime/context.go#L23:

```go
const MetadataHeaderPrefix = "Grpc-Metadata-"
[...]
const MetadataPrefix = "grpcgateway-"
[...]
const MetadataTrailerPrefix = "Grpc-Trailer-"
[...]
const xForwardedFor = "X-Forwarded-For"
const xForwardedHost = "X-Forwarded-Host"
```

I am not 100% sure what are those prefixes used for, but I'll dig in depeer.

We could define a new `runtime.AnnotationContextOption` for method type, which would append to response headers something like `X-Grpc-Method-Type: Unary`, etc.

johanbrandhorst:
I believe the header prefixes you found are used for parsing incoming headers, I don't know that we set any specific headers in the response. You can write a custom ForwardResponseOption that sets headers in the response (such as cookies), and maybe we could have some way for users to use that to inform the websocket of the type? I also think it's OK for us to require websocket users to set the type of the RPC explicitly. I expect doing it manually won't be that hard because who maintains more than 1-2 websockets around the gateway anyway?

johanbrandhorst (approved):
Thanks again for all your work on this, lets tackle the exact websocket wrapper work in the original issue. I can happily approve this PR knowing that there's nothing particularly bad about draining the request body.

PR #5340, opened 2025-03-10
Add support for x-go-type flag

### LINKED ISSUE #4927: Automatically annotate using x-go-type
## 🚀 Feature

In a way I'm a bit surprised this hasn't been done / hasn't been requested.

Given the content of the proto file, it would be quite easy to generate the `x-go-type` annotation automatically for the OpenAPIv2 spec file, such that go-swagger can re-use the go type and not regenerate a new one. We use a monorepo, so for us this is super powerful.

I've managed to get it going using

```protobuf
option go_package = "<path/to/my/package>";

message <TypeName> {
   option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_schema) = {
        json_schema: {
          extensions: {
            key: "x-go-type"
            value: {
                struct_value: { 
                    fields: {
                        key: "type",
                        value: { string_value: "<TypeName>" }
                    },
                    fields: {
                        key: "import",
                        value: { 
                            struct_value: {
                                fields: {
                                    key: "package",
                                    value: { string_value: "<path/to/my/package>" }
                                } 
                            }
                        }
                    }
                }
            }
          }
        }
      }
}
```

So I guess I'm curious as to whether this was considered / rejected, and if PRs would be welcome here.

We wouldn't want to pollute JsonSchema of course, so I'd suggest this could go through another annotation.

rohitlohar45:
**References to other Issues or PRs**

Resolves #4927

Yes

This pull request introduces support for the `x-go-type` extension in the generated OpenAPIv2 specifications. This extension allows tools such as go-swagger to reuse existing Go types rather than generating new ones.

**Other comments**

### DIFF
--- a/examples/internal/proto/examplepb/unannotated_echo_service.buf.gen.yaml
+++ b/examples/internal/proto/examplepb/unannotated_echo_service.buf.gen.yaml
@@ -10,3 +10,4 @@ plugins:
     opt:
       - grpc_api_configuration=examples/internal/proto/examplepb/unannotated_echo_service.yaml
       - openapi_configuration=examples/internal/proto/examplepb/unannotated_echo_service.swagger.yaml
+      - generate_x_go_type=true
--- a/protoc-gen-openapiv2/internal/genopenapi/template.go
+++ b/protoc-gen-openapiv2/internal/genopenapi/template.go
@@ -523,13 +523,16 @@ func renderMessageAsDefinition(msg *descriptor.Message, reg *descriptor.Registry
 		if schema.extensions == nil {
 			schema.extensions = []extension{}
 		}
+		goTypeName := msg.GetName()
+
+		goTypeName = casing.JSONCamelCase(goTypeName)
 		schema.extensions = append(schema.extensions, extension{
 			key: "x-go-type",
 			value: json.RawMessage(`{
                 "import": {
                     "package": "` + msg.File.GoPkg.Path + `"
                 },
-                "type": "` + msg.GetName() + `"
+                "type": "` + goTypeName + `"
             }`),
 		})
 	}
--- a/protoc-gen-openapiv2/internal/genopenapi/template.go
+++ b/protoc-gen-openapiv2/internal/genopenapi/template.go
@@ -2,6 +2,7 @@ package genopenapi
 
 import (
 	"bytes"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"math"
@@ -517,6 +518,22 @@ func renderMessageAsDefinition(msg *descriptor.Message, reg *descriptor.Registry
 			Type: "object",
 		},
 	}
+
+	if reg.GetGenerateXGoType() && msg.File.GoPkg.Path != "" {
+		if schema.extensions == nil {
+			schema.extensions = []extension{}
+		}
+		schema.extensions = append(schema.extensions, extension{
+			key: "x-go-type",
+			value: json.RawMessage(`{
+                "import": {
+                    "package": "` + msg.File.GoPkg.Path + `"
+                },
+                "type": "` + msg.GetName() + `"
+            }`),
+		})
+	}
+
 	msgComments := protoComments(reg, msg.File, msg.Outers, "MessageType", int32(msg.Index))
 	if err := updateOpenAPIDataFromComments(reg, &schema, msg, msgComments, false); err != nil {
 		return openapiSchemaObject{}, err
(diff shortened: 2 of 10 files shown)


protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -517,6 +518,22 @@ func renderMessageAsDefinition(msg *descriptor.Message, reg *descriptor.Registry
...
+		if schema.extensions == nil {
+			schema.extensions = []extension{}
+		}
+		schema.extensions = append(schema.extensions, extension{
+			key: "x-go-type",
+			value: json.RawMessage(`{
+                "import": {
+                    "package": "` + msg.File.GoPkg.Path + `"
+                },
+                "type": "` + msg.GetName() + `"
```
johanbrandhorst, line 532: I think this might not be enough to get the Go message name. Here is the logic for doing this in `protoc-gen-go`:

https://github.com/protocolbuffers/protobuf-go/blob/f7fcf5ba06e9fddb63b924d31704eca1eeabde88/compiler/protogen/protogen.go#L894

What is `msg.Name()` here? Is it already camelcased? We have the `internal/casing` package for doing this sort of thing elsewhere.

johanbrandhorst (changes requested):
A few things:

* Please add a new proto file that uses this option, or perhaps add it to one of the existing `buf.gen.yaml` files (not a_bit_of_everything.proto, it's too big). I want to see th effect this has on the protobuf generation.
* Please add this new flag to the bazel flags.

rohitlohar45:
@johanbrandhorst can you specify the bazel files to add the flag to, I am little bit confused in it

johanbrandhorst:
> @johanbrandhorst can you specify the bazel files to add the flag to, I am little bit confused in it

As an example: https://github.com/grpc-ecosystem/grpc-gateway/commit/f1c8b2a92c34bd2fe022e7fedfa8f300b02270ff

protoc-gen-openapiv2/internal/genopenapi/template.go
rohitlohar45: I have used internal/casing now, referring existing code

rohitlohar45:
@johanbrandhorst I have added this flag in bazel file and also added a test for the same

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -517,6 +518,25 @@ func renderMessageAsDefinition(msg *descriptor.Message, reg *descriptor.Registry
...
+			schema.extensions = []extension{}
+		}
+		goTypeName := msg.GetName()
+
+		goTypeName = casing.JSONCamelCase(goTypeName)
+		schema.extensions = append(schema.extensions, extension{
+			key: "x-go-type",
+			value: json.RawMessage(`{
+                "import": {
+                    "package": "` + msg.File.GoPkg.Path + `"
```
johanbrandhorst, line 533: I wonder if we should error if `msg.File.GoPkg.Path` is empty? Otherwise this is just not going to work.

examples/internal/proto/examplepb/unannotated_echo_service.buf.gen.yaml
```diff
@@ -10,3 +10,4 @@ plugins:
     opt:
       - grpc_api_configuration=examples/internal/proto/examplepb/unannotated_echo_service.yaml
       - openapi_configuration=examples/internal/proto/examplepb/unannotated_echo_service.swagger.yaml
+      - generate_x_go_type=true
```
johanbrandhorst, line 13: I would expect this to require us to regenerate the corresponding swagger file, were there no changes?

johanbrandhorst (changes requested):
Thanks for adding the flag, I had another thought.

protoc-gen-openapiv2/internal/genopenapi/template.go
rohitlohar45: If `msg.File.GoPkg.Path` is empty, the `x-go-type` won't be generated due to the condition `if reg.GetGenerateXGoType() && msg.File.GoPkg.Path != ""`. Would it be better to show an error in this case, or should we just display a warning to inform the user?

examples/internal/proto/examplepb/unannotated_echo_service.buf.gen.yaml
rohitlohar45: I have added the generated swagger file, sorry missed it

protoc-gen-openapiv2/internal/genopenapi/template.go
johanbrandhorst: Nah this is fine

johanbrandhorst (approved):

johanbrandhorst:
Looks like the yaml file still needs to be generated. See CONTRIBUTING.mb in the root of the repo for steps to regenerate all the files.

rohitlohar45:
@johanbrandhorst added the generated YAML file

PR #5510, opened 2025-04-25
Fix response_body

ndtretyak:

Yes

#4622 replaced `proto.Message` with a concrete response type in `response_*` structs and removed the line that cast `proto.Message`, since it seemed that we could now access `m`'s fields directly
https://github.com/grpc-ecosystem/grpc-gateway/pull/4622/files#diff-9648d8792ce733c8bbce0d613a620fffdb73cff98a91763507a1984eb520d5feL766 

However, if the response type has a field with the same name (e.g., `type Response struct {Response string}`) it does not work as expected.


johanbrandhorst:
Looks like we need to regenerate some bazel files, you can see the steps in CONTRIBUTING.md

johanbrandhorst:

Issue #3517, opened 2023-08-21
Wrong Path Parameter Naming When Using ':' in Path


arthaali:
## 🐛 Bug Report

Plugin version
- grpc-gateway/v2/protoc-gen-grpc-gateway@v2.15.0
- grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2@v2.15.0

The parameters are not correctly named when the path contains `:`

## To Reproduce

Given a proto
```
syntax = "proto3";

  rpc SearchItems(item.SearchItemsRequest) returns (item.SearchItemsResponse) {
    option (google.api.http) = {get: "/item/search:item/{item_no_query}"};
  }
```

2. buf generate openapiv2.swagger file `api.swagger.json`

 ```
 {
  "swagger": "2.0",
  "info": {
    "title": "api-rest-gateway/proto/api.proto",
    "version": "version not set"
  },
 "paths": {
  "/item/search:items/{item_no_query}": {
      ....
    },
   }
}
```

## Expected behavior

Expect to convert the parameter `item_no_query` to camel case `itemNoQuery`

 ```
 {
  "swagger": "2.0",
  "info": {
    "title": "api-rest-gateway/proto/api.proto",
    "version": "version not set"
  },
 "paths": {
  "/item/search:items/{itemNoQuery}": {
      ....
    },
   }
}
```

## Actual Behavior

The path parameter remains in snake case instead of being converted to the camel case.

```
{
  "swagger": "2.0",
  "info": {
    "title": "api-rest-gateway/proto/api.proto",
    "version": "version not set"
  },
  "paths": {
    "/item/search:items/{item_no_query}": {
      ...
    }
  }
}

```
## Your Environment

-Ubuntu
-Buf
-Plugin version: `protoc-gen-openapiv2@v2.15.0`

johanbrandhorst:
Thanks for your issue. Do you mean to say that other query parameters are correctly converted? It only happens when there's a `:` in the path?

johanbrandhorst:
[...]

PR #5601, opened 2025-05-24
fix(openapi): Convert path params after colon to `camelCase`

iamrajiv:

Fixes: #3517

`Core fix`: Process the verb segment for parameter `camelCase` conversion instead of ignoring it completely.

### DIFF
--- a/protoc-gen-openapiv2/internal/genopenapi/template.go
+++ b/protoc-gen-openapiv2/internal/genopenapi/template.go
@@ -1089,18 +1089,18 @@ func templateToParts(path string, reg *descriptor.Registry, fields []*descriptor
 			buffer += string(char)
 		case ':':
 			if depth == 0 {
-				// As soon as we find a ":" outside a variable,
-				// everything following is a verb. But we need to continue processing
-				// the remaining path to handle parameter camelCase conversion.
-				parts = append(parts, buffer)
-				verbSegment := path[i:]
-
-				if reg.GetUseJSONNamesForFields() {
-					verbSegment = processParametersInSegment(verbSegment, fields, msgs)
+				// Only treat this as a verb if we're at the end of the path or
+				// if there are no more path segments (only more literals after the colon)
+				remainingPath := path[i:]
+				if !strings.Contains(remainingPath, "/") {
+					parts = append(parts, buffer)
+					verbSegment := remainingPath
+					if reg.GetUseJSONNamesForFields() {
+						verbSegment = processParametersInSegment(verbSegment, fields, msgs)
+					}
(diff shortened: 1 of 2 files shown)


protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -1091,10 +1090,17 @@ pathLoop:
...
-				buffer = path[i:]
-				break pathLoop
+				verbSegment := path[i:]
+
+				if reg.GetUseJSONNamesForFields() {
+					verbSegment = processParametersInSegment(verbSegment, fields, msgs)
+				}
+
+				parts = append(parts, verbSegment)
+				return parts
```
johanbrandhorst, line 1103: It might be worth updating the comment here. I think a `verb`, per the [http.proto spec](https://github.com/googleapis/googleapis/blob/9bed6f077867e387960082891ec0e1b8bb254a47/google/api/http.proto#L219-L226) can only appear in the last path segment (after the last `/`), and can only contain `LITERAL`s (i.e., no variables). Therefore, this comment is not correct, as it handles anything after the first `:` as a verb. I would be curious to see what the logic in https://github.com/grpc-ecosystem/grpc-gateway/blob/d02adaf59b69676abffe58c6adac6b09eab33b2d/internal/httprule/parse.go does to the example added in the latest test. Does it identify `item/{item_no_query}` as a verb? Probably not. Could you please check by adding the string to https://github.com/grpc-ecosystem/grpc-gateway/blob/d02adaf59b69676abffe58c6adac6b09eab33b2d/internal/httprule/parse_test.go? Anyway, I think we may want to rewrite this logic to take into account that a `:` can appear anywhere, including outside the last path segment.

johanbrandhorst (changes requested):

protoc-gen-openapiv2/internal/genopenapi/template.go
iamrajiv: Thanks, @johanbrandhorst, for the review. So, what I understand is that you’re saying only colons after the last / and after all {} should be treated as verb markers, right? So, I guess I need to refactor in such a way that only segments containing parameters are processed for parameter conversion. Path segments with colons should be left alone, correct?

iamrajiv: Yes, @johanbrandhorst, I used parse in the httprule and created some test cases to verify it. I have made the changes and pushed them. Could you please check now?

johanbrandhorst (approved):
Alright, this seems to work as expected, thank you!

PR #5679, opened 2025-06-24
fix: Use type name instead of full type declaration

### LINKED ISSUE #5678: Invalid syntax when using standalone with response_body bindings
## 🐛 Bug Report

When using `standalone` with response body bindings, the generated code was incorrectly trying to refer to a field with the full type definition syntax (object.<import_package>.ObjectName). 

## To Reproduce

Generate a grpc-gateway file with the following buf.gen.yaml: 
```yaml
version: v1
plugins:
  - plugin: grpc-gateway
    out: .
    opt:
      - paths=source_relative
      - standalone=true
```
Where your proto has a response body binding in the http option, e.g.:
```proto
 rpc GetResponseBody(ResponseBodyIn) returns (ResponseBodyOut) {
    option (google.api.http) = {
      get: "/responsebody/{data}"
      response_body: "response"
    };
  }
```

## Expected behavior

Using the response_body_service.proto example:
```go
...
import (
	extExamplepb "github.com/grpc-ecosystem/grpc-gateway/v2/examples/internal/proto/examplepb"
)
...
type response_ResponseBodyService_GetResponseBody_0 struct {
	*extExamplepb.ResponseBodyOut
}

func (m response_ResponseBodyService_GetResponseBody_0) XXX_ResponseBody() interface{} {
	response := m.ResponseBodyOut
	return response.Response
}
```

## Actual Behavior

```go
...
import (
	extExamplepb "github.com/grpc-ecosystem/grpc-gateway/v2/examples/internal/proto/examplepb"
)
...
type response_ResponseBodyService_GetResponseBody_0 struct {
	*extExamplepb.ResponseBodyOut
}

func (m response_ResponseBodyService_GetResponseBody_0) XXX_ResponseBody() interface{} {
	response := m.extExamplepb.ResponseBodyOut // the "extExamplepb is the package import which should not be there
	return response.Response
}
```

## Your Environment

(Environment name, version and operating system.)

cabeedelhertz:

Fixes #5678 

When using `standalone=true`, this would end up as `response := m.extPackageNamepb.ResponseObject` where extPackageNamepb is the import alias for the main proto type package. This resulted in invalid syntax and failed to compile.

Seems to have been introduced here #5510 which went out in the latest release. Noticed this issue when bumping protos, and the registering of my gateway handler failed to compile
#### Other comments

for reference, current behavior vs with this fix, using the response_body_service.proto example:

### Current Behavior

```go
...
import (
	extExamplepb "github.com/grpc-ecosystem/grpc-gateway/v2/examples/internal/proto/examplepb"
)
...
type response_ResponseBodyService_GetResponseBody_0 struct {
	*extExamplepb.ResponseBodyOut
}

func (m response_ResponseBodyService_GetResponseBody_0) XXX_ResponseBody() interface{} {
	response := m.extExamplepb.ResponseBodyOut // the "extExamplepb is the package import which should not be there
	return response.Response
}
```
### New behavior

```go
...
import (
	extExamplepb "github.com/grpc-ecosystem/grpc-gateway/v2/examples/internal/proto/examplepb"
)
...
type response_ResponseBodyService_GetResponseBody_0 struct {
	*extExamplepb.ResponseBodyOut
}

func (m response_ResponseBodyService_GetResponseBody_0) XXX_ResponseBody() interface{} {
	response := m.ResponseBodyOut
	return response.Response
}
```


examples/internal/proto/standalone/unannotated_echo_service.pb.gw.go
```diff
@@ -768,9 +815,35 @@ func RegisterUnannotatedEchoServiceHandlerClient(ctx context.Context, mux *runti
...
+	})
 	return nil
 }
 
+type response_UnannotatedEchoService_EchoNested_0 struct {
+	*extExamplepb.UnannotatedSimpleMessage
+}
+
+func (m response_UnannotatedEchoService_EchoNested_0) XXX_ResponseBody() interface{} {
+	response := m.UnannotatedSimpleMessage
```
cabeedelhertz, line 843: example of `standalone=true`, the type defined on line 839 is the type defined from the external package, but the reference to the object uses the actual type name because thats how we'd reference it from `m`

PR #5818, opened 2025-08-08
Scaffold openapiv3 plugin

lootag:


celaus:
superseeded by #6010 so this one can be closed 😄

PR #5887, opened 2025-09-03
fix(runtime): annotate context even if no incoming headers

### LINKED ISSUE #5885: metadataAnnotators not called when no incoming headers
## 🐛 Bug Report

I'm not sure if this is a bug or wanted (if unwanted I can fix it in a PR); but right now metadataAnnotators are not called if there are no incoming headers in this piece of code:

https://github.com/grpc-ecosystem/grpc-gateway/blob/main/runtime/context.go#L204-L210

## To Reproduce

(Write your steps here:)

1. Register an annotator with runtime.WithMetadata
2. Send a request without incoming headers, we can set for it:
```go
runtime.WithIncomingHeaderMatcher(func(s string) (string, bool) {
     return s, false
})
```
3. The annotator is not called

## Expected behavior

Annotator gets called

## Actual Behavior

Annotator not called

## Your Environment

Mac OS

boolangery:

Fixes #5885

Metadata annotator was not called when no incoming headers. In practice it's not happening in production since most requests will contain some sort of header. But it happens in some unit tests.

### DIFF
--- a/runtime/context.go
+++ b/runtime/context.go
@@ -201,12 +201,12 @@ func annotateContext(ctx context.Context, mux *ServeMux, req *http.Request, rpcM
 	if timeout != 0 {
 		ctx, _ = context.WithTimeout(ctx, timeout)
 	}
-	if len(pairs) == 0 {
-		return ctx, nil, nil
-	}
 	md := metadata.Pairs(pairs...)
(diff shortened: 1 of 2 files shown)


johanbrandhorst (approved):
LGTM!

johanbrandhorst:
Thanks for your contribution!

Issue #5684, opened 2025-06-25
Unexpected/missing gRPC messages published in Swagger JSON/API doc


boseabhishek:
## 🐛 Bug Report

I'm trying to generate Swagger based OpenAPI docs for my grpc services, proxied to REST endpoints via `grpc-gateway`. Please see details below:

1. I have an accounts.proto whoch has two RPC services - `Add` and `Show`. I want to ONLY expose `Show`( `GET /v1/account) which is achieved by using annotation `google.api.http` and not using teh same on `Add`.

**accounts.proto**

```
syntax = "proto3";

package account;

import "google/api/annotations.proto";
import "protoc-gen-openapiv2/options/annotations.proto";

option go_package = "accounts/pkg/account;account";

option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger) = {
  info: {
    title: "Account Service";
    version: "1.0";
    contact: {
      name: "some-name";
      url: "https://github.com/some-name";
      email: "some-name@example.com";
    };
  };
  schemes: [HTTP, HTTPS];
  consumes: ["application/json"];
  produces: ["application/json"];
  responses: {
    key: "default";
    value: {
      description: "An unexpected error occurred.";
    }
  }
};


service AccountService {
  rpc Add(AddRequest) returns (AddResponse);
  rpc Show(ShowRequest) returns (ShowResponse) {
    option (google.api.http) = {
      get: "/v1/account"
    };
  }
}

message Money {
  option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_schema) = {
      json_schema: {
          title: "Money"
          description: "Money data structure"
      };
  };

  int64 amount = 1;
}


message AddRequest {
  Money money = 1;
}

message AddResponse {
}

message ShowRequest {
}

message ShowResponse {
  Money money = 1;
} 
```

2. But on the generated swagger.json, I can see unexpected messages in definitions

**swagger.json** (excerpt)

```
"definitions": {
    "accountAddResponse": {
      "type": "object"
    },
    "accountMoney": {
      "type": "object",
      "properties": {
        "amount": {
          "type": "string",
          "format": "int64"
        }
      },
      "description": "Money data structure",
      "title": "Money"
    },
    "accountShowResponse": {
      "type": "object",
      "properties": {
        "money": {
          "$ref": "#/definitions/accountMoney"
        }
      }
    },
    "protobufAny": {
      "type": "object",
      "properties": {
        "@type": {
[...]

PR #6030, opened 2025-10-23
fix: Exclude methods without HTTP bindings from OpenAPI definitions

iamrajiv:

Fixes: #5684

### DIFF
--- a/protoc-gen-openapiv2/internal/genopenapi/generator_test.go
+++ b/protoc-gen-openapiv2/internal/genopenapi/generator_test.go
@@ -1942,3 +1942,136 @@ func TestGenerateXGoType(t *testing.T) {
 		})
 	}
 }
+
+// TestIssue5684_UnusedMethodsNotInOpenAPI tests that methods without HTTP bindings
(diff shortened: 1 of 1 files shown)


iamrajiv:
Hi @johanbrandhorst, thanks for the review. I tested the generate_unbound_methods flow and it works fine. OpenAPIv2 already had this flag. When enabled it auto-creates default POST bindings for methods without HTTP annotations (descriptor/services.go:46–52). When disabled those methods just have empty Bindings arrays. My len(meth.Bindings) == 0 check correctly handles that.

johanbrandhorst (commented):
Thanks for the response, were you able to confirm that this change fixes the issue reported by the user in #5684?

iamrajiv:
Yes, @johanbrandhorst, I’ve confirmed that this change fixes the issue reported in #5684. I added a test case that reproduces the exact scenario from the issue. Should I push that local test case here?

johanbrandhorst:
Yes that would be great if it's not too big of a change

PR #6076, opened 2025-11-06
Feature: Add strict URL pattern validation to prevent routing issues

Suryanshu185:
Implement comprehensive URL pattern validation in grpc-gateway to prevent common routing conflicts and performance issues. Validation is enabled by default with strict enforcement to ensure API reliability.

Key features:
- Validates maximum 4 static segments at URL pattern start
- Prevents consecutive static segments after dynamic segments
- Fails compilation on invalid patterns with clear error messages

The validator integrates into the HTTP rule parsing pipeline and validates patterns during protobuf code generation. Invalid patterns cause compilation to fail, forcing developers to use proper URL design patterns.

### DIFF
--- a/internal/httprule/parse.go
+++ b/internal/httprule/parse.go
@@ -21,6 +21,12 @@ func Parse(tmpl string) (Compiler, error) {
 	if !strings.HasPrefix(tmpl, "/") {
 		return template{}, InvalidTemplateError{tmpl: tmpl, msg: "no leading /"}
 	}
+
+	// Validate URL pattern will prevent invalid patterns issues
+	if err := ValidateTemplate(tmpl); err != nil {
+		return template{}, InvalidTemplateError{tmpl: tmpl, msg: err.Error()}
+	}
+
 	tokens, verb := tokenize(tmpl[1:])
 
 	p := parser{tokens: tokens}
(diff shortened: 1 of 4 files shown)


johanbrandhorst:
Hi, I'm a little confused, what kind of issues are we trying to prevent here? Could you create an issue with some example problems that this fixes?

Also.. I hate to ask this, but I'm getting some AI vibes from this. Did you use AI when creating this PR? Please disclose any use of AI.

Suryanshu185:
If someone wants to use a TRIE to structure their API endpoint URLs, it may not work correctly unless the URLs follow the specific structure and the way i have strcutured it is flexible to be bent according to use case this PR has the rules bent according to my use case in my project (thought others may face it someday) If it does seem like a real issue that someone else might also face i can create an issue and i can make it more generic than it already is :).
Also i did in fact write the commit message with AI (i just believe it provides a much cleaner representation of information if i use AI to polish and rewrite the commit message for me)

@johanbrandhorst

johanbrandhorst:
Please open an issue first. I need a concrete use case. Show me your minimized proto file that is currently failing which requires this additional "validation". I don't really understand why we need to restrict the number of static segments. We can come back to the PR if the outcome of the issue is to make a change.

Issue #2635, opened 2022-04-10
protoc-gen-openapiv2: `google.api.field_behavior = REQUIRED` annotates the wrong schema object


cloneable:
## 🐛 Bug Report

protoc-gen-openapiv2 offers to mark properties as required based on the `google.api.field_behavior` field option, but it adds the `"required": ["<property>"],` annotation to the property itself in addition to adding it to the object containing the property.

https://grpc-ecosystem.github.io/grpc-gateway/docs/mapping/customizing_openapi_output/#using-googleapifield_behavior

## To Reproduce

```proto
import "protoc-gen-openapiv2/options/annotations.proto";
import "google/api/field_behavior.proto";

message TestMessage {
  option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_schema) = {
    json_schema: { required: [ "fieldOne", "fieldTwo" ] }
  };

  string field_one = 1;

  string field_two = 2 [(google.api.field_behavior) = REQUIRED];

  string field_three = 3 [(google.api.field_behavior) = REQUIRED];
}
```

Run `protoc-gen-openapiv2`.

## Expected Output

```json
    "testingTestMessage": {
      "type": "object",
      "properties": {
        "fieldOne": {
          "type": "string"
        },
        "fieldTwo": {
          "type": "string"
        },
        "fieldThree": {
          "type": "string"
        }
      },
      "required": [
        "fieldOne",
        "fieldTwo",
        "fieldThree"
      ]
    }
```

## Actual Output

```json
    "testingTestMessage": {
      "type": "object",
      "properties": {
        "fieldOne": {
          "type": "string"
        },
        "fieldTwo": {
          "type": "string",
          "required": [     <-- wrong
            "fieldTwo"
          ]
        },
        "fieldThree": {
          "type": "string",
          "required": [     <-- wrong
            "fieldThree"
          ]
        }
      },
      "required": [
        "fieldOne",
        "fieldTwo",
        "fieldTwo",     <-- duplicate
        "fieldThree"
      ]
    }
```

## Your Environment

github.com/grpc-ecosystem/grpc-gateway/v2 v2.10.0

johanbrandhorst:
Hi, thanks for the issue! That definitely looks wrong. If you're interested in helping to contribute a fix, the first step is usually to create a failing test. It looks like with what you've got here that shouldn't be too hard at all. Would you be interested in contributing a fix?

CC @ewhauser

[...]

PR #6078, opened 2025-11-06
fix(openapiv2): prevent nested required fields hoisting to parent schema

rdark:
# Bug: Nested Required Fields Incorrectly Hoisted to Parent Body Schema

## Summary

`protoc-gen-openapiv2` incorrectly hoists nested message field names marked as `REQUIRED` into the parent request body's `required` array when those fields are also used in path parameters. This creates invalid OpenAPI schemas where the `required` array references field names that don't exist at the top level.

## Version

- grpc-gateway v2.27.3 (also affects earlier versions)

## Prerequisites

The bug occurs when ALL these conditions are met:

1. **Update RPC** with a request message containing a nested message field
2. **Path parameter extraction**: At least one field from the nested message is used in the URL path (e.g., `{thing.name}`)
3. **REQUIRED annotations**: The nested message has fields marked with `(google.api.field_behavior) = REQUIRED`
4. **Body parameter generation**: The RPC uses `body: "*"` or similar

### Proto Definition

```protobuf
message Foo {
  string name = 1 [(google.api.field_behavior) = REQUIRED];
  string value = 2 [(google.api.field_behavior) = REQUIRED];
}

message UpdateFooRequest {
  Foo thing = 1 [(google.api.field_behavior) = REQUIRED];
  google.protobuf.FieldMask update_mask = 2;
}

service FooService {
  rpc UpdateFoo(UpdateFooRequest) returns (UpdateFooResponse) {
    option (google.api.http) = {
      patch: "/api/v1/{thing.name}"
      body: "*"
    };
  }
}
```

### Current (Incorrect) Output

```json
{
  "definitions": {
    "FooServiceUpdateFooBody": {
      "type": "object",
      "properties": {
        "thing": {
          "type": "object",
          "properties": {
            "value": { "type": "string" }
          }
        },
        "updateMask": { "type": "string" }
      },
      "required": [
        "value",      // WRONG: "value" is nested inside "thing"!
        "thing"
      ]
    }
  }
}
```

### Expected (Correct) Output

```json
{
  "definitions": {
    "FooServiceUpdateFooBody": {
      "type": "object",
      "properties": {
        "thing": {
          "type": "object",
          "properties": {
            "value": { "type": "string" }
          },
          "required": ["value"]  // Nested required stays here
        },
        "updateMask": { "type": "string" }
      },
      "required": ["thing"]  // Only top-level fields
    }
  }
}
```

## Root Cause

In `protoc-gen-openapiv2/internal/genopenapi/template.go` (around line 609-615), when processing nested message fields with path parameters, ALL required field names from the nested schema are hoisted to the parent's `required` array without distinguishing between:
- The field name itself (which should be in parent's required if marked REQUIRED)
- Nested field names within that field (which should stay in nested schema)

possibly originated in https://github.com/grpc-ecosystem/grpc-gateway/issues/2635

### DIFF
--- a/examples/internal/clients/abe/api/swagger.yaml
+++ b/examples/internal/clients/abe/api/swagger.yaml
@@ -6500,6 +6500,15 @@ definitions:
     description: "Nested is nested type."
     example: "{\"ok\":\"TRUE\"}"
   A bit of everything_2:
+    required:
+    - "doubleValue"
+    - "floatValue"
+    - "int64Value"
+    - "requiredStringField1"
+    - "requiredStringField2"
+    - "requiredStringViaFieldBehaviorAnnotation"
+    - "required_field_behavior_json_name_custom"
+    - "required_field_schema_json_name_custom"
     properties:
       singleNested:
         $ref: "#/definitions/ABitOfEverythingNested"
--- a/examples/internal/clients/abe/model_a_bit_of_everything_2.go
+++ b/examples/internal/clients/abe/model_a_bit_of_everything_2.go
@@ -19,9 +19,9 @@ type ABitOfEverything2 struct {
 	SingleNested *ABitOfEverythingNested `json:"singleNested,omitempty"`
 	Nested []ABitOfEverythingNested `json:"nested,omitempty"`
 	// Float value field
-	FloatValue float32 `json:"floatValue,omitempty"`
-	DoubleValue float64 `json:"doubleValue,omitempty"`
-	Int64Value string `json:"int64Value,omitempty"`
+	FloatValue float32 `json:"floatValue"`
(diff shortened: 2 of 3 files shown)


johanbrandhorst (changes requested):
Hi, thanks for this high quality PR! It looks like this change fixed some issues in the existing generated examples, could you regenerate the files and commit them? See CONTRIBUTING.md for a guide. Thanks!

Note: think the proto_lint failure is a red herring, a rebase should fix it.

johanbrandhorst (approved):
Thanks a lot!

johanbrandhorst:
Please rebase, thank you!

johanbrandhorst:
Thank you for contributing :)

Issue #6173, opened 2025-12-15
Hope to upgrade the version of Gulp


rainmenzhao:
## 🚀 Feature

(A clear and concise description of what the feature is.)
The current version of Gulp is 3.9.1 in examples/internal/browser. The highest supported Node version for it is 11, while the Node ITS is 24,thanks!

johanbrandhorst:
Yeah we should update our node dependencies, but not that we only use it for the integration tests, no part of the released binary contains or relies on node code. Would you be able to help us upgrade these dependencies? The biggest challenge will be ensuring that the integration tests and the swagger generator can continue running.

essapp:
> Yeah we should update our node dependencies, but not that we only use it for the integration tests, no part of the released binary contains or relies on node code. Would you be able to help us upgrade these dependencies? The biggest challenge will be ensuring that the integration tests and the swagger generator can continue running.

After analysis and testing, I can upgrade Gulp and optimize the related operations. You can assign the task to me. Thank you very much!

PR #6191, opened 2025-12-22
feat: #6173

essapp:
Upgrade gulp. The main focus is to conduct tests on the gateway server and the grpc server. Therefore, Swagger UI has been used to replace the previous testing method. At the same time, try to utilize the original functions of the node to facilitate future upgrades.


johanbrandhorst:
Looks like the node tests are failing after this change. Where did the tests go? Looks like we deleted the generated echo_service and a_bit_of_everything JS files. How are we supposed to interact with the API servers without the generated files?

Issue #6234, opened 2026-01-13
use_opaque_api generates invalid code


jpeach:
## 🐛 Bug Report

Enabling `use_opaque_api` can generate code that attempts to pass protobuf messages by value, which fails to compile.

Sorry that I didn't test this out earlier :-(

## To Reproduce

Given the definitions:
```protobuf
  rpc UpdateAppointment(UpdateAppointmentRequest) returns (Appointment) {
    option (google.api.http) = {
    patch: "/{appointment.stylist=stylists/*}/{appointment.name=appointments/*}"
      body: "appointment"
    };
    option (google.api.http).additional_bindings = {
    patch: "/{appointment.salon=salons/*}/{appointment.name=appointments/*}"
      body: "appointment"
    };
    option (google.api.method_signature) = "appointment,update_mask";
  }
```

The following code is generated:
```Go
func request_AppointmentService_UpdateAppointment_0(ctx context.Context, marshaler runtime.Marshaler, client AppointmentServiceClient, req *http.Request, pathParams map[string]string) (proto.Message, runtime.ServerMetadata, error) {        var (
                protoReq UpdateAppointmentRequest
                metadata runtime.ServerMetadata
                err      error
        )
        newReader, berr := utilities.IOReaderFactory(req.Body)
        if berr != nil {
                return nil, metadata, status.Errorf(codes.InvalidArgument, "%v", berr)
        }
        var bodyData Appointment
        if err := marshaler.NewDecoder(newReader()).Decode(&bodyData); err != nil && !errors.Is(err, io.EOF) {
                return nil, metadata, status.Errorf(codes.InvalidArgument, "%v", err)
        }
        protoReq.SetAppointment(bodyData)
        if req.Body != nil {
                _, _ = io.Copy(io.Discard, req.Body)
        }
...
}
```

The `protoReq.SetAppointment(bodyData)` call fails to compile because in this case `bodyData` is not a pointer (it is generated as a pointer in other cases).

## Expected behavior

Should compile :)

## Actual Behavior

Doesn't compile :)

## Your Environment

grpc-gateway v2.27.4

jpeach:
cc @kop

kop:
@jpeach, fixed by https://github.com/grpc-ecosystem/grpc-gateway/pull/6246

jpeach:
[...]

PR #6246, opened 2026-01-15
fix(gengateway): use pointer for bodyData in OpaqueAPI PATCH requests

kop:
When generating code for the OpaqueAPI with a PATCH request that includes a FieldMask, the generated code incorrectly declared `bodyData` as a value type. This caused a compilation error when passing it to the `SetXXX` method, which expects a pointer for message types.

This change updates the templates to declare `bodyData` as a pointer and decode directly into it.

Additionally, extended `examples/internal/proto/examplepb/opaque.proto` with an `OpaqueUpdateProduct` method to serve as a test and reproduction case for this scenario.

Fixes: #6234

### DIFF
--- a/protoc-gen-grpc-gateway/internal/gengateway/template.go
+++ b/protoc-gen-grpc-gateway/internal/gengateway/template.go
@@ -8,12 +8,11 @@ import (
 	"strings"
 	"text/template"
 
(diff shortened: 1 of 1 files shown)


johanbrandhorst:
Thank you!

PR #6335, opened 2026-02-06
fix(openapiv2): exclude oneof fields from required with proto3 field semantics

### LINKED ISSUE #6334: protoc-gen-openapiv2: oneof fields incorrectly marked as required with use_proto3_field_semantics=true
## Bug Report

When using `use_proto3_field_semantics=true`, `oneof` member fields are incorrectly added to the `required` array of the parent message's OpenAPI schema. This is semantically wrong because a `oneof` means **at most one** of the fields can be set — they should never all be required simultaneously.

## Steps to Reproduce

Given a proto message with a `oneof`:

```proto
message Filter {
  oneof filter_group_or_value {
    FilterList list = 1;
    FilterRule rule = 2;
  }
}
```

Run `protoc-gen-openapiv2` with `use_proto3_field_semantics=true`.

## Expected Output

`oneof` fields should be **excluded** from the `required` array, since only one can be set at a time:

```json
"Filter": {
  "type": "object",
  "properties": {
    "list": {
      "$ref": "#/definitions/FilterList"
    },
    "rule": {
      "$ref": "#/definitions/FilterRule"
    }
  }
}
```

## Actual Output

Both `oneof` members are marked as required:

```json
"Filter": {
  "type": "object",
  "properties": {
    "list": {
      "$ref": "#/definitions/FilterList"
    },
    "rule": {
      "$ref": "#/definitions/FilterRule"
    }
  },
  "required": [
    "list",
    "rule"
  ]
}
```

## Analysis

The `use_proto3_field_semantics=true` option marks all proto3 fields that are not `optional` as required. However, it does not check whether a field is part of a `oneof` group. Fields within a `oneof` should be treated as implicitly optional (only one can be present), so they should be excluded from the `required` list.

A potential fix would be to check `field.OneofIndex` in `updateSwaggerObjectFromFieldBehavior` and skip adding `oneof` members to the `required` array when `use_proto3_field_semantics` is enabled.

## Environment

- Plugin: `buf.build/grpc-ecosystem/openapiv2:v2.27.3`
- Configuration:
  ```
  opt:
    - use_proto3_field_semantics=true
    - json_names_for_fields=true
    - output_format=json
  ```

## Workaround

Currently the only workarounds are:
1. Remove `use_proto3_field_semantics=true` (makes all fields optional)
2. Post-process the generated swagger to strip `oneof` fields from `required` arrays
3. Use per-field `google.api.field_behavior = REQUIRED` annotations instead of the blanket flag

sessa:
## Summary

- Fixes `oneof` fields being incorrectly marked as `required` when `use_proto3_field_semantics=true` is set
- Adds a `field.OneofIndex == nil` check in `updateSwaggerObjectFromFieldBehavior` so that oneof members are excluded from the required array
- Explicit `REQUIRED` field behavior annotations still override this correctly

## Problem

When `use_proto3_field_semantics=true`, all non-optional proto3 fields are added to the OpenAPI `required` array. This includes `oneof` members, which is semantically wrong -- a `oneof` means at most one field can be set, so they should never all be required.

For example, this proto:

```proto
message Filter {
  oneof filter_group_or_value {
    FilterList list = 1;
    FilterRule rule = 2;
  }
}
```

Generates both `list` and `rule` as required, when neither should be.

## Fix

One-line change in `updateSwaggerObjectFromFieldBehavior`:

```go
// Before
required = !field.GetProto3Optional()

// After
required = !field.GetProto3Optional() && field.OneofIndex == nil
```

Proto3 `optional` fields use synthetic oneofs (`OneofIndex != nil`), but `GetProto3Optional()` already handles that case. The `OneofIndex == nil` check catches real `oneof` groups.

## Test plan

- [x] Added 3 new test cases to `Test_updateSwaggerObjectFromFieldBehavior`:
  - Oneof field not required with proto3 field semantics
  - Oneof field not required without proto3 field semantics
  - Oneof field still required when explicitly annotated with `REQUIRED`
- [x] All existing tests pass with no regressions

Fixes #6334

Made with [Cursor](https://cursor.com)

### DIFF
--- a/examples/internal/proto/examplepb/proto3_field_semantics.proto
+++ b/examples/internal/proto/examplepb/proto3_field_semantics.proto
@@ -8,9 +8,7 @@ option go_package = "github.com/grpc-ecosystem/grpc-gateway/v2/examples/internal
 
 service Proto3FieldSemanticsService {
   rpc GetFilter(GetFilterRequest) returns (GetFilterResponse) {
-    option (google.api.http) = {
-      get: "/v1/filter"
-    };
+    option (google.api.http) = {get: "/v1/filter"};
   }
 }
 


sessa:
@johanbrandhorst, hope you had a good weekend -- thank you for the feedback (and this tool). Let me know if I you need anything else from me.

sessa:
Added it to the ignore sections like the other example use cases, tested locally, `buf lint` should pass now.

johanbrandhorst:
> Added it to the ignore sections like the other example use cases, tested locally, `buf lint` should pass now.

Busted by the formatter this time 😂. Sorry!

sessa:
My fault.  I think that should do it. Thank you!

sessa:
@johanbrandhorst -- what's the releasee process like for this?

Thank you!

johanbrandhorst:
I'll usually put one out every few months or if there is a significant bug fix, but I think we can put one out for this and other fixes, so I did: https://github.com/grpc-ecosystem/grpc-gateway/releases/tag/v2.27.8.

bleblanc-dev:
Hi @johanbrandhorst what steps are required to publish `2.27.8` to the Buf Schema Registry ? we are relying on it to fetch the new version.

https://buf.build/grpc-ecosystem/openapiv2

johanbrandhorst:
The buf.build plugin version is maintained by buf: https://github.com/bufbuild/plugins/tree/main/plugins/grpc-ecosystem/openapiv2. I'm not sure why it hasn't been updated yet, maybe a problem with their automation? You could try reaching out on their slack or opening an issue on the repo.

PR #6384, opened 2026-02-19
fix: use proto.Merge instead of value assignment in opaque API mode

### LINKED ISSUE #6380: use_opaque_api=true: generated code copies proto message by value (go vet copylocks violation)
## Summary

When `use_opaque_api: true` is set in the protoc-gen-grpc-gateway options,
the generated `*.pb.gw.go` files contain a value assignment that triggers
`go vet copylocks`:

```go
var bodyData XxxRequest
marshaler.NewDecoder(req.Body).Decode(&bodyData)
protoReq = bodyData  // ← copies protoimpl.MessageState (contains sync.Mutex)
```

## To Reproduce

1. Add use_opaque_api: true to buf.gen.yaml:

```yaml
plugins:
  - plugin: grpc-gateway
    out: gen
    opt:
      - use_opaque_api=true
```

2. Define a proto with an HTTP body binding:

```proto
rpc CreateFoo(CreateFooRequest) returns (CreateFooResponse) {
  option (google.api.http) = {
    post: "/v1/foos"
    body: "*"
  };
}
```

3. Run buf generate and then exec `go vet`

## Expected behavior

No go vet errors. The generated code should use proto.Merge to avoid copying the mutex embedded in protoimpl.MessageState.

## Actual Behavior

`./gen/foo/foo.pb.gw.go:XX:2: assignment copies lock value to protoReq: google.golang.org/protobuf/internal/impl.MessageState contains google.golang.org/protobuf/internal/impl.noCopy contains sync.Mutex`

### Root Cause

In opaque API mode, proto messages embed protoimpl.MessageState, which contains sync.Mutex via noCopy.
Value assignment (protoReq = bodyData) copies this mutex, violating go vet copylocks.

The fix is to replace the value assignment with proto.Merge:

```go
// Before (generated, broken):
protoReq = bodyData

// After (correct):
proto.Merge(&protoReq, &bodyData)
```

The relevant template is likely in: `protoc-gen-grpc-gateway/internal/gengateway/template.go`

### Workaround

We currently post-process generated files with a sed-like patch to replace `protoReq = bodyData` → `proto.Merge(&protoReq, &bodyData)` .

## Your Environment

- grpc-gateway: v2.28.0
- buf options: use_opaque_api: true
- Go: 1.23.x
- OS: macOS (darwin)

clarabennettdev:
Fixes #6380

## Summary

When `use_opaque_api=true`, the generated `*.pb.gw.go` files contain value assignments (`protoReq = bodyData`) that copy `protoimpl.MessageState`, which embeds `sync.Mutex` via `noCopy`. This triggers `go vet copylocks` violations.

## Changes

Replaced all 4 occurrences of `protoReq = bodyData` with `proto.Merge(&protoReq, &bodyData)` in `protoc-gen-grpc-gateway/internal/gengateway/template.go`:

- Request handler (non-fieldmask, body="*")
- Request handler (fieldmask, body="*")
- Local request handler (non-fieldmask, body="*")
- Local request handler (fieldmask, body="*")

`proto` is already imported in generated code via `generator.go`, so no import changes needed.

All existing tests pass.

### DIFF
--- a/protoc-gen-grpc-gateway/internal/gengateway/template.go
+++ b/protoc-gen-grpc-gateway/internal/gengateway/template.go
@@ -405,7 +405,7 @@ var filter_{{ .Method.Service.GetName }}_{{ .Method.GetName }}_{{ .Index }} = {{
 	if err := marshaler.NewDecoder(req.Body).Decode(&bodyData); err != nil && !errors.Is(err, io.EOF) {
 		return nil, metadata, status.Errorf(codes.InvalidArgument, "%v", err)
 	}
-	protoReq = bodyData
+	proto.Merge(&protoReq, &bodyData)
 	{{- else }}
 	bodyData := &{{ .GetBodyFieldType }}{}
 	if err := marshaler.NewDecoder(req.Body).Decode(bodyData); err != nil && !errors.Is(err, io.EOF) {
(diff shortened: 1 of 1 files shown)


johanbrandhorst:
Thanks for this, looks like a dupe of #6383?

Issue #6352, opened 2026-02-11
Allow optional fields in field path


sloretz:
## 🚀 Feature

I'm annotating an existing gRPC service, and one of the request messages has an `optional` field that's only used in a specific use case. I want to annotate the RPC with an `additional_binding` for that specific use case. Following [Resource-oriented Design](https://google.aip.dev/121) as best I can, the `optional` field would fit better in the path than as a query parameter.

Here's a made-up example:

```proto
service AnimalService {
  rpc Feed(FeedRequest) returns (FeedResponse) {
    option (google.api.http) = {
      post: "/v1/animals/{name}:feed"
      body: "*"
      additional_bindings {
        // ERROR: Can't use owner_name in the path because it is marked as 'optional'.
        post: "/v1/owners/{metadata.owner_name}/pets/{name}:feed"
        body: "*"
      }
    };
  }
}

message FeedRequest {
  string name = 1;
  Metadata metadata = 2;
}

message Metadata {
  // Only pets have a named owner
  optional string owner_name = 1;
  // ...
}

message FeedResponse {
  // ...
}
```

However, gRPC Gateway #1951 added logic that forbids putting optional fields into the path.

```
--grpc-gateway_out: optional field not allowed in field path: owner_name in metadata.owner_name
```

johanbrandhorst:
Hm, I see how this is useful in this special case, because the path is only allowed as an alternative to another path. The [`http.proto` spec](https://github.com/googleapis/googleapis/blob/master/google/api/http.proto) doesn't explicitly state that the path parameter must not be an optional field. I'm not sure it's clear to me what it means to make the field optional though. Is

```
POST /v1/owners//pets/fido:feed
```

A valid request?

JackBogart:
Funny that this issue just got created, I've been running into issues with marking path parameters as `optional` too and was about to create an issue.

It's weird that we can't mark path fields as optional, especially when the [protobuf documentation recommends](https://protobuf.dev/programming-guides/field_presence/#background) adding the label for proto3 basic types:

```
We recommend always adding the optional label for proto3 basic types. This provides a smoother path to editions, which uses explicit presence by default.
```

Imo marking a field with the `optional` label shouldn't mean the field in optional in the request schema. It's purely for tracking field presence. Marking a field as optional via the `field_behavior` option seems more appropriate + recommended in google's AIP: https://google.aip.dev/203#optional.

johanbrandhorst:
[...]

PR #6416, opened 2026-02-27
fix: allow proto3 optional fields in path parameters

susanachl:

Fixes [#6352](https://github.com/grpc-ecosystem/grpc-gateway/issues/6352)

#### Motivation & Context

This PR removes the compiler restriction (originally introduced in [#1951](https://github.com/grpc-ecosystem/grpc-gateway/pull/1951)) that prevents `proto3 optional` fields from being mapped to HTTP path parameters.

**Addressing the Routing Ambiguity Concern:**
By removing this check, we decouple the Protobuf schema validation from the HTTP routing layer. If an optional field is mapped to a path parameter like `/owners/{owner_name}/pets/`, the structural requirement of the URL template remains intact. 

If a client omits the optional field, resulting in an empty path segment (`POST /v1/owners//pets/fido:feed`), the underlying gRPC-Gateway HTTP router will naturally fail to match the template and return a `404 Not Found`. This correctly shifts the responsibility of validating path completeness from the proto compiler to the HTTP router, which is the expected behavior for malformed requests.

#### Changes

- Removed `proto3_optional` check in `internal/descriptor/services.go`.
- Refactored `TestOptionalProto3URLPathMappingError` to `TestOptionalProto3URLPathMappingSuccess` in `internal/descriptor/services_test.go`.
- Added an `additional_binding` to the `Custom` RPC in `examples/internal/proto/examplepb/a_bit_of_everything.proto` using an optional path parameter to prove end-to-end generation success.

### DIFF
--- a/examples/internal/clients/abe/api/swagger.yaml
+++ b/examples/internal/clients/abe/api/swagger.yaml
@@ -82,6 +82,432 @@ paths:
           description: "An unexpected error response."
           schema:
             $ref: "#/definitions/googleRpcStatus"
+  /v1/example/a_bit_of_everything/custom/{optionalStringValue}:
+    get:
+      tags:
+      - "ABitOfEverything"
+      operationId: "ABitOfEverythingService_Custom2"
+      parameters:
+      - name: "optionalStringValue"
+        in: "path"
+        required: true
+        type: "string"
+        x-exportParamName: "OptionalStringValue"
+      - name: "singleNested.name"
+        in: "query"
+        description: "name is nested field."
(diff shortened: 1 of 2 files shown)


johanbrandhorst (approved):
LGTM, thank you! I can't think of a reason this would break any existing users, and since this is something multiple users want, lets try it and see how it goes :).

johanbrandhorst:
Looks like you need to run the regenerate command, see CONTRIBUTING.md for instructions. Thanks!

PR #5627, opened 2025-06-03
Add openapi v3

amirsalarsafaei:
#### References to other Issues or PRs

This feature was request in #441.

#### Brief description of what is fixed or changed
This is a minimal implementation (as I'm writing this description a draft) for openapiv3 protoc plugin. 

#### Other comments
I plan to add tests and complete other features. I just want to get this look at before I go all in.
Right now it generates OpenAPI spec version 3, using `kin-openapi` library.

PR #6431, opened 2026-03-05
feat: add protoc-gen-openapiv3 plugin for OpenAPI v3 spec generation

### LINKED ISSUE #441: Feature Request: OpenAPI 3 support

abhipranay:
## Description

This PR adds `protoc-gen-openapiv3`, a new protoc plugin that generates OpenAPI v3 specifications from Protocol Buffer definitions.

Fixes #441

## Background

I'm aware of the existing PR #5627 for OpenAPI v3 support. I chose to start from scratch rather than contribute to that PR because:
1. I had already solved this problem at work and had a working implementation
2. Starting fresh allowed me to follow the established patterns from `protoc-gen-openapiv2` more closely
3. It was easier to ensure feature completeness with a clean implementation

I'm happy to collaborate with the maintainers on merging the best aspects of both implementations if needed.

## Motivation

OpenAPI v3 has been the standard for several years and offers significant improvements over v2:
- Native `oneOf`/`anyOf`/`allOf` composition types
- Better support for nullable fields
- Improved request body handling
- More precise schema definitions

This implementation follows the patterns established by `protoc-gen-openapiv2` while taking advantage of v3-specific features.

### New Files

```
protoc-gen-openapiv3/
├── main.go                           # Plugin entry point
├── BUILD.bazel                       # Bazel build support
├── options/
│   ├── openapiv3.proto              # OpenAPI v3 message types (Schema, Operation, etc.)
│   ├── annotations.proto            # Proto extensions for customization
│   ├── openapiv3.pb.go              # Generated Go code
│   ├── annotations.pb.go            # Generated Go code
│   └── BUILD.bazel
└── internal/genopenapiv3/
    ├── generator.go                  # Main generation logic
    ├── types.go                      # OpenAPI v3 type definitions
    ├── annotations.go                # Annotation reading helpers
    ├── apply_annotations.go          # Annotation application logic
    ├── comments.go                   # Proto comment extraction
    ├── format.go                     # JSON/YAML output formatting
    ├── well_known_types.go           # Protobuf WKT mappings
    ├── *_test.go                     # Unit tests
    └── BUILD.bazel
```

### Key Features

1. **Full OpenAPI v3.0.3 Support**
   - Generates valid OpenAPI v3 specs from proto files with HTTP annotations
   - Supports JSON and YAML output formats

2. **Automatic `oneOf` Detection**
   - Proto `oneof` fields are automatically converted to OpenAPI `oneOf` schemas
   - Skips synthetic oneofs from proto3 optional fields

3. **Schema Composition via Annotations**
   - `allOf`, `anyOf`, `oneOf`, `not` support through proto options
   - Discriminator support for polymorphic schemas

4. **Proto3 Field Semantics**
   - Non-optional, non-repeated fields marked as required by default
   - Configurable via `use_proto3_field_semantics` flag

5. **Comment Processing**
   - Extracts proto comments as descriptions
   - Removes internal comments `(-- ... --)` per AIP-192
   - Removes `buf:lint:ignore` directives

6. **Annotation Support** (similar to v2)
   - File-level: `openapiv3_document` for info, servers, security
   - Method-level: `openapiv3_operation` for operation customization
   - Message-level: `openapiv3_schema` for schema customization
   - Field-level: `openapiv3_field` for field customization
   - Service-level: `openapiv3_tag` for tag customization
   - Enum-level: `openapiv3_enum` for enum customization

### Usage

```yaml
# buf.gen.yaml
plugins:
  - local: protoc-gen-openapiv3
    out: gen/openapi/v3
    opt:
      - output_format=yaml
    strategy: all
```

### Example Proto Annotation

```protobuf
import "protoc-gen-openapiv3/options/annotations.proto";

option (grpc.gateway.protoc_gen_openapiv3.options.openapiv3_document) = {
  info: {
    title: "My API"
    version: "1.0.0"
  }
  servers: [{url: "https://api.example.com"}]
};

message User {
  option (grpc.gateway.protoc_gen_openapiv3.options.openapiv3_schema) = {
    description: "A user in the system"
  };
  
  string id = 1 [(grpc.gateway.protoc_gen_openapiv3.options.openapiv3_field) = {
    description: "Unique identifier"
    example: "usr_123"
  }];
}
```

### CLI Flags

| Flag | Default | Description |
|------|---------|-------------|
| `output_format` | `json` | Output format: `json` or `yaml` |
| `json_names_for_fields` | `true` | Use JSON names for field names |
| `remove_internal_comments` | `true` | Remove `(-- ... --)` comments |
| `use_proto3_field_semantics` | `true` | Mark non-optional fields as required |
| `include_package_in_tags` | `false` | Include package in service tags |
| `disable_default_errors` | `false` | Disable default error responses |
| `enums_as_ints` | `false` | Render enums as integers |
| `simple_operation_ids` | `false` | Simpler operation ID generation |
| `generate_unbound_methods` | `false` | Generate for methods without HTTP rules |

## Testing

- Unit tests for all major components
- Tested against real-world proto files

```bash
go test ./protoc-gen-openapiv3/...
```

## Checklist

- [x] I have read the [contributing guidelines](https://github.com/grpc-ecosystem/grpc-gateway/blob/main/CONTRIBUTING.md)
- [x] I have added unit tests
- [x] I have added Bazel build support
- [x] I have updated the Makefile
- [x] Documentation is included in code comments

## Breaking Changes

None. This is a new plugin that does not affect existing functionality.

## Future Work

- Add more examples in the examples directory
- Integration with existing grpc-gateway examples
- Additional test coverage for edge cases


abhipranay:
@johanbrandhorst I did tested this locally with different files but I would appreciate if you can provide pointers which can help me test the changes better.

aidandj:
A few comments

* ~~the `allow_merge` option seems to not be working, it still just generates one file, although the file name option does seem to be working, it causes issues because it just renames the files for different services to the same thing.~~ buf local plugins don't run with `strategy: all` by default, this 
* `disable_default_errors` seems to not be working for me, still getting the default error, but `disable_default_responses` seems to correctly remove the error response, while not removing the default 200 response. Fixed in https://github.com/abhipranay/grpc-gateway/pull/2 as well
* ~~When i set an api version using the `grpc.gateway.protoc_gen_openapiv3.options.openapiv3_document` it doesn't seem to propagate to the generated spec~~ This only happens if i dont have any services in the proto file defining the document. I think this is a simple fix. https://github.com/abhipranay/grpc-gateway/pull/2
* `openapi_naming_strategy` seems to not affect anything, seems to always be `fqn`
* ~~`preserve_rpc_order` seems to not have any effect~~ This seems to be working fine, not sure why i thought it wasn't
* `proto3_optional_nullable` seems to not have any effect

Otherwise looking good. Happy to help contribute, is there anything on that list that youd want help with? For me the `allow_merge`, `disable_default_errors` and api version are blockers for us to use it, but the others seem like we should either remove the options, or document that they don't work.

aidandj:
Also, by adding `protoc-gen-openapiv3` to the `buf.gen.yaml` you could generate a lot of examples and validate them against an openapi spec validator.

It would be a good idea to update the examples with more v3 spec examples as well, i may be able to look at that as well.

https://github.com/aidandj/grpc-gateway/tree/aidan/v3-examples

Its looking good, I put the `a_bit_of_everything` in a swagger editor and only got one error:

```
6735 | 10071 | nullable has no special meaning, if not set on purpose use `type="null"` instead
-- | -- | --
```

aidandj:
Also, a new extension will likely need to be reserved: https://chromium.googlesource.com/chromium/src/+/master/third_party/protobuf/docs/options.md

abhipranay:
> Also, by adding `protoc-gen-openapiv3` to the `buf.gen.yaml` you could generate a lot of examples and validate them against an openapi spec validator.
> 
> It would be a good idea to update the examples with more v3 spec examples as well, i may be able to look at that as well.
> 
> https://github.com/aidandj/grpc-gateway/tree/aidan/v3-examples
> 
> Its looking good, I put the `a_bit_of_everything` in a swagger editor and only got one error:
> 
> ```
> 6735 | 10071 | nullable has no special meaning, if not set on purpose use `type="null"` instead
> -- | -- | --
> ```

@aidandj Thanks for suggestion. ~Let me take a look.~

I am not able to replicate. [Here](https://github.com/abhipranay/grpc-gateway/pull/3) are the generated openapi v3 json

abhipranay:
> A few comments
> 
> * ~the `allow_merge` option seems to not be working, it still just generates one file, although the file name option does seem to be working, it causes issues because it just renames the files for different services to the same thing.~ buf local plugins don't run with `strategy: all` by default, this
> * `disable_default_errors` seems to not be working for me, still getting the default error, but `disable_default_responses` seems to correctly remove the error response, while not removing the default 200 response. Fixed in [Copy leader logic over from v2 to properly merge spec abhipranay/grpc-gateway#2](https://github.com/abhipranay/grpc-gateway/pull/2) as well
> * ~When i set an api version using the `grpc.gateway.protoc_gen_openapiv3.options.openapiv3_document` it doesn't seem to propagate to the generated spec~ This only happens if i dont have any services in the proto file defining the document. I think this is a simple fix. [Copy leader logic over from v2 to properly merge spec abhipranay/grpc-gateway#2](https://github.com/abhipranay/grpc-gateway/pull/2)
> * `openapi_naming_strategy` seems to not affect anything, seems to always be `fqn`
> * ~`preserve_rpc_order` seems to not have any effect~ This seems to be working fine, not sure why i thought it wasn't
> * `proto3_optional_nullable` seems to not have any effect
> 
> Otherwise looking good. Happy to help contribute, is there anything on that list that youd want help with? For me the `allow_merge`, `disable_default_errors` and api version are blockers for us to use it, but the others seem like we should either remove the options, or document that they don't work.

Let me address below two:

> openapi_naming_strategy seems to not affect anything, seems to always be fqn
> proto3_optional_nullable seems to not have any effect

Seems like i need to mostly copy paste from v2.

protoc-gen-openapiv3/internal/genopenapiv3/generator.go
```diff
@@ -0,0 +1,1000 @@
...
+	// Oneof fields are mutually exclusive, not required individually
+	if field.OneofIndex != nil {
+		return false
+	}
+	// Repeated fields default to empty, so are effectively optional
+	if field.GetLabel() == descriptorpb.FieldDescriptorProto_LABEL_REPEATED {
+		return false
+	}
+	return true
+}
```
JackBogart: This is a common misconception w/ the `optional` label in protobuf. Fields using the `optional` label are NOT required. In fact, no fields are required as it caused so many issues in proto2 the label had to be removed. If a field uses implicit field presence tracking, it'll default to the type's default value when not supplied.

I highly recommend reading AIP-149: https://google.aip.dev/149
and subsequently AIP-203: https://google.aip.dev/203

Marking fields as required in openAPI descriptions should be dictated by the `google.api.field_behavior` option as explained in the linked AIP. The `optional` label should be strictly left for code generation that involves field presence tracking.

Otherwise I'm not too sure how you could mark a field as optional in editions as the `optional` label is gone, as all fields automatically track field presence.

abhipranay: TIL. 
Let me fix this.

aidandj: > Otherwise I'm not too sure how you could mark a field as optional in editions as the optional label is gone, as all fields automatically track field presence.
Editions has a `field_presence` setting

> In fact, no fields are required as it caused so many issues in proto2 the label had to be removed.
It was added back to proto3

While I agree with the above, the existing openapiv2 behavior should be supported in my opinion. There could be an option for only marking required based on the field_behavior option as an improvement, it could even be default.

aidandj:
> Also, a new extension will likely need to be reserved: https://chromium.googlesource.com/chromium/src/+/master/third_party/protobuf/docs/options.md


@johanbrandhorst would you know who the best person to request this would be? Seems like something should get reserved even if this isn't the PR that gets used to implement the v3 spec.

protoc-gen-openapiv3/internal/genopenapiv3/generator.go
aidandj: So is the current implementation as follows?
* Non-optional proto scalar field (editions implicit prescense) -> non-required openapi field
* Optional proto3 scalar field (editions explicit presence) -> nullable openapi field (if nullable option is set)
* Behavior optioned proto field -> required openapi field

johanbrandhorst:
> > Also, a new extension will likely need to be reserved: https://chromium.googlesource.com/chromium/src/+/master/third_party/protobuf/docs/options.md
> 
> @johanbrandhorst would you know who the best person to request this would be? Seems like something should get reserved even if this isn't the PR that gets used to implement the v3 spec.

I expect that would be me. I'm a bit busy right now and will not be able to give this the attention it deserves in the short term, but it's very exciting to see multiple members of the community coming together to test it, thank you everyone! I will email the registry to ask for an extension number.

I would also say that even if this doesn't support all the flags that openapiv2 does, we should still try to ship it, so if you find some flags don't work yet, just remove them and we can add them back later with proper testing.

Regarding the attempt to keep this implementation similar to openapiv2, that's okay, I would've preferred an independent implementation since openapiv2 is such a mess, but the more important part will be to get this into the hands of users, so beggars can't be choosers 😁.

aidandj:
> > > Also, a new extension will likely need to be reserved: https://chromium.googlesource.com/chromium/src/+/master/third_party/protobuf/docs/options.md
> > 
> > 
> > @johanbrandhorst would you know who the best person to request this would be? Seems like something should get reserved even if this isn't the PR that gets used to implement the v3 spec.
> 
> I expect that would be me. I'm a bit busy right now and will not be able to give this the attention it deserves in the short term, but it's very exciting to see multiple members of the community coming together to test it, thank you everyone! I will email the registry to ask for an extension number.
> 
> I would also say that even if this doesn't support all the flags that openapiv2 does, we should still try to ship it, so if you find some flags don't work yet, just remove them and we can add them back later with proper testing.
> 
> Regarding the attempt to keep this implementation similar to openapiv2, that's okay, I would've preferred an independent implementation since openapiv2 is such a mess, but the more important part will be to get this into the hands of users, so beggars can't be choosers 😁.

Thanks Johan!

If a group of the community can arrive to something that seems ready to at least get out to users would you be willing to release it in some form of alpha stage? When using tools like buf it is quite challenging to work with branches of plugins outside of testing. Integrating it with CI is complex.

JackBogart (commented):
gonna try to give ya a big review within the next week, little busy atm but more than happy to help with getting this through

johanbrandhorst:
Yes I'd be happy to put out a release with an early version of the plugin, if a few users here can confirm that it is already helping them. We can always continue to iterate on the functionality as we go :). As I said, I won't have time to review this soon, so I'm very grateful for your help. When you all think this is good for a review I will try to get to it in the next few weeks.

protoc-gen-openapiv3/options/openapiv3.proto
```diff
@@ -0,0 +1,640 @@
...
+  string summary = 2;
+  // A verbose explanation of the operation behavior. CommonMark syntax MAY be
+  // used for rich text representation.
+  string description = 3;
+  // Additional external documentation for this operation.
+  ExternalDocumentation external_docs = 4;
+  // Unique string used to identify the operation.
+  string operation_id = 5;
+  // field 6 is reserved for 'parameters'.
+  reserved 6;
```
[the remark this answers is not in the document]
aidandj: Came up with an implementation here: https://github.com/abhipranay/grpc-gateway/pull/11

It worked in the swagger editor and passed all the existing tests.

abhipranay:
## Design Discussion

I've opened #6475 to discuss the proto schema design for `openapiv3.proto`, specifically around adopting a `SchemaOrReference` pattern for better type safety and OpenAPI 3.1 compliance.

While reviewing the PR changes for supporting HEADER parameters, I realized that the current schema allows invalid states (e.g., duplicate headers). Similarly, the current proto schema design makes assumptions about how the implementation will work rather than enforcing correctness at the type level.

Would appreciate maintainer input on the design before this PR is merged, as changing the proto structure later would be a breaking change.

cc @johanbrandhorst

FYI @aidandj @JackBogart

abhipranay:
Moved to draft to implement improvements proposed [here](https://github.com/grpc-ecosystem/grpc-gateway/issues/6475)

aidandj:
Got some time to do a little bit of testing today.

Named header references seem weird. In my original implementation you would reference a header component like this:
```
    option (grpc.gateway.protoc_gen_openapiv3.options.openapiv3_operation) = {
      parameters: {
        headers: [
          {ref: "#/components/parameters/PreferHeader"}]
      }
    };
```
And the component has a name defined already
```
  components: {
    parameters: {
      key: "PreferHeader"
      value: {
        name: "Prefer"
        in: "header"
        description: "Validation strictness per RFC 7240. Use 'handling=strict' to reject unknown fields."
        schema: {
          type: "string"
          enum: [
            "handling=strict",
            "handling=lenient"
          ]
        }
      }
    }
  }
```
Now it is done like this:
```
    option (grpc.gateway.protoc_gen_openapiv3.options.openapiv3_operation) = {
      parameters: {
        headers: [
          {
            name: "Prefer"
            value: {
              reference: {ref: "#/components/parameters/PreferHeader"}
            }
          }
        ]
      }
    };
```
There is now a duplication of the name, and the name is actually ignored and overridden by the component.

Another note, the output yaml file is now out of order it seems, maybe this is ok, but having components before the info block seems odd, maybe i'm missing something from the spec though.

We should also revisit all places that Schema is used and decide if SchemaOrReference should be used. For example, in MediaType it should be SchemaOrReference, to support this pattern.

```
      responses: {
        key: "207"
        value: {
          description: "Mixed response. Some orders were accepted while others were rejected. See the SubmitOrdersResponse details for more information."
          content: {
            key: "application/json"
            value: {
              schema: {
                reference: {ref: "#/components/schemas/api.v1.ResponseMessage"}
              }
            }
          }
        }
      }
      responses: {
        key: "400"
        value: {
          description: "All orders were rejected, see the SubmitOrdersResponse details for more information."
          content: {
            key: "application/json"
            value: {
              schema: {
                reference: {ref: "#/components/schemas/api.v1.ResponseMessage"}
              }
            }
          }
        }
      }
```

I pushed that change here: https://github.com/abhipranay/grpc-gateway/pull/17

The other places I see that would make sense to check are:

- [ ] Parameter.schema. (Does it make sense to have a component with a schema that is a ref? I think it might?)
- [ ] Header.schema (I could see a use case in the Response message header specifier to have a ref instead of a schema)

Those types could either become `SchemaOrRef` or where they are used could take a ref option (like in components we have a `HeaderOrRef`, but in `Response` we don't)

aidandj:
I found a bug with yaml formatting, but forgot to push my fix before leaving for a long weekend.

The response keys are getting formatted as yaml numbers instead of strings. I fixed it by updating the Responses MarshalYaml function and adding a tag to the codes key of `!!str`

I started a yaml golden file test to match the json golden files test as well.

Can push it up on Monday

abhipranay:
> Got some time to do a little bit of testing today.
> 
> Named header references seem weird. In my original implementation you would reference a header component like this:
> 
> ```
>     option (grpc.gateway.protoc_gen_openapiv3.options.openapiv3_operation) = {
>       parameters: {
>         headers: [
>           {ref: "#/components/parameters/PreferHeader"}]
>       }
>     };
> ```
> 
> And the component has a name defined already
> 
> ```
>   components: {
>     parameters: {
>       key: "PreferHeader"
>       value: {
>         name: "Prefer"
>         in: "header"
>         description: "Validation strictness per RFC 7240. Use 'handling=strict' to reject unknown fields."
>         schema: {
>           type: "string"
>           enum: [
>             "handling=strict",
>             "handling=lenient"
>           ]
>         }
>       }
>     }
>   }
> ```
> 
> Now it is done like this:
> 
> ```
>     option (grpc.gateway.protoc_gen_openapiv3.options.openapiv3_operation) = {
>       parameters: {
>         headers: [
>           {
>             name: "Prefer"
>             value: {
>               reference: {ref: "#/components/parameters/PreferHeader"}
>             }
>           }
>         ]
>       }
>     };
> ```
> 
> There is now a duplication of the name, and the name is actually ignored and overridden by the component.
> 
> Another note, the output yaml file is now out of order it seems, maybe this is ok, but having components before the info block seems odd, maybe i'm missing something from the spec though.
> 
> We should also revisit all places that Schema is used and decide if SchemaOrReference should be used. For example, in MediaType it should be SchemaOrReference, to support this pattern.
> 
> ```
>       responses: {
>         key: "207"
>         value: {
>           description: "Mixed response. Some orders were accepted while others were rejected. See the SubmitOrdersResponse details for more information."
>           content: {
>             key: "application/json"
>             value: {
>               schema: {
>                 reference: {ref: "#/components/schemas/api.v1.ResponseMessage"}
>               }
>             }
>           }
>         }
>       }
>       responses: {
>         key: "400"
>         value: {
>           description: "All orders were rejected, see the SubmitOrdersResponse details for more information."
>           content: {
>             key: "application/json"
>             value: {
>               schema: {
>                 reference: {ref: "#/components/schemas/api.v1.ResponseMessage"}
>               }
>             }
>           }
>         }
>       }
> ```
> 
> I pushed that change here: [abhipranay#17](https://github.com/abhipranay/grpc-gateway/pull/17)
> 
> The other places I see that would make sense to check are:
> 
> * [ ]  Parameter.schema. (Does it make sense to have a component with a schema that is a ref? I think it might?)
> * [ ]  Header.schema (I could see a use case in the Response message header specifier to have a ref instead of a schema)
> 
> Those types could either become `SchemaOrRef` or where they are used could take a ref option (like in components we have a `HeaderOrRef`, but in `Response` we don't)

@aidandj I implemented the changes mentioned as those make sense to me.

abhipranay:
> I found a bug with yaml formatting, but forgot to push my fix before leaving for a long weekend.
> 
> The response keys are getting formatted as yaml numbers instead of strings. I fixed it by updating the Responses MarshalYaml function and adding a tag to the codes key of `!!str`
> 
> I started a yaml golden file test to match the json golden files test as well.
> 
> Can push it up on Monday

@aidandj adding `!!str` doesn't seem clean approach. The implementation already have ability to control formatting the yaml node.

See if we can just use `yaml.DoubleQuotedStyle` for nodes related to status codes in response.

protoc-gen-openapiv3/internal/genopenapiv3/model/canonical.go
```diff
@@ -0,0 +1,838 @@
...
+
+import (
+	"bytes"
+	"encoding/json"
+	"sort"
+
+	"go.yaml.in/yaml/v3"
+)
+
+// Canonical Model - Internal representation of OpenAPI documents.
```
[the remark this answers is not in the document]
aidandj: I like the design. I'll give it another review with that in mind tomorrow. Otherwise I think we're close to getting something ready for review.

I'll try and do another pass on the protos as well, those are important to get it right first try.

aidandj:
Alright here is an interesting one.
Nested timestamp, it gets put in the spec like this:
```
            properties:
                nestedTimestamp:
                    type: string
                    format: date-time
```
But there is an unused schema for google.protobuf.Timestamp. If you just have a top level timestamp, this isn't  a problem.

proto:
```
// Ping service definition.
syntax = "proto3";

package api.ping.v1;

import "google/api/visibility.proto";
import "google/protobuf/timestamp.proto";

message Nested {
  .google.protobuf.Timestamp nested_timestamp = 1;
}

// Not internal comment
//
// (-- This is a abstraction on a task_scheduler.proto Order --)
message PingRequest {
  // This is a field comment.
  // (-- This is a abstraction on a task_scheduler.proto Order --)
  optional string field_hello = 1 [json_name = "fieldH_llo"];
  google.protobuf.Timestamp timestamp = 2;
  Nested nested = 3;
}

message PingResponse {
  optional string field_world = 1 [json_name = "fieldW_rld"];
}

service PingService {
  option (google.api.api_visibility).restriction = "PUBLIC";
  rpc Ping(PingRequest) returns (PingResponse) {
    option (google.api.http) = {get: "/api/v1/ping"};
    option deprecated = true;
  }
  rpc APing(PingRequest) returns (stream PingResponse) {
    option (google.api.http) = {get: "/api/v1/a_ping"};
  }
}

```

openapi
```
openapi: 3.1.0
info:
    title: api/ping/v1/ping.proto
    version: version not set
paths:
    /api/v1/a_ping:
        get:
            tags:
                - PingService
            operationId: PingService_APing
            parameters:
                - name: fieldH_llo
                  in: query
                  schema:
                    type: string
                - name: timestamp
                  in: query
                  required: true
                  schema:
                    type: string
                    format: date-time
                - name: nested
                  in: query
                  required: true
                  schema:
                    $ref: '#/components/schemas/api.ping.v1.Nested'
            responses:
                "200":
                    description: A successful response
                    content:
                        application/json:
                            schema:
                                $ref: '#/components/schemas/api.ping.v1.PingResponse'
    /api/v1/ping:
        get:
            tags:
                - PingService
            operationId: PingService_Ping
            parameters:
                - name: fieldH_llo
                  in: query
                  schema:
                    type: string
                - name: timestamp
                  in: query
                  required: true
                  schema:
                    type: string
                    format: date-time
                - name: nested
                  in: query
                  required: true
                  schema:
                    $ref: '#/components/schemas/api.ping.v1.Nested'
            responses:
                "200":
                    description: A successful response
                    content:
                        application/json:
                            schema:
                                $ref: '#/components/schemas/api.ping.v1.PingResponse'
            deprecated: true
components:
    schemas:
        api.ping.v1.Nested:
            type: object
            required:
                - nestedTimestamp
            properties:
                nestedTimestamp:
                    type: string
                    format: date-time
        api.ping.v1.PingResponse:
            type: object
            properties:
                fieldW_rld:
                    type: string
        google.protobuf.Timestamp:
            type: string
            format: date-time
tags:
    - name: PingService

```

aidandj:
I found one other error that is going to take me a while to isolate (found it running our company internal protos through the generator).

In the swagger editor it is tagging certain refs with

```
Reference Error - Ul: Invalid object key "package.Message" at position 2 in "/components/schemas/package.Message": key not found in object
```

I doubt this is a blocker for releasing an alpha version though, and I may not get to isolate it until next week.

abhipranay:
> I found one other error that is going to take me a while to isolate (found it running our company internal protos through the generator).
> 
> In the swagger editor it is tagging certain refs with
> 
> ```
> Reference Error - Ul: Invalid object key "package.Message" at position 2 in "/components/schemas/package.Message": key not found in object
> ```
> 
> I doubt this is a blocker for releasing an alpha version though, and I may not get to isolate it until next week.

This one seems like issue because some messages definitions are not present. This can be a config issue at source. Eg:

message A {
 SomeCustomMessageType B = 1; // SomeCustomMessageType definition is not available in same file but is in imports or we forgot to import or file cannot be 
}

abhipranay:
> Alright here is an interesting one. Nested timestamp, it gets put in the spec like this:
> 
> ```
>             properties:
>                 nestedTimestamp:
>                     type: string
>                     format: date-time
> ```
> 
> But there is an unused schema for google.protobuf.Timestamp. If you just have a top level timestamp, this isn't a problem.
> 
> proto:
> 
> ```
> // Ping service definition.
> syntax = "proto3";
> 
> package api.ping.v1;
> 
> import "google/api/annotations.proto";
> import "google/api/visibility.proto";
> import "google/protobuf/timestamp.proto";
> 
> message Nested {
>   .google.protobuf.Timestamp nested_timestamp = 1;
> }
> 
> // Not internal comment
> //
> // (-- This is a abstraction on a task_scheduler.proto Order --)
> message PingRequest {
>   // This is a field comment.
>   // (-- This is a abstraction on a task_scheduler.proto Order --)
>   optional string field_hello = 1 [json_name = "fieldH_llo"];
>   google.protobuf.Timestamp timestamp = 2;
>   Nested nested = 3;
> }
> 
> message PingResponse {
>   optional string field_world = 1 [json_name = "fieldW_rld"];
> }
> 
> service PingService {
>   option (google.api.api_visibility).restriction = "PUBLIC";
>   rpc Ping(PingRequest) returns (PingResponse) {
>     option (google.api.http) = {get: "/api/v1/ping"};
>     option deprecated = true;
>   }
>   rpc APing(PingRequest) returns (stream PingResponse) {
>     option (google.api.http) = {get: "/api/v1/a_ping"};
>   }
> }
> ```
> 
> openapi
> 
> ```
> openapi: 3.1.0
> info:
>     title: api/ping/v1/ping.proto
>     version: version not set
> paths:
>     /api/v1/a_ping:
>         get:
>             tags:
>                 - PingService
>             operationId: PingService_APing
>             parameters:
>                 - name: fieldH_llo
>                   in: query
>                   schema:
>                     type: string
>                 - name: timestamp
>                   in: query
>                   required: true
>                   schema:
>                     type: string
>                     format: date-time
>                 - name: nested
>                   in: query
>                   required: true
>                   schema:
>                     $ref: '#/components/schemas/api.ping.v1.Nested'
>             responses:
>                 "200":
>                     description: A successful response
>                     content:
>                         application/json:
>                             schema:
>                                 $ref: '#/components/schemas/api.ping.v1.PingResponse'
>     /api/v1/ping:
>         get:
>             tags:
>                 - PingService
>             operationId: PingService_Ping
>             parameters:
>                 - name: fieldH_llo
>                   in: query
>                   schema:
>                     type: string
>                 - name: timestamp
>                   in: query
>                   required: true
>                   schema:
>                     type: string
>                     format: date-time
>                 - name: nested
>                   in: query
>                   required: true
>                   schema:
>                     $ref: '#/components/schemas/api.ping.v1.Nested'
>             responses:
>                 "200":
>                     description: A successful response
>                     content:
>                         application/json:
>                             schema:
>                                 $ref: '#/components/schemas/api.ping.v1.PingResponse'
>             deprecated: true
> components:
>     schemas:
>         api.ping.v1.Nested:
>             type: object
>             required:
>                 - nestedTimestamp
>             properties:
>                 nestedTimestamp:
>                     type: string
>                     format: date-time
>         api.ping.v1.PingResponse:
>             type: object
>             properties:
>                 fieldW_rld:
>                     type: string
>         google.protobuf.Timestamp:
>             type: string
>             format: date-time
> tags:
>     - name: PingService
> ```

This should be fixed now.

abhipranay:
@johanbrandhorst The PR is ready for alpha release.

johanbrandhorst (changes requested):
Thanks again to the community for testing this out! I will be trying to do a code review over the next week or so and making suggestions for any changes. I've started the CI job and it seems like we may have some minor bazel file formatting errors - please fix those. You can rerun the bazel job locally by following the steps in `CONTRIBUTING.md`. It also looks like `make generate` timed out, is there a good reason for that or is that just a CI failure?

Another thing we'll need for an alpha release are docs! Please add some new docs to the `docs/` folder, we can just have something basic like a simple example for how to run it, but we want to make sure people can discover this new tool. Also please update the README.md. Thanks again for all your work!

johanbrandhorst:
I will say, it still pains me to see this so big. My hope for a first version of `protoc-gen-openapiv3` was something extremely simple, implementing only the most basic features and built on a new foundation that doesn't carry all the baggage of `protoc-gen-openapiv2`, that could be extended over time. I'm worried that this is not the right direction right now. While I appreciate your work on this so far, I want to be clear that there is no promise that this gets merged in this state. I need to keep the interests of the project, maintainers and all of its users in mind. I don't know how much you've used AI in your work on this PR, but if you want to experiment with a from-scratch implementation based on the learnings of the testing of this PR, I wouldn't necessarily be opposed to that. My fundamental requirement is simplicity, both in the code and in how the user uses it.

aidandj:
> I will say, it still pains me to see this so big. My hope for a first version of `protoc-gen-openapiv3` was something extremely simple, implementing only the most basic features and built on a new foundation that doesn't carry all the baggage of `protoc-gen-openapiv2`, that could be extended over time. I'm worried that this is not the right direction right now. While I appreciate your work on this so far, I want to be clear that there is no promise that this gets merged in this state. I need to keep the interests of the project, maintainers and all of its users in mind. I don't know how much you've used AI in your work on this PR, but if you want to experiment with a from-scratch implementation based on the learnings of the testing of this PR, I wouldn't necessarily be opposed to that. My fundamental requirement is simplicity, both in the code and in how the user uses it.

openapiv3 itself is very complex, and what abhipranay has done I think isn't held back by the baggage of v2. The openapi 3.0 json schema is over 1600 lines long, and the spec documentation is over 70 pages long. I'm not sure what else could be dropped, maybe some of the options, but a lot of those are valid options with different uses.

What part in particular feels overly large?

johanbrandhorst:
The desire to support different versions of OpenAPIv3 (3.0 vs 3.1) adds a huge amount of duplication and maintenance overhead. Maybe it would be better if we just started with support for 3.1? It looks like the PR doesn't actually support v3.0 (`--openapi_version` supports v3.1 only). I do value the work here and I think we can use it, but it would be nice to compact it to the absolute minimum viable product.

johanbrandhorst:
Again, I'm not sure how much AI was used here and I don't want to disparage the work, but I might suggest asking an agent to take this PR and narrow it down to the absolute MVP features we need and some basic options (keeping as much of the valuable test and user feedback we have here, of course).

aidandj:
3.2.0 is on the horizon, and offers first class support for server sent events. Which when combined with grpc gateway makes for a pretty powerful streaming declaration. I could see dropping support for 3.0, but being able to support multiple versions is important.

JackBogart:
> 3.2.0 is on the horizon, and offers first class support for server sent events. Which when combined with grpc gateway makes for a pretty powerful streaming declaration. I could see dropping support for 3.0, but being able to support multiple versions is important.

it would probably be easier to get support for one version as a proof of concept, then work on the others. this PR is already pretty big, to the point where its hard to review and I dont envy repo owners trying to review.

although it would be funky to start with 3.1 instead of 3.0

abhipranay:
@johanbrandhorst Appreciate the honest feedback.

On the size, a big chunk is tests (~12K lines) and generated `.pb.go` files (~12K lines). The actual implementation to review is ~6,500 lines. But I hear you, it's still a lot.

What I learned from community testing: @aidandj, @JackBogart and others found real issues: optional field semantics, oneOf edge cases, header parameters. That feedback is baked in now. I'd not want to lose it.

Here's what I can do:

1. 3.1 only - Drop the version abstraction layer entirely. Why 3.1 first:
    - The current implementation already targets 3.1 (--openapi_version only supports 3.1)
    - 3.1 aligns with JSON Schema draft 2020-12 - better tooling compatibility
    - 3.2 is on the horizon with SSE support, so 3.1 is a better stepping stone
    - Users who need 3.0 can continue using protoc-gen-openapiv2 - no gap in coverage
  2. Strip most flags - Keep only output_format and json_names_for_fields. Add others later based on demand.
  3. Minimal tests - Just enough to prevent regressions. Expand coverage in follow-ups.

This might get the PR down to ~16K lines total, with only ~7K lines of actual code to review (rest is generated). 

I am not sure whether it will be possible to go much lower as v3 is still a decent amount of work.

I can try to split this into 2-3 smaller PRs if that's easier to digest.

Let me know if that sounds like a plan ?

aidandj:
> 2. Strip most flags - Keep only output_format and json_names_for_fields. Add others later based on demand.

These are the flags I currently use:
```
      - use_proto3_field_semantics=true
      - output_format=yaml
      - enable_rpc_deprecation=true
      - preserve_rpc_order=true
      - visibility_restriction_selectors=PUBLIC
      - allow_merge=true,merge_file_name=openapi
      - disable_default_errors=true
      - remove_internal_comments=true
      - openapi_naming_strategy=fqn
 ```

I think all of those are vital options. Maybe we could default the naming strategy to something reasonable, but I don't see a point in dropping any of the others.

Count this as demand for the other flags 😄

**Comment by johanbrandhorst** [comment]
As I've said before, I really appreciate the work here, and I definitely want us to get something into the hands of users, because openapiv3 is something people have asked for for almost a decade at this point. However, I'll pose it to you (readers of this thread and you in particular @abhipranay): what do you think about starting with something like this: https://github.com/grpc-ecosystem/grpc-gateway/pull/6623? It's just the basic "translate http.proto annotations to openapiv3", no flags, no nothing, not even any custom annotations. Obviously we will layer all of that on top as and when users demand it. It sounds like we already have a few users here with practical need for a few of these flags, but lets add them one by one. The annotations too are reasonably easy to add on top of the initial code. I've added you as a co-author, since you've made significant contributions to the effort.

File: protoc-gen-openapiv3/options/annotations.proto [hunk 4]
```diff
@@ -0,0 +1,138 @@
...
+  //      description: "A great API";
+  //    };
+  //    servers: [
+  //      { url: "https://api.example.com"; }
+  //    ];
+  //  };
+  //
+  // Registered extension number 1285 in protobuf global extension registry.
+  // See: https://github.com/protocolbuffers/protobuf/blob/main/docs/options.md
+  OpenAPI openapiv3_document = 1285;
```
johanbrandhorst, line 29: FYI we've [officially registered](https://github.com/protocolbuffers/protobuf/commit/5e0f68530ae6ad47b78714e60861321e1d1c04ae) 1295 as the openapiv3 extension

abhipranay:
> As I've said before, I really appreciate the work here, and I definitely want us to get something into the hands of users, because openapiv3 is something people have asked for for almost a decade at this point. However, I'll pose it to you (readers of this thread and you in particular @abhipranay): what do you think about starting with something like this: #6623? It's just the basic "translate http.proto annotations to openapiv3", no flags, no nothing, not even any custom annotations. Obviously we will layer all of that on top as and when users demand it. It sounds like we already have a few users here with practical need for a few of these flags, but lets add them one by one. The annotations too are reasonably easy to add on top of the initial code. I've added you as a co-author, since you've made significant contributions to the effort.

Sounds good to me. Happy to contribute.

abhipranay:
I will close this as we aligned to continue with https://github.com/grpc-ecosystem/grpc-gateway/pull/6623

Will contribute by working on issues created under `openapiv3`

Thanks folks.
Especially @aidandj @JackBogart and @johanbrandhorst .

PR #6613, opened 2026-04-14
Deprecate fields and methods if file is deprecated

aidandj:
Being able to deprecate an entire file and mark methods and fields as deprecated is nice when deprecating an entire API. I started this MR by passing a file descriptor around to the methods in template, so I could check for file level deprecation, but because of how the files are merged, individual file options are lost before templating. So instead if a file is deprecated, I propagate that option to the fields and methods, which is technically less correct, but simplifies the logic a lot.

The prior art for this is the golang code generator deprecates all fields, methods, services, etc if a file is deprecated.

### DIFF
--- a/protoc-gen-openapiv2/internal/genopenapi/generator.go
+++ b/protoc-gen-openapiv2/internal/genopenapi/generator.go
@@ -369,8 +369,33 @@ func encodeOpenAPI(file *wrapper, format Format) (*descriptor.ResponseFile, erro
 	}, nil
 }
 
+func deprecateFieldsAndMethods(file *descriptor.File) {
+	if opts := file.GetOptions(); opts != nil && opts.GetDeprecated() {
+		for _, msg := range file.Messages {
+			for _, field := range msg.GetField() {
+				if field.Options == nil {
+					field.Options = &descriptorpb.FieldOptions{}
+				}
+				field.Options.Deprecated = proto.Bool(true)
+			}
+		}
+		for _, svc := range file.Services {
+			for _, method := range svc.GetMethod() {
+				if method.Options == nil {
+					method.Options = &descriptorpb.MethodOptions{}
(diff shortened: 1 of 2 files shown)


protoc-gen-openapiv2/internal/genopenapi/generator.go
```diff
@@ -369,8 +369,33 @@ func encodeOpenAPI(file *wrapper, format Format) (*descriptor.ResponseFile, erro
 	}, nil
 }
 
+func deprecateFieldsAndMethods(file *descriptor.File) {
+	if opts := file.GetOptions(); opts != nil && opts.GetDeprecated() {
```
johanbrandhorst: Could we lift this check outside of this function so that it's clearer in the caller what's happening?

protoc-gen-openapiv2/internal/genopenapi/generator_test.go
```diff
@@ -2661,3 +2667,317 @@ func TestGenerateCrossPackageMessageReference(t *testing.T) {
...
+		}
+
+		resp := requireGenerate(t, &req, genopenapi.FormatJSON, false, false, true)
+		if len(resp) != 1 {
+			t.Fatalf("expected 1 response file, got %d", len(resp))
+		}
+
+		content := resp[0].GetContent()
+
+		var doc map[string]interface{}
```
johanbrandhorst: Please use `map[string]any`

johanbrandhorst (changes requested):
Some minor comments, thanks for this!

aidandj:
> Some minor comments, thanks for this!

Good feedback, done!

johanbrandhorst (approved):
LGTM!

johanbrandhorst:
Thanks for your contribution!

PR #6623, opened 2026-04-15
openapiv3: add minimal protoc-gen-openapiv3 generator

### LINKED ISSUE #441: Feature Request: OpenAPI 3 support

johanbrandhorst:
Adds a minimal protoc-gen-openapiv3 generator that can generate OpenAPI v3.1 specifications from protobuf service definitions. No annotations or options are supported yet.

Builds on the work of @abhipranay in https://github.com/grpc-ecosystem/grpc-gateway/pull/6431.

Fixes #441


johanbrandhorst:
CC @abhipranay @JackBogart @aidandj @commiterate. I think even if you try this and find it isn't suitable for you right now that I'd rather merge this and create issues for each outstanding request to work on as a follow-up. What do you think?

aidandj:
I guess I don't totally see the difference between merging this, immediately following up with a bunch of changes to support additional features. When they are already supported in another PR. I get that the initial review may be simpler, but when a working set of options is already implemented it seems a bit redundant.

What about stripping the adapter pattern in the other PR, so it's just 3.1. I think that won't be that challenging to review, while still supporting all the options.

johanbrandhorst:
Thanks for your comment, to me there is a pretty big difference, since this code base is small enough that I (and hopefully others) can review it and agree on it as a basis for the new plugin. I think we should carefully consider every new flag we want to add. What I am looking for from practitioners such as yourself right now is:

1. Does this generator cause any issues with your current set of files?
2. Do the new defaults make sense? E.g. we might add flags for `omit_default_error`, etc, but I'm hoping the defaults chosen here (which are different from the openapiv2 defaults) make sense in isolation.

I have to take a long view on the maintenance of this new tool and I am far more comfortable with this smaller step-by-step attempt at adding this new logic. If the end result is the same over the course of a few weeks of adding features that users are asking for, I think that gives us the best of both worlds.

aidandj:
Oneof's don't seem to be working.

Given: 
```
message PingResponse {
  optional string field_world = 1 [json_name = "fieldW_rld"];

  message Inner1 {
    optional string field_inner1 = 1;
  }

  message Inner2 {
    optional string field_inner2 = 1;
  }

  message Inner3 {
    optional string field_inner3 = 1;
  }

  oneof inner_message {
    Inner1 inner1 = 4;
    Inner2 inner2 = 5;
    Inner3 inner3 = 6;
  }
}
```

I get:
image

With the other branch:
image

johanbrandhorst:
Thanks for trying it! I can't reproduce the `any` nested message rendering - could you share a more complete example? I'm seeing:
```json
{
  "example.v1.PingResponse": {
    "type": "object",
    "properties": {
      "fieldW_rld": {
        "type": "string"
      },
      "inner1": {
        "$ref": "#/components/schemas/example.v1.PingResponse.Inner1"
      },
      "inner2": {
        "$ref": "#/components/schemas/example.v1.PingResponse.Inner2"
      },
      "inner3": {
        "$ref": "#/components/schemas/example.v1.PingResponse.Inner3"
      }
    },
    "oneOf": [
      {
        "not": {
          "anyOf": [
            {
              "required": [
                "inner1"
              ]
            },
            {
              "required": [
                "inner2"
              ]
            },
            {
              "required": [
                "inner3"
              ]
            }
          ]
        }
      },
      {
        "required": [
          "inner1"
        ]
      },
      {
        "required": [
          "inner2"
        ]
      },
      {
        "required": [
          "inner3"
        ]
      }
    ]
  },
  "example.v1.PingResponse.Inner1": {
    "type": "object",
    "properties": {
      "fieldInner1": {
        "type": "string"
      }
    }
  },
  "example.v1.PingResponse.Inner2": {
    "type": "object",
    "properties": {
      "fieldInner2": {
        "type": "string"
      }
    }
  },
  "example.v1.PingResponse.Inner3": {
    "type": "object",
    "properties": {
      "fieldInner3": {
        "type": "string"
      }
    }
  }
}
```
With regards to the `not: anyOf` wrapping, that's what I believe to be a more correct way of handling the case where none of the `oneof` messages is set. It seems like the other generator inserts a `none` message in this case, but that's not logically correct AFAICT.

aidandj:
If you load the `a_bit_of_everything.openapi.json` into a swagger editor you can see it
image

johanbrandhorst:
> If you load the `a_bit_of_everything.openapi.json` into a swagger editor you can see it image

Thanks, I see it now too! I _think_ this is a bug in the SwaggerEditor rendering, because the schema of the `oneOf` types is defined in the `properties`, but I've updated the generator to include the schema in each `oneOf` definition as well, and it's now rendering correctly for me.

image

abhipranay:
> CC @abhipranay @JackBogart @aidandj @commiterate. I think even if you try this and find it isn't suitable for you right now that I'd rather merge this and create issues for each outstanding request to work on as a follow-up. What do you think?

Yea sounds like a good plan to me. As long as we are motivated to make this a reality ( even in mvp form ) it's a win.
I am damn sure people needed this and switching to other solutions like connectrpc etc is not that easy for every team. 🧑‍💻

protoc-gen-openapiv3/internal/genopenapi/schema.go
```diff
@@ -0,0 +1,443 @@
...
+		schema.Properties = make(map[string]*SchemaOrRef, total)
+	}
+	for _, field := range regular {
+		b.addProperty(schema, field)
+	}
+	for _, g := range oneofs {
+		for _, field := range g.fields {
+			b.addProperty(schema, field)
+		}
+	}
```
abhipranay, line 201: @johanbrandhorst This seems like a bug. Only regular fields should be added to top level properties. Adding oneOf fields at top level will change the semantic meaning.

```proto
syntax = "proto3";

  package example.v1;

  import "google/type/date.proto";

  message Expiration {
    oneof expiration {
      int32 countdown_days = 1;
      google.type.Date deadline = 2;
    }
  }
```

Correct output should be:
```json
  {
    "example.v1.Expiration": {
      "type": "object",
      "oneOf": [
        {
          "type": "object",
          "not": { "anyOf": [{ "required": ["countdownDays"] }, { "required": ["deadline"] }] }
        },
        {
          "type": "object",
          "properties": { "countdownDays": { "type": "integer", "format": "int32" } },
          "required": ["countdownDays"]
        },
        {
          "type": "object",
          "properties": { "deadline": { "$ref": "#/components/schemas/google.type.Date" } },
          "required": ["deadline"]
        }
      ]
    }
  }

```
Below should fix it.

```suggestion
	total := len(regular)
	if len(regular) > 0 {
		schema.Properties = make(map[string]*SchemaOrRef, total)
	}
	for _, field := range regular {
		b.addProperty(schema, field)
	}
```

aidandj:
Should we create an issue (issues?) for the next round of features?

High priority for me include
- merging files (new option)
- enable deprecation (imo default true)
- omit internal comments (imo default true)
- custom annotations 
- disable default errors (option?)
- visibility restrictors (needs to be an option)

Yaml output is a nice to have

johanbrandhorst:
> Should we create an issue (issues?) for the next round of features?
> 
> High priority for me include
> 
>     * merging files (new option)
> 
>     * enable deprecation (imo default true)
> 
>     * omit internal comments (imo default true)
> 
>     * custom annotations
> 
>     * disable default errors (option?)
> 
>     * visibility restrictors (needs to be an option)
> 
> 
> Yaml output is a nice to have

Lets try and decide on reasonable defaults here, and where we want options, yes, lets create an issue for each option. The current implementation has deprecation and internal comments (per AIP-192) handled already. The other 4 I think we could do 4 separate issues for. I'd like each issue to have an opportunity for users to discuss the need and ideally express their own support for it. I'd like at least 2 users supporting an option before we add it, though when we're this early in development I expect we can just add back some obvious ones, like the ones you suggested, without needing 2 individual supporters. If you'd like to get started on those issues, please do!

aidandj:
> > Should we create an issue (issues?) for the next round of features?
> > 
> > High priority for me include
> > 
> >     * merging files (new option)
> > 
> >     * enable deprecation (imo default true)
> > 
> >     * omit internal comments (imo default true)
> > 
> >     * custom annotations
> > 
> >     * disable default errors (option?)
> > 
> >     * visibility restrictors (needs to be an option)
> > 
> > 
> > Yaml output is a nice to have
> 
> Lets try and decide on reasonable defaults here, and where we want options, yes, lets create an issue for each option. The current implementation has deprecation and internal comments (per AIP-192) handled already. The other 4 I think we could do 4 separate issues for. I'd like each issue to have an opportunity for users to discuss the need and ideally express their own support for it. I'd like at least 2 users supporting an option before we add it, though when we're this early in development I expect we can just add back some obvious ones, like the ones you suggested, without needing 2 individual supporters. If you'd like to get started on those issues, please do!

Will do.

I'm on mobile and can't search through the PR. Is file level deprecation supported (like the change we just added to the v2 generator)

johanbrandhorst:
> I'm on mobile and can't search through the PR. Is file level deprecation supported (like the change we just added to the v2 generator)

I'll make sure to get it in if it isn't, no reason it shouldn'd be.

protoc-gen-openapiv3/internal/genopenapi/schema.go
johanbrandhorst: I had a look at the spec, specifically the [Schema Object](https://spec.openapis.org/oas/v3.1.0.html#schema-object), which references the [JSON Schema Draft 2020-12](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-01). Sections [§7.7](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-01#name-annotations) and [§10.1](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-01#name-keyword-independence) establishes that adjacent keywords are independently evaluated and their results combined, which means that having both `properties` and `oneOf` at the same level simply means the instance must satisfy both. The JSON Schema example page [contains an example (device type)](https://json-schema.org/learn/json-schema-examples#device-type) with the `oneOf` schema defined in both the `properties` and the `oneOf` sections.

Is there a particular semantic difference you are aware of that this will trigger? To me it sounds like we can safely define the field schema in both places.

aidandj:
Issues created

aidandj:
Should we mark the v3 generator as alpha. And indicate that output may change as we work through options?

I had some contributions to the original PR, if I could also be marked as a co-author.

johanbrandhorst:
Yes, it would make sense to have an alpha phase for early adopters. I will add you to the contributors on the commit.

aidandj:
OpenAPI 3.1 allows deprecating parameters. When I deprecate a field in a request message I'm not seeing the parameter get deprecated.

Small change: https://github.com/grpc-ecosystem/grpc-gateway/pull/6646

protoc-gen-openapiv3/internal/genopenapi/operation.go
```diff
@@ -0,0 +1,414 @@
...
+			// Should not happen: the path template referenced a field that
+			// the binding did not record. Emit an open string parameter so
+			// the operation is at least syntactically valid.
+			params = append(params, &ParameterRef{Value: &Parameter{
+				Name:     pp.openAPIName,
+				In:       "path",
+				Required: true,
+				Schema:   &SchemaOrRef{Value: &Schema{Type: SchemaType{"string"}}},
+			}})
+			log.Printf("unexpected path parameter not found in binding: %q", pp.fieldName)
```
aidandj, line 74: can log.Printf be used here? should it be grpclog?

protoc-gen-openapiv3/internal/genopenapi/operation.go
```diff
@@ -0,0 +1,414 @@
...
+
+	if field.GetType() == descriptorpb.FieldDescriptorProto_TYPE_MESSAGE &&
+		wellKnownTypeSchema(field.GetTypeName()) == nil {
+		// Non-WKT message: distinguish maps, repeated messages, and regular
+		// nested messages. An unresolved message falls through to the
+		// single-parameter path so the operation stays syntactically valid.
+		msg, err := b.reg.LookupMsg("", field.GetTypeName())
+		switch {
+		case err != nil:
+			log.Printf("protoc-gen-openapiv3: cannot resolve message %q for query parameter %q: %v; emitting opaque parameter",
```
aidandj, line 136: Here, and other places in this file

JackBogart:
> Should we create an issue (issues?) for the next round of features?
> 
> High priority for me include
> 
>     * merging files (new option)
> 
>     * enable deprecation (imo default true)
> 
>     * omit internal comments (imo default true)
> 
>     * custom annotations
> 
>     * disable default errors (option?)
> 
>     * visibility restrictors (needs to be an option)
> 
> 
> Yaml output is a nice to have

Just wanna say that I'd definitely like most of these options as well. I actively use visibility restrictors, omitting internal comments, and deprecation. I can also easily see the use case of the others.

Not saying that the first iteration of the generator needs these, but wanted to show there's a general desire for them eventually.

protoc-gen-openapiv3/internal/genopenapi/operation.go
johanbrandhorst: Yeah, fair question, I kinda like the output of log.Printf better, but I see we're using grpclog already in `descriptor` package, so we should probably just use grpclog. I'll experiment a bit.

protoc-gen-openapiv3/internal/genopenapi/schema.go
abhipranay: Make sense.

abhipranay: I re-read the references in detail. You are right. 

`adjacent keywords are independently evaluated and their results combined, which means that having both properties and oneOf at the same level simply means the instance must satisfy both.` is true.

johanbrandhorst:
Thanks everyone for your contribution to this PR 🎉. Now we can start adding back some of the other features everyone wants 😄.

abhipranay:
Thanks @johanbrandhorst for proactiveness on this long pending feature.
Now we have a starting point. 🥳

PR #6681, opened 2026-04-22
protoc-gen-openapiv3: add annotations

### LINKED ISSUE #6674: Spec override

johanbrandhorst:
Fixes #6674


johanbrandhorst:
CC @aidandj

aidandj:
This one may take me a bit to review. How much of it comes from the old PR? We worked through a number of possibilities there.

protoc-gen-openapiv3/options/openapiv3.proto
```diff
@@ -0,0 +1,169 @@
+syntax = "proto3";
+
+// Package grpc.gateway.protoc_gen_openapiv3.options defines the minimal set of
+// OpenAPI 3.1.0 override messages consumed by protoc-gen-openapiv3.
```
aidandj, line 4: Should we be differentiating 3.1 and 3.x anywhere?

protoc-gen-openapiv3/options/annotations.proto
```diff
@@ -0,0 +1,40 @@
...
+  // Ref: https://github.com/protocolbuffers/protobuf/commit/5e0f68530ae6ad47b78714e60861321e1d1c04ae
+  Document openapiv3_document = 1295;
+}
+
+extend google.protobuf.MethodOptions {
+  // ID 1295 assigned by protobuf-global-extension-registry@google.com for the
+  // gRPC-Gateway OpenAPI v3 generator. All extensions in this file share the
+  // same ID because they each extend a different descriptor message.
+  // Ref: https://github.com/protocolbuffers/protobuf/commit/5e0f68530ae6ad47b78714e60861321e1d1c04ae
+  Operation openapiv3_operation = 1295;
```
aidandj, line 23: Random thought. But if the v3_operation message wrapped a v3_1_operation message. Then we could support incompatible changes in 3.x without a new extension number.

johanbrandhorst:
> This one may take me a bit to review. How much of it comes from the old PR? We worked through a number of possibilities there.

None of it as far as I know - I wasn't aware, sorry. This is an intentionally minimal list.

protoc-gen-openapiv3/options/openapiv3.proto
johanbrandhorst: I think we can keep this 3.1.0 specific right now and evaluate the need for 3.x down the road. It might even be a different generator, maybe we rename this openapiv31 or something.

protoc-gen-openapiv3/options/annotations.proto
johanbrandhorst: Eh new extensions are not that big of a deal, I assume users will only care about targeting one version of OpenAPI at a time?

aidandj:
Ok I see it now. The title made me think it was all of them.

protoc-gen-openapiv3/options/openapiv3.proto
```diff
@@ -0,0 +1,169 @@
...
+  // Additional external documentation for this operation.
+  ExternalDocs external_docs = 4;
+  // Unique string used to identify the operation. The id MUST be unique among
+  // all operations described in the API.
+  string operation_id = 5;
+  // Declares this operation to be deprecated. Consumers SHOULD refrain from
+  // usage of the declared operation. Default value is false. Setting this to
+  // false does not un-deprecate a method that is marked deprecated in proto
+  // (`option deprecated = true` on the method, service, or file).
+  bool deprecated = 6;
```
aidandj, line 127: I know you are going for simple initial implementation. But operation has a repeated `servers` parameter. And since you already defined `Servers` for the main doc, it would be a simple addition.

protoc-gen-openapiv3/options/openapiv3.proto
```diff
@@ -0,0 +1,169 @@
...
+//
+// Spec: https://spec.openapis.org/oas/v3.1.0#schema-object
+// Underlying dialect: https://json-schema.org/draft/2020-12/json-schema-core
+message Schema {
+  // A preferably short string describing the purpose of the instance
+  // described by the schema.
+  string title = 1;
+  // A string providing explanation about the purpose of the instance
+  // described by the schema.
+  string description = 2;
```
aidandj, line 149: If adding deprecated to operation, it should be added here as well.

protoc-gen-openapiv3/options/openapiv3.proto
```diff
@@ -0,0 +1,169 @@
...
+//
+// Spec: https://spec.openapis.org/oas/v3.1.0#info-object
+message Info {
+  // The title of the API.
+  string title = 1;
+  // A short summary of the API.
+  string summary = 2;
+  // A description of the API. CommonMark syntax MAY be used for rich text
+  // representation.
+  string description = 3;
```
aidandj, line 45: Looking at the openapi spec, only title and version are required. Should other fields be marked optional so that they can be omitted from the json if not set?

protoc-gen-openapiv3/options/openapiv3.proto
```diff
@@ -0,0 +1,169 @@
...
+// mutually exclusive in the OpenAPI 3.1.0 spec; the generator will error
+// if both are set.
+//
+// Spec: https://spec.openapis.org/oas/v3.1.0#license-object
+message License {
+  // The license name used for the API.
+  string name = 1;
+  // An SPDX expression for the API. The identifier field is mutually
+  // exclusive of the url field.
+  string identifier = 2;
```
aidandj: These could be a oneof. I like the idea of enforcing these things at the proto level, although I understand theres some reason to keep it simpler.

protoc-gen-openapiv3/options/openapiv3.proto
johanbrandhorst: The override semantic is "non-empty replaces, empty is no-op," so on the proto side `optional` doesn't change generator behavior — an unset `string description` and `optional string description = ""` both mean "no override." JSON omission of unset fields in the generated OpenAPI doc is handled by `omitempty` on the Go types, not by the proto definition, so changing it here wouldn't change anything, and these fields are all already `omitempty` in the Go JSON type definitions.

protoc-gen-openapiv3/options/openapiv3.proto
johanbrandhorst: Done, thanks!

protoc-gen-openapiv3/options/openapiv3.proto
johanbrandhorst: Done, thanks!

protoc-gen-openapiv3/options/openapiv3.proto
johanbrandhorst: Sounds good, done!

PR #6732, opened 2026-05-01
fix(openapiv2): sort messages before rendering definitions for determinism

### LINKED ISSUE #6652: Nondeterministic swagger generation when nested message flattens to same OpenAPI name as top-level message
## Title

  Nondeterministic swagger generation when nested message flattens to same OpenAPI name as top-level message

  ## Bug Description

  When a nested message name combines to the same OpenAPI definition name as a top-level message in the same package, `protoc-gen-openapiv2` produces **different swagger output on each invocation** even though the `.proto` source
  files have not changed.

  This happens because `renderMessagesAsDefinition` iterates over a Go `map[string]*descriptor.Message`, and Go map iteration order is randomized. When two messages map to the same swagger definition name, whichever message happe
  ns to be visited last silently overwrites the previous one. Since the visit order changes from run to run, the generated swagger file changes non-deterministically.

  ## To Reproduce

  ### `example.proto`

  ```protobuf
  syntax = "proto3";

  package example.v1;

  option go_package = "example/v1";

  // --- Case 1: Nested message flattens to same name as top-level message ---

  message PostMeta {
    string id = 1;
    Plugin plugin = 2;

    message Plugin {
      string id = 1;
      uint64 template_id = 2;
    }
  }

  // Flattened OpenAPI name: PostMeta + Plugin = "PostMetaPlugin"
  message PostMetaPlugin {
    string id = 1;
    string name = 2;
    string link = 3;
  }

  // --- Case 2: Another common collision pattern ---

  message TypeDetail {
    FilterConfig filter_config = 1;

    message FilterConfig {
      string filter_bar = 1;
    }
  }

  // Flattened OpenAPI name: TypeDetail + FilterConfig = "TypeDetailFilterConfig"
  message TypeDetailFilterConfig {
    uint64 id = 1;
    uint64 type_id = 2;
    string scene = 3;
  }

  // --- Service that references both colliding messages ---

  message GetExampleRequest {
    string id = 1;
  }

  message GetExampleResponse {
    PostMeta post_meta = 1;
    PostMetaPlugin post_meta_plugin = 2;
    TypeDetail type_detail = 3;
    TypeDetailFilterConfig type_detail_filter_config = 4;
  }

  service ExampleService {
    rpc GetExample(GetExampleRequest) returns (GetExampleResponse) {
      option (google.api.http) = {
        get: "/examples/{id}"
      };
    }
  }
  Reproduction script
  #!/bin/bash
  set -e

  go install github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2@latest

  mkdir -p /tmp/swagger_run1 /tmp/swagger_run2

  # Run 1
  protoc \
    -I . \
    -I $(go env GOPATH)/pkg/mod/github.com/grpc-ecosystem/grpc-gateway/v2@$(protoc-gen-openapiv2 --version 2>/dev/null || echo latest) \
    --openapiv2_out /tmp/swagger_run1 \
    example.proto

  # Run 2
  protoc \
    -I . \
    -I $(go env GOPATH)/pkg/mod/github.com/grpc-ecosystem/grpc-gateway/v2@$(protoc-gen-openapiv2 --version 2>/dev/null || echo latest) \
    --openapiv2_out /tmp/swagger_run2 \
    example.proto

  echo "=== Comparing generated swagger files ==="
  if diff -q /tmp/swagger_run1/example.swagger.json /tmp/swagger_run2/example.swagger.json > /dev/null 2>&1; then
    echo "Files are identical"
  else
    echo "Files DIFFER despite identical proto input!"
    echo ""
    echo "--- examplePostMetaPlugin definition in run 1 ---"
    python3 -c "import json; d=json.load(open('/tmp/swagger_run1/example.swagger.json')); print(json.dumps(d['definitions'].get('examplePostMetaPlugin', {}), indent=2, sort_keys=True))"
    echo ""
    echo "--- examplePostMetaPlugin definition in run 2 ---"
    python3 -c "import json; d=json.load(open('/tmp/swagger_run2/example.swagger.json')); print(json.dumps(d['definitions'].get('examplePostMetaPlugin', {}), indent=2, sort_keys=True))"
  fi
  Expected Behavior
  Running protoc-gen-openapiv2 multiple times against the same unchanged .proto files should always produce bit-for-bit identical swagger output.
  Actual Behavior
  The generated examplePostMetaPlugin definition randomly alternates between:
  • The nested PostMeta.Plugin schema (2 fields: id, templateId)
  • The top-level PostMetaPlugin schema (3 fields: id, name, link)
  Similarly, exampleTypeDetailFilterConfig randomly alternates between:
  • The nested TypeDetail.FilterConfig schema (1 field: filterBar)
  • The top-level TypeDetailFilterConfig schema (3 fields: id, typeId, scene)
  Root Cause Analysis
  In protoc-gen-openapiv2/internal/genopenapi/template.go, the function renderMessagesAsDefinition iterates over messages using for name, msg := range messages:
  func renderMessagesAsDefinition(messages messageMap, d openapiDefinitionsObject, reg *descriptor.Registry, customRefs refMap, pathParams []descriptor.Parameter) error {
      for name, msg := range messages { // <-- map iteration is randomized in Go
          swgName, ok := fullyQualifiedNameToOpenAPIName(msg.FQMN(), reg)
          // ...
          d[swgName] = renderMessageAsDefinition(...) // <-- last write wins
      }
      return nil
  }
  Since messageMap is defined as map[string]*descriptor.Message, the iteration order is non-deterministic. When two distinct protobuf messages flatten to the same swgName (e.g. PostMeta.Plugin → PostMetaPlugin and PostMetaPlugin
  → PostMetaPlugin), the entry written to d[swgName] depends on whichever message happens to be visited last, which changes from run to run.
  The same issue also exists in renderEnumerationsAsDefinition, which iterates over enumMap (also a Go map).
  Environment
  • Plugin version: v2.19.1 through v2.29.0 (latest main branch as of 2026-04-20)
  • Protoc version: v29.2
  • Go version: go1.23+
  • OS: macOS / Linux
  Proposed Fix
  Sort the map keys before iterating in both renderMessagesAsDefinition and renderEnumerationsAsDefinition. This makes the "last write wins" behavior deterministic. For example:
  func renderMessagesAsDefinition(messages messageMap, d openapiDefinitionsObject, reg *descriptor.Registry, customRefs refMap, pathParams []descriptor.Parameter) error {
      names := make([]string, 0, len(messages))
      for name := range messages {
          names = append(names, name)
      }
      sort.Strings(names)

      for _, name := range names {
          msg := messages[name]
          // ... existing logic
      }
      return nil
  }
  ▌ Note: Sorting ensures deterministic output, but it still means one message silently overwrites the other. An even better long-term fix might be to detect the collision and emit a warning/error, or to use a disambiguated nam
  ▌ ing scheme for nested messages that would otherwise collide with top-level messages.

alliasgher:
Fixes #6652

`renderMessagesAsDefinition` iterates a Go map whose order is randomized per run. When two messages resolve to the same OpenAPI definition name (e.g. a nested `PostMeta.Plugin` and a top-level `PostMetaPlugin` both flatten to `"PostMetaPlugin"`), whichever is visited last silently overwrites the other — and because map order varies, the generated swagger file changes non-deterministically between invocations.

**Fix:** sort the message keys before iterating so that any collision is resolved in a predictable, stable way (lexicographically last wins), making successive invocations produce identical output.

The existing test suite passes without changes.

### DIFF
--- a/protoc-gen-openapiv2/internal/genopenapi/template.go
+++ b/protoc-gen-openapiv2/internal/genopenapi/template.go
@@ -834,7 +834,16 @@ func transformAnyForJSON(schema *openapiSchemaObject, useJSONNames bool) {
 }
 
 func renderMessagesAsDefinition(messages messageMap, d openapiDefinitionsObject, reg *descriptor.Registry, customRefs refMap, pathParams []descriptor.Parameter) error {
-	for name, msg := range messages {
+	// Sort keys so that when two messages flatten to the same OpenAPI definition
+	// name the winner is deterministic (last in sorted order wins) rather than
+	// varying with Go's random map iteration order.
+	names := make([]string, 0, len(messages))
+	for name := range messages {
+		names = append(names, name)
+	}
+	sort.Strings(names)
+	for _, name := range names {
+		msg := messages[name]
 		swgName, ok := fullyQualifiedNameToOpenAPIName(msg.FQMN(), reg)
 		if !ok {
 			return fmt.Errorf("can't resolve OpenAPI name from %q", msg.FQMN())
--- a/protoc-gen-openapiv2/internal/genopenapi/template.go
+++ b/protoc-gen-openapiv2/internal/genopenapi/template.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"reflect"
 	"regexp"
+	"maps"
 	"slices"
 	"sort"
 	"strconv"
@@ -837,12 +838,7 @@ func renderMessagesAsDefinition(messages messageMap, d openapiDefinitionsObject,
 	// Sort keys so that when two messages flatten to the same OpenAPI definition
 	// name the winner is deterministic (last in sorted order wins) rather than
 	// varying with Go's random map iteration order.
-	names := make([]string, 0, len(messages))
-	for name := range messages {
-		names = append(names, name)
-	}
-	sort.Strings(names)
-	for _, name := range names {
+	for _, name := range slices.Sorted(maps.Keys(messages)) {
 		msg := messages[name]
 		swgName, ok := fullyQualifiedNameToOpenAPIName(msg.FQMN(), reg)
 		if !ok {
@@ -854,6 +850,9 @@ func renderMessagesAsDefinition(messages messageMap, d openapiDefinitionsObject,
 
 		if opt := msg.GetOptions(); opt != nil && opt.MapEntry != nil && *opt.MapEntry {
 			continue
+		}
+		if _, exists := d[swgName]; exists {
+			grpclog.Warningf("Collision: multiple messages map to OpenAPI definition name %q; the definition will be overwritten", swgName)
 		}
 		var err error
 		d[swgName], err = renderMessageAsDefinition(msg, reg, customRefs, pathParams)
@@ -1079,7 +1078,8 @@ func primitiveSchema(t descriptorpb.FieldDescriptorProto_Type) (ftype, format st
 
 // renderEnumerationsAsDefinition inserts enums into the definitions object.
 func renderEnumerationsAsDefinition(enums enumMap, d openapiDefinitionsObject, reg *descriptor.Registry, customRefs refMap) {
-	for _, enum := range enums {
+	for _, key := range slices.Sorted(maps.Keys(enums)) {
+		enum := enums[key]
 		swgName, ok := fullyQualifiedNameToOpenAPIName(enum.FQEN(), reg)
 		if !ok {
 			panic(fmt.Sprintf("can't resolve OpenAPI name from FQEN %q", enum.FQEN()))
@@ -1135,6 +1135,9 @@ func renderEnumerationsAsDefinition(enums enumMap, d openapiDefinitionsObject, r
 			panic(err)
 		}
 
+		if _, exists := d[swgName]; exists {
+			grpclog.Warningf("Collision: multiple enums map to OpenAPI definition name %q; the definition will be overwritten", swgName)
+		}
 		d[swgName] = enumSchemaObject
 	}
 }


protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -834,7 +834,16 @@ func transformAnyForJSON(schema *openapiSchemaObject, useJSONNames bool) {
...
-	for name, msg := range messages {
+	// Sort keys so that when two messages flatten to the same OpenAPI definition
+	// name the winner is deterministic (last in sorted order wins) rather than
+	// varying with Go's random map iteration order.
+	names := make([]string, 0, len(messages))
+	for name := range messages {
+		names = append(names, name)
+	}
+	sort.Strings(names)
+	for _, name := range names {
```
johanbrandhorst: This is exactly what `slices.Sorted` and `maps.Keys` is for 😁 
```suggestion
	for _, name := range slices.Sorted(maps.Keys(m)) {
```

protoc-gen-openapiv2/internal/genopenapi/template.go
johanbrandhorst, line 858: Can we add a check here if `d[swgName]` is already set, and log a warning if so? At least then it's not just silently overwriting the data.

johanbrandhorst, line 1: Can we also fix the same bug in `renderEnumerationsAsDefinition`, as mentioned in the issue?

johanbrandhorst (changes requested):
Thanks!

protoc-gen-openapiv2/internal/genopenapi/template.go
alliasgher: Done in 48e829a.

protoc-gen-openapiv2/internal/genopenapi/template.go
alliasgher: Done in 48e829a.

alliasgher: Done in 48e829a.

johanbrandhorst (approved):
Thanks!

PR #6825, opened 2026-05-20
feat(runtime): add per-route message validation for local handler mode

kamalyes:
RegisterXxxHandlerServer bypasses the gRPC interceptor chain because it calls the service implementation directly without creating a real gRPC connection. This means unary/stream interceptors (including validation) are never invoked for HTTP requests in local mode.

This adds a generic validation interface that works at the HTTP middleware layer instead:

- MessageFactory type for per-route message type registration
- HandleWithMessage() to register handlers with message factories
- ValidatorFunc type for injectable validation (go-argus, protovalidate, etc.)
- WithValidator() ServeMuxOption to enable automatic validation middleware
- protoc-gen-grpc-gateway now generates HandleWithMessage for unary methods

### DIFF
--- a/protoc-gen-grpc-gateway/internal/gengateway/template.go
+++ b/protoc-gen-grpc-gateway/internal/gengateway/template.go
@@ -799,7 +799,7 @@ func Register{{ $svc.GetName }}{{ $.RegisterFuncSuffix }}Server(ctx context.Cont
 		return
 	})
 	{{- else -}}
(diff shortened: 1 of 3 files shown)


johanbrandhorst:
Hi, thanks for your PR! Would you mind creating an issue where we can discuss the need for this feature, before going into implementation? There are several points of discussion to be had before we can proceed with merging something like this, for example, what exactly is your use case, and why is the regular HTTP middleware insufficient? I'd also ideally like to see this implemented in a way that doesn't break existing users, or those with a different version for the generator and runtime library. As it stands, I don't want to merge this until we have had a discussion about the need for it first.

Issue #2557, opened 2022-02-28
Transcoding streaming HTTPBody responses leads to file downloads of size greater than expected


mcquackers:
## 🐛 Bug Report

Using ForwardResponseStream with a stream of HttpBody leads to extra bytes appended to data in Go

## To Reproduce

(Write your steps here:)

1. Create a Unary Request - Streaming Response grpc proto (FooServer (GetFooRequest) returns (stream google.api.HttpBody))
1. Generate go bindings with GRPC Gateway and go_grpc
1. In the implementation, retrieve a remote tar.gz (eg, gitlab project archive) and return the reader.
1. Read from the reader, and return stream responses with the read data
1. Invoke the handler and note that the final file download's size is different from the amount of bytes read from the reader.

## Expected behavior

I would expect that the downloaded file would have accurate byte size and not be corrupted (tar files fail to extract, for example)

## Actual Behavior

The downloaded file differs from the expected size by the number of stream responses sent

## Your Environment

MacOS Monterey 12.1
go version go1.17.6 darwin/arm64

This is the code I'm using to write from the reader to the stream:
```
func readOntoGetProjectZipStream(reader io.Reader, server proxyV1.ConfigurationProxyService_GetProjectZipServer) error {
	b := make([]byte, 16384)
	var total int
	var n int
	var nr int
	var err error
	for nr, err = reader.Read(b); nr > 0; nr, err = reader.Read(b) {
		total += nr
		body := &httpbody.HttpBody{
			ContentType: "application/x-tar",
			Data:        make([]byte, nr),
		} //made these allocations and copying very explicit to remove any chance of off-data
		copy(body.Data, b[:nr])
		server.Send(body)
		n++
	}
	fmt.Println(total) //6479278 - as expected
	fmt.Println(n)     //398 or 424 usually; final downloaded file is greater than expected 6479278 by this amount
	if err != nil && err != io.EOF {
		return err
	}
	return nil
}```

johanbrandhorst:
[...]

PR #6858, opened 2026-05-27
docs: add binary file downloads guide

angelin-mary:

Related to #2557 (HTTPBodyMarshaler delimiter with streaming HttpBody)

Yes

Adds a new "Binary file downloads" page under the Mapping section, as a companion to the existing "Binary file uploads" page.

The current docs recommend using a custom `mux.HandlePath` route for binary uploads but have no equivalent guidance for file downloads. This gap leads developers to use custom handlers (e.g. Gin) for download endpoints, which works locally but fails silently in environments behind a reverse proxy — because the proxy only knows about routes registered through gRPC-Gateway.

This page documents the correct approach using `google.api.HttpBody` as the return type combined with `grpc.SetHeader` to set `Content-Disposition`, which registers the route natively in gRPC-Gateway and works consistently across all environments.

What's included:
- Proto definition example using `google.api.HttpBody`
- Go handler implementation with `Content-Disposition` via `grpc.SetHeader`
- Explanation of why custom handlers break in reverse-proxied environments
- Frontend JS example using File System Access API with fallback
- RFC 5987 encoding guidance for non-ASCII filenames

#### Other comments

None

### DIFF
--- a/docs/docs/mapping/binary_file_downloads.md
+++ b/docs/docs/mapping/binary_file_downloads.md
@@ -1,0 +1,181 @@
+---
+layout: default
+title: Binary file downloads
+nav_order: 4
+parent: Mapping
+permalink: /docs/mapping/binary_file_downloads/
+---
+
+# Binary file downloads
+
+If you need to return a file download from your gRPC-Gateway endpoint (e.g. a CSV
+export, PDF, or any binary file), the recommended approach is to use
+`google.api.HttpBody` as your response type. This ensures the route is registered
+natively in gRPC-Gateway and works reliably in all environments — including those
+behind a reverse proxy or API gateway.
+
+> **Why not use a custom `mux.HandlePath` route for downloads?**
+>
+> For file **uploads**, a custom route on the mux is necessary because multipart
+> form data cannot be modelled in gRPC (see [Binary file uploads](binary_file_uploads.md)).
+>
+> For file **downloads**, using a custom route (e.g. a Gin catch-all handler) can
+> work in local development but will silently fail in environments behind a reverse
+> proxy or API gateway — because the proxy only knows about routes registered
+> through gRPC-Gateway. Using `google.api.HttpBody` registers the route properly
+> and avoids this inconsistency.
+
+## Define your proto
+
+Add `google/api/httpbody.proto` to your imports and use `google.api.HttpBody` as
+the return type:
+
+```protobuf
+syntax = "proto3";
+
+import "google/api/annotations.proto";
+import "google/api/httpbody.proto";
+
+service ReportService {
+  rpc ExportReport(ExportReportRequest) returns (google.api.HttpBody) {
+    option (google.api.http) = {
+      get: "/v1/reports/export"
+    };
+  }
+}
+
+message ExportReportRequest {
+  int64 date_from = 1;
+  int64 date_to   = 2;
+}
+```
+
+## Implement the handler in Go
+
+In your gRPC server implementation, build your file content, set the
+`Content-Disposition` header via `grpc.SetHeader`, and return an `HttpBody`:
+
+```go
+import (
+    "bytes"
+    "context"
+    "fmt"
+    "time"
+
+    "google.golang.org/grpc"
+    "google.golang.org/grpc/codes"
+    "google.golang.org/grpc/metadata"
+    "google.golang.org/grpc/status"
+    httpbody "google.golang.org/genproto/googleapis/api/httpbody"
+
+    pb "your/generated/proto/package"
+)
+
+func (s *Server) ExportReport(
+    ctx context.Context,
+    req *pb.ExportReportRequest,
+) (*httpbody.HttpBody, error) {
+
+    // Build your file content — CSV, PDF, etc.
+    var buf bytes.Buffer
+    buf.WriteString("col1,col2,col3\n")
+    buf.WriteString("val1,val2,val3\n")
+
+    // Set Content-Disposition so the browser triggers a file save dialog.
+    filename := fmt.Sprintf("report-%s.csv", time.Now().Format("20060102"))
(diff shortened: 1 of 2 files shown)


docs/docs/mapping/binary_file_downloads.md
```diff
@@ -0,0 +1,181 @@
...
+behind a reverse proxy or API gateway.
+
+> **Why not use a custom `mux.HandlePath` route for downloads?**
+>
+> For file **uploads**, a custom route on the mux is necessary because multipart
+> form data cannot be modelled in gRPC (see [Binary file uploads](binary_file_uploads.md)).
+>
+> For file **downloads**, using a custom route (e.g. a Gin catch-all handler) can
+> work in local development but will silently fail in environments behind a reverse
+> proxy or API gateway — because the proxy only knows about routes registered
```
johanbrandhorst, line 24: Could you give an example of a reverse proxy or API gateway that exhibits this problem? I wasn't aware of any http.proto-aware proxies in use with the grpc-gateway. This section feels generally quite specific to this use case (a grpc-gateway behind another API gateway/reverse proxy), and I might suggest we remove it.

docs/docs/mapping/binary_file_downloads.md
```diff
@@ -0,0 +1,181 @@
...
+  rpc ExportReport(ExportReportRequest) returns (google.api.HttpBody) {
+    option (google.api.http) = {
+      get: "/v1/reports/export"
+    };
+  }
+}
+
+message ExportReportRequest {
+  int64 date_from = 1;
+  int64 date_to   = 2;
```
johanbrandhorst, line 49: As an example, this feels poor. Maybe just make it something like `string report_id = 1;`?

docs/docs/mapping/binary_file_downloads.md
```diff
@@ -0,0 +1,181 @@
...
+}
+```
+
+## Implement the handler in Go
+
+In your gRPC server implementation, build your file content, set the
+`Content-Disposition` header via `grpc.SetHeader`, and return an `HttpBody`:
```
johanbrandhorst, line 56: Add a reference to the MDN docs for `Content-Disposition` here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Disposition.

docs/docs/mapping/binary_file_downloads.md
```diff
@@ -0,0 +1,181 @@
...
+    httpbody "google.golang.org/genproto/googleapis/api/httpbody"
+
+    pb "your/generated/proto/package"
+)
+
+func (s *Server) ExportReport(
+    ctx context.Context,
+    req *pb.ExportReportRequest,
+) (*httpbody.HttpBody, error) {
+
```
johanbrandhorst, line 78: Nit: remove this empty line

docs/docs/mapping/binary_file_downloads.md
```diff
@@ -0,0 +1,181 @@
...
+    "context"
+    "fmt"
+    "time"
+
+    "google.golang.org/grpc"
+    "google.golang.org/grpc/codes"
+    "google.golang.org/grpc/metadata"
+    "google.golang.org/grpc/status"
+    httpbody "google.golang.org/genproto/googleapis/api/httpbody"
+
```
johanbrandhorst, line 70: Nit: I generally just use 2 import groups but I can't remember if we use 3 in other examples. In any case, lets just be consistent with other examples.

docs/docs/mapping/binary_file_downloads.md
```diff
@@ -0,0 +1,181 @@
...
+        ContentType: "text/csv; charset=utf-8",
+        Data:        buf.Bytes(),
+    }, nil
+}
+```
+
+## How the browser receives it
+
+gRPC-Gateway forwards the `Content-Disposition` metadata header to the HTTP
+response. The browser sees a standard file download response:
```
johanbrandhorst, line 100: Doesn't this require you to set up a custom [outgoing HTTPHeaderMatcher](https://pkg.go.dev/github.com/grpc-ecosystem/grpc-gateway/v2/runtime#WithOutgoingHeaderMatcher)? Otherwise the gateway will prefix the response header with `Grpc-Gateway-`? Something like

```
mux := runtime.NewServeMux(
    runtime.WithOutgoingHeaderMatcher(func(key string) (string, bool) {
        if key == "content-disposition" {
            return "Content-Disposition", true
        }
        return runtime.DefaultHeaderMatcher(key)
    }),
)
```

docs/docs/mapping/binary_file_downloads.md
```diff
@@ -0,0 +1,181 @@
...
+
+If your service runs behind a reverse proxy (Nginx, Envoy, etc.), the proxy
+routes requests based on the paths registered with gRPC-Gateway. A custom
+handler added outside of gRPC-Gateway (e.g. via a separate Gin router) will
+**not** be known to the proxy and requests will fail with 404 or 502 in staging
+and production environments, even if it works on localhost.
+
+Using `google.api.HttpBody` as shown above registers the download route through
+gRPC-Gateway itself, so it is visible to the proxy and behaves consistently
+across all environments.
```
johanbrandhorst, line 159: Again, this feels specific to a deployment scenario, we can just remove it. If they're following this guide they won't have this problem.

johanbrandhorst (changes requested):
Thanks for this!

docs/docs/mapping/index.md
```diff
@@ -1,6 +1,7 @@
 ---
 layout: default
-title: Mapping
-nav_order: 2
-has_children: true
+title: Binary file downloads
+nav_order: 4
+parent: Mapping
+permalink: /docs/mapping/binary_file_downloads/
```
johanbrandhorst, line 6: Is this change correct? Why is it editing an existing file instead of creating a new one?

PR #6948, opened 2026-06-16
Fix nested body path initialization for oneof fields

### LINKED ISSUE #6938: Nil pointer dereference for body path through message to oneof field
## 🐛 Bug Report

Generated .pb.gw.go code panics with nil pointer dereference when google.api.http annotation has body referencing a nested path through a message field to a oneof field (e.g. body: "input.input_value"). The generated code checks if the oneof wrapper is nil, but does not check if the parent message pointer is nil.

## To Reproduce

  1. Define a proto message with a message field containing a oneof:
```
  message ExecutionInput {
    oneof input {
      string input_json = 1;
      google.protobuf.Value input_value = 2;
    }
  }

  message StartExecutionRequest {
    string workflow_id = 1;
    ExecutionInput input = 3;
  }
```
  2. Define an RPC with additional_bindings where body points to the nested oneof field:
```
  service ExecutionService {
    rpc Start(StartExecutionRequest) returns (StartExecutionResponse) {                                                   
      option (google.api.http) = {
        additional_bindings: {
          body: "input.input_value"
          post: "/workflows/v1/execution/{workflow_id}/start"
        }
        body: "*"
        post: "/workflows/v1/execution/start"                                                                             
      };
    };
  }
```
  3. Generate the gateway code with protoc-gen-grpc-gateway.
  4. Send an HTTP request to the additional_bindings endpoint:
```
  curl -X POST -H "Content-Type: application/json" -d '{}' \
    http://localhost/workflows/v1/execution/some-id/start
```
  5. The server panics.

## Expected behavior

  The generated code initializes intermediate message pointer fields before accessing nested fields:
```
  if protoReq.Input == nil {
      protoReq.Input = &ExecutionInput{}
  }
  if protoReq.Input.Input == nil {
      protoReq.Input.Input = &ExecutionInput_InputValue{}
  }
```

## Actual Behavior

  The generated code in request_ExecutionService_Start_1 skips the nil-check for the intermediate message field:
```
  // execution_service.pb.gw.go:73
  if protoReq.Input.Input == nil {  // PANIC: protoReq.Input is nil (*ExecutionInput)
      protoReq.Input.Input = &ExecutionInput_InputValue{}
  }
  
  Panic:
  http2: panic serving 127.0.0.1:41710: runtime error: invalid memory address or nil pointer dereference
  goroutine ... [running]:
  ...request_ExecutionService_Start_1(...)
      execution_service.pb.gw.go:73 +0x88
```      
  Root cause: FieldPath.AssignableExprPrep() in internal/descriptor/types.go generates nil-checks only for oneof fields (c.Target.OneofIndex != nil). For intermediate message pointer fields it calls c.ValueExpr() without a nil-check, leaving the pointer uninitialized.

## Your Environment

  - grpc-gateway: v2.28.0 (also confirmed with v2.29.0)
  - protoc: v28.3
  - Go: 1.24
  - OS: Linux (also reproduced on macOS)

cyphercodes:

Fixes #6938

Yes.

`AssignableExprPrep` now initializes singular intermediate message fields before descending into nested body/path field expressions. This prevents generated gateway handlers from dereferencing a nil parent message when a body selector traverses through a message field to a oneof field.

A regression test covers the `input.input_value` shape from the issue and verifies the parent message initialization is emitted before the oneof wrapper initialization.

#### Other comments

Tests run:

- `go test ./internal/descriptor ./protoc-gen-grpc-gateway/...`


johanbrandhorst (changes requested):
Thanks for this PR! Could you make a change to one of our example protobuf files that would reproduce this bug, so we can protect against it breaking again in the future? See examples/internal/proto.

cyphercodes:
Updated to address the review request by adding a protobuf fixture that reproduces the nested body path through an intermediate message to a oneof field. I also excluded `internal/descriptor/testdata` from the default generation template so the fixture can live in-tree without being emitted by the OpenAPI generator, which currently requires top-level body fields.

Verification:
- `buf lint internal/descriptor/testdata/v1/nested_body_oneof.proto`
- `make proto`
- `go test ./internal/descriptor`
- `buf generate --path internal/descriptor/testdata/v1/nested_body_oneof.proto --template /tmp/grpc_gateway_only.buf.gen.yaml` plus assertion that generated gateway code contains `protoReq.Input == nil`, `protoReq.Input = &ExecutionInput{}`, and `protoReq.Input.Input == nil`
- `git diff --check`

johanbrandhorst (changes requested):
This... is not what I asked for? I asked for you to ideally update one of the existing protobuf example files, but you've added a new protobuf file as testdata, and then promptly excluded it from the generation path? Please remove the changes to `buf.gen.yaml` and the new protobuf file, it would be much better if you just edited `a_bit_of_everthing.proto` or one of the others.

dantedoyl:
@cyphercodes Hi! Any updates?

cyphercodes:
Thanks for the ping. I checked the existing-example route requested in the last review: adding a nested body path such as `body: "single_nested.nested_value"` in `examples/internal/proto/examplepb/a_bit_of_everything.proto` makes `make proto` fail earlier in `protoc-gen-openapiv2` with:

```text
panic: body of request "ABitOfEverythingService" is not a top level field: 'single_nested.nested_value'
```

The current Bazel failure is the expected generated `BUILD.bazel` diff for the descriptor testdata fixture that was rejected, not a separate source failure.

I don't want to broaden this PR into OpenAPI nested-body support without maintainer direction. Which direction would you prefer?

1. allow a descriptor/gateway-only fixture outside default OpenAPI generation,
2. point me at an existing proto path that avoids OpenAPI generation, or
3. expand this PR to include the OpenAPI generator support too.

johanbrandhorst:
I'm confused by your answer. This fixes an issue in the gateway generator, right? If so, we should definitely just go ahead and fix that same issue in the openapiv2 generator and include an example in `a_bit_of_everything.proto`.

dantedoyl:
@cyphercodes Hi! Any updates?

cyphercodes:
Updated this to match the requested existing-example direction.

Changes:
- removed the `buf.gen.yaml` exclusion and deleted the descriptor-only testdata proto
- added a nested body/oneof case to `examples/internal/proto/examplepb/a_bit_of_everything.proto`
- taught `protoc-gen-openapiv2` to render nested body field paths instead of rejecting them as non-top-level
- regenerated the gateway/openapi artifacts and added the example server stub

Verification:
- `go test ./internal/descriptor ./protoc-gen-openapiv2/internal/genopenapi`
- `go install ./protoc-gen-grpc-gateway ./protoc-gen-openapiv2 ./protoc-gen-openapiv3`
- `buf generate --path examples/internal/proto/examplepb/a_bit_of_everything.proto`
- `buf generate --template ./protoc-gen-openapiv3/buf.gen.yaml --path ./examples/internal/proto/examplepb/a_bit_of_everything.proto`
- `go test ./examples/internal/proto/examplepb ./examples/internal/server ./protoc-gen-grpc-gateway/internal/gengateway ./protoc-gen-openapiv2/internal/genopenapi`
- `go test ./...`
- `git diff --check`

dantedoyl:
@johanbrandhorst Hi! Could you review the changes?

dantedoyl:
@johanbrandhorst ping)

protoc-gen-openapiv2/internal/genopenapi/template.go
```diff
@@ -1719,20 +1719,20 @@ func renderServices(services []*descriptor.Service, paths *openapiPathsObject, r
 							}
 						}
 					} else {
-						// Body field path is limited to one path component. From google.api.HttpRule.body:
-						// "NOTE: the referred field must be present at the top-level of the request message type."
-						// Ref: https://github.com/googleapis/googleapis/blob/b3397f5febbf21dfc69b875ddabaf76bee765058/google/api/http.proto#L350-L352
```
johanbrandhorst, line 1724: Is this no longer correct? It would be a shame to lose this explanatory comment.

johanbrandhorst (changes requested):

cyphercodes:
Updated to address the latest review and generated-artifact failure. I restored the explanatory `HttpRule.body` comment with an added note for nested gateway body paths, then refreshed the generated OpenAPI/clients artifacts from `make generate`.

Verification:
- `make generate`
- `go test ./protoc-gen-openapiv2/internal/genopenapi ./examples/internal/proto/examplepb ./examples/internal/server ./protoc-gen-grpc-gateway/internal/gengateway`
- `git diff --check --cached`

johanbrandhorst:
Looks like there's still a bazel error

cyphercodes:
Updated the Bazel BUILD files that `bazel run //:gazelle` reported as stale for the generated nested-body oneof client/model files.

Verification:
- `go test ./protoc-gen-openapiv2/internal/genopenapi ./examples/internal/proto/examplepb ./examples/internal/server ./protoc-gen-grpc-gateway/internal/gengateway`
- `/root/go/bin/bazelisk run //:gazelle && git add -N . && git diff --exit-code -- . ':(exclude)bazel-grpc-gateway-6948'`
- `git diff --check HEAD~1..HEAD`

johanbrandhorst (approved):
Thanks!

PR #6973, opened 2026-06-20
Fix: prioritize OpenAPI config options over proto options

### LINKED ISSUE #2891: Options defined in the OpenAPI config file should take precedence over configs in the Proto file
## 🚀 Feature

As pointed out in #2858, users usually use the openapi configuration file when they can't edit the protobuf file for some reason. So it would be better for the options defined in the openapi configuration file to take precedence.

Currently, only `Service` has been modified to give priority to the options defined in the OpenAPI configuration file, while the `Method`, `Message`, `File` and `Field` still give priority to the options defined in the Proto file.

### What we need to do

Fix each `getXxxOpenAPIOption` method so that it first retrieves the options defined in the OpenAPI configuration file, and uses the values defined in the Proto file only if the OpenAPI config options are empty. 

It would be nice to check if the automatically generated examples do not change and add some test cases if needed.

For `Method`, we may need to fix the function below.
https://github.com/grpc-ecosystem/grpc-gateway/blob/031bc11c5e7a7ed791f0325cbc01f715ae6714bf/protoc-gen-openapiv2/internal/genopenapi/template.go#L2488

`Service` has already been fixed. You can refer to the following implementation when fixing the others.
https://github.com/grpc-ecosystem/grpc-gateway/blob/031bc11c5e7a7ed791f0325cbc01f715ae6714bf/protoc-gen-openapiv2/internal/genopenapi/template.go#L2518

I think we need to add notes to the doc, as changing the configuration files' priority may impact existing users.

maskedmaxx:

Fixes #2891.

Yes.

This updates OpenAPI option lookup so registered config-file options take precedence over proto annotations for method, message, file, and field options.

This matches the existing service option behavior and fixes the inconsistency described in #2891.

A regression test was added for the case where both proto annotations and config-file options are present.

#### Other comments

Tested with:

```sh
go test ./protoc-gen-openapiv2/internal/genopenapi -run 'Test(OpenAPIConfigOptionsTakePrecedenceOverProtoOptions|ApplyTemplateOverrideWithOperation|SchemaOfField|RenderMessagesAsDefinition)$' -count=1

### DIFF
--- a/protoc-gen-openapiv2/internal/genopenapi/template.go
+++ b/protoc-gen-openapiv2/internal/genopenapi/template.go
@@ -3313,31 +3313,23 @@ func getEnumValueVisibilityOption(fd *descriptorpb.EnumValueDescriptorProto) *vi
 }
 
 func getMethodOpenAPIOption(reg *descriptor.Registry, meth *descriptor.Method) (*openapi_options.Operation, error) {
+	if opts, ok := reg.GetOpenAPIMethodOption(meth.FQMN()); ok {
+		return opts, nil
+	}
 	opts, err := extractOperationOptionFromMethodDescriptor(meth.MethodDescriptorProto)
 	if err != nil {
 		return nil, err
-	}
-	if opts != nil {
-		return opts, nil
-	}
-	opts, ok := reg.GetOpenAPIMethodOption(meth.FQMN())
-	if !ok {
-		return nil, nil
 	}
 	return opts, nil
 }
 
 func getMessageOpenAPIOption(reg *descriptor.Registry, msg *descriptor.Message) (*openapi_options.Schema, error) {
+	if opts, ok := reg.GetOpenAPIMessageOption(msg.FQMN()); ok {
+		return opts, nil
+	}
 	opts, err := extractSchemaOptionFromMessageDescriptor(msg.DescriptorProto)
 	if err != nil {
 		return nil, err
-	}
-	if opts != nil {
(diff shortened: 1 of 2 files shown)


johanbrandhorst (commented):
This looks good at a first glance, but to be clear, does this prioritize options from the OpenAPI configuration file, or does it prioritize any sort of OpenAPI annotations over protobuf options? I think we want the former, but the latter may be a breaking change we want to consider more carefully.

maskedmaxx:
Good question @johanbrandhorst, and thanks for calling that out. This PR is intended to prioritize the OpenAPI configuration file options specifically.

The changed lookups only check the registry first, e.g. `reg.GetOpenAPIMethodOption`, `reg.GetOpenAPIMessageOption`, `reg.GetOpenAPIFileOption`, and `reg.GetOpenAPIFieldOption`. Those values are populated through `RegisterOpenAPIOptions`, which is used for the external OpenAPI configuration file.

If no registered config option exists, the code still falls back to the proto OpenAPI annotation extraction, like `extractOperationOptionFromMethodDescriptor`, `extractSchemaOptionFromMessageDescriptor`, etc.

So the precedence becomes:

OpenAPI config file option > proto OpenAPI annotation

That matches the existing service option behavior, and my understanding is that it should not change precedence for unrelated protobuf options.

johanbrandhorst (approved):
Great, thanks for answering!

PR #6988, opened 2026-06-23
feat(protoc-gen-openapiv2): move to protoc toolchain

### LINKED ISSUE #5896: protoc toolchain support for bazel protoc_gen_openapiv2 rules
## 🚀 Feature

We have a Bazel+nix monorepo and are working on adopting grpc-gateway. Recently, the build team stiched over to using protoc toolchains with the protoc version from nix to avoid the continuous protobuf rebuilds we were experiencing with the native Bazel behavior.

Unfortunately, this broke our use of the protoc_gen_openapiv2 Bazel rule since that uses protoc directly rather than through a toolchain.

It would be super nice to be able to get protoc toolchain support for this rule.

EliSauder:

Fixes #5896

Move to using the protoc toolchain for the openapiv2 bazel rule rather than a direct reference. This should enable utilizing pre-built protoc.

#### Other comments

I don't know if there are existing usage tests in this repo, but I tested my fork as a submodule and everything built successfully. I was also able to successfully `bazel build //... && bazel test //...` in this branch.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,2 +1,5 @@
+common --@com_google_protobuf//bazel/flags:prefer_prebuilt_protoc
+common --incompatible_enable_proto_toolchain_resolution
+
 build --cxxopt=-std=c++17 --host_cxxopt=-std=c++17
 build --test_output=errors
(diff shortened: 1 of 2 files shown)


johanbrandhorst (approved):
LGTM, thanks!

PR #7026, opened 2026-07-01
fix(openapiv2): preserve x-nullable when use_allof_for_refs wraps $ref

pkhanaljc:

Yes.

- When both `proto3_optional_nullable=true` and `use_allof_for_refs=true` are enabled, optional **message** fields now keep `x-nullable: true` after the `allOf` `$ref` rewrite.
- `XNullable` is included in the wrap condition and copied into the wrapper `schemaCore`, matching how `extensions` were preserved in #3100.
- Adds `TestRenderMessagesAsDefinition` cases for optional proto3 message fields with and without a field title.

## Problem

`schemaOfField` sets `XNullable: true` for proto3 optional fields when `proto3_optional_nullable` is enabled. The `use_allof_for_refs` block in `renderMessageAsDefinition` rebuilds message `$ref` fields into an `allOf` wrapper but previously copied `Title`, `Description`, `Example`, `readOnly`, and `extensions` only—dropping `XNullable` because it lives on `schemaCore`, not in `extensions`.

Example broken output:

```yaml
items:
  title: Support for array item definitions
  allOf:
    - $ref: '#/definitions/...'
```

Expected:

```yaml
items:
  title: Support for array item definitions
  x-nullable: true
  allOf:
    - $ref: '#/definitions/...'
```

## Test plan

- [x] `go test ./protoc-gen-openapiv2/internal/genopenapi/ -run TestRenderMessagesAsDefinition/optional`
- [x] `go test ./protoc-gen-openapiv2/internal/genopenapi/ -run TestRenderMessagesAsDefinition`

#### Other comments


johanbrandhorst (changes requested):
Thanks for this PR, could you add an example that exhibits this bug to one of our example protobuf files, so we can see what the generated file changes look like? Thanks!

pkhanaljc:
@johanbrandhorst I have added an example. Thanks.

johanbrandhorst (approved):
Very nice, thanks!

Issue #7007, opened 2026-06-28
Clarify that `tools.go` in Installation is an example for user projects


hawk-roy:
In README.md -> Installation -> Compile from source, the documentation shows an example Go file with build tags and blank imports. This file is commonly named `tools.go`.

As a new user, I initially looked for `tools.go` in this repository and was confused because it does not exist.

Could the README clarify that this file should be created in the user's own Go module/project, and is not expected to already exist in this repository?

For example, adding a sentence like:

"Create a `tools.go` file in your own module to track these tool dependencies."

This would make the installation steps clearer for beginners.

hawk-roy:
Hi @johanbrandhorst I opened a small documentation PR to clarify the relationship between the Go 1.24 `tool` directive and the legacy blank-import tool tracking pattern: #7042.

PR #7042, opened 2026-07-04
docs: clarify Go tool dependency options

hawk-roy:
References to other Issues or PRs

Related to #7007

Have you read the Contributing Guidelines?

Yes.

Brief description of what is fixed or changed

This PR adds a short clarification before the legacy blank-import tool tracking example.

The README shows both the Go 1.24 `tool` directive approach and the older blank-import pattern. This clarification helps readers understand that the `tool` directive should be preferred for Go 1.24 and later, while the blank-import pattern remains useful for projects that are not using Go 1.24 yet.

Other comments

Docs-only change. No tests were run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,8 @@ management. Use a
 [tool dependency](https://go.dev/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module)
 to track the versions of the following executable packages:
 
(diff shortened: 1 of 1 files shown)


johanbrandhorst (approved):
Thanks!

PR #7059, opened 2026-07-08
fix(gengateway): remove redundant enum assignment for nested proto3 f…

### LINKED ISSUE #2901: Embedded struct / message is not correctly populated, causing panic
## 🐛 Bug Report

Embedded struct / message is not correctly populated, causing panic

## To Reproduce 

Use this particular proto file:

```proto3
syntax = "proto3";

package test;

import "protoc-gen-openapiv2/options/annotations.proto";

option go_package = "github.com/alexadhy/protos-gen-go/test;test";

option (grpc.gateway.protoc_gen_openapiv2.options.openapiv2_swagger) = {
	responses: {
		key: "200";
	}
	responses: {
		key: "404";
		value: {
			description: "Returned when the resource does not exist.";
		}
	}
	responses: {
		key: "500";
		value: {
			description: "Returned when the service is temporarily unavailable.";
		}
	}
	responses: {
		key: "503";
		value: {
			description: "Returned when the resource is temporarily unavailable.";
		}
	}
};

service XService {
	rpc UpdateX (UpdateXRequest) returns (X) {
		option (google.api.http) = {
			put: "/v1/exes/{x.type=*}"
			body: "x"
		};
	}
}

message UpdateXRequest {
	// The parent resource
	string parent = 1;

	// The X to set.
	X x = 2;
}

enum XType {
	TYPE_UNDEFINED = 0;
	TYPE_A = 1;
	TYPE_B = 2;
}

message X {
	XType type = 1;
	string name = 2;
}

```

on generated code, it generates (snippet):
```proto3
// Suppress "imported and not used" errors
var _ codes.Code
var _ io.Reader
var _ status.Status
var _ = runtime.String
var _ = utilities.NewDoubleArray
var _ = metadata.Join

var (
	filter_XService_UpdateX_0 = &utilities.DoubleArray{Encoding: map[string]int{"x": 0, "type": 1}, Base: []int{1, 2, 1, 0, 0}, Check: []int{0, 1, 2, 3, 2}}
)

func request_XService_UpdateX_0(ctx context.Context, marshaler runtime.Marshaler, client XServiceClient, req *http.Request, pathParams map[string]string) (proto.Message, runtime.ServerMetadata, error) {
	var protoReq UpdateXRequest

	newReader, berr := utilities.IOReaderFactory(req.Body)
	if berr != nil {
		return nil, metadata, status.Errorf(codes.InvalidArgument, "%v", berr)
	}
	if err := marshaler.NewDecoder(newReader()).Decode(&protoReq.X); err != nil && err != io.EOF {
	}

	var (
		val string
		e   int32
		ok  bool
		err error
		_   = err
	)

	val, ok = pathParams["x.type"]
	if !ok {
		return nil, metadata, status.Errorf(codes.InvalidArgument, "missing parameter %s", "x.type")
	}

	err = runtime.PopulateFieldFromPath(&protoReq, "x.type", val)
	if err != nil {
		return nil, metadata, status.Errorf(codes.InvalidArgument, "type mismatch, parameter: %s, error: %v", "x.type", err)
	}

	e, err = runtime.Enum(val, XType_value)
	if err != nil {
		return nil, metadata, status.Errorf(codes.InvalidArgument, "could not parse path as enum value, parameter: %s, error: %v", "x.type", err)
	}

	protoReq.X.Type = XType(e) // this line is causing issue

	if err := req.ParseForm(); err != nil {
	}
	if err := runtime.PopulateQueryParameters(&protoReq, req.Form, filter_XService_UpdateX_0); err != nil {
	}

	msg, err := client.UpdateX(ctx, &protoReq, grpc.Header(&metadata.HeaderMD), grpc.Trailer(&metadata.TrailerMD))
	return msg, metadata, err

}

```

I added a comment where the generated code creates an issue, because the `X` object is not yet initialized, access / assignment to `X` via `protoreq.X.type = XType(e)` will panic

## Expected behavior

I expect that it should be populated correctly before assigning something, perhaps something like `protoreq.X = &X{}` 

## Actual Behavior

Panic

## Your Environment

OS: Mac
Generated via buf with the following buf.lock:
# Generated by buf. DO NOT EDIT.
version: v1
deps:
  - remote: buf.build
    owner: googleapis
    repository: googleapis
    commit: 8d7204855ec14631a499bd7393ce1970
  - remote: buf.build
    owner: grpc-ecosystem
    repository: grpc-gateway
    commit: bc28b723cd774c32b6fbc77621518765

```

### PR DESCRIPTION
**Pull request by abhishek-dalbanjan**
Fixes #2901

Yes

Removes the redundant enum assignment block in `protoc-gen-grpc-gateway/internal/gengateway/template.go` for nested proto3 fields (`IsNestedProto3 = true`). 

For nested proto3 fields, `runtime.PopulateFieldFromPath` completely handles the initialization of nested structs and sets the enum value properly based on the path parameters. However, the generator was redundantly emitting code to parse the enum again and directly assign it to `protoReq.Nested.Field`. This caused a nil pointer dereference panic when the nested struct was uninitialized prior to this manual assignment.

Additionally, updated `hasEnumPathParam` to omit nested proto3 enums, which prevents generating unused variables (`e int32`) in the generated output.

#### Other comments
All unit tests and integration tests have been run locally, including those testing repeated path enums and opaque API fields, and they pass successfully without any regressions.

### DIFF
--- a/examples/internal/proto/examplepb/a_bit_of_everything.pb.gw.go
+++ b/examples/internal/proto/examplepb/a_bit_of_everything.pb.gw.go
@@ -1706,7 +1706,6 @@ func request_ABitOfEverythingService_CheckNestedEnumGetQueryParams_0(ctx context
 	var (
 		protoReq ABitOfEverything
 		metadata runtime.ServerMetadata
-		e        int32
 		err      error
 	)
 	if req.Body != nil {
@@ -1720,11 +1719,6 @@ func request_ABitOfEverythingService_CheckNestedEnumGetQueryParams_0(ctx context
 	if err != nil {
 		return nil, metadata, status.Errorf(codes.InvalidArgument, "type mismatch, parameter: %s, error: %v", "single_nested.ok", err)
 	}
-	e, err = runtime.Enum(val, ABitOfEverything_Nested_DeepEnum_value)
-	if err != nil {
-		return nil, metadata, status.Errorf(codes.InvalidArgument, "could not parse path as enum value, parameter: %s, error: %v", "single_nested.ok", err)
-	}
-	protoReq.SingleNested.Ok = ABitOfEverything_Nested_DeepEnum(e)
 	if err := req.ParseForm(); err != nil {
 		return nil, metadata, status.Errorf(codes.InvalidArgument, "%v", err)
 	}
@@ -1739,7 +1733,6 @@ func local_request_ABitOfEverythingService_CheckNestedEnumGetQueryParams_0(ctx c
 	var (
 		protoReq ABitOfEverything
 		metadata runtime.ServerMetadata
-		e        int32
 		err      error
 	)
 	val, ok := pathParams["single_nested.ok"]
@@ -1750,11 +1743,6 @@ func local_request_ABitOfEverythingService_CheckNestedEnumGetQueryParams_0(ctx c
 	if err != nil {
 		return nil, metadata, status.Errorf(codes.InvalidArgument, "type mismatch, parameter: %s, error: %v", "single_nested.ok", err)
 	}
-	e, err = runtime.Enum(val, ABitOfEverything_Nested_DeepEnum_value)
-	if err != nil {
-		return nil, metadata, status.Errorf(codes.InvalidArgument, "could not parse path as enum value, parameter: %s, error: %v", "single_nested.ok", err)
-	}
-	protoReq.SingleNested.Ok = ABitOfEverything_Nested_DeepEnum(e)
 	if err := req.ParseForm(); err != nil {
 		return nil, metadata, status.Errorf(codes.InvalidArgument, "%v", err)
 	}
--- a/examples/internal/integration/integration_test.go
+++ b/examples/internal/integration/integration_test.go
@@ -2882,3 +2882,23 @@ func testExcessBodyStreamWithBodyUnexpected(t *testing.T, port int) {
 		t.Errorf("server context not done")
 	}
 }
+
+func TestNestedEnumGetQueryParams(t *testing.T) {
+	if testing.Short() {
+		t.Skip()
+		return
+	}
+
+	// 1 corresponds to DeepEnum.TRUE
+	url := "http://localhost:8088/v1/example/a_bit_of_everything/params/get/nested_enum/1"
+	
+	resp, err := http.Get(url)
+	if err != nil {
+		t.Fatalf("http.Get(%q) failed with %v; want success", url, err)
+	}
+	defer resp.Body.Close()
+
+	if got, want := resp.StatusCode, http.StatusOK; got != want {
+		t.Fatalf("resp.StatusCode = %d; want %d", got, want)
+	}
+}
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Review by johanbrandhorst — CHANGES_REQUESTED** [review]
Thanks for this PR, could you update one of the example protobuf files with a message that exhibits this issue? That way we can see the effect of the change in the generated files and will know if we break this in the future again.

**Comment by abhishek-dalbanjan** [comment]
> Thanks for this PR, could you update one of the example protobuf files with a message that exhibits this issue? That way we can see the effect of the change in the generated files and will know if we break this in the future again.


I've run make generate and pushed the updated .pb.gw.go stubs. It turns out a_bit_of_everything.proto already contained an endpoint (CheckNestedEnumGetQueryParams) that maps a nested proto3 enum (single_nested.ok) as a path parameter, which perfectly exhibits this issue. You can see in the latest commit that the redundant enum assignment block and unused variables have been cleanly removed from the generated stubs."

File: examples/internal/proto/examplepb/a_bit_of_everything.pb.gw.go [hunk 1]
```diff
@@ -1720,11 +1719,6 @@ func request_ABitOfEverythingService_CheckNestedEnumGetQueryParams_0(ctx context
 	if err != nil {
 		return nil, metadata, status.Errorf(codes.InvalidArgument, "type mismatch, parameter: %s, error: %v", "single_nested.ok", err)
 	}
-	e, err = runtime.Enum(val, ABitOfEverything_Nested_DeepEnum_value)
-	if err != nil {
-		return nil, metadata, status.Errorf(codes.InvalidArgument, "could not parse path as enum value, parameter: %s, error: %v", "single_nested.ok", err)
-	}
-	protoReq.SingleNested.Ok = ABitOfEverything_Nested_DeepEnum(e)
```
johanbrandhorst, line 1727: Where is this value being assigned now, if not here?

abhishek-dalbanjan: The value is already being safely assigned just a few lines above this block by runtime.PopulateFieldFromPath(&protoReq, "single_nested.ok", val).

For nested proto3 fields, the generator correctly emits a call to PopulateFieldFromPath. This runtime function traverses the path, safely allocates any intermediate nested structs (protoReq.SingleNested in this case) if they are nil, and uses reflection to correctly parse and assign the enum value.

Before this PR, the generator was redundantly emitting code that would parse the enum again (via runtime.Enum) and perform a hardcoded, direct assignment (protoReq.SingleNested.Ok = ...). This was not only duplicate work, but that direct assignment was exactly what caused the nil pointer dereference panic if the intermediate struct was uninitialized. By removing this redundant manual assignment for nested fields, we let PopulateFieldFromPath handle both the safe allocation and the assignment.

johanbrandhorst: Thanks for the explanation!

johanbrandhorst (approved):
LGTM

PR #7065, opened 2026-07-09
fix(openapiv2): omit empty body objects for nested path parameters

### LINKED ISSUE #2624: protoc-gen-openapiv2: don't emit body objects containing only nested path parameters
## 🚀 Feature

When path parameters are all part of a message in the body protoc-gen-openapiv2 emits a body property that is an empty object. While technically the field `id` is selected by `body: "*"` its sole field `value` is a path parameter and `id` is emitted as body object without properties.

```proto
service EntityService {
  rpc DeleteEntity(DeleteEntityRequest) returns (DeleteEntityResponse) {
    option (google.api.http) = {
      delete: "/entities/{id.value}"
      body: "*"
    };
  }
}

// message Entity {
//      EntityId id = 1;
//      etc.
// }

message EntityId {
        string value = 1;
}

message DeleteEntityRequest {
        EntityId id = 1;
        string other_field = 2;
}

message DeleteEntityResponse {}
```

```json
    "/entities/{id.value}": {
      "delete": {
        "operationId": "EntityService_DeleteEntity",
        "responses": {
          "200": {
            "schema": {
              "$ref": "#/definitions/testingDeleteEntityResponse"
            }
          }
        },
        "parameters": [
          {
            "name": "id.value",
            "in": "path",
            "required": true,
            "type": "string"
          },
          {
            "name": "body",
            "in": "body",
            "required": true,
            "schema": {
              "type": "object",
              "properties": {
                "id": {
                  "type": "object"     <---
                },
                "otherField": {
                  "type": "string"
                }
              }
            }
          }
        ]
      }
    }
```

alliasgher:

Fixes #2624

Yes.

When every sub-field of a body message field is bound to a path parameter, `protoc-gen-openapiv2` stripped those sub-fields but still emitted the field as an empty object property, e.g. for `DeleteEntity({id.value}, body: "*")`:

```json
"schema": {
  "type": "object",
  "properties": {
    "id": { "type": "object" },   // <-- empty, id.value is a path param
    "otherField": { "type": "string" }
  }
}
```

Following the approach discussed on the issue (Option 1), `renderMessageAsDefinition` now skips a field when it contains path parameters (`subPathParams` is non-empty) **and** renders to an empty object, removing it from the body schema and from the `required` list. The skip is gated on the presence of path parameters, so messages that are legitimately empty for other reasons (e.g. `google.protobuf.Empty`, empty response bodies) are unaffected.

Added `TestRenderServicesOpenapiOmitsBodyFieldOnlyPathParams`, which reproduces the issue's exact scenario and asserts the empty `id` object is omitted while `other_field` remains.

#### Other comments

I verified this changes no existing generated fixtures: building the plugin before and after this change and generating all example `.swagger.json` with an identical template produces an identical result, so no example currently exercises this path and none needed regeneration. The behavior is covered by the new unit test.


johanbrandhorst (changes requested):
This is great, thanks! Please add an RPC to our example protos that reproduces this issue so we can see it rendered into our swagger files after generation.

alliasgher:
Added an UpdateEntity RPC to a_bit_of_everything in 27453bc. Its generated body definition (ABitOfEverythingServiceUpdateEntityBody) now contains only other_field, with the empty id object omitted. Used PUT since the gateway rejects body on DELETE without allow_delete_body.

johanbrandhorst:
It's still failing

johanbrandhorst (changes requested):
CI is failing

alliasgher:
Added the `UpdateEntity` RPC to `a_bit_of_everything` (`PUT /v1/example/a_bit_of_everything/entity/{id.value}` with `body: "*"`), so the empty-body-path-param case is exercised in the generated swagger.

CI was red because the generated service now requires an `UpdateEntity` method on the hand-written example server (`require_unimplemented_servers=false`) and the regen was partial. Fixed in 5f839157: implemented the server method (unblocks build/test/lint/node_test) and ran the full `make generate` so every artifact is consistent (openapi.json, pb.gw.go, the abev3 oapi-codegen client, and the abe swagger client plus its BUILD.bazel entries).

examples/internal/proto/examplepb/a_bit_of_everything.openapi.json
```diff
@@ -738,6 +738,65 @@
...
+        ],
+        "requestBody": {
+          "content": {
+            "application/json": {
+              "schema": {
+                "type": "object",
+                "properties": {
+                  "id": {
+                    "$ref": "#/components/schemas/grpc.gateway.examples.internal.proto.examplepb.EntityId"
+                  },
```
johanbrandhorst, line 766: Now we just need to fix this in openapiv3 too ;). Next PR?

johanbrandhorst (approved):
Thanks!

examples/internal/proto/examplepb/a_bit_of_everything.openapi.json
alliasgher: Good catch. Yes, I'll open a follow-up PR to apply the same fix in protoc-gen-openapiv3.

alliasgher: Opened #7068 for the openapiv3 fix.

PR #7068, opened 2026-07-09
fix(openapiv3): omit empty body objects for nested path parameters

alliasgher:

Follow-up to #7065 (which fixed the openapiv2 side of #2624). Requested by @johanbrandhorst in https://github.com/grpc-ecosystem/grpc-gateway/pull/7065#discussion_r3555387973.

#### Have you read the Contributing Guidelines?

Yes.

`protoc-gen-openapiv2` was fixed in #7065 to drop body fields whose only sub-fields are bound to path parameters. `protoc-gen-openapiv3` had the same bug: for `body: "*"`, `buildRequestBody` only skipped fields that were *directly* path parameters (`isPathParam`), so a field like `id` (an `EntityId` whose only member `value` is bound to the `{id.value}` path parameter) was still emitted as a `$ref` to a component that carries nothing but a path parameter.

`buildRequestBody` now skips a `body: "*"` field when it is fully consumed by path parameters. The check is recursive: `subPathParams` strips one field-path component per level of message nesting, so it handles arbitrarily deep nested path params and terminates. A field with any remaining non-path-parameter content, a repeated field, or a legitimately empty message is unaffected.

#### Notes

The example rendering (`a_bit_of_everything.openapi.json`) will reflect this once #7065 merges (it adds the `UpdateEntity` RPC that exercises the case); this PR is the generator fix plus a unit test, and produces no change to the current example output. Added `TestFieldConsumedByPathParams`.

### DIFF
--- a/protoc-gen-openapiv3/internal/genopenapi/operation.go
+++ b/protoc-gen-openapiv3/internal/genopenapi/operation.go
@@ -310,7 +310,7 @@ func buildRequestBody(b *schemaBuilder, m *descriptor.Method, binding *descripto
 				// Field is hidden by visibility rules, skip.
 				continue
 			}
-			if isPathParam(field, binding.PathParams) {
+			if b.fieldConsumedByPathParams(field, binding.PathParams) {
 				continue
 			}
 			b.addProperty(bodySchema, field)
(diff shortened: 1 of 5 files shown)


protoc-gen-openapiv3/internal/genopenapi/operation.go
```diff
@@ -424,6 +424,54 @@ func isPathParam(field *descriptor.Field, params []descriptor.Parameter) bool {
...
+	return inner
+}
+
+// fieldConsumedByPathParams reports whether every leaf value of field is bound
+// to a path parameter, so the field would render as an empty object in a
+// body="*" request body and should be omitted rather than emitted as
+// `{"type":"object"}` or a $ref whose only members are path parameters. A
+// scalar is consumed only when it is itself a path parameter; a singular
+// message is consumed when all of its sub-fields are (recursively) consumed.
+// Mirrors the openapiv2 fix for #2624.
```
johanbrandhorst: Remove this comment

johanbrandhorst (changes requested):
You'll need to re-run the generators

protoc-gen-openapiv3/internal/genopenapi/operation.go
alliasgher: Removed in 91893439.

alliasgher:
Thanks. Re-ran the generators (the a_bit_of_everything openapi.json and abev3 client now drop the empty id object) and removed the comment, in 91893439.

johanbrandhorst (approved):
Thanks!

PR #7147, opened 2026-07-24
docs: explain Buf strategy for shared API config

dlanov:
## Summary

Document how to use one gRPC API Configuration YAML file for services defined in multiple directories when generating code with Buf.

Buf invokes local plugins once per directory by default. When a shared configuration contains selectors for services in other directories, each plugin invocation sees only part of the schema and reports the other selectors as unmatched.

The documentation now shows that affected local plugins should use `strategy: all`, ensuring all service descriptors are sent in one plugin request.

## Verification

Reproduced with two unannotated services in separate directories using one shared configuration file:

- `strategy: directory`: failed with `HTTP rules without a matching selector`
- `strategy: all`: successfully generated both `.pb.gw.go` files

No generator behavior was changed.

Addresses #3059.


johanbrandhorst (approved):
Thanks!

PR #7149, opened 2026-07-25
docs: fix SLSA link in README

ZayanKhan-12:
In the "How to use" section, the README links `[SLSA3 signatures](slsa.dev)`. Because the destination has no scheme, Markdown treats `slsa.dev` as a repository-relative path, so the link resolves to `github.com/grpc-ecosystem/grpc-gateway/blob/main/slsa.dev` and 404s instead of going to the SLSA site.

Changed to the absolute `https://slsa.dev`. Docs-only.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


johanbrandhorst (approved):
Thanks1

PR #7188, opened 2026-08-03
feat: add option to disable request body draining

dlanov:

Fixes #6495.

#### Have you read the Contributing Guidelines?

Yes.

Adds a `disable_request_body_draining` option to `protoc-gen-grpc-gateway`.

Request-body draining remains enabled by default, preserving existing generated behavior. When enabled, the option omits the explicit request-body drain in generated handlers for RPCs without client streaming.

The option covers:

* handlers with a mapped request body;
* PATCH handlers using automatic field-mask processing; and
* handlers without a body mapping.

Normal request processing, including protobuf decoding and `req.ParseForm()`, may still consume the request body.

This change also:

* documents Buf, `protoc`, and Bazel usage; and
* adds a public Bazel compiler target that enables the option while preserving
  the existing default compiler behavior.

Regression tests verify both the default behavior and the opt-out behavior for all three request paths.

#### Testing

* `go test ./protoc-gen-grpc-gateway/internal/gengateway -run '^TestDisableRequestBodyDraining$' -count=50`
* `go test ./protoc-gen-grpc-gateway/internal/gengateway -count=20`
* `go test ./protoc-gen-grpc-gateway/... -count=1`
* `go vet ./protoc-gen-grpc-gateway/...`
* `bazel build //protoc-gen-grpc-gateway:go_gen_grpc_gateway_disable_request_body_draining`
* `bazel test //protoc-gen-grpc-gateway/internal/gengateway:gengateway_test`
* `bazel run :gazelle`


johanbrandhorst (changes requested):
Could you please add a new protobuf file and generator which uses this option to our example/proto protofiles? We have a bunch of examples in there of flag-specific generators, just add it alongside those. Thanks!

dlanov:
Added a dedicated `disable_request_body_draining.proto` example and Buf generator configuration under `examples/internal/proto/examplepb`. I also wired it into `make proto` and added a Bazel `go_proto_library` that uses the new compiler target. The checked-in generated gateway contains the expected `ParseForm` paths without the explicit request-body drain.

johanbrandhorst:
I think this is superceded by https://github.com/grpc-ecosystem/grpc-gateway/pull/7192. Thanks for your contribution regardless!

PR #7207, opened 2026-08-08
fix(openapiv2): omit oneof siblings of path parameters from query parameters

### LINKED ISSUE #2494: An RPC proto oneof field documents the field not used in the path as an optional request parameter

hsdfat:
When a member of a oneof group is bound to a path parameter, the other members of that group can never be set, so they should not be emitted as query parameters. This mirrors the existing isBodySameOneOf handling for request bodies.

The check is done in messageToQueryParameters, where every field is a direct field of the request message, and compares against the first component of the path parameter's field path. Comparing oneof indices further down the recursion is unsound because oneof_index is only unique within its containing message.

Fixes #2494

### DIFF
--- a/examples/internal/clients/echo/client/echo_service/echo_service_echo3_parameters.go
+++ b/examples/internal/clients/echo/client/echo_service/echo_service_echo3_parameters.go
@@ -72,11 +72,6 @@ type EchoServiceEcho3Params struct {
 	// Lang.
 	Lang string
 
-	// LineNum.
-	//
-	// Format: int64
-	LineNum *string
-
 	// NIDnID.
 	NIDnID *string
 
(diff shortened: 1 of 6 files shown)


johanbrandhorst (approved):
LGTM, thanks!

PR #7215, opened 2026-08-10
fix(openapiv2): render Any additional properties as true

### LINKED ISSUE #4724: google.protobuf.Any generated OpenAPI v2 yaml that is potentially incorrect
## 🐛 Bug Report
I believe the generated OpenAPI v2 yaml for a google.protobuf.Any is incorrect.
This may also occur on other objects that use `additionalProperties: {}`

## To Reproduce
1. Have a proto file that includes a google.protobuf.Any somewhere
2. Generate the OpenAPI v2 yaml
3. Look at the OpenAPI v2 yaml, try generating OpenAPI client sdk's based on it.

## Expected behavior
I expect the generated yaml to look like this:
```yaml
  protobufAny:
    type: object
    properties:
      '@type':
        type: string
    additionalProperties: true
```

When turned into OpenAPI client sdk for golang, it would become something like this:
```golang
type ProtobufAny struct {
	Type *string `json:"@type"`
	AdditionalProperties map[string]any
}
```

## Actual Behavior
```yaml
  protobufAny:
    type: object
    properties:
      '@type':
        type: string
    additionalProperties: {}
```

I believe this is wrong, because `additionalProperties: {}` can either imply a second level map, or an object without fields.  It should be `additionalProperties: true`

When turned into OpenAPI client sdk for golang, it becomes this:
```golang
type ProtobufAny struct {
	Type *string `json:"@type"`
	AdditionalProps map[string]struct{}
}
```

or

```golang
type ProtobufAny struct {
	Type *string `json:"@type"`
	AdditionalProperties map[string]map[string]interface{}
}
```

In both cases, the client fails on trying to unmarshal into this struct.

## Your Environment
v2.22.0
Linux and Mac

Boulea7:
## Summary

- represent Swagger 2.0 `additionalProperties` as either a boolean or a schema object in the OpenAPI v2 generator
- emit `additionalProperties: true` for `google.protobuf.Any` while preserving schema-valued `additionalProperties` for protobuf maps
- regenerate the affected OpenAPI fixtures and example clients

Fixes #4724.

## Generated files

The 47 changed files consist of three generator/test files and 44 generated files:

- 39 `*.swagger.json` fixtures, each changing `protobufAny.additionalProperties` from `{}` to `true`
- five go-swagger `protobuf_any.go` example clients, which use the generated open-ended-object field name

These files were produced by the repository generation workflow. Repeating generation produced the same diff.

## Verification

- `go test ./protoc-gen-openapiv2/internal/genopenapi -run '^(TestApplyTemplateProtobufAny|TestSchemaOfField)$' -count=1`
- `go test ./protoc-gen-openapiv2/internal/genopenapi -count=1`
- `bazelisk test //protoc-gen-openapiv2/internal/genopenapi:genopenapi_test`
- `staticcheck ./...`
- `go vet -copylocks=false ./protoc-gen-openapiv2/internal/genopenapi`
- `buf lint`
- `make generate`
- `bazelisk run :gazelle`
- `bazelisk run :buildifier`
- `gofmt` and `git diff --check`

`go test ./...` reached every other package but could not start the integration package because `127.0.0.1:9090` was already in use. The unchanged `origin/main` commit failed in the same way. Running the complete integration package on an unused port passed:

```text
go test ./examples/internal/integration -count=1 -args -network=tcp -endpoint=127.0.0.1:19090
```

`go vet ./...` also reports existing warnings in `runtime/context.go` and an unchanged section of `template_test.go`; the focused vet command above passes.

## AI assistance

OpenAI Codex assisted with implementation and verification.


examples/internal/helloworld/helloworld.swagger.json
```diff
@@ -943,7 +943,7 @@
           "type": "string"
         }
       },
-      "additionalProperties": {}
```
johanbrandhorst, line 946: This seems to be invalid OpenAPIv2. According to https://swagger.io/specification/v2/ the additionalProperties must be an object?

johanbrandhorst (changes requested):

examples/internal/helloworld/helloworld.swagger.json
veqryn: Swagger says that these follow the json schema spec:
https://datatracker.ietf.org/doc/html/draft-fge-json-schema-validation-00#section-5.4.4
The spec says `additionalProperties` can be a boolean.

johanbrandhorst: Are you sure? My reading of this is that "instead of the JSON schema definition of these properties (fields), use the OpenAPI Schema Object definition"
image
What's your read?

veqryn: Honestly, the spec seems lacking in what they actually mean here.
But I do know that both OpenAPI and Swagger own code-gen tools that take a yaml v2 spec and turn it into v3 specs.
Both of them do the correct thing for `"additionalProperties": true`, and do the wrong thing for `"additionalProperties": {}` and when it is missing completely (despite the spec appearing to say that all 3 situations are to be treated as identical).

https://github.com/swagger-api/swagger-codegen/issues/12468
and
https://github.com/OpenAPITools/openapi-generator/issues/19596

johanbrandhorst: Yikes, fair enough, that paints a picture. Yeah alright lets try with a boolean I guess?

johanbrandhorst (changes requested):
Please rebase your change on main and fix the CI failures

PR #7237, opened 2026-08-14
openapiv3: add extension (x-*) support

### LINKED ISSUE #7211: `openapiv3`: extension support
Hello,

It could be nice to be able to declare extensions (or `x-*` custom fields) on openapi-v3, in a similar way as [openapi-v2](https://grpc-ecosystem.github.io/grpc-gateway/docs/mapping/customizing_openapi_output/#using-proto-options). 

Example:
```protobuf
service AnyService {
  rpc AnyMethod(AnyMessage) returns (AnyResponse) {
    option (google.api.http) = {
      get: "/v1/anything",
    };
    option (grpc.gateway.protoc_gen_openapiv3.options.openapiv3_operation) = {
      extensions: [
        { key: "x-custom-field", value: "anything" }
      ]
    };
  }
}
```

Resources:
- [OpenAPI v3.x Extension spec](https://spec.openapis.org/oas/v3.1.2.html#specification-extensions)

Thanks!

Abmun:
 
Fixes #7211
 
Yes.
 
Adds OpenAPI specification extension (`x-*`) support to `protoc-gen-openapiv3`, matching the feature `protoc-gen-openapiv2` already has (see `openapiv2.proto`'s `extensions` fields).
 
Concretely:
 
- Adds a `map<string, google.protobuf.Value> extensions` field to the `Document`, `Info`, `Operation`, `Schema`, and `Tag` messages in `protoc-gen-openapiv3/options/openapiv3.proto`, so extensions can be set via the existing `openapiv3_document`, `openapiv3_operation`, `openapiv3_schema`, and `openapiv3_field` annotations.
- Wires these through the generator so `x-*` keys render inline on the corresponding object in the generated OpenAPI 3.1.0 JSON.
- Extension keys must start with `x-` per the [OpenAPI spec](https://spec.openapis.org/oas/v3.1.0#specification-extensions). A key that doesn't is logged and dropped rather than failing generation, consistent with how this generator already degrades other malformed annotation input (e.g. an unresolvable field type) elsewhere.
- A field-level `openapiv3_field` annotation that sets only `extensions` on a `$ref`-typed field now forces an `allOf` wrapper, the same way `title` and `deprecated` already do - extensions can't sit as a `$ref` sibling.
- Regenerated `openapiv3.pb.go` / `openapiv3_protoopaque.pb.go` via `buf generate` with the pinned `protoc-gen-go v1.36.0`.
- Added test coverage for all five annotation sites (including the invalid-key-is-dropped case) in `generator_test.go` / `testdata/annotations.prototext`.
Example:
 
```protobuf
option (grpc.gateway.protoc_gen_openapiv3.options.openapiv3_operation) = {
  extensions: [
    { key: "x-custom-field", value: "anything" }
  ]
};
```
 
#### Other comments
 
`go build`, `go vet`, and `go test ./protoc-gen-openapiv3/...` all pass. No other packages depend on `protoc-gen-openapiv3/options`, so the change is fully contained to that generator.


protoc-gen-openapiv3/internal/genopenapi/annotations.go
```diff
@@ -56,6 +61,37 @@ func fieldSchemaAnnotation(field *descriptor.Field) (*options.Schema, bool) {
 	return s, true
 }
 
+// processExtensions converts a proto options extensions map (as attached to
+// openapiv3_document, openapiv3_operation, openapiv3_schema, and
+// openapiv3_field annotations) into the sorted, JSON-ready form the
+// generator's internal types render inline.
+//
+// Per the OpenAPI 3.1.0 spec, extension keys must start with "x-"; an entry
```
johanbrandhorst: Add a reference to the relevant part of the spec here, as you did in your PR description, please.

protoc-gen-openapiv3/internal/genopenapi/types.go
```diff
@@ -8,6 +8,55 @@ import (
...
+// struct so its own MarshalJSON is not re-entered) and splices the given
+// extensions into the resulting JSON object as additional top-level
+// "x-*" keys. extensions must already be sorted by key; processExtensions
+// guarantees this.
+//
+// This avoids hand-maintaining a parallel struct definition per extension-
+// bearing type: it marshals the real fields via the normal struct tags,
+// then does a small amount of byte-level surgery to append the extension
+// keys before the closing brace.
+func marshalWithExtensions(v any, extensions []extension) ([]byte, error) {
```
johanbrandhorst, line 57: Can't we just have a `MarshalJSON` method on the `extension` type and avoid this byte level surgery?

johanbrandhorst (changes requested):
This is great, thanks! A few comments.

protoc-gen-openapiv3/internal/genopenapi/annotations.go
Abmun: @johanbrandhorst Good call, added the spec link (https://spec.openapis.org/oas/v3.1.0#specification-extensions) to the doc comment, matching what's in the PR description and the .proto comments. Pushed.

protoc-gen-openapiv3/internal/genopenapi/types.go
Abmun: @johanbrandhorst  Moved the per-extension key/value encoding into extension.MarshalJSON(), it now renders a complete valid single-key JSON object ({"x-foo":"bar"}) rather than being hand-formatted inline. marshalWithExtensions still does a small amount of splicing to inline multiple extensions as top-level siblings of the parent object's own fields (rather than nesting them under an "extensions" key, which the spec doesn't want). I don't think that outer splice can be fully eliminated without a heavier reflection-based approach like protoc-gen-openapiv2 uses, but happy to go that route too if you'd prefer consistency with v2's extensionMarshalJSON. Pushed either way for now.

protoc-gen-openapiv3/internal/genopenapi/types.go
```diff
@@ -8,6 +8,74 @@ import (
...
+}
+
+// MarshalJSON renders the extension as a single-key JSON object, e.g.
+// {"x-foo":"bar"}. It is a complete, valid JSON value on its own — even
+// though marshalWithExtensions only ever uses it as a fragment, stripping
+// the surrounding braces to splice multiple extensions inline as
+// top-level siblings of a parent object's own fields (OpenAPI extensions
+// are not nested under an "extensions" key in the rendered document).
+func (e extension) MarshalJSON() ([]byte, error) {
+	key, err := json.Marshal(e.key)
```
johanbrandhorst: What does it mean to `json.Marshal` a value of type `string`? Is it just quoting it? We can just do that manually here instead.

protoc-gen-openapiv3/internal/genopenapi/types.go
johanbrandhorst: I definitely would prefer to avoid the reflection based approach, and I think once Go 1.28 is out and we can use json/v2 we could update this logic to work at the JSON token level instead of byte level.

johanbrandhorst (changes requested):

protoc-gen-openapiv3/internal/genopenapi/types.go
Abmun: @johanbrandhorst Agreed, and glad the byte-splicing approach works for now. I'll leave this as-is rather than switch to the reflection-based approach v2 uses. Happy to revisit once json/v2 lands and this can move to token-level instead of byte-level.

protoc-gen-openapiv3/internal/genopenapi/types.go
Abmun: @johanbrandhorst Replaced json.Marshal(e.key) with strconv.Quote(e.key). It's just quoting/escaping the string, and since the key is already validated to start with "x-" by processExtensions, that path can never actually error. Pushed.

johanbrandhorst (approved):

Issue #7220, opened 2026-08-10
JSON body fields not present in the target message are silently dropped by default — is the lenient default intended, and can the drop be surfaced?


cleitonaugusto:
Hi, and thanks for grpc-gateway — this is a question about a default, not a bug report.

When a JSON request carries a field the target protobuf message does not define, the gateway drops that field while transcoding to protobuf and the RPC proceeds normally. The client receives `200 OK` and the backend receives a well-formed message; neither side is told a field was discarded.

### Minimal reproduction

Service with a message that names two fields:

```proto
message Command {
  string action = 1;
  int32  amount = 2;
}
service Gateway {
  rpc Send(Command) returns (Ack) {
    option (google.api.http) = { post: "/v1/send" body: "*" };
  }
}
```

A default `runtime.NewServeMux()` in front of a gRPC backend. Then:

```
POST /v1/send  {"action":"transfer","amount":1,"auth":"...."}
```

The backend receives only `action` and `amount`; the `auth` field is gone, and the response is `200 OK`. Measured on **grpc-gateway v2.30.0**, protobuf **v1.36.12**, Go **1.25**.

### The part that surprised me

The default marshaler resolves to `protojson` with `DiscardUnknown: true`, which is **more lenient than `protojson`'s own default** (`DiscardUnknown: false`, which rejects unknown fields). Opting back into strictness works and is easy once you know to look:

```go
runtime.WithMarshalerOption(runtime.MIMEWildcard, &runtime.JSONPb{
    UnmarshalOptions: protojson.UnmarshalOptions{DiscardUnknown: false},
})
```

With that, the same request returns `400` with `unknown field "auth"`. So a strict mode is fully available — the question is about the default and about discoverability.

### Questions

1. Is the lenient default (`DiscardUnknown: true`) intentional, given that the `protojson` library it wraps defaults to strict?
2. Would you consider documenting the security-relevant implication — that any material a client places in a field the schema does not name (for example an authentication or authorization token added by a client that expects the server to see it) is discarded silently, invisibly to both ends?
3. Is there interest in making the strict option more discoverable, or in surfacing the dropped-field event (a log line or a metric), so an operator can notice it without knowing to configure it in advance?

[...]

PR #7282, opened 2026-08-25
docs: add a security considerations page

cleitonaugusto:
Follows up on #7220, where @johanbrandhorst suggested a docs page on security considerations.

It covers the behaviour from that issue: JSON fields the target message does not define are dropped while transcoding, and neither end is told. The page says when that matters, how to reject unknown fields instead, and how to log them without rejecting for people who cannot break existing clients.

I put it under Operations rather than Mapping, on the guess that it is something you check before putting a gateway in front of a service rather than something you look up while writing protos. That is one line of front matter to change if you would rather it sat next to the marshaler docs.

Everything in it is either from the source or from what I measured for the issue: the default is in `runtime/marshaler_registry.go`, and the `400` with `unknown field` is what I got after flipping the setting on v2.30.0.

First draft, so cut anything that reads as too long. I left out the comparison with other transcoders, since that felt more like my notes than something a reader of these docs needs.

### DIFF
--- a/docs/docs/operations/security_considerations.md
+++ b/docs/docs/operations/security_considerations.md
@@ -1,0 +1,102 @@
+---
+layout: default
+title: Security considerations
+nav_order: 6
+parent: Operations
+---
+
+# Security considerations
+
+Things worth knowing about the defaults before you put a gateway in front of a
+service that matters.
+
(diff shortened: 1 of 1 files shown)


johanbrandhorst (approved):
This is great!

PR #7323, opened 2026-09-03
Fix template_test format strings

skitt:
#### References to other Issues or PRs

N/A

#### Have you read the [Contributing Guidelines](https://github.com/grpc-ecosystem/grpc-gateway/blob/main/CONTRIBUTING.md)?

Yes.

#### Brief description of what is fixed or changed

This fixes `template_test.go` format strings: when `is` and `want` aren't strings, this uses `%#v` rather than `%s` (which fails the build with Go 1.27).

PR #7325, opened 2026-09-03
Move examples to their own module

### LINKED ISSUE #7317: Move the examples to a separate Go module
## 🚀 Feature

With the recent switch to go-swagger (#6735), the grpc-gateway dependency tree has become rather large; would it be possible to move the examples to a separate submodule? This would drop 38 dependencies from the main `go.mod`, helping all dependants.

(This shows up in Kubernetes where we prefer our own OpenAPI modules to `go-openapi`; see https://github.com/kubernetes/kubernetes/pull/141360.)

skitt:

Fixes #7317

Yes.

This moves examples to a separate module, pruning many dependencies from the main module, which is beneficial for dependants.

The build and CI are adjusted to match: builds, tests, and tidy operations operate on any directory containing a go.mod file. Tests run specifically on examples are run from the examples module.

#### Other comments

I’m submitting this as a draft because I don’t know how to configure a new Bazel module.

This includes #7323 which should be merged separately.


johanbrandhorst:
I ran the CI jobs but it seems like there's a problem. Honestly I don't really know how to carve out a separate Bazel module either, I'd recommend getting advice from your favorite AI assistant 😁.

skitt:
> I ran the CI jobs but it seems like there's a problem. Honestly I don't really know how to carve out a separate Bazel module either, I'd recommend getting advice from your favorite AI assistant 😁.

Right, I’m missing an escape on the `find` commands throughout. I’ll figure out the Bazel side of things too :wink:.

skitt:
@johanbrandhorst it seems that dealing with multiple modules in Bazel with Gazelle is rather difficult (it needs references to commits, which then need constantly updating). Would dropping the Bazel build be at all a possibility? Kubernetes did it a few years ago, essentially because the benefits of the Bazel build were no longer relevant thanks to improvements in the Go ecosystem, because maintaining two build systems in parallel was difficult, and because the Bazel infrastructure caused a fair amount of friction. See https://github.com/kubernetes/kubernetes/issues/88553 for details.

johanbrandhorst:
Hm, I don't have any clear numbers on how many users are using the Bazel integration, but I'm not sure that we should remove it just because setting up a nested testing module seems hard. We already have various CI automations that run bazel jobs for dependabot/renovate PRs, if it's something we can just script, I'd rather we do that following the patterns we use today. Could you explore if we can use something similar here?

PR #7351, opened 2026-09-10
fix(openapiv2): emit a primitive type for well-known object query parameters

### LINKED ISSUE #4758: Incorrect google.protobuf.Empty representation with oneof in GET request
## 🐛 Bug Report

Hi guys,

I have the following files

`AdminAPI.proto`:
```proto
syntax = "proto3";
option go_package = "./admin";

package admin;

import "google/protobuf/empty.proto";

service AdminService {
  rpc GetInstruments(GetInstrumentsRequest) returns (InstrumentList);
}

message InstrumentNameList {
  repeated string names = 1;
}

message GetInstrumentsRequest {
  oneof instrument_name_filter {
    google.protobuf.Empty all_instrument_names = 1;
    InstrumentNameList instrument_name_list = 2;
  }
}

message InstrumentList {
  repeated Instrument instruments = 1;
}

message Instrument {
  // long definition here
}
```

`grpc_api_configuration.yaml`:
```yaml
type: google.api.Service
config_version: 3

http:
  rules:
    - selector: admin.AdminService.GetInstruments
      get: /v1/instruments
```

`buf.gen.yaml`:
```yaml
version: v1
plugins:
  - plugin: go
    out: gen
    opt:
      - paths=source_relative
  - plugin: go-grpc
    out: gen
    opt:
      - paths=source_relative
  - plugin: grpc-gateway
    out: gen
    opt:
      - paths=source_relative
      - grpc_api_configuration=grpc_api_configuration.yaml
  - plugin: openapiv2
    out: docs
    opt:
      - grpc_api_configuration=grpc_api_configuration.yaml
      - output_format=yaml

```

After generating code with `buf generate` I get the following in my swagger:
```yaml
paths:
  /v1/instruments:
    get:
      operationId: AdminService_GetInstruments
      responses:
        "200":
          description: A successful response.
          schema:
            $ref: '#/definitions/adminInstrumentList'
        default:
          description: An unexpected error response.
          schema:
            $ref: '#/definitions/rpcStatus'
      parameters:
        - name: allInstrumentNames
          in: query
          required: false
          type: object
        - name: instrumentNameList.names
          in: query
          required: false
          type: array
          items:
            type: string
          collectionFormat: multi
      tags:
        - AdminService
```

Seems it's impossible to provide `allInstrumentNames` in query. I tried different options, but always get an error
```
{
    "code": 3,
    "message": "parsing field \"all_instrument_names\": unsupported message type: \"google.protobuf.Empty\"",
    "details": []
}
```

Underlying gRPC service expects whether `all_instrument_names` or `instrument_name_list` to be filled, and we cannot change it.

If proto service method accepts just `google.protobuf.Empty`, everything works fine since REST server doesn't except any query parameters.

Also, if I change method to POST, everything becomes ok with the following request:
```json
{
    "allInstrumentNames": {}
}
```

So are there any ways to fix it?

## Expected behavior

There is a way to pass `google.protobuf.Empty` struct in query parameter.

## Actual Behavior

REST server returns error
```json
{
    "code": 3,
    "message": "parsing field \"all_instrument_names\": unsupported message type: \"google.protobuf.Empty\"",
    "details": []
}
```

## Your Environment

`grpc-gateway` version is v2.22.0

iamrajiv:

Fixes #4758

Yes.

Option A from the discussion in #4758.

`google.protobuf.Empty` and `google.protobuf.Struct` render as `type: object`, which is not
valid for a parameter outside the body, and Empty could not be parsed from a query string at
all (`unsupported message type: "google.protobuf.Empty"`), so the parameter we document in
10 ABE operations could not be sent by anyone.

- Generator: non-body parameters whose well-known type renders as an object are
  `type: string` — Empty, Struct, and the `structValue` member of `google.protobuf.Value`.
  The runtime already read Struct and Value as JSON text, so this just describe
- Runtime: `parseMessage` gains a `google.protobuf.Empty` case, accepting an empty value or
  `{}` and rejecting anything else.

```
?oneof_empty=      -> 200, oneofValue set to oneofEmpty {}
?oneof_empty={}    -> 200, same (?oneofEmpty= works too)
?oneof_empty=true  -> parsing field "oneof_empty": expected an empty value or "{}", got "true"
```

Nothing breaks — the old behaviour was a hard INVALID_ARGUMENT.

#### Other comments

- go-swagger validation of the ABE spec drops from 84 errors to 24; all 60 remo
  those 10 parameters, no new errors. The rest are the pre-existing `floatValue.default` ones.
- No spec in the repo now has a non-body parameter with a missing or `object` t
  Makefile `jq` pass no longer needs its coercion branches. Removed them; all five clients
  regenerate byte-identically. Happy to drop that commit if you'd rather keep t
- Tests at three levels, each failing without the fix: parameter type, query parsing, and a
  round trip through a real gateway (400 before, 200 after).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,24 +17,17 @@ GENERATE_UNBOUND_METHODS_EXAMPLE_SPEC=examples/internal/proto/examplepb/generate
 # Targets are PHONY because the file set isn't fully predictable from the
 # spec without parsing it.
 #
-# The jq pass coerces non-body parameters with no `type`/`schema` (map<K,V>
-# request fields) and non-body parameters typed as `object` (oneof-of-empty
-# fields) to `string`. They can't round-trip through a query string in a
-# meaningful way, but go-swagger needs *some* primitive to chew on. The
-# accompanying default-value rewrite turns string defaults on numeric/integer
-# parameters into the corresponding number — protoc-gen-openapiv2 emits e.g.
-# `"default": "0"` on integer params, which go-swagger refuses.
+# The jq pass turns string defaults on numeric/integer parameters into the
+# corresponding number — protoc-gen-openapiv2 emits e.g. `"default": "0"` on
+# integer params, which go-swagger refuses.
 define swagger_client
 	rm -rf $(1)/client $(1)/models
 	mkdir -p $(1)
 	jq 'walk( \
-		if type == "object" and has("in") and (.in != "body") then \
-			(if (has("type") | not) and (has("schema") | not) then . + {type: "string"} \
-			 elif .type == "object" then . + {type: "string"} \
-			 else . end) \
-			| (if (.type == "number" or .type == "integer") and (.default | type) == "string" \
-			   then .default |= tonumber else . end) \
-		else . end \
+		if type == "object" and has("in") and (.in != "body") \
+		   and (.type == "number" or .type == "integer") \
+		   and (.default | type) == "string" \
+		then .default |= tonumber else . end \
 	)' $(2) > $(1)/.spec.json
 	swagger generate client -q -f $(1)/.spec.json -t $(1) \
 		--client-package=client --model-package=models --skip-validation
--- a/examples/internal/integration/integration_test.go
+++ b/examples/internal/integration/integration_test.go
@@ -2159,6 +2159,20 @@ func testRequestQueryParams(t *testing.T, port int) {
 			},
 		},
 		{
+			name:        "get oneof of empty url query value",
+			httpMethod:  "GET",
+			contentType: "application/json",
+			apiURL:      fmt.Sprintf("http://localhost:%d/v1/example/a_bit_of_everything/params/get/foo?oneof_empty=", port),
+			wantContent: &examplepb.ABitOfEverything{
+				SingleNested: &examplepb.ABitOfEverything_Nested{
+					Name: "foo",
+				},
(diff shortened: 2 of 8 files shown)


Makefile
```diff
@@ -17,24 +17,19 @@ GENERATE_UNBOUND_METHODS_EXAMPLE_SPEC=examples/internal/proto/examplepb/generate
...
-# fields) to `string`. They can't round-trip through a query string in a
-# meaningful way, but go-swagger needs *some* primitive to chew on. The
-# accompanying default-value rewrite turns string defaults on numeric/integer
-# parameters into the corresponding number — protoc-gen-openapiv2 emits e.g.
-# `"default": "0"` on integer params, which go-swagger refuses.
+# The jq pass turns string defaults on numeric/integer parameters into the
+# corresponding number — protoc-gen-openapiv2 emits e.g. `"default": "0"` on
+# integer params, which go-swagger refuses. It used to also coerce non-body
+# parameters that had no type, or a type of `object`, to `string`; the
+# generator no longer emits either, so that part is gone.
```
johanbrandhorst: We don't need to include historic information in these comments, remove the last sentence.

johanbrandhorst (changes requested):
Looks like we need a minor bazel file fix too, but otherwise LGTM!

johanbrandhorst (approved):

PR #7369, opened 2026-09-16
reject out-of-range numeric enum query values in parseField

itecz26:

Yes.

numeric enum query params are parsed with `strconv.Atoi` and then converted to `protoreflect.EnumNumber` (int32) in `parseField`, so a value outside the int32 range is silently truncated rather than rejected. on 64-bit builds `?enum_value=4294967296` (2^32) wraps to 0 and is accepted as the zero enum value; on 32-bit builds `Atoi` already errors, so the behavior is platform dependent. switched to `strconv.ParseInt(value, 10, 32)` so the value is bounded like the `Int32Kind` path and out-of-range numeric enums are rejected consistently.

added a case to `TestPopulateParameters` for the `2^32` input.


johanbrandhorst (approved):
Thanks!

PR #7371, opened 2026-09-16
feat(protoc-gen-openapiv3): support response_body binding

tobiaskohlbau:

The current implementation of protoc-gen-openapiv3 does not support response_body binding similiar to how v2 already does it. If the proto definition maps the response_body onto a specific field it should also be rendered correctly in the openapi specification.

#### Other comments

This is me first time contributing, let me know if something is off or I should add.


johanbrandhorst (changes requested):
Thanks for your contribution! Please add a test for this new functionality, ideally with a prototext test file similar to what we have for the existing tests.

tobiaskohlbau:
I've tried to add a test like described by you. I'm not familiar with prototext and tried to do my best in there. Let me know if something is not up on par.

johanbrandhorst (approved):
LGTM, thanks!

PR #7382, opened 2026-09-18
clamp overflowing grpc-timeout to max duration

itecz26:

none

yes

`timeoutDecode` computes `d * time.Duration(t)` from the `Grpc-Timeout` request header with no overflow guard, so a large value (even a spec-valid 8-digit hour count like `99999999H`) wraps int64 and yields a bogus, often negative, duration that then becomes the request context deadline. switched the value parse to `ParseUint` and clamp to `math.MaxInt64` when the multiply would overflow, matching how grpc-go's `decodeTimeout` handles the same input.

#### Other comments

added an internal test for the overflow/clamp and negative-input cases; the existing `Grpc-Timeout` tests (including the 9-digit values) still pass.


johanbrandhorst (approved):
Nice!

PR #7389, opened 2026-09-20
fix(openapiv2): emit minItems and maxItems on query parameters

### LINKED ISSUE #7283: openapiv2: min_items/max_items not emitted on query parameters for repeated fields
## 🐛 Bug Report

For repeated fields used as query parameters, openapiv2_field min_items / max_items are not written into the OpenAPI parameter object. unique_items is emitted correctly, but min_items / max_items are dropped.

Cause: openapiParameterObject lacked MaxItems, and query-param mapping in nestedQueryParams only copies UniqueItems from the field schema (not MinItems / MaxItems). These constraints already work for schema/body properties.

## To Reproduce
1. Define a repeated field with array constraints:
```protobuf
repeated string tags = 1 [(grpc.gateway.protoc_gen_openapiv2.options.openapiv2_field) = {
  min_items: 1
  max_items: 10
  unique_items: true
}];
```
2. Expose the field as a query parameter (e.g. a GET method with no body, or a field not covered by body).
3. Generate OpenAPI with protoc-gen-openapiv2 and inspect the parameter for tags.

## Expected behavior

The query parameter object includes all three constraints:

```json
{
  "name": "tags",
  "in": "query",
  "type": "array",
  "collectionFormat": "multi",
  "minItems": 1,
  "maxItems": 10,
  "uniqueItems": true
}
```

## Actual Behavior

Only uniqueItems appears on the parameter. minItems and maxItems are missing (even though the same annotation correctly emits them on schema/body properties).

## Your Environment

grpc-gateway / protoc-gen-openapiv2 (main)
macOS (darwin)

wofiporia:
﻿#### References to other Issues or PRs

Fixes #7283

yes

When repeated fields with `min_items` or `max_items` schema options are mapped to query parameters, `protoc-gen-openapiv2` omitted these constraints in the resulting OpenAPI parameter object (only `unique_items` was copied).

This change adds `MaxItems` to `openapiParameterObject` and copies `MinItems` and `MaxItems` from the field schema when constructing query parameters for repeated fields in `nestedQueryParams`.

#### Other comments

Added unit test `TestQueryParametersIncludeRepeatedFieldMinMaxItems` in `template_test.go` covering both constrained and unconstrained repeated query parameters.


johanbrandhorst (changes requested):
Could you add this annotation to one of our example protobuf files so we can see end-to-end how it propagates into the appropriate place in a generated swagger.json file?

wofiporia:
Added `min_items: 1` and `max_items: 5` to `repeated_string_annotation` in `a_bit_of_everything.proto` and regenerated the swagger specs across the example pb definitions.

johanbrandhorst:
Bazel error looks like a blip but there's a diff in the generated Go files too I think?

wofiporia:
You were right — the generated Go files were stale. I'd regenerated the swagger specs but not run the full `make generate`, so `a_bit_of_everything.pb.go` (the annotations are serialized into the `rawDesc` bytes) and the `abe` go-swagger client (now emitting `validateRepeatedStringAnnotation` for both bounds) were missing.

Regenerated in the `build-env` image and checked in — that's the 7 files the `generate` job was flagging. And yes, `bazel/fuzz` was the `rules_cc` 500 blip.

johanbrandhorst (approved):

PR #7390, opened 2026-09-20
validate forwarded header values before adding to grpc metadata

itecz26:

N/A

Yes

annotateContext runs isValidGRPCMetadataTextValue on every header in the forwarding loop and skips values with bytes outside 0x20-0x7e, but the special-cased authorization, x-forwarded-host and x-forwarded-for values are appended to the metadata without that check. so a client-supplied `authorization: tok\x01en` or `x-forwarded-host: bad\x7fhost` reaches the backend as gRPC metadata even though an identical `Grpc-Metadata-*` header with the same bytes is dropped. this applies the same validation at those three sites; valid values are unaffected, and an invalid x-forwarded-host now falls back to req.Host the same way an absent one does.

#### Other comments

added TestAnnotateContext_SkipsInvalidForwardedValues, which fails on current main (the invalid values are forwarded) and passes with the fix.


johanbrandhorst (approved):
Thanks!

PR #7397, opened 2026-09-22
reject out-of-range enum numbers in jsonpb non-proto decoder

itecz26:

yes

when an http rule binds the request body straight to an enum field (`body: "enum_field"`, or a repeated enum), decoding goes through `decodeNonProtoField` instead of protojson. there a json number was run through `int32(v)` on the decoded `float64`, so `4294967297` silently wrapped to `-2147483648` and `1.9` truncated to `1`, whereas protojson rejects both with "invalid value for enum field". guard the conversion with an integral + int32-range check so those inputs error out the same way protojson does.

#### Other comments

added a regression test covering scalar and repeated enum decoding.


johanbrandhorst (approved):
Thanks!

